PR #29, opened 2023-09-24
cashu Js Bindings

thesimplekid:


thesimplekid:
https://github.com/thesimplekid/cashu-crab/issues/20

thesimplekid:
Places where i manually use serde should use wasm serde

https://rustwasm.github.io/docs/wasm-bindgen/reference/arbitrary-data-with-serde.html

- [ ] nut06

PR #40, opened 2023-11-07
fix: compile error with optional nuts

ngutech21:
running "just build" local results in multiple compile errors. This is a fix for the first one:

`error[E0599]: no method named `check_spendable` found for struct `cashu_sdk::client::blocking::Client` in the current scope
  --> bindings/cashu-sdk-ffi/src/client.rs:97:18
   |
96 | / ...   self.inner
97 | | ...       .check_spendable(&proofs.iter().map(|p| p.as_ref().deref().clone()).collect())?
   | |           -^^^^^^^^^^^^^^^ method not found in `Client`
   | |___________|
   | 

error[E0599]: no method named `get_info` found for struct `cashu_sdk::client::blocking::Client` in the current scope
   --> bindings/cashu-sdk-ffi/src/client.rs:103:32
    |
103 |         Ok(Arc::new(self.inner.get_info()?.into()))
    |                                ^^^^^^^^ method not found in `Client`
`  
I think CI didn't catch this one, because CI never builds the full workspace without any feature flags like "just build" does.

### DIFF
--- a/bindings/cashu-sdk-ffi/Cargo.toml
+++ b/bindings/cashu-sdk-ffi/Cargo.toml
@@ -9,14 +9,12 @@ rust-version.workspace = true
 name = "cashu_sdk_ffi"
 crate-type = ["cdylib", "staticlib"]
 
-
 [dependencies]
 cashu-ffi = { path = "../cashu-ffi" }
-cashu-sdk = { path = "../../crates/cashu-sdk", default-features = false, features = ["wallet", "mint", "blocking"] }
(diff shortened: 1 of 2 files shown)


thesimplekid:
I'm working on fixing up optional features as I'm working on NUT11 and NUT12. The blocking client one is a weird one that has pop up recently that i haven't' figured out myself. Eventually I will make the FFI bindings async and wont need the blocking client.

thesimplekid:
`just build` is working for me on main now after refactoring the client. Let me know if you're still running into issues

ngutech21:
Everything works fine. Thank you

PR #42, opened 2023-11-07
consolidate split-amount

ngutech21:
There are two implementations of a split amount function that can be consolidated. The function utils::split_amount() is only used once in nut00::random() and could be replaced by Amount.split(). The only difference of the two implementations is the order in which the amounts occur, but I think this doesn't matter for the one usage in the random() function.

The tests have been moved to the amount module and adopted to the new ordering of the implementation.


thesimplekid:
Thank you, good catch think this is left over from when i originally had multiple amount types in the crate but is not needed anymore.

PR #59, opened 2024-03-12
fix(nut11): use 32 byte pubkey

thesimplekid:


crates/cashu/src/nuts/nut11.rs
```diff
@@ -215,7 +214,7 @@ impl TryFrom<P2PKConditions> for Secret {
             return Err(Error::Amount);
         }
 
-        let data: PublicKey = pubkeys[0].clone().into();
+        let data = pubkeys[0].clone();
 
         let data = data.to_string();
```
lorenzolfm, line 219: Throws a runtime error if no item at index 0. Might want to do something like

```suggestion
        let data = match pubkeys.get(0) {
            Some(data) => data.to_string(),
            None => return Err(Error::CustomError(String::from("No pubkey at index 0"))),
        };
```

crates/cashu/src/nuts/nut11.rs
```diff
@@ -237,7 +236,11 @@ impl TryFrom<P2PKConditions> for Secret {
             tags.push(Tag::Refund(refund_keys).as_vec())
         }
 
-        tags.push(Tag::SigFlag(sig_flag).as_vec());
+        if !sig_flag.is_default() {
+            tags.push(Tag::SigFlag(sig_flag).as_vec());
+        }
+
+        let tags = if !tags.is_empty() { Some(tags) } else { None };
```
lorenzolfm, line 243: you can use `then` to avoid the if/else:

```suggestion
        let tags = (!tags.is_empty()).then(|| tags);
```

crates/cashu/src/nuts/nut11.rs
thesimplekid: There is a check that there is at least one pubkey on the line above, but this is a cleaner way to do it thanks

thesimplekid:
Closing as cashu will use 33 byte pubkeys

PR #73, opened 2024-04-08
Decode and encode Token V4

### LINKED ISSUE #68: Implement TokenV4
https://github.com/cashubtc/nuts/pull/109

davidcaseria:
Closes #68 

I have an open question on the draft NUT if secrets should be bytes instead of strings: https://github.com/cashubtc/nuts/pull/109/files#r1555859864. However, as it stands, this successfully decodes and encodes v4 tokens.

### DIFF
--- a/crates/cashu/Cargo.toml
+++ b/crates/cashu/Cargo.toml
@@ -7,7 +7,7 @@ readme = "README.md"
 homepage.workspace = true
 repository.workspace = true
 license.workspace = true
-rust-version.workspace = true # MSRV
+rust-version.workspace = true                      # MSRV
 description = "Cashu rust wallet and mint library"
 
 
@@ -27,12 +27,12 @@ nut13 = ["dep:bip39", "dep:bip32", "nut09"]
 
 [dependencies]
 base64 = "0.22.0"
-bitcoin = { version = "0.31.0", features=["serde",  "rand"] }
+bitcoin = { version = "0.31.0", features = ["serde", "rand"] }
 bip39 = { version = "2.0.0", optional = true }
 bip32 = { version = "0.5.1", optional = true }
 hex = "0.4.3"
-k256 = { version = "0.13.1", features=["arithmetic", "serde", "schnorr"] }
-lightning-invoice = { version = "0.29.0", features=["serde"] }
+k256 = { version = "0.13.1", features = ["arithmetic", "serde", "schnorr"] }
+lightning-invoice = { version = "0.29.0", features = ["serde"] }
 log = "0.4.2"
 rand = "0.8.5"
 serde = { workspace = true }
@@ -42,7 +42,7 @@ url = { workspace = true }
 itertools = "0.12.0"
 thiserror = { workspace = true }
 uuid = { version = "1.6.1", features = ["v4"] }
-ciborium = "0.2.2"
+ciborium = { version = "0.2.2", default-features = false, features = ["std"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { workspace = true }
--- a/crates/cashu/Cargo.toml
+++ b/crates/cashu/Cargo.toml
@@ -42,6 +42,7 @@ url = { workspace = true }
 itertools = "0.12.0"
 thiserror = { workspace = true }
 uuid = { version = "1.6.1", features = ["v4"] }
+ciborium = "0.2.2"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { workspace = true }
--- a/crates/cashu/src/nuts/nut00.rs
+++ b/crates/cashu/src/nuts/nut00.rs
@@ -118,6 +118,7 @@ pub mod wallet {
 
     use base64::engine::{general_purpose, GeneralPurpose};
     use base64::{alphabet, Engine as _};
+    use serde::ser::Error;
     use serde::{Deserialize, Serialize};
     use url::Url;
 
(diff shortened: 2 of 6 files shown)


thesimplekid:
Thanks for this. Going to leave open for now as the NUT is still a draft

crates/cashu/Cargo.toml
```diff
@@ -42,6 +42,7 @@ url = { workspace = true }
 itertools = "0.12.0"
 thiserror = { workspace = true }
 uuid = { version = "1.6.1", features = ["v4"] }
+ciborium = "0.2.2"
```
lorenzolfm, line 45: I would recommend always adding a dependency with the default-features set to `false` and than manually importing the features that you want to use in the project. This helps with:

* Better control over dependencies
* Less dependency conflicts
* Reduced binary size 
* Improved compile times

For this specific one:
```suggestion
ciborium = { version = "0.2.2", default-features = false, features = ["std"] }
```

crates/cashu/src/nuts/nut00.rs
```diff
@@ -366,34 +366,89 @@ pub mod wallet {
...
-            let decoded_str = String::from_utf8(decoded)?;
-            let token: Token = serde_json::from_str(&decoded_str)?;
-            Ok(token)
+            let decoded = GeneralPurpose::new(&alphabet::URL_SAFE, decode_config).decode(s)?;
+            if is_v3 {
+                let decoded_str = String::from_utf8(decoded)?;
+                let token: Token = serde_json::from_str(&decoded_str)?;
+                Ok(token)
+            } else {
+                println!("{}", hex::encode(decoded.clone()));
```
lorenzolfm, line 397: should this println! be here?

crates/cashu/src/nuts/nut00.rs
```diff
@@ -366,34 +366,89 @@ pub mod wallet {
...
-            let s = if s.starts_with("cashuA") {
+            let is_v3 = s.starts_with("cashuA");
+            let is_v4 = s.starts_with("cashuB");
+            let s = if is_v3 {
                 s.replace("cashuA", "")
+            } else if is_v4 {
+                s.replace("cashuB", "")
             } else {
                 return Err(wallet::Error::UnsupportedToken);
             };
```
lorenzolfm, line 387:
```suggestion
            let (is_v3, s) = match (s.strip_prefix("cashuA"), s.strip_prefix("cashuB")) {
                (Some(s), None) => (true, s),
                (None, Some(s)) => (false, s),
                _ => return Err(wallet::Error::UnsupportedToken),
            };
```

crates/cashu/src/nuts/nut00.rs
```diff
@@ -366,34 +366,89 @@ pub mod wallet {
...
+        pub memo: Option<String>,
+        /// Token Unit
+        #[serde(rename = "u", skip_serializing_if = "Option::is_none")]
+        pub unit: Option<CurrencyUnit>,
+    }
+
+    impl fmt::Display for TokenV4 {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            let mut data = Vec::new();
+            ciborium::into_writer(self, &mut data).unwrap();
```
lorenzolfm, line 437: i highly discourage using `unwrap`.

Recommend going with `expect` and adding useful log to the user

crates/cashu/src/nuts/nut02.rs
```diff
@@ -45,6 +45,15 @@ pub struct Id {
 
 impl Id {
     const STRLEN: usize = 14;
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        hex::decode(self.to_string()).unwrap()
```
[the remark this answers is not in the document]
thesimplekid: i would rather we change the function signature to a result

crates/cashu/src/nuts/nut00.rs
thesimplekid: I agree unwrap should only be allowed within tests. Ideally a specific and useful error is returned

crates/cashu/src/nuts/nut00.rs
```diff
@@ -486,6 +567,118 @@ impl PartialOrd for Proof {
...
+
+// fn deserialize_v4_secret<'de, D>(deserializer: D) -> Result<Secret, D::Error>
+// where
+//     D: serde::Deserializer<'de>,
+// {
+//     let bytes = Vec::<u8>::deserialize(deserializer)?;
+//     Ok(Secret(
+//         String::from_utf8(bytes).map_err(serde::de::Error::custom)?,
+//     ))
+// }
```
thesimplekid, line 650: I think these can be removed now that your question on the Nut is resolved?

crates/cashu/src/nuts/nut02.rs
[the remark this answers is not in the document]
lorenzolfm: then maybe the function name should be `try_into_bytes` :thinking:

crates/cashu/src/nuts/nut00.rs
```diff
@@ -366,34 +366,89 @@ pub mod wallet {
...
-            let token: Token = serde_json::from_str(&decoded_str)?;
-            Ok(token)
+            let decoded = GeneralPurpose::new(&alphabet::URL_SAFE, decode_config).decode(s)?;
+            if is_v3 {
+                let decoded_str = String::from_utf8(decoded)?;
+                let token: Token = serde_json::from_str(&decoded_str)?;
+                Ok(token)
+            } else {
+                println!("{}", hex::encode(decoded.clone()));
+                let token: TokenV4 = ciborium::from_reader(&decoded[..])?;
```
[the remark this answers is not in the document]
davidcaseria: The decoded bytes are CBOR not JSON.

crates/cashu/src/nuts/nut02.rs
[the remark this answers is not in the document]
davidcaseria: I don't think we should change the signature because this _should_ never fail. However, the only reason it can fail now is that we store the hex string instead of the bytes in this object. Ideally, the `Id` struct should hold the bytes. Should I mark this as a TODO for now and follow up with a refactor shortly?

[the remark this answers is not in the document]
thesimplekid: Yes that makes sense

davidcaseria:
Closing in favor of new PR given the major code refactor.

thesimplekid:
> Closing in favor of new PR given the major code refactor.

Yeah sorry about that

davidcaseria:
No worries!

PR #81, opened 2024-04-10
consolidate to one CDK crate

thesimplekid:
Having the two crates adds complexity without benefit since features can be used instead.

Credit: Thought about this after seeing Yuki's fork


thesimplekid:
@yukibtc  @lorenzolfm @davidcaseria thoughts?

lorenzolfm:
I agree with this approach. Seems much simpler :)

.github/workflows/ci.yml
```diff
@@ -26,14 +26,12 @@ jobs:
...
-            -p cashu --no-default-features --features mint,
-            -p cashu --no-default-features --features all-nuts,
-            -p cashu-sdk,
-            -p cashu-sdk --no-default-features,
-            -p cashu-sdk --no-default-features --features all-nuts,
+            -p cdk,
+            -p cdk --no-default-features,
+            -p cdk --no-default-features --features wallet,
+            -p cdk --no-default-features --features mint --features redb,
+            -p cdk --no-default-features --features mint --features redb,
```
yukibtc, line 33: This line should be removed

misc/scripts/check-crates.sh
```diff
@@ -24,11 +24,10 @@ fi
 
 buildargs=(
     "-p cashu"
```
yukibtc, line 26: This should be renamed to `cdk`

yukibtc (changes requested):
Looks good. I found only 2 things to fix

Issue #88, opened 2024-04-14
Remove serde utils mod


thesimplekid:
https://github.com/thesimplekid/cashu-crab/blob/664776f1224840c2f7ab944348e6bf534ff993f2/crates/cdk/src/serde_utils.rs#L1-L7

Issue #87, opened 2024-04-14
Module should be moved to file for code style


thesimplekid:
https://github.com/thesimplekid/cashu-crab/blob/664776f1224840c2f7ab944348e6bf534ff993f2/crates/cdk/src/nuts/nut13.rs#L61-L63

yukibtc:
I can take both this and the #88 if no one is working on them

thesimplekid:
Thanks. All yours :)

PR #91, opened 2024-04-17
cdk: adj. code according to `CODE_STYLE` and remove `serde_utils` mod

yukibtc:
* Remove `wallet` and `mint` features from `nuts` module
* Sort dependencies in ASC order
* Fix `check-docks.sh` script

Closes https://github.com/cashubtc/cdk/issues/88
Closes https://github.com/cashubtc/cdk/issues/87


crates/cdk/src/nuts/nut00.rs
```diff
@@ -46,18 +57,26 @@ pub enum Error {
...
+    ///
+    /// ID from which we expect a signature.
     #[serde(rename = "id")]
     pub keyset_id: Id,
     /// Blinded secret message (B_)
+    ///
+    /// The blinded secret message generated by the sender.
     #[serde(rename = "B_")]
-    pub b: PublicKey,
+    pub secret: PublicKey,
```
thesimplekid, line 76: I'm not sure that we should rename this to secret as its actually the blinded secret and not the secret itself. Though a more descriptive name maybe a good idea `blinded_secret`?

PR #98, opened 2024-04-24
Fix precommit scrips check

thesimplekid:
This removes rustup from the scripts. As rust versions should be managed through the nix shell and not rust up.


thesimplekid:
closes #76

PR #96, opened 2024-04-24
feat: use bip32 DerivationPath for mint keyset

EthnTuttle:

PR #100, opened 2024-04-24
Refactor Mint to use bip32 derivation and not store priv keys

### LINKED ISSUE #97: Use seed instead of Mnemonic for creating Mint

### LINKED ISSUE #99: Only store MintKeySetInfo in the DB
KeySetInfo plus the seed is enough to regenerate the  MintKeySet on start up

https://github.com/cashubtc/cdk/blob/c8e61b34b8014946114823c435b8a086e559efb6/crates/cdk/src/mint.rs#L746

davidcaseria:
Closes #97 
Closes #99


crates/cdk/src/mint.rs
```diff
@@ -762,3 +761,33 @@ impl From<MintKeySetInfo> for KeySetInfo {
...
+            .derive_priv(secp, &derivation_path)
+            .expect("RNG busted"),
+        unit,
+        max_order,
+    );
+    let keyset_info = MintKeySetInfo {
+        id: keyset.id,
+        unit: keyset.unit.clone(),
+        active: true,
+        valid_from: SystemTime::now()
```
thesimplekid, line 784: We can use the util function here 


https://github.com/cashubtc/cdk/blob/c8e61b34b8014946114823c435b8a086e559efb6/crates/cdk/src/util/mod.rs#L36-L41

crates/cdk/src/mint.rs
```diff
@@ -328,23 +321,24 @@ impl Mint {
...
             .await?
             .ok_or(Error::InactiveKeyset)?;
 
         // Check that the keyset is active and should be used to sign
-        if keyset.id.ne(&active) {
+        if keyset_info.id.ne(&active) {
             return Err(Error::InactiveKeyset);
         }
 
+        let keyset = self.generate_keyset(keyset_info.clone());
```
thesimplekid, line 341: I don't think we should regenerate the full `keyset` for every blind sign, it would negatively effect performance. Maybe we can keep the active `keysets` in memory but not persisted to the db. This way the active `keysets` can be generated from the seed only once on mint start up and not on each mint and swap request.

crates/cdk/src/cdk_database/mod.rs
```diff
@@ -102,9 +102,9 @@ pub trait MintDatabase {
     async fn get_melt_quotes(&self) -> Result<Vec<MeltQuote>, Self::Err>;
     async fn remove_melt_quote(&self, quote_id: &str) -> Result<(), Self::Err>;
 
-    async fn add_keyset(&self, keyset: MintKeySet) -> Result<(), Self::Err>;
-    async fn get_keyset(&self, id: &Id) -> Result<Option<MintKeySet>, Self::Err>;
-    async fn get_keysets(&self) -> Result<Vec<MintKeySet>, Self::Err>;
+    async fn add_keyset(&self, keyset: MintKeySetInfo) -> Result<(), Self::Err>;
+    async fn get_keyset(&self, id: &Id) -> Result<Option<MintKeySetInfo>, Self::Err>;
+    async fn get_keysets(&self) -> Result<Vec<MintKeySetInfo>, Self::Err>;
```
thesimplekid, line 107: Should these be renamed to `add_keyset_info` etc or think its okay to leave them?

crates/cdk/src/mint.rs
davidcaseria: Yes, will do. I wasn't sure how invasive this change should be.

crates/cdk/src/cdk_database/mod.rs
davidcaseria: I can add `_info` to make it more explicit.

thesimplekid:
Hey the CI failed because the traits were not the mint and wallet features as they should have been, my fault. I fixed it here if you want to cherry-pick this commit onto your branch https://github.com/thesimplekid/cdk/commit/e709ce7efcd5b26d18971087a13989db1a5d12c5

crates/cdk/src/cdk_database/mod.rs
```diff
@@ -5,12 +5,17 @@ use std::collections::HashMap;
 use async_trait::async_trait;
```
thesimplekid, line 5: ``
```suggestion
#[cfg(any(feature = "wallet", feature = "mint"))]
use std::collections::HashMap;
#[cfg(any(feature = "wallet", feature = "mint"))]
use async_trait::async_trait;
```

crates/cdk/src/cdk_database/mod.rs
```diff
@@ -5,12 +5,17 @@ use std::collections::HashMap;
...
 
-use crate::nuts::{
-    BlindSignature, CurrencyUnit, Id, KeySetInfo, Keys, MintInfo, MintKeySet, Proof, Proofs,
-    PublicKey,
-};
+#[cfg(feature = "mint")]
+use crate::mint::MintKeySetInfo;
+#[cfg(feature = "mint")]
+use crate::nuts::{BlindSignature, CurrencyUnit, Proof, PublicKey};
+use crate::nuts::{Id, MintInfo};
```
thesimplekid, line 12:
```suggestion
#[cfg(any(feature = "wallet", feature = "mint"))]
use crate::nuts::{Id, MintInfo};
```

crates/cdk/src/cdk_database/mod.rs
```diff
@@ -5,12 +5,17 @@ use std::collections::HashMap;
...
+#[cfg(feature = "wallet")]
+use crate::nuts::{KeySetInfo, Keys, Proofs};
+#[cfg(feature = "mint")]
 use crate::secret::Secret;
 use crate::types::{MeltQuote, MintQuote};
```
thesimplekid, line 17:
```suggestion
#[cfg(any(feature = "wallet", feature = "mint"))]
use crate::types::{MeltQuote, MintQuote};
```

thesimplekid (commented):
Sorry a few more. This should be checked by precommit now that I think those scrips are fixed on main.

thesimplekid (approved):
ACK fe103750014bc2ca94fc3af8b51f14be95818d33

@EthnTuttle are you happy to go with this over #96

@davidcaseria no need to worry about the CI flake is updated on main

thesimplekid:
Merged in a5a50f281c8ad0bb429da1eebe137a11f9086d1d

PR #104, opened 2024-04-29
NUT-14

### LINKED ISSUE #101: Implement NUT14: HTLCs
https://github.com/cashubtc/nuts/blob/main/14.md

thesimplekid:
- [x] Info endpoint for nut14 support
closes #101

### DIFF
--- a/crates/cdk/src/nuts/nut00.rs
+++ b/crates/cdk/src/nuts/nut00.rs
@@ -14,11 +14,14 @@ use serde::{Deserialize, Deserializer, Serialize};
 use thiserror::Error;
 use url::Url;
 
+use super::nut10;
+use super::nut11::SpendingConditions;
 use crate::dhke::blind_message;
 use crate::nuts::nut01::{PublicKey, SecretKey};
-use crate::nuts::nut11::{witness_deserialize, witness_serialize, Signatures};
+use crate::nuts::nut11::{serde_p2pk_witness, P2PKWitness};
 use crate::nuts::nut12::BlindSignatureDleq;
-use crate::nuts::{Id, P2PKConditions, ProofDleq};
+use crate::nuts::nut14::{serde_htlc_witness, HTLCWitness};
+use crate::nuts::{Id, ProofDleq};
 use crate::secret::Secret;
 use crate::url::UncheckedUrl;
 use crate::Amount;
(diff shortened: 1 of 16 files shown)


crates/cdk/src/nuts/nut00.rs
```diff
@@ -123,11 +122,40 @@ pub struct BlindSignature {
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum Witness {
+    /// P2PK Witness
+    #[serde(serialize_with = "p2pk_witness_serialize")]
+    #[serde(deserialize_with = "p2pk_witness_deserialize")]
+    P2PKWitness(P2PKWitness),
+    /// HTLC Witness
+    #[serde(serialize_with = "htlc_witness_serialize")]
+    #[serde(deserialize_with = "htlc_witness_deserialize")]
```
davidcaseria, line 139: My personal preference is to have a module that has a `serialize` and `deserialize` function, which then can be used as `#[serde(with = “module_name”)]`. If you prefer this way, that's valid, but I just wanted to share if you also like that shorthand.

crates/cdk/src/nuts/nut14.rs
```diff
@@ -0,0 +1,131 @@
...
+    D: Deserializer<'de>,
+{
+    let s: String = String::deserialize(deserializer)?;
+    serde_json::from_str(&s).map_err(de::Error::custom)
+}
+
+impl Proof {
+    /// Verify HTLC
+    pub fn verify_htlc(&self) -> Result<(), Error> {
+        let secret: Secret = self.secret.clone().try_into().unwrap();
```
lorenzolfm, line 64: should you handle the error instead of panic! with unwrap()?

crates/cdk/src/nuts/nut14.rs
```diff
@@ -0,0 +1,131 @@
...
+        if secret.kind.ne(&super::Kind::HTLC) {
+            return Err(Error::IncorrectSecretKind);
+        }
+
+        let htlc_witness = match &self.witness {
+            Some(Witness::HTLCWitness(witness)) => witness,
+            _ => return Err(Error::IncorrectSecretKind),
+        };
+
+        let hash_lock = Sha256Hash::from_str(&secret.secret_data.data).unwrap();
```
lorenzolfm, line 98: same for unwrap here

PR #126, opened 2024-05-13
Add wallet tracing and fix receive bug

davidcaseria:
This follows up on a Discord conversation to add better tracing to the `Wallet`. Please verify that the proper arguments are skipped for security purposes.

Additionally, it fixes a bug where new mints were not added to the store from a received token, causing balance calculations to be incorrect.

### DIFF
--- a/crates/cdk-redb/src/wallet.rs
+++ b/crates/cdk-redb/src/wallet.rs
@@ -302,7 +302,7 @@ impl WalletDatabase for RedbWalletDatabase {
         Ok(())
     }
 
-    #[instrument(skip(self))]
+    #[instrument(skip_all)]
     async fn add_keys(&self, keys: Keys) -> Result<(), Self::Err> {
         let db = self.db.lock().await;
         let write_txn = db.begin_write().map_err(Error::from)?;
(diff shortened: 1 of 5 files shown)


crates/cdk/src/client.rs
```diff
@@ -74,6 +75,7 @@ impl HttpClient {
     }
 
     /// Get Active Mint Keys [NUT-01]
+    #[instrument(skip(self))]
```
thesimplekid, line 78:
```suggestion
    #[instrument(skip(self), fields(mint_url = %mint_url))]
```

I think we should format the url as display. its a bit verbose displaying it as debug like default. 
image
vs
image

crates/cdk-redb/src/wallet.rs
```diff
@@ -290,6 +302,7 @@ impl WalletDatabase for RedbWalletDatabase {
         Ok(())
     }
 
+    #[instrument(skip(self))]
```
thesimplekid, line 305:
```suggestion
    #[instrument(skip_all)]
```
Logging the full `Keys` map will be too verbose I think

thesimplekid:
Merged in 303204d8585b223b21464720564f23e5cba515ea

PR #127, opened 2024-05-13
NUT13-only Wallet with provided seed

### LINKED ISSUE #111: Use Xpriv in Wallet

### LINKED ISSUE #125: Wallet: Support only nut13 deterministic wallets
Currently both randomly generated secrets and NUT-13 derived secrets are supported. While the spec does not require supporting only NUT-13 doing so would reduce code complexity as the two options don't have to be handled separately. This change can also be made transparent to the lib user as if no secret is passed one can be generated for them, that they can choose to never use. Going this route leads to a more sensible default implementation of deterministic secrets allowing for CDK based wallets to be recovered by default instead of an extra feature as they are now.

davidcaseria:
The Wallet will now only support NUT -13-based secret derivation. The cdk user must provide a seed, which is used to generate an internal Xpriv.

Also, the database counter is changed to a u32 since the maximum value is `2^31-1`.

Closes #111 
Closes #125 
Relies on #126

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,6 @@ jobs:
             -p cdk --no-default-features,
             -p cdk --no-default-features --features wallet,
             -p cdk --no-default-features --features mint,
-            -p cdk --no-default-features --features all-nuts,
             -p cdk-redb
           ]
     steps:
@@ -65,7 +64,6 @@ jobs:
             -p cdk,
             -p cdk --no-default-features,
             -p cdk --no-default-features --features wallet,
-            -p cdk --no-default-features --features all-nuts,
             -p cdk-js
           ]
     steps:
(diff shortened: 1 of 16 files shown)


thesimplekid:
@davidcaseria Mind just rebaseing on main and squashing the commits. I can do it if not just let me know

thesimplekid (commented):
Thanks for the PR looks good. Think just remove the feature and good to go

davidcaseria:
I simplified your suggestion to a one-liner: `let count = count.map_or(0, |c| c + 1);`. I think that's still pretty clear if you know what the `map_or` function does, but let me know if you'd like me to change it to your suggestion.

thesimplekid:
> I simplified your suggestion to a one-liner: let count = count.map_or(0, |c| c + 1);

Even better thanks

thesimplekid (approved):

PR #84, opened 2024-04-12
Decode and encode Token V4

davidcaseria:
Refactor of #73 given the crate changes.

Closes #68.

PR #158, opened 2024-06-06
feat(wallet): token v4

thesimplekid:
supersedes #84


davidcaseria:
Are V4 tokens still going to support multiple mints?

thesimplekid:
> Are V4 tokens still going to support multiple mints?

Not sure, I would prefer they don't. I left a comment on the nuts [PR](https://github.com/cashubtc/nuts/pull/109#discussion_r1629740064) . This is just you're work from the other PR I copied onto main since there are been a lot of changes merging was a mess. Hope that 's okay. Still gonna wait on the spec to merge this, should be an easy enough change if the multi mints are removed.

davidcaseria:
Sounds good. I was holding off because I knew there was an open discussion and didn't want to keep having to make changes.

I also prefer if multi-mints were no longer supported.

crates/cdk/src/nuts/nut00/token.rs
```diff
@@ -0,0 +1,501 @@
...
+            _ => return Err(Error::UnsupportedToken),
+        };
+
+        let decode_config = general_purpose::GeneralPurposeConfig::new()
+            .with_decode_padding_mode(base64::engine::DecodePaddingMode::Indifferent);
+        let decoded = GeneralPurpose::new(&alphabet::URL_SAFE, decode_config).decode(s)?;
+        let token: TokenV4 = ciborium::from_reader(&decoded[..])?;
+        Ok(token)
+    }
+}
```
thesimplekid, line 337: Remove v3

davidcaseria (approved):

PR #181, opened 2024-06-18
feat(NUT05): MeltQuote state

thesimplekid:


crates/cdk-sqlite/src/mint/migrations/20240618195700_quote_state.sql
thesimplekid, line 1: Remove paid column

thesimplekid:
close for #201

PR #199, opened 2024-06-25
feat(wallet): make wallet single mint and unit

thesimplekid:
This changes the `wallet` to be for a single mint and unit, simplifying the core library logic and allowing it to be more flexible for users. Multiple `wallet` can be created by the user to enable multi mint and multi unit.


crates/cdk/src/wallet/mod.rs
```diff
@@ -26,123 +25,59 @@ use crate::nuts::{
...
-    #[cfg(feature = "nostr")]
-    nostr_client: nostr_sdk::Client,
 }
 
 impl Wallet {
     pub fn new(
+        mint_url: &str,
+        unit: CurrencyUnit,
         localstore: Arc<dyn WalletDatabase<Err = cdk_database::Error> + Send + Sync>,
         seed: &[u8],
```
davidcaseria, line 48: Is it safe to re-use the same seed for different mint URLs?

thesimplekid: Yes, wallets shouldn't really care about the mint_url other then how to access the mint. When generating secrets the keyset_id is used so as long as the mints have different keysets different secrets will be generated. Similarly if a mint is exposed on two endpoints the wallet should treat it has the same mint in terms of deriving secrets.

thesimplekid:
> I didn't do a thorough review, but I am generally supportive of this effort to simplify. Before releasing 0.1, I think we should add a multi-mint wallet struct. That is how my wallet works right now and I imagine that will be a more common use case.

I'll add it today

crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -0,0 +1,304 @@
...
+
+use super::Error;
+use crate::amount::SplitTarget;
+use crate::nuts::{CurrencyUnit, SecretKey, SpendingConditions, Token};
+use crate::types::{Melted, MintQuote};
+use crate::{Amount, UncheckedUrl, Wallet};
+
+#[derive(Debug, Clone)]
+pub struct MultiMintWallet {
+    pub wallets: Arc<Mutex<HashMap<WalletKey, Arc<Wallet>>>>,
```
davidcaseria, line 22: I don't think `Wallet` needs to be wrapped in an `Arc` since it is already `Clone`.

davidcaseria (approved):

PR #228, opened 2024-07-09
Provide Integrated LDK Node

davidcaseria:


crates/cdk/src/amount.rs
```diff
@@ -29,6 +29,11 @@ impl Amount {
         self.0 / 1000
     }
 
+    /// To satoshis Amount
```
thesimplekid, line 32: I've thought about adding these a few times but didn't because i was concerted that it could be misleading since we do not guarantee that the amount is something we can actually convert to msat. The 3 options I've thought about are not having the helper functions and leaving conversion up to the user. Which is pretty annoying to we probably should have them. Two adding them like you've done here and leave it up to the user to make sure its only used when it is an appropriate unit. Finally, adding a unit to the amount struct and then having `try_to_msats` that errors if not a convertible unit. However, this gets a bit awkward when serialized/deserialized since the unit in not in the amount of keyset for example. So I think I lean towards this option of having the functions and leaving it up to the user/us in other parts of the lib to not be used on an inappropriate unit. Be interested if you thought about this and have other ideas.

davidcaseria:
Superseded by #904

PR #232, opened 2024-07-10
Use arg to set log level for cli

davidcaseria:
Defaults to error logging.


thesimplekid (approved):

thesimplekid:
Thanks. I also added an env_filter because the sqlx crate logging is really annoying that every query is info.

https://github.com/cashubtc/cdk/blob/22fd5427fbda60a137187d150c6de0d661e4cd0f/crates/cdk-cli/src/main.rs#L72-L80

PR #266, opened 2024-07-28
feat(tests): integration test

thesimplekid:


crates/cdk/src/mint/mod.rs
```diff
@@ -655,11 +655,20 @@ impl Mint {
...
+                .await?;
             return Err(Error::DuplicateProofs);
         }
 
         for proof in &swap_request.inputs {
-            self.verify_proof(proof).await?
+            if let Err(err) = self.verify_proof(proof).await {
+                tracing::info!("Error verifying proof in swap");
+                self.localstore
+                    .update_proofs_states(&input_ys, State::Unspent)
```
vnprc, line 668: If I am understanding correctly this would be a good test to write:
- attempt to spend a p2pk locked token with no signature or incorrect signature
  - should get an error
- attempt to spend it again with the correct signature
  - should get a successful spend

vnprc: I guess I need to improve my terminology. This is my current understanding, please correct any inaccuracies or misunderstandings I might have.

A swap is an exchange of one or more ecash tokens for one or more newly generated tokens, potentially with different spend conditions.

A proof is a signed, unblinded secret, aka an ecash token that has not yet been redeemed.

Is this right?

thesimplekid: Yes, that would be a good test. 

> A swap is an exchange of one or more ecash tokens for one or more newly generated tokens, potentially with different spend conditions.

Yes

> A proof is a signed, unblinded secret, aka an ecash token that has not yet been redeemed.

Yes, the [proof](https://github.com/cashubtc/nuts/blob/main/00.md#proof) is the struct that includes the unblinded signature and the secret.

PR #268, opened 2024-07-30
fix: insufficient funds bug in proof_selection example

vnprc:
For some reason this example was minting a token worth 64 sats and attempting to generate proofs for 65 sats worth of tokens. We can either decrement the proof request to match the amount for the one ecash token or mint 1 more sat to match the proof request amount.

### DIFF
--- a/crates/cdk/examples/proof_selection.rs
+++ b/crates/cdk/examples/proof_selection.rs
@@ -51,7 +51,7 @@ async fn main() {
     let proofs = wallet.get_proofs().await.unwrap();
 
     let selected = wallet
-        .select_proofs_to_send(Amount::from(65), proofs, false)
+        .select_proofs_to_send(Amount::from(64), proofs, false)
         .await
         .unwrap();
 


crates/cdk/examples/proof_selection.rs
vnprc, line 1: Alternative fix:
```diff --git a/crates/cdk/examples/proof_selection.rs b/crates/cdk/examples/proof_selection.rs
index 3c7f7b0..3270722 100644
--- a/crates/cdk/examples/proof_selection.rs
+++ b/crates/cdk/examples/proof_selection.rs
@@ -22,7 +22,7 @@ async fn main() {
 
     let wallet = Wallet::new(mint_url, unit, Arc::new(localstore), &seed, None);
 
-    for amount in [64] {
+    for amount in [64, 1] {
         let amount = Amount::from(amount);
         let quote = wallet.mint_quote(amount).await.unwrap();
```

thesimplekid:
Ah good catch, I was using this code to test the proof selection so had different amount in that list to mint and when i removed them for the example didn't select correctly. I didn't realize the examples weren't getting run in CI (https://github.com/cashubtc/cdk/issues/269).

thesimplekid (approved):

PR #270, opened 2024-08-01
fix: send proofs with conditions

thesimplekid:


crates/cdk/examples/p2pk.rs
vnprc, line 1: Typo at the bottom of the file: "Redeamed" should be "Redeemed".

vnprc (approved):
LGTM!

PR #272, opened 2024-08-01
Add proxy support

davidcaseria:
Allows using mints accessible by [NWS](https://github.com/asmogo/nws/).


thesimplekid (commented):
Can this be used with any proxy or only nostr. I think we should and can support both? Also are there any mints with an nprofile I can test this with?

davidcaseria:
Updated to be a generic proxy. I added a `host_matcher` to take a regex so that I do not have to send all requests to the proxy. I figure this will be useful for either TOR or NWS services.

PR #274, opened 2024-08-02
Nostr NIP-60

davidcaseria:
[NIP-60](https://github.com/nostr-protocol/nips/pull/1369)

This implementation takes a hybrid approach: the wallet info is stored as a kind `37375` event but keeps the proofs stored locally. A CDK application developer can optionally save the `7375` events by providing a `TxInfo` object for transactions. If a wallet should be accessible on another device or application, the restore endpoint for each mint should be called to retrieve any missing proofs.


thesimplekid (commented):
What do you think of making this a separate crate like we do for the sqlite and redb. This is a bit of a smaller implantation then those require so maybe just a feature flag makes sense, but may be best to stick to that pattern?

davidcaseria:
Sounds good. I will work on that this week.

crates/cdk/src/mint_url.rs
```diff
@@ -91,6 +91,12 @@ impl fmt::Display for MintUrl {
     }
 }
 
+impl From<Url> for MintUrl {
+    fn from(url: Url) -> Self {
+        Self(url.to_string())
```
ok300, line 96: IMO it's better to use `MintUrl::from_str` here because it does some extra validation and processing.

davidcaseria:
This is not compatible with the merged version NIP-60, so closing for now.

PR #276, opened 2024-08-06
test: first mint unit tests

vnprc:
First attempt to add some unit tests. I tried to start with a few fundamental operations. I hope to build up to more complex tests. This was mostly a learning exercise but I'm opening a draft PR to get some initial feedback on my approach. Is this useful? Is it redundant with existing tests? What additional assertions would be helpful?

I will probably want to move the new mint code to a helper function for use in other tests.

I think I uncovered an issue with `MintKeySet::generate_from_seed`. It requires a `MintKeysetInfo` struct, which includes an ID that is derived from the keysets but we haven't generated the keysets yet. It might be a good idea to change the function signature to accept `max_order`, `unit`, and `derivation_path` individually instead of the entire keyset info struct.

### DIFF
--- a/crates/cdk/src/mint/mod.rs
+++ b/crates/cdk/src/mint/mod.rs
@@ -98,7 +98,9 @@ impl Mint {
                         let keyset = MintKeySet::generate_from_xpriv(
                             &secp_ctx,
                             xpriv,
-                            highest_index_keyset.clone(),
+                            highest_index_keyset.max_order,
+                            highest_index_keyset.unit,
+                            highest_index_keyset.derivation_path.clone(),
                         );
                         active_keysets.insert(id, keyset);
                         let mut keyset_info = highest_index_keyset;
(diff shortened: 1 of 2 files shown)


thesimplekid:
>  Is this useful? Is it redundant with existing tests? What additional assertions would be helpful?

This is very useful as I would like to have more testing. No this is not redundant. 

> I think I uncovered an issue with MintKeySet::generate_from_seed. It requires a MintKeysetInfo struct, which includes an ID that is derived from the keysets but we haven't generated the keysets yet. It might be a good idea to change the function signature to accept max_order, unit, and derivation_path individually instead of the entire keyset info struct.

The way I currently use this is I use https://github.com/cashubtc/cdk/blob/6adf3267c02e1f4e9e15f8c4be0a548a6ecc5584/crates/cdk/src/nuts/nut02.rs#L289-L294 `generate` when creating a new keyset and then `from_seed` when regenerate from `MintKeyset info` I already have, as that's what is persisted in the database. https://github.com/cashubtc/cdk/blob/6adf3267c02e1f4e9e15f8c4be0a548a6ecc5584/crates/cdk/src/nuts/nut02.rs#L324-L328

However, there is no reason to enforce this and your suggested changes would make the api more flexible so I think its a good idea.

thesimplekid (approved):

Issue #258, opened 2024-07-26
NUT06: Mint Icon Url


thesimplekid:
https://github.com/cashubtc/nuts/pull/130

PR #282, opened 2024-08-10
Added mint_icon_url to mint details.

cjbeery24:


crates/cdk-sqlite/src/wallet/migrations/20240612132920_init.sql
```diff
@@ -8,6 +8,7 @@ CREATE TABLE IF NOT EXISTS mint (
     description_long TEXT,
     contact TEXT,
     nuts TEXT,
+    mint_icon_url TEXT,
```
thesimplekid, line 11:
```suggestion
```

Since we want existing mints to be able to upgrade to the new db scheme we don't want to change the init migration. Instead we need to create a new migration with the sqlx cli from within the [wallet folder](https://github.com/cashubtc/cdk/tree/main/crates/cdk-sqlite/src/wallet).

```sh
sqlx migrate add mint_icon_url
```
This will crate a new blank migration file where we can add the migration sql similar to the [input_fee](https://github.com/cashubtc/cdk/blob/main/crates/cdk-sqlite/src/wallet/migrations/20240710144711_input_fee.sql) one. 

```sql
ALTER TABLE mint ADD mint_icon_url TEXT;
```

We don't need to worry about migrating the redb database as its a new optional field it will simply be `None`. The migrations are a bit weird so let me know if you have any questions or would prefer me to push a commit for it.

thesimplekid (changes requested):
Hey thank you for this just a small not on the SQL migration. Don't worry about the fmt ci failure it doesn't error for me locally and I don't see an issue if it keeps erroring ill investigate it.

thesimplekid:
closes #258

thesimplekid (approved):
ACK 57b110bd0f23fb5bb01132198227d06b3c024388

Issue #267, opened 2024-07-29
wallet: check how mint url is handled


thesimplekid:
https://github.com/cashubtc/nuts/pull/151/files

PR #283, opened 2024-08-10
Implemented new struct 'MintUrl' which trims trailing slashes.

cjbeery24:
I wasn't sure if I should insert the trim functionality into the existing "UncheckedUrl" implementation, so I created a similar but new "MintUrl" which seems more appropriately named anyway.

For:
https://github.com/cashubtc/cdk/issues/267


crates/cdk/src/mint_url.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct MintUrl(String);
+
+impl MintUrl {
+    /// New mint url
+    pub fn new<S>(url: S) -> Self
+    where
+        S: Into<String>,
+    {
+        let url: String = url.into();
+        Self(url.trim_end_matches("/").to_string())
```
thesimplekid, line 32:
```suggestion
        Self(url.trim_end_matches('/').to_string())
```

crates/cdk/src/mint_url.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Join onto url
+    pub fn join(&self, path: &str) -> Result<Url, Error> {
+        let url: Url = self.try_into()?;
+        Ok(url.join(path)?)
+    }
+
+    /// Remove trailing slashes from url
+    pub fn trim_trailing_slashes(&self) -> Self {
+        Self(self.to_string().trim_end_matches("/").to_string())
```
thesimplekid, line 48:
```suggestion
        Self(self.to_string().trim_end_matches('/').to_string())
```

crates/cdk/src/mint_url.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+impl<S> From<S> for MintUrl
+where
+    S: Into<String>,
+{
+    fn from(url: S) -> Self {
+        let url: String = url.into();
+        Self(url.trim_end_matches("/").to_string())
```
thesimplekid, line 58:
```suggestion
        Self(url.trim_end_matches('/').to_string())
```

thesimplekid (commented):
Thanks for this. With this change we can completely remove the `UncheckedUrl` in favor or `MintUrl`. 

Also since we need to account for databases that already have mints with trailing slashes in them since with this change we wont be able to get them since the lookup id will be the url without a `/` I suggest we add a migration that trims the trailing `/` from the db. In sql this can be done by adding a migration file via the cli and using sql `trim`. In redb we can bump the db version number and then write a function to trip urls with trailing `/`.

cjbeery24:
@thesimplekid I added SQL migrations to remove trailing slashes from existing mint_urls. I'm not as familiar with redb, would you be able to make the necessary changes for that?

crates/cdk-sqlite/src/mint/migrations/20240811031111_update_mint_url.sql
```diff
@@ -0,0 +1 @@
+UPDATE `mint_quote` SET `mint_url`=TRIM(TRAILING '/' FROM `mint_url`);
```
thesimplekid, line 1:
```suggestion
UPDATE `mint_quote` SET `mint_url` = RTRIM(`mint_url`, '/');
```

crates/cdk-sqlite/src/wallet/migrations/20240810233905_update_mint_url.sql
```diff
@@ -0,0 +1,4 @@
+UPDATE `mint` SET `mint_url`=TRIM(TRAILING '/' FROM `mint_url`);
+UPDATE `keyset` SET `mint_url`=TRIM(TRAILING '/' FROM `mint_url`);
+UPDATE `mint_quote` SET `mint_url`=TRIM(TRAILING '/' FROM `mint_url`);
+UPDATE `proof` SET `mint_url`=TRIM(TRAILING '/' FROM `mint_url`);
```
thesimplekid, line 4:
```suggestion
-- Delete duplicates from `mint`
DELETE FROM `mint`
WHERE `mint_url` IN (
    SELECT `mint_url`
    FROM (
        SELECT RTRIM(`mint_url`, '/') AS trimmed_url, MIN(rowid) AS keep_id
        FROM `mint`
        GROUP BY trimmed_url
        HAVING COUNT(*) > 1
    )
)
AND rowid NOT IN (
    SELECT MIN(rowid)
    FROM `mint`
    GROUP BY RTRIM(`mint_url`, '/')
);

UPDATE `mint` SET `mint_url` = RTRIM(`mint_url`, '/');
UPDATE `keyset` SET `mint_url` = RTRIM(`mint_url`, '/');
UPDATE `mint_quote` SET `mint_url` = RTRIM(`mint_url`, '/');
UPDATE `proof` SET `mint_url` = RTRIM(`mint_url`, '/');
```

thesimplekid (changes requested):
Looks like since its sqlite we need to use `RTRIM` and since there is a unique constraint on the `mint_url` we need to delete the duplicate. 

Yes no problem I will do the redb.

crates/cdk/src/mint_url.rs
```diff
@@ -0,0 +1,112 @@
...
+        let url: Url = self.try_into()?;
+        Ok(url.join(path)?)
+    }
+
+    /// Remove trailing slashes from url
+    pub fn trim_trailing_slashes(&self) -> Self {
+        Self(self.to_string().trim_end_matches('/').to_string())
+    }
+}
+
```
thesimplekid, line 51:
```suggestion
impl<'de> Deserialize<'de> for MintUrl {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let s = String::deserialize(deserializer)?;
        MintUrl::from_str(&s).map_err(serde::de::Error::custom)
    }
}
```

I think we should add a custom deserlizer so that if we get a token that does not trim the slash we trim it.

thesimplekid:
redb migration 9992356aa1071aac05fe78681417bee015812ae7

thesimplekid (approved):
I'm happy with this if you are. If you could just rebase and squash the commits

PR #285, opened 2024-08-12
Add Transaction ID for a set of proofs

davidcaseria:


davidcaseria:
This was included in #686

PR #294, opened 2024-08-16
fix: typos

prusnak:
* there were a copy-paste errors in README.md files of cdk-rexie and cdk-sqlite which claimed these are a redb storage backends
* Sqlite -> SQLite


thesimplekid (approved):
Thank you

PR #302, opened 2024-08-21
Modify WalletDatabase trait to better support db transactions

davidcaseria:
The SQLite crate doesn't currently use transactions but can be modified to do so in the future.

### DIFF
--- a/crates/cdk/src/cdk_database/mod.rs
+++ b/crates/cdk/src/cdk_database/mod.rs
@@ -134,8 +134,8 @@ pub trait WalletDatabase: Debug {
     async fn set_pending_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
     /// Reserve proofs in storage. Proofs are identified by their Y value.
     async fn reserve_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
-    /// Unreserve proofs in storage. Proofs are identified by their Y value.
-    async fn unreserve_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
+    /// Set proofs as unspent in storage. Proofs are identified by their Y value.
+    async fn set_unspent_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
     /// Get proofs from storage
     async fn get_proofs(
         &self,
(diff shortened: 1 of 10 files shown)


crates/cdk/src/cdk_database/mod.rs
```diff
@@ -124,8 +124,18 @@ pub trait WalletDatabase: Debug {
...
+        added: Vec<ProofInfo>,
+        removed_ys: Vec<PublicKey>,
+    ) -> Result<(), Self::Err>;
+    /// Set proofs as pending in storage. Proofs are identified by their Y value.
+    async fn set_pending_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
+    /// Reserve proofs in storage. Proofs are identified by their Y value.
+    async fn reserve_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
+    /// Unreserve proofs in storage. Proofs are identified by their Y value.
+    async fn unreserve_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
     /// Get proofs from storage
```
thesimplekid, line 139: Proofs can got from state `UNSPENT`` -> `PENDING` as well as the reverse the same is true for the `RESERVED` state `UNSPENT` -> `RESERVED` and back. The `unreserve_proof` fn here can bring a proof from either `RESERVED` or `PENDING` to `UNSPENT` so i suggest we just change the name to reflect that.

thesimplekid:
> The SQLite crate doesn't currently use transactions but can be modified to do so in the future.

Yes, I also want to improve these to use the `query!` macro to support compile time type checking. But lets make that in a separate specific pr and keep this one small.

crates/cdk/src/cdk_database/mod.rs
davidcaseria: Do you have a suggested name? `set_unspent_proofs`?

thesimplekid (changes requested):
Rexie needs to update to use the new name 4856142f35eca265b212381c0000b19c0f505147

Since the proofs from a token are unknown when receiving a token they are unknown to the wallet database when we attempt to update the state in `create_swap` to solve this we need to add the proof to the database before this. I think the pending state makes the most sense for this 8a8293e.

davidcaseria:
:man_facepalming: I always forget about rexie.

thesimplekid:
> 🤦‍♂️ I always forget about rexie.

Yeah since it only compiles to wasm. If you use `just final-check` it will catch it.

thesimplekid (approved):

thesimplekid:
Merged in d0d7281c77dbc696e8830775b3015dcea473326a

PR #309, opened 2024-08-30
Get blinded secrets

vnprc:
Get blinded secrets in a dedicated function. This enables developers to provide their own network transport to the cashu mint.

Not sure why it wanted to reformat the entire codebase but I split that into a separate commit to make reviewing this PR easier.

### DIFF
--- a/bindings/cdk-js/src/nuts/nut00/currency_unit.rs
+++ b/bindings/cdk-js/src/nuts/nut00/currency_unit.rs
@@ -9,6 +9,7 @@ pub enum JsCurrencyUnit {
     Msat,
     Usd,
     Eur,
+    Hash,
 }
 
 impl From<CurrencyUnit> for JsCurrencyUnit {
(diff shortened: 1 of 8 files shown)


thesimplekid:
> Not sure why it wanted to reformat the entire codebase but I split that into a separate commit to make reviewing this PR easier.

Yeah that happened to me as well. But these changes should be on `main` in https://github.com/cashubtc/cdk/commit/847eab8e077dc6609993adac0810a6ad3896f605 can you try resetting to [e9742eb](https://github.com/cashubtc/cdk/pull/309/commits/e9742eb51ada49f16705288aee7d9c8f720d8dce) then rebasing onto main

thesimplekid:
@vnprc Cdk now supports a custom currency unit so the hash variant in the enum isn't needed. Looks like there is something else in this but since its been stale for awhile I'm going to close this. Feel free to reopen or open a new updated pr with changes you would like.

vnprc:
Yep. I was planning on it.

PR #311, opened 2024-09-02
Format mint URL (lowercase hostname)

callebtc:
- Convert hostname to lowercase, keep path in original case
- Remove `new` (not used? Use `from` instead)

### DIFF
--- a/crates/cdk/src/mint_url.rs
+++ b/crates/cdk/src/mint_url.rs
@@ -24,6 +24,9 @@ pub struct MintUrl(String);
 
 impl MintUrl {
     fn format_url(url: &str) -> String {
+        if url.len() == 0 {
+            return String::new();
+        }
         let url = url.trim_end_matches('/');
         // https://URL.com/path/TO/resource -> https://url.com/path/TO/resource
         let protocol = url.split("://").nth(0).unwrap().to_lowercase();
--- a/crates/cdk/src/mint_url.rs
+++ b/crates/cdk/src/mint_url.rs
@@ -23,13 +23,31 @@ pub enum Error {
 pub struct MintUrl(String);
 
 impl MintUrl {
-    /// New mint url
-    pub fn new<S>(url: S) -> Self
-    where
-        S: Into<String>,
-    {
-        let url: String = url.into();
-        Self(url.trim_end_matches('/').to_string())
+    fn format_url(url: &str) -> String {
+        let url = url.trim_end_matches('/');
+        // https://URL.com/path/TO/resource -> https://url.com/path/TO/resource
+        let protocol = url.split("://").nth(0).unwrap().to_lowercase();
+        let host = url
+            .split("://")
+            .nth(1)
+            .unwrap()
+            .split('/')
+            .nth(0)
+            .unwrap()
+            .to_lowercase();
+        let path = url
+            .split("://")
+            .nth(1)
+            .unwrap()
+            .split('/')
+            .skip(1)
+            .collect::<Vec<&str>>()
+            .join("/");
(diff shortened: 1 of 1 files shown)


crates/cdk/src/mint_url.rs
```diff
@@ -42,11 +63,6 @@ impl MintUrl {
         let url: Url = self.try_into()?;
         Ok(url.join(path)?)
     }
-
-    /// Remove trailing slashes from url
-    pub fn trim_trailing_slashes(&self) -> Self {
-        Self(self.to_string().trim_end_matches('/').to_string())
-    }
```
callebtc, line 49: Old unused code?

crates/cdk/src/mint_url.rs
```diff
@@ -23,13 +23,34 @@ pub enum Error {
 pub struct MintUrl(String);
 
 impl MintUrl {
-    /// New mint url
-    pub fn new<S>(url: S) -> Self
-    where
-        S: Into<String>,
-    {
-        let url: String = url.into();
-        Self(url.trim_end_matches('/').to_string())
```
callebtc, line 32: Unused code?

crates/cdk/src/mint_url.rs
thesimplekid: Yeah we don't need this since we strip the `/` when creating the `MintUrl`

crates/cdk/src/mint_url.rs
thesimplekid: Yes, redundant due to `from`

crates/cdk/src/mint_url.rs
```diff
@@ -16,20 +16,48 @@ pub enum Error {
...
+    fn format_url(url: &str) -> Result<String, Error> {
+        if url.is_empty() {
+            return Ok(String::new());
+        }
```
thesimplekid, line 32: Should this return an error instead of an empty string?

crates/cdk/src/mint_url.rs
```diff
@@ -65,7 +88,11 @@ where
...
     fn from(url: S) -> Self {
         let url: String = url.into();
-        Self(url.trim_end_matches('/').to_string())
+        let formatted_url = Self::format_url(&url);
+        match formatted_url {
+            Ok(url) => Self(url),
+            Err(_) => Self::default(),
+        }
     }
 }
```
thesimplekid, line 97: I think we should just remove this `From` and move this logic into the `FromStr` below having this `from` doesn't add much as if a user has a type `Into<String>` they can just call `MintUrl::from_str(&their_type.to_string())?` this avoids hiding the error case here by returning an empty string.

crates/cdk/src/mint_url.rs
callebtc: The test in `mod_rotate_keyset` passes an empty string, so this part just reproduces the old code. What do you think?

thesimplekid: I think it should return an error. We can change the test MintConfg to have a valid url with: 
```rust

        let config = MintConfig::<'_> {
            mint_url: "http://example.com",
            ..Default::default()
        };
```

https://github.com/cashubtc/cdk/blob/1cd80deaca28a63094931020bd90a26d9300998d/crates/cdk/src/mint/mod.rs#L1572

like we do in the test above it:

https://github.com/cashubtc/cdk/blob/1cd80deaca28a63094931020bd90a26d9300998d/crates/cdk/src/mint/mod.rs#L1532-L1536

thesimplekid (approved):

PR #313, opened 2024-09-02
feat: implement nut-06 time

prusnak:
implements https://github.com/cashubtc/nuts/issues/138

### DIFF
--- a/crates/cdk-sqlite/src/wallet/mod.rs
+++ b/crates/cdk-sqlite/src/wallet/mod.rs
@@ -92,6 +92,7 @@ impl WalletDatabase for WalletSqliteDatabase {
             nuts,
             mint_icon_url,
             motd,
+            time,
         ) = match mint_info {
             Some(mint_info) => {
                 let MintInfo {
@@ -104,6 +105,7 @@ impl WalletDatabase for WalletSqliteDatabase {
                     nuts,
                     mint_icon_url,
                     motd,
+                    time,
                 } = mint_info;
 
                 (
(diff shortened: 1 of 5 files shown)


prusnak:
> not sure if this is the most idiomatic way how to implement this, advice welcome

I tried different approach in https://github.com/cashubtc/cdk/pull/313

crates/cdk-sqlite/src/wallet/mod.rs
```diff
@@ -104,6 +104,7 @@ impl WalletDatabase for WalletSqliteDatabase {
                     nuts,
                     mint_icon_url,
                     motd,
+                    time: _,
```
thesimplekid, line 107: I think we should save the time here as it serves as an indication of when the `MintInfo` was last fetched.

crates/cdk-sqlite/src/wallet/mod.rs
```diff
@@ -781,6 +783,7 @@ fn sqlite_row_to_mint_info(row: &SqliteRow) -> Result<MintInfo, Error> {
             .unwrap_or_default(),
         mint_icon_url,
         motd,
+        time: Some(unix_time()),
```
thesimplekid, line 786: Adding it here won't tell us what the unix time the Mint last told us it was it will tell us what the unix time currently is for the wallet.

prusnak: addressed in 347b8a19caf955f87174e3b389e3069794875078

crates/cdk-sqlite/src/wallet/mod.rs
prusnak: addressed in 347b8a19caf955f87174e3b389e3069794875078

thesimplekid (approved):
ACK 347b8a19caf955f87174e3b389e3069794875078

Great thank you.

PR #314, opened 2024-09-02
feat: implement nut-06 time (alternative 2)

prusnak:
alternative approach to https://github.com/cashubtc/cdk/pull/313

not sure if this is the most idiomatic way how to implement this, advice welcome

### DIFF
--- a/crates/cdk/src/nuts/nut06.rs
+++ b/crates/cdk/src/nuts/nut06.rs
@@ -2,6 +2,8 @@
 //!
 //! <https://github.com/cashubtc/nuts/blob/main/06.md>
 
+use std::time::{SystemTime, UNIX_EPOCH};
+
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use super::nut01::PublicKey;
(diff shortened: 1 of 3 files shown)


crates/cdk/src/nuts/nut06.rs
```diff
@@ -79,6 +81,20 @@ pub struct MintInfo {
...
+
+fn serialize_time<S>(_: &Option<u64>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let epoch = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_secs();
+    serializer.serialize_u64(epoch)
```
thesimplekid, line 97: I don't think we should use a serialization function to change or add data only to serialize existing data. So I actually prefer your first approach.

crates/cdk/src/nuts/nut06.rs
```diff
@@ -170,6 +186,16 @@ impl MintInfo {
...
+
+    /// Get time
+    pub fn time(&self) -> Option<u64> {
+        Some(
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_secs(),
+        )
+    }
```
thesimplekid, line 198: We have this function in the util module we should be able to import it and reuse it here


https://github.com/cashubtc/cdk/blob/1cd80deaca28a63094931020bd90a26d9300998d/crates/cdk/src/util/mod.rs#L25

prusnak:
closing in favor of https://github.com/cashubtc/cdk/pull/313

Issue #322, opened 2024-09-04
Check these are all same len


thesimplekid:
https://github.com/cashubtc/cdk/blob/eb0ba7da43802d6de2206c749e613518dc918934/crates/cdk/src/dhke.rs#L100-L102

PR #326, opened 2024-09-05
return error in construct_proofs if arguments are not the same length

prusnak:
fixes https://github.com/cashubtc/cdk/issues/322


thesimplekid (approved):
Thank you

PR #337, opened 2024-09-09
NUT-04 Mint Quote Description

lollerfirst:


thesimplekid (commented):
Thanks left a few comments.

crates/cdk/src/nuts/nut04.rs
```diff
@@ -214,6 +214,8 @@ pub struct MintMethodSettings {
     /// Max Amount
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_amount: Option<Amount>,
+    /// Quote Description
+    pub quote_description: bool,
```
prusnak, line 218: Shouldn't this field be just called `description`?

The example here just shows `"description": true` - https://github.com/cashubtc/nuts/blob/main/04.md#settings

crates/cdk/src/nuts/nut04.rs
```diff
@@ -254,6 +256,7 @@ impl Default for Settings {
             unit: CurrencyUnit::Sat,
             min_amount: Some(Amount::from(1)),
             max_amount: Some(Amount::from(1000000)),
+            quote_description: true,
```
prusnak, line 259: Maybe we could return here the following depending on whether the backend supports descriptions?

``` rust
quote_description: ln.get_settings().invoice_description
```

Of course, we would need to get the value here, the code above is just a pseudo-code.

thesimplekid: We won't be able to do that here in `default()` so I think the default should be `false` and then [here](https://github.com/cashubtc/cdk/blob/0367e0cd4ee043a2d96f8db82de94c36737fb794/crates/cdk-mintd/src/main.rs#L153-L154) when setting `MintMethofSettings` for each backend in `mintd` we shouldn't use the default but actually set it based on what the backend supports.

crates/cdk-integration-tests/src/lib.rs
```diff
@@ -185,7 +185,7 @@ pub async fn mint_proofs(
     let wallet_client = HttpClient::new();
 
     let mint_quote = wallet_client
-        .post_mint_quote(mint_url.parse()?, 1.into(), CurrencyUnit::Sat)
+        .post_mint_quote(mint_url.parse()?, 1.into(), CurrencyUnit::Sat, None)
```
prusnak, line 188: should we maybe try to pass description here if it is supported by the mint?

crates/cdk/src/wallet/mod.rs
```diff
@@ -510,7 +510,7 @@ impl Wallet {
         let unit = self.unit;
         let quote_res = self
             .client
-            .post_mint_quote(mint_url.clone().try_into()?, amount, unit)
+            .post_mint_quote(mint_url.clone().try_into()?, amount, unit, None)
```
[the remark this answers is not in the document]
lollerfirst: Do we let the wallet fail the mint request if the user specifies a description and the backend does not support it or do we check before by calling the info endpoint?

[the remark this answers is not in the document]
thesimplekid: The wallet should check that its supported. We should already to be able to get this from the wallet database by calling [`get_mint`](https://github.com/cashubtc/cdk/blob/5b954d25fde1d6e6842c7b25b72819c0ea88ef84/crates/cdk-sqlite/src/wallet/mod.rs#L169) that will return the [`MIntInfo`](https://github.com/cashubtc/cdk/blob/0367e0cd4ee043a2d96f8db82de94c36737fb794/crates/cdk/src/nuts/nut06.rs#L55) then on that nuts `mint_info.nuts.nut04.description`

crates/cdk/src/nuts/nut04.rs
lollerfirst: Shouldn't the default be `true` as this is the case for the vast majority of backends?

thesimplekid: I think false since even if a backend can support it the mint operator may not want to and it should be explicitly enabled.

thesimplekid:
Sorry about the conflicts. If you rebase onto main it shouldn't be too bad all the ones with conflicts should just be deleted except `crates/cdk-integration-tests/tests/mint.rs` and I think that should just be as is on `main`

crates/cdk/src/wallet/mod.rs
```diff
@@ -500,17 +497,34 @@ impl Wallet {
...
+        // If we have a description, we check that the mint supports it.
+        if description.is_some() {
+            let mint_method_settings = self.localstore
+                .get_mint(mint_url.clone())
+                .await?
+                .expect("Could not retrieve mint information")
+                .nuts
+                .nut04
+                .get_settings(&unit, &PaymentMethod::Bolt11)
+                .expect("Could not find mint unit and method");
```
thesimplekid, line 518: Can we remove the expect here and return an error `Error::UnitUnsupported` expect will panic like unwrap does so it should be avoided, unless it checked in someway so we know it won't panic.

crates/cdk-integration-tests/tests/mint.rs
thesimplekid, line 1: I looks like this is the old mint.rs file. It should be the one currently on main

lollerfirst: good now?

crates/cdk-integration-tests/tests/mint.rs
```diff
@@ -367,4 +367,4 @@ async fn test_mint_enforce_fee() -> Result<()> {
     let _ = mint.process_swap_request(swap_request).await?;
 
     Ok(())
-}
+}
```
prusnak, line 370: missing newline at end of file

crates/cdk/src/cdk_lightning/mod.rs
```diff
@@ -121,16 +121,18 @@ pub struct PaymentQuoteResponse {
 }
 
 /// Ln backend settings
-#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Serialize, Deserialize)]
```
prusnak, line 124: why drop `Copy`?

thesimplekid: Mint and Melt settings have `PaymentMethod` which one variant is a custom string so it can't be `copy` but I think we can remove the custom or make it a static str and should be able to add copy back. Could make that its own pr or add it here I don't mind.

thesimplekid, line 124: > @prusnak I think i accidental delete your comment about removing copy. My response was showing up twice and i tried to delete the duplicate and deleted your comment as well I guess. 

Mint and Melt settings have `PaymentMethod` has one variant that is a custom string so it can't be copy but I think we can remove the custom or make it a static str and should be able to add copy back. Could make that its own pr or add it here I don't mind.

prusnak: No, it was me who deleted the comment. I figured out why the copy was removed, so I deleted it. I had no idea, you'd react so quickly :-)

What you write wrt `PaymentMethod` makes sense.

crates/cdk-cli/src/sub_commands/mint.rs
```diff
@@ -22,6 +22,9 @@ pub struct MintSubCommand {
     /// Currency unit e.g. sat
     #[arg(default_value = "sat")]
     unit: String,
+    /// Quote description
+    #[arg(default_value = None)]
```
thesimplekid, line 26: T
```suggestion
 #[serde(skip_serializing_if = "Option::is_none")]
```

Default None is implicit we can replace it with skip if none.

crates/cdk/src/wallet/mod.rs
lollerfirst: I tried using the `match` construct like you showed in the code style guide, but it wouldn't let me. So i used `if let`.

crates/cdk/src/wallet/mod.rs
```diff
@@ -500,17 +497,35 @@ impl Wallet {
...
+                .nuts
+                .nut04
+                .get_settings(&unit, &PaymentMethod::Bolt11) {
+                if !mint_method_settings.description {
+                    return Err(Error::InvoiceDescriptionUnsupported);
+                }
+            } else {
+                return Err(Error::UnitUnsupported);
+            }
+        }
```
thesimplekid, line 524:
```suggestion
        // If we have a description, we check that the mint supports it.
        if description.is_some() {
            let mint_method_settings = self
                .localstore
                .get_mint(mint_url.clone())
                .await?
                .ok_or(Error::IncorrectMint)?
                .nuts
                .nut04
                .get_settings(&unit, &PaymentMethod::Bolt11)
                .ok_or(Error::UnsupportedUnit)?;

            if !mint_method_settings.description {
                return Err(Error::InvoiceDescriptionUnsupported);
            }
        }

```

We can do it without the match in this case since the error should just be bubbled up.

lollerfirst: Oh wow ok 😅

prusnak:
these two changes need to be applied to fix failures in the CI:

``` patch
diff --git a/bindings/cdk-js/src/wallet.rs b/bindings/cdk-js/src/wallet.rs
index 3c91f71..5b65556 100644
--- a/bindings/cdk-js/src/wallet.rs
+++ b/bindings/cdk-js/src/wallet.rs
@@ -89,7 +89,7 @@ impl JsWallet {
     pub async fn mint_quote(&mut self, amount: u64) -> Result<JsMintQuote> {
         let quote = self
             .inner
-            .mint_quote(amount.into())
+            .mint_quote(amount.into(), None)
             .await
             .map_err(into_err)?;
 
diff --git a/crates/cdk-cli/src/sub_commands/mint.rs b/crates/cdk-cli/src/sub_commands/mint.rs
index ac87ec1..46ce6a2 100644
--- a/crates/cdk-cli/src/sub_commands/mint.rs
+++ b/crates/cdk-cli/src/sub_commands/mint.rs
@@ -11,9 +11,10 @@ use cdk::wallet::multi_mint_wallet::WalletKey;
 use cdk::wallet::{MultiMintWallet, Wallet};
 use cdk::Amount;
 use clap::Args;
+use serde::{Deserialize, Serialize};
 use tokio::time::sleep;
 
-#[derive(Args)]
+#[derive(Args, Serialize, Deserialize)]
 pub struct MintSubCommand {
     /// Mint url
     mint_url: MintUrl,
```

lollerfirst:
Don't we want the JS binding also use the description?

prusnak (approved):

prusnak:
I think we can also remove (now unused) `MintMeltSettings` from `crates/cdk/src/cdk_lightning/mod.rs`

thesimplekid:
> I think we can also remove (now unused) MintMeltSettings from crates/cdk/src/cdk_lightning/mod.rs

Yes, don't believe there is any need for it now.

prusnak (approved):
LGTM f4a0652

thesimplekid (approved):
ACK f4a0652d240ea3f061b11c1b5c6aa9c7ab194e55

LGTM. Thanks for this

crates/cdk/src/cdk_lightning/mod.rs
thesimplekid: https://github.com/cashubtc/cdk/pull/347

PR #340, opened 2024-09-10
Return amount and fee_paid for Melted response

davidcaseria:
This allows a `Wallet` to determine the total amount paid during a melt.


thesimplekid (approved):
LGTM

PR #347, opened 2024-09-14
Payment method

thesimplekid:


crates/cdk/src/nuts/nut05.rs
```diff
@@ -242,7 +242,7 @@ impl From<MeltQuoteBolt11Response> for MeltBolt11Response {
 }
 
 /// Melt Method Settings
-#[derive(Debug, Default, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
```
prusnak, line 245: we should add `Copy` to `MintMethodSettings` in `crates/cdk/src/nuts/nut04.rs` and then do

``` patch
-        mint_settings: self.mint_settings.clone(),
+        mint_settings: self.mint_settings,
```

everywhere

thesimplekid: good catch, I was thinking `MintMethodSettings` has a string so couldn't be copy but thats not the case.

prusnak (approved):

Issue #351, opened 2024-09-15
Check if we can update sqlx to v0.8.0


thesimplekid:
I used v0.6 of sqlx because there were some circular deps that I suspected were from the older version of rust-bitcoin now that thats been updated we may be able to update sqlx. Ideally keeping msrv where it is but sqlx does keep a strict msrv so it may or may not be possible.

prusnak:
my attempt in https://github.com/cashubtc/cdk/pull/352

PR #352, opened 2024-09-16
bump sqlx to 0.8.2

prusnak:
fixes https://github.com/cashubtc/cdk/issues/351

I have not tested this, just tried that the code compiles - hence the draft status


thesimplekid:
Thanks for getting it this far I'll take a look.

thesimplekid:
Looks like if we bump the db_msrv to `1.74.0` it compiles. sqlx-core introduced some syntax that relies on that so I don't think we can avoid bumping it unless we stay on the old version of sqlx and i don't think there is a need for that. Its unfortunate that sqlx doesn't stick enforce as MSRV even if its a newer one. But seems to be the same for https://github.com/rusqlite/rusqlite/issues/566

thesimplekid:
close for 354

PR #361, opened 2024-09-22
Cache `SwapResponse`, `MeltBolt11Response` and `MintBolt11Response` 

lollerfirst:
Cache responses containing critical information, so that clients might be able to replay the same requests.
Caching stores the responses, using the requests as key.

Modified:
 * Added `moka`, `serde_json`, `paste` as dependencies;
 * Added `cache` as  a private field to the `MintState` (also moved the struct up in the file from its og position)
 * Cache is built with a standard hasher and fixed parameters for capacity, ttl and time to idle (probably best to parametrize those from config, not yet sure how to do that)
 * `post_cache_wrapper` is a macro that creates async wrappers that do exactly what you might expect (check requests with the cache, cache positive responses). **I would have preferred to use async closures but apparently Rust has a grudge against them and they cannot be used(?)**. If anybody knows a better way to do this you're welcome to change it.

Tests:
 * There is a integration test for the mint operation.


crates/cdk-axum/Cargo.toml
```diff
@@ -16,3 +16,7 @@ cdk = { path = "../cdk", version = "0.4.0", default-features = false, features =
 tokio = { version = "1", default-features = false }
 tracing = { version = "0.1", default-features = false, features = ["attributes", "log"] }
 futures = { version = "0.3.28", default-features = false }
+moka = { version = "0.11.1", features = ["future"] }
+serde_json = "1.0.128"
+paste = "1.0.15"
+triomphe = "0.1.12"
```
thesimplekid, line 22:
```suggestion
```

We don't need this here we can just pin it in the nix file. I think it will case the msrv build to fail since we pin to a version lower then the one here.

crates/cdk-axum/src/lib.rs
```diff
@@ -14,10 +14,22 @@ use cdk::cdk_lightning::{self, MintLightning};
...
 mod router_handlers;
 
+/// CDK Mint State
+#[derive(Clone)]
+pub struct MintState {
+    ln: HashMap<LnKey, Arc<dyn MintLightning<Err = cdk_lightning::Error> + Send + Sync>>,
+    mint: Arc<Mint>,
+    mint_url: MintUrl,
+    quote_ttl: u64,
+    cache: Cache<String, String, std::hash::RandomState>,
```
thesimplekid, line 30:
```suggestion
    cache: Cache<String, String>,
```

crates/cdk-axum/src/lib.rs
```diff
@@ -30,25 +42,30 @@ pub async fn create_mint_router(
         mint,
         mint_url: MintUrl::from_str(mint_url)?,
         quote_ttl,
+        cache: Cache::builder()
+            .max_capacity(10_000)
+            .time_to_live(Duration::from_secs(60 * 60 * 24))
+            .time_to_idle(Duration::from_secs(60 * 60 * 24))
+            .build_with_hasher(std::hash::RandomState::default()),
```
thesimplekid, line 49:
```suggestion
            .build(),
```

thesimplekid (commented):
Locally got it to build for MSRV and Stable with these changes

crates/cdk-axum/src/router_handlers.rs
```diff
@@ -11,9 +11,38 @@ use cdk::nuts::{
...
+                let Json(json_extracted_payload) = payload.clone();
+                let State(mint_state) = state.clone();
+                let cache_key = serde_json::to_string(&json_extracted_payload).unwrap();
+                if let Some(cached_response) = mint_state.cache.get(&cache_key) {
+                    return Ok(Json(serde_json::from_str(&cached_response).unwrap()));
+                }
+
+                let Json(response) = $handler(state, payload).await?;
+                mint_state.cache.insert(cache_key, serde_json::to_string(&response).unwrap()).await;
+                Ok(Json(response))
```
thesimplekid, line 36: Can we remove the use of `unwrap` and return an error response. The use of `unwrap` should be avoided since it will panic at runtime. In cases where we are sure that it won't panic we can use `expect("Reason why were know it wont panic")`. But I think here since we already are returning a response we can return a error response,

crates/cdk-axum/Cargo.toml
```diff
@@ -16,3 +16,6 @@ cdk = { path = "../cdk", version = "0.4.0", default-features = false, features =
 tokio = { version = "1", default-features = false }
 tracing = { version = "0.1", default-features = false, features = ["attributes", "log"] }
 futures = { version = "0.3.28", default-features = false }
+moka = { version = "0.11.1", features = ["future"] }
+serde_json = "1.0.128"
```
thesimplekid, line 20:
```suggestion
serde_json = "1"
```

crates/cdk-axum/src/router_handlers.rs
```diff
@@ -11,9 +11,38 @@ use cdk::nuts::{
...
+        paste! {
+            /// Cache wrapper function for $handler:
+            /// Wrap $handler into a function that caches responses using the request as key
+            pub async fn [<cache_ $handler>](
+                state: State<MintState>,
+                payload: Json<$request_type>
+            ) -> Result<Json<$response_type>, Response> {
+                let Json(json_extracted_payload) = payload.clone();
+                let State(mint_state) = state.clone();
+                let cache_key = serde_json::to_string(&json_extracted_payload).unwrap();
```
thesimplekid, line 29:
```suggestion
               let cache_key = serde_json::to_string(&json_extracted_payload).map_err(|err| {
                    into_response(Error::from(err))
                })?;

```

thesimplekid (approved):

PR #368, opened 2024-09-26
refactor: ln backends within mint

thesimplekid:


davidcaseria (approved):
I like this a lot better. Good work. :+1:

PR #375, opened 2024-09-30
Remove received proofs from db after swap

davidcaseria:
Fixes an issue where the received proofs remain in the wallet after being swapped.

### DIFF
--- a/crates/cdk/src/wallet/receive.rs
+++ b/crates/cdk/src/wallet/receive.rs
@@ -117,7 +117,9 @@ impl Wallet {
             .into_iter()
             .map(|p| ProofInfo::new(p, self.mint_url.clone(), State::Pending, self.unit))
             .collect::<Result<Vec<ProofInfo>, _>>()?;
-        self.localstore.update_proofs(proofs_info, vec![]).await?;
+        self.localstore
+            .update_proofs(proofs_info.clone(), vec![])
+            .await?;
 
         let mut pre_swap = self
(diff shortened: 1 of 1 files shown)


davidcaseria:
@thesimplekid for this itest failure, it seems like a flakey test: the order is wrong for the equality check. I can fix it if you'd like, but is there a specific order expected, or should I sort the change blind signatures by the amount so that any returned order will pass?

thesimplekid:
Its not flaky on my machine :wink:. I think just sort it for the test. https://github.com/cashubtc/cdk/pull/376 other then that I don't think we care.

crates/cdk/src/wallet/receive.rs
```diff
@@ -152,13 +154,21 @@ impl Wallet {
         mint_proofs.extend(p);
 
         let mut total_amount = Amount::ZERO;
-        for (mint, proofs) in received_proofs {
-            total_amount += Amount::try_sum(proofs.iter().map(|p| p.amount))?;
-            let proofs = proofs
+        for (mint, recv_proofs) in received_proofs {
```
thesimplekid, line 157: Unrelated to this actual PR so can do as a separate one., but i believe this hashmap and loop are unnecessary now that we only handle single mint tokens here. Opened an issue for it here https://github.com/cashubtc/cdk/issues/377

thesimplekid (approved):
LGTM

PR #381, opened 2024-10-01
Refactor to_unit into amount module

davidcaseria:
In a couple of my codebases, I've found it helpful to expose the `to_unit` outside of a mint setting. Since `cdk_lightning` is behind the `mint` feature, I think moving it into the `amount` module would be beneficial so that the `mint` feature does not need to be enabled.


thesimplekid (approved):
Nice

PR #382, opened 2024-10-01
chore: bump bitcoin dependency to 0.32.3

prusnak:


thesimplekid:
We should keep the minimum Minor version that we support. Cargo with automatically pull in the latest patch version even if it is not in our Cargo.toml. Upgrading the patch version will enforce us and our deps to have an higher version then we actually need.

prusnak:
Ah. I did not know this. Thanks for the explanation. Closing.

PR #387, opened 2024-10-03
Provide sorting method for selecting proofs to swap

davidcaseria:
Following up on a conversation with @thesimplekid about proof selection.

The `Wallet` now has a single `select_proofs` that uses a `SelectProofsOptions` struct to control which proofs are selected.

One notable change is the method of selecting **active** proofs. The four methods are now exposed:
- `Largest`: the highest value first
- `Closest`: the closest in value first
- `Smallest`: the smallest in value first
- `Fewest`: the fewest number of proofs over the amount

### DIFF
--- a/crates/cdk/src/wallet/proofs.rs
+++ b/crates/cdk/src/wallet/proofs.rs
@@ -167,113 +167,47 @@ impl Wallet {
         Ok(balance)
     }
 
-    /// Select proofs to send
+    /// Select proofs to send or swap for a specific amount.
     #[instrument(skip_all)]
-    pub async fn select_proofs_to_send(
+    pub async fn select_proofs(
         &self,
         amount: Amount,
         proofs: Proofs,
-        include_fees: bool,
+        opts: SelectProofsOptions,
     ) -> Result<Proofs, Error> {
-        // TODO: Check all proofs are same unit
-
-        if Amount::try_sum(proofs.iter().map(|p| p.amount))? < amount {
-            return Err(Error::InsufficientFunds);
-        }
-
-        let (mut proofs_larger, mut proofs_smaller): (Proofs, Proofs) =
-            proofs.into_iter().partition(|p| p.amount > amount);
-
-        let next_bigger_proof = proofs_larger.first().cloned();
-
         let mut selected_proofs: Proofs = Vec::new();
-        let mut remaining_amount = amount;
 
-        while remaining_amount > Amount::ZERO {
-            proofs_larger.sort();
-            // Sort smaller proofs in descending order
-            proofs_smaller.sort_by(|a: &Proof, b: &Proof| b.cmp(a));
-
-            let selected_proof = if let Some(next_small) = proofs_smaller.clone().first() {
-                next_small.clone()
-            } else if let Some(next_bigger) = proofs_larger.first() {
-                next_bigger.clone()
-            } else {
-                break;
-            };
-
-            let proof_amount = selected_proof.amount;
-
-            selected_proofs.push(selected_proof);
-
-            let fees = match include_fees {
-                true => self.get_proofs_fee(&selected_proofs).await?,
-                false => Amount::ZERO,
-            };
-
-            if proof_amount >= remaining_amount + fees {
-                remaining_amount = Amount::ZERO;
-                break;
-            }
-
-            remaining_amount = amount.checked_add(fees).ok_or(Error::AmountOverflow)?
-                - Amount::try_sum(selected_proofs.iter().map(|p| p.amount))?;
-            (proofs_larger, proofs_smaller) = proofs_smaller
-                .into_iter()
-                .skip(1)
-                .partition(|p| p.amount > remaining_amount);
-        }
-
-        if remaining_amount > Amount::ZERO {
-            if let Some(next_bigger) = next_bigger_proof {
-                return Ok(vec![next_bigger.clone()]);
-            }
-
-            return Err(Error::InsufficientFunds);
-        }
-
-        Ok(selected_proofs)
-    }
-
-    /// Select proofs to send
-    ///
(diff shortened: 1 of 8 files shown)


crates/cdk/src/wallet/proofs.rs
```diff
@@ -166,112 +167,393 @@ impl Wallet {
...
+            } else {
+                amount - a.amount
+            };
+            let b_diff = if b.amount > amount {
+                b.amount - amount
+            } else {
+                amount - b.amount
+            };
+            a_diff.cmp(&b_diff)
+        }),
```
ok300, line 260: Maybe this can be simplified to
```rust
ProofSelectionMethod::Closest => proofs.sort_by_key(|p| match p.amount > amount {
    true => p.amount - amount,
    false => amount - p.amount,
}),
```

crates/cdk/src/wallet/proofs.rs
```diff
@@ -166,112 +167,393 @@ impl Wallet {
...
+    proofs: &Proofs,
+    amount: Amount,
+    fees: HashMap<Id, u64>,
+) -> Option<Vec<Proof>> {
+    tracing::trace!(
+        "Selecting LEAST proofs over amount {} with fees {:?}",
+        amount,
+        fees
+    );
+    let max_sum = Amount::try_sum(proofs.iter().map(|p| p.amount)).ok()? + 1.into();
```
ok300, line 275: Since `1` is added outside of `try_sum`, it could overflow.

IMO it should be put in `try_sum`:
```rust
let mut addends: Vec<Amount> = proofs.iter().map(|p| p.amount).collect();
addends.push(1.into());
let max_sum = Amount::try_sum(addends).ok()?;
```

thesimplekid: It will panic, not overflow since the amount type internally does a checked add with an expect. https://github.com/cashubtc/cdk/blob/260f262c395056ae243ccc8cfa0e9de793c5fd20/crates/cdk/src/amount.rs#L169-L175. But could use a checked add if we want to be explicit about it.

crates/cdk/src/wallet/proofs.rs
davidcaseria: I'm not a huge fan of matching booleans since I think it reads less clearly, but I like using `sort_by_key`. Thanks!

crates/cdk/src/wallet/proofs.rs
```diff
@@ -166,112 +167,389 @@ impl Wallet {
...
+}
 
-        for inactive_proof in inactive_proofs {
-            selected_proofs.push(inactive_proof);
-            let selected_total = Amount::try_sum(selected_proofs.iter().map(|p| p.amount))?;
-            let fees = self.get_proofs_fee(&selected_proofs).await?;
+impl Default for SelectProofsOptions {
+    fn default() -> Self {
+        Self {
+            allow_inactive_keys: false,
```
thesimplekid, line 386: I think we want to always prefer inactive keys key. We want to encourage the behavior to rotate away from inactive keysets as quickly as possible. I'm okay with leaving it as option and not completely enforcing it like before but I think the default should be true, and maybe rename it to `prefer_inactive_keysets` to make what it controls a bit more clear. Open push back if there was a reason you defaulted to false?

crates/cdk/src/wallet/proofs.rs
```diff
@@ -166,112 +167,389 @@ impl Wallet {
...
+    for proof in proofs {
+        let max_other_amounts: u64 = (max_sum - proof.amount).into();
+        for t in (0..=max_other_amounts).rev() {
+            if let Some(current_sum) = dp[t as usize] {
+                let new_sum = current_sum + proof.amount;
+                let target_index = (t + Into::<u64>::into(proof.amount)) as usize;
+
+                // If this sum has not been reached yet, or if the new sum is smaller, or if the new path is shorter
+                if dp[target_index].is_none()
+                    || dp[target_index].unwrap() > new_sum
```
thesimplekid, line 291: Do we know that this will always unwrap without panicking? If so can we change it to an `.expect()` with a reason why it wont panic. 

In general unwrap should only be used in tests in places where we're conformable with `unwrap` behavior because its checked in some other way lets use `expect`so we know why its safe to use.

crates/cdk/src/wallet/proofs.rs
```diff
@@ -166,112 +167,389 @@ impl Wallet {
...
+
+    dp[0] = Some(Amount::ZERO);
+
+    // Fill DP table and track paths
+    for proof in proofs {
+        let max_other_amounts: u64 = (max_sum - proof.amount).into();
+        for t in (0..=max_other_amounts).rev() {
+            if let Some(current_sum) = dp[t as usize] {
+                let new_sum = current_sum + proof.amount;
+                let target_index = (t + Into::<u64>::into(proof.amount)) as usize;
```
thesimplekid, line 287: Don't feel the need to change this just caught my eye, but I usually use u64::from(Amount) in places i can't just use `.into()` just a bit shorter does the same thing.

thesimplekid (commented):
Nice work on this. Just a few comments

crates/cdk/src/wallet/proofs.rs
davidcaseria: Good idea. Thanks!

crates/cdk/src/wallet/proofs.rs
davidcaseria: Yes, because the `None` check is right before it. I'll change to expect though.

crates/cdk/src/wallet/proofs.rs
davidcaseria: Sounds good to me!

PR #388, opened 2024-10-03
Consolidate workspace metadata

ok300:
Since the crates are organized in a workspace, we can define most metadata fields in the workspace `Cargo.toml` and simply reference them in the member crates `Cargo.toml`s.

This makes maintenance easier. For example, `version` needs to be updated only in one place when releasing a new version. At the same time, it also gives us the flexibility to set a custom value for a member crate and overwrite the workspace value, if needed.

Note: In dependencies that were specified by `path`, like
```toml
cdk = { path = "../cdk", version = "0.4.0", default-features = false, features = ["mint"] }
```
the `version` was removed, since it's ignored by Cargo anyway. Instead, Cargo uses the code under that path, regardless of specified version.


ok300:
Too bad, looks like the `workspace-inheritance` feature is only available starting with MSRV `1.64.0`, which is higher than the current `1.63.0`.

PR #389, opened 2024-10-05
`mintd`: Add Swagger UI for mint APIs

ok300:
This PR adds a Swagger UI for the mint's API endpoints, like https://testnut.cashu.space/docs . The Swagger UI documents the mint's APIs, as well as exposes an easy interface to explore and call the APIs.

To accomplish this:
- two new dependencies were added, `utoipa` and `utoipa-swagger-ui`, at versions that are compatible with our current `axum` version (`0.6.20`)
- the structs exposed by the mint API, either as arguments or as responses, derive a new trait `ToSchema`
- `utoipa::path` annotations were added to every exposed API path
- `ApiDocV1` was added, which is a struct that holds all OpenAPI definitions, which the Swagger UI relies on

Using the default mint config, the Swagger UI is available at http://localhost:8085/swagger-ui/

Open TODOs
- [x] Expose all available APIs (feature parity with https://testnut.cashu.space/docs )
- [x] Place new dependencies and imports behind a feature flag (`mint` or `mint/utoipa`)
- [x] Clarify and consolidate: `quote` vs `quote_id` in args


thesimplekid:
Really nice work on this pr. Could we just make add config in the mintd config.toml whether this is exposed or not. And then do you mind rebasing a squashing the commits. 


Also no need to do this now but it seems like it would be possible to add a `utopia` feature and not use use the `mint` feature do you think thats the case?

thesimplekid (changes requested):
Really nice work on this pr. Could we just make add config in the mintd config.toml whether this is exposed or not. And then do you mind rebasing a squashing the commits.

Also no need to do this now but it seems like it would be possible to add a utopia feature and not use use the mint feature do you think thats the case?

ok300:
> it seems like it would be possible to add a utopia feature and not use use the mint feature do you think thats the case?

Would something like 345f3ad87e2c49f11ab2a0d79c4297dda0063297 work, adding an optional feature to the mint?

I named the feature `swagger`, since `utoipa` is just the library but [the interface is called Swagger](https://swagger.io/tools/swagger-ui/).

ok300:
> Really nice work on this pr. Could we just make add config in the mintd config.toml whether this is exposed or not. And then do you mind rebasing a squashing the commits.

Done.

> Also no need to do this now but it seems like it would be possible to add a utopia feature and not use use the mint feature do you think thats the case?

Done.

thesimplekid (approved):
Awesome, LGTM

PR #392, opened 2024-10-06
CI yml: add missing commas in build-args list

ok300:
This PR fixes a few typos in the `ci.yml`, where some `build-arg` values were not separated by comma.

Without commas between the `build-args` values, the values are concatenated and treated as one.


thesimplekid (approved):
Nice catch

Issue #203, opened 2024-06-27
NUT-17: Websockets


thesimplekid:
https://github.com/cashubtc/nuts/pull/98#discussion_r1655550190

thesimplekid:
Hey @crodas,

Sure go for it thank you

thesimplekid:
@crodas Have you started on this how is it coming?

thesimplekid:
Yes thats fine. Feel free to publish a draft pr if you want me to take a look at anything even before its working.

thesimplekid:
@crodas I'm going to take this back.

prusnak:
PR https://github.com/cashubtc/cdk/pull/394 has been merged

Should this issue be closed or is there still something left to be done?

thesimplekid:
> Should this issue be closed or is there still something left to be done?

#394 Is mint only so we do still need wallets support, @crodas said he would work on that so lets close this and ill open another tracking issue for that. 

Thanks tracking this.

PR #394, opened 2024-10-08
Mint Websockets (NUT-17)

crodas:
Very early stage nonfunctional prototype to add support for any subscriptions, aiming for NUT-17 support, #203 .

TODO:
* [x] Write tests for event generation inside the existing Mint tests (84b2a6a)
* [x] Write unit tests for the websocket client and external tests (if possible)
* [x] Does it worth to make it opt-in or possible to disable subscriptions?
* [x] Add [Signaling Support via NUT-06](https://github.com/cashubtc/nuts/blob/main/17.md#signaling-support-via-nut-06) (6fafbfdd70158f2c5f40a65e5505d7719379fa60)

### DIFF
--- a/.helix/languages.toml
+++ b/.helix/languages.toml
@@ -1,2 +1,2 @@
 [language-server.rust-analyzer.config]
-cargo = { features = ["wallet", "mint"] }
+cargo = { features = ["wallet", "mint", "swagger"] }
--- a/bindings/cdk-js/src/nuts/nut06.rs
+++ b/bindings/cdk-js/src/nuts/nut06.rs
@@ -78,6 +78,7 @@ impl JsMintInfo {
         contact: Option<Vec<JsContactInfo>>,
         nuts: JsValue,
         icon_url: Option<String>,
+        urls: Option<Vec<String>>,
         motd: Option<String>,
         time: Option<u64>,
     ) -> Result<JsMintInfo> {
@@ -92,6 +93,7 @@ impl JsMintInfo {
                     .map(|contacts| contacts.iter().map(|c| c.deref().clone()).collect()),
                 nuts: serde_wasm_bindgen::from_value(nuts).map_err(into_err)?,
                 icon_url,
+                urls,
                 motd,
                 time,
             },
--- a/crates/cdk-cln/src/lib.rs
+++ b/crates/cdk-cln/src/lib.rs
@@ -16,10 +16,7 @@ use cdk::cdk_lightning::{
     self, CreateInvoiceResponse, MintLightning, PayInvoiceResponse, PaymentQuoteResponse, Settings,
 };
 use cdk::mint::FeeReserve;
-use cdk::nuts::{
-    CurrencyUnit, MeltMethodSettings, MeltQuoteBolt11Request, MeltQuoteState, MintMethodSettings,
-    MintQuoteState,
-};
+use cdk::nuts::{CurrencyUnit, MeltQuoteBolt11Request, MeltQuoteState, MintQuoteState};
 use cdk::util::{hex, unix_time};
 use cdk::{mint, Bolt11Invoice};
 use cln_rpc::model::requests::{
@@ -45,28 +42,19 @@ pub struct Cln {
     rpc_socket: PathBuf,
     cln_client: Arc<Mutex<cln_rpc::ClnRpc>>,
     fee_reserve: FeeReserve,
-    mint_settings: MintMethodSettings,
-    melt_settings: MeltMethodSettings,
     wait_invoice_cancel_token: CancellationToken,
     wait_invoice_is_active: Arc<AtomicBool>,
 }
 
 impl Cln {
     /// Create new [`Cln`]
-    pub async fn new(
-        rpc_socket: PathBuf,
-        fee_reserve: FeeReserve,
-        mint_settings: MintMethodSettings,
-        melt_settings: MeltMethodSettings,
-    ) -> Result<Self, Error> {
+    pub async fn new(rpc_socket: PathBuf, fee_reserve: FeeReserve) -> Result<Self, Error> {
         let cln_client = cln_rpc::ClnRpc::new(&rpc_socket).await?;
 
         Ok(Self {
             rpc_socket,
             cln_client: Arc::new(Mutex::new(cln_client)),
             fee_reserve,
-            mint_settings,
-            melt_settings,
             wait_invoice_cancel_token: CancellationToken::new(),
             wait_invoice_is_active: Arc::new(AtomicBool::new(false)),
         })
@@ -81,8 +69,6 @@ impl MintLightning for Cln {
         Settings {
             mpp: true,
             unit: CurrencyUnit::Msat,
-            mint_settings: self.mint_settings.clone(),
-            melt_settings: self.melt_settings.clone(),
             invoice_description: true,
         }
     }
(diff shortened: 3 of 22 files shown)


thesimplekid:
Hey, haven't done a very through review but it looks like a good approach so far, thank you for you work. If there is a specific part you would like me to look at let me know.

thesimplekid (changes requested):
Nice work, think were just missing sending updates in a few places for the proofs some I was able to note in line. This one i couldn't note inline so will note it here, when we swap we need to send an update for the state of the proofs that are spent as part of the swap. Think line 171 of `mint/swap.rs`

crates/cdk-axum/src/ws/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+                tracing::info!("WebSocket closed");
+                break;
+            }
+            Either::Left((None, _)) => {
+                tracing::info!("WebSocket closed");
+                break;
+            }
+            _ => {}
+        }
+    }
```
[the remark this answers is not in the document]
thesimplekid: Yeah for some reason clippy thinks it breaks msrv when it doesn't, so we can use ` #[allow(clippy::incompatible_msrv)]
` to suppress the warning

thesimplekid:
In the `verify_melt` fn we need to broadcast the proof state and the quote state when they are changed to pending

thesimplekid (changes requested):
In the `verify_melt` fn we need to broadcast the proof state and the quote state when they are changed to pending.

I think you have all the other state changes. What else do you feel is missing from this PR before you would like to see it merged? 

>  Does it worth to make it opt-in or possible to disable subscriptions?

Do you mean like as a feature flag? I don't think it is and we should just have the mint always support it.

crates/cdk-axum/src/ws/unsubscribe.rs
```diff
@@ -0,0 +1,29 @@
+use super::{handler::WsHandle, WsContext, WsError};
+use cdk::pub_sub::SubId;
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct Method {
+    pub sub_id: SubId,
+}
```
thesimplekid, line 7:
```suggestion
pub struct Method {
    #[serde(rename = "subId")]
    pub sub_id: SubId,
}
```

sub id should be snake case

crates/cdk-axum/src/ws/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+                let message = if let Ok(message) = serde_json::to_string(&notification) {
+                    message
+                } else {
+                    tracing::error!("Could not serialize notification");
+                    continue;
+                };
+
+                if socket.send(Message::Text(message)).await.is_err() {
+                    break;
+                }
```
thesimplekid, line 89:
```suggestion
          if let Err(err)= socket.send(Message::Text(message)).await {
                   tracing::error!("Could not send websocket message: {}", err);
                     break;
          }
```

What do you think of using an if let to log the error here?

crates/cdk-axum/src/ws/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+
+                match request.method.process(request.id, &mut context).await {
+                    Ok(result) => {
+                        if socket
+                            .send(Message::Text(result.to_string()))
+                            .await
+                            .is_err()
+                        {
+                            break;
+                        }
```
thesimplekid, line 108: Same as above use an If let to log the error?

crates/cdk/src/pub_sub/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    /// Return number of active subscriptions
+    pub fn active_subscriptions(&self) -> usize {
+        self.active_subscriptions.load(atomic::Ordering::SeqCst)
+    }
+
+    /// Task to remove dropped subscriptions from the storage struct
+    ///
+    /// This task will run in the background (and will dropped when the Manager
+    /// is ) and will remove subscriptions from the storage struct it is dropped.
```
thesimplekid, line 162:
```suggestion
    /// This task will run in the background (and will be dropped when the [`Manager`]
    /// is) and will remove subscriptions from the storage struct it is dropped.
```

crates/cdk/src/nuts/nut17.rs
```diff
@@ -0,0 +1,358 @@
...
+            state: melt_quote.state,
+            paid: Some(melt_quote.state == MeltQuoteState::Paid),
+            expiry: melt_quote.expiry,
+            amount: melt_quote.amount,
+            fee_reserve: melt_quote.fee_reserve,
+        }
+    }
+}
+
+impl PubSubManager {
```
thesimplekid, line 193: Should we also have a proof state helper here for constancy?

crates/cdk-integration-tests/tests/regtest.rs
```diff
@@ -39,11 +83,40 @@ async fn test_regtest_mint_melt_round_trip() -> Result<()> {
...
+                  "subId": "test-sub",
+                }
+
+        }))?))
+        .await?;
 
-    assert!(melt.state == MeltQuoteState::Paid);
+    assert_eq!(
+        reader.next().await.unwrap().unwrap().to_text().unwrap(),
+        r#"{"jsonrpc":"2.0","result":{"status":"OK","sub_id":"test-sub"},"id":2}"#
```
thesimplekid, line 104:
```suggestion
        r#"{"jsonrpc":"2.0","result":{"status":"OK","subId":"test-sub"},"id":2}"#
```

crates/cdk-axum/src/ws/subscribe.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub_sub::SubId,
+};
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct Method(Params);
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct Response {
+    status: String,
+    sub_id: SubId,
```
thesimplekid, line 16:
```suggestion
    #[serde(rename = "subId")]  
    sub_id: SubId,
```

thesimplekid (changes requested):
Overall really nice work. Just a few small things i noticed when testing vs Nutshell

crates/cdk/src/nuts/nut17.rs
```diff
@@ -0,0 +1,358 @@
...
+    fn from(mint_quote: &MintQuote) -> MintQuoteBolt11Response {
+        MintQuoteBolt11Response {
+            quote: mint_quote.id.clone(),
+            request: mint_quote.request.clone(),
+            state: mint_quote.state,
+            expiry: Some(mint_quote.expiry),
+        }
+    }
+}
+
```
thesimplekid, line 176: We have this in NUT04 lets move this there.


https://github.com/cashubtc/cdk/blob/d9fb5f814a745098c8a3c4ac8784ce2d28f52139/crates/cdk/src/nuts/nut04.rs#L95-L105

crates/cdk-axum/src/lib.rs
```diff
@@ -153,7 +154,7 @@ pub async fn create_mint_router(mint: Arc<Mint>, cache_ttl: u64, cache_tti: u64)
             get(get_check_mint_bolt11_quote),
         )
         .route("/mint/bolt11", post(cache_post_mint_bolt11))
-        .route("/melt/quote/bolt11", post(post_melt_bolt11_quote))
```
thesimplekid, line 156: Think the merge got messed up here this route shouldn't be deleted

thesimplekid (approved):
Great work thanks for this.

PR #404, opened 2024-10-13
Bolt12 with support for multiple payments per mint quote

### LINKED ISSUE #355: CDK-CLI: Add command/flag to claim mints paid after exit
Currently in the cli if the command exits after getting a mint quote but before actually minting there is no way to restart the minting if the invoice is paid. This is supported by the library it is just not exposed by the cli.

thesimplekid:
This PR included bolt12 support for both the mint and the wallet. It also includes a large refactor of mintd, these really should have been separated but too late now (sorry). 

| Backend  | Send | Receive |
|----------|------|---------|
| CLN      | :heavy_check_mark:    | :heavy_check_mark:       |
| LND\*      | :construction:    | :x:  | 
| PHD**     | :heavy_check_mark:     | :x:       |
| LNBITS   | :x:    | :x:   |
| Strike***   | :question:    | :question:       |  

\* With use of LNDK as lnd does not directly support bolt12
** PHD doesnt allow creation of one time use bolt12, so receive cannot be supported
*** Strike api doesn't seem to support bolt12, coming soon?

### Related Issues should be closed by this PR
- [ ] Should also close #355, this logic is handled for bolt12 use same logic for bolt11
- [ ] https://github.com/cashubtc/cdk/issues/430


crates/cdk/src/mint/mint_20.rs
```diff
@@ -0,0 +1,102 @@
...
+            .nut18
+            .as_ref()
+            .ok_or(Error::UnsupportedUnit)?;
+
+        if nut18.disabled {
+            return Err(Error::MintingDisabled);
+        }
+
+        let ln = self.bolt12_backends.get(&unit).ok_or_else(|| {
+            tracing::info!("Bolt11 mint request for unsupported unit");
```
mubarak23, line 36: can we change this to reflect Bolt12 -> ` Bolt12 mint request for unsupported unit`

thesimplekid:
> Dropping this comment to see if there will/should be a Bolt12 specific event for this? or are the Quotes relatively the same for 11 and 12 that this works for both? For context, I'm interested in subbing to events that occur when Bolt12 offers are paid and this seems like the right place. If I'm off-base in my understand, perhaps there's a better way of being alerted to a Bolt12 offer being paid?

I haven't thought about this since i did this before websockets but I think this is a good point I think the mint should send a notification each time the mint quote is paid for quote thats payable multiple times. I'm picking this pack up and will add this as well.

thesimplekid:
superseded by https://github.com/cashubtc/cdk/pull/709

PR #406, opened 2024-10-14
`mintd`: sort pks by amount in `/v1/keys`

ok300:
This PR brings two related changes:
- it ensures the mint keys exposed in `/v1/keys/{keyset_id}` are sorted by amount. Before, they were sorted alphabetically for the `String` representation of `Amounts` (`"123", "2", "33"` vs now `"2", "33", "123"`).
- it simplifies the first step in `Id` generation, namely the sorting of pubkeys. Before, there were two conversions (`Amount` to `String` to `u64`) before sorting, for each `PublicKey`. Now, the two conversions are skipped and the keys are directly sorted by `Amount`.

### DIFF
--- a/crates/cdk/src/amount.rs
+++ b/crates/cdk/src/amount.rs
@@ -4,8 +4,9 @@
 
 use std::cmp::Ordering;
 use std::fmt;
+use std::str::FromStr;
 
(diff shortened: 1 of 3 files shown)


thesimplekid (approved):
LGTM

PR #408, opened 2024-10-16
Add `Proofs` trait to consolidate aggregate fns

ok300:
This PR adds the `ProofsMethods` trait for `Proofs`. The goal of this trait is to have a place where such methods can be consolidated. This simplifies the codebase, makes it more readable and maintainable.

There are many operations that have to be done on a `Vec<Proof>` (a.k.a. `Proofs`), which involve iterating, mapping, and re-collecting the result. This is usually boilerplate that's repeated. One such example is summing up the amounts of all proofs involved, which this PR consolidates in `ProofsMethods::total_amount`.


thesimplekid (approved):
Nice, thank you

PR #410, opened 2024-10-20
Simplify `MintUrl`, use it directly in `wallet/client.rs`

ok300:
This PR simplifies how `MintUrl` is handled:
- args that represent a mint URL are given the explicit `MintUrl` type, instead of `Url` and a later `try_into` conversion
- as a consequence, `TryInto` implementations were not used anymore to convert between `MintUrl` and `Url` and were removed
- a few unused error types were also removed (in fact they were duplicated)
- the class method `join_url` was converted to a method of `MintUrl` for better readability


thesimplekid (approved):
Great this is much cleaner. Love a negative git diff

PR #417, opened 2024-10-23
Keyset ID: fix deserialization edge-case, add unit tests

ok300:
This PR fixes an edge-case where the keyset ID could be deserialized from a string starting with a non `00` byte into a valid keyset ID with the version byte `00`.

The PR also
- replaces the custom serializer and deserializer with `serde` annotations
- adds more serialization and deserialization unit tests

### DIFF
--- a/crates/cdk/src/nuts/nut02.rs
+++ b/crates/cdk/src/nuts/nut02.rs
@@ -18,7 +18,7 @@ use bitcoin::hashes::Hash;
 use bitcoin::key::Secp256k1;
 #[cfg(feature = "mint")]
 use bitcoin::secp256k1;
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, VecSkipError};
 use thiserror::Error;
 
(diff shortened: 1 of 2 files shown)


thesimplekid (approved):

crates/cdk/src/nuts/nut02.rs
```diff
@@ -130,81 +131,46 @@ impl fmt::Display for Id {
...
 }
 
-impl FromStr for Id {
-    type Err = Error;
+impl TryFrom<String> for Id {
+    type Error = Error;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        // Check if the string length is valid
+    fn try_from(s: String) -> Result<Self, Self::Error> {
```
thesimplekid, line 137: Is there any benefit to having a `TryFrom<String>` should we just remove it in favor of the FromStr`?

crates/cdk/src/nuts/nut02.rs
```diff
@@ -130,81 +131,46 @@ impl fmt::Display for Id {
...
-                    )),
-                    _ => E::custom(e),
-                })
-            }
-        }
-
-        deserializer.deserialize_str(IdVisitor)
+impl From<Id> for String {
+    fn from(value: Id) -> Self {
+        value.to_string()
```
thesimplekid, line 161: Similar question as above should we just use the `fmt::Display` impl

crates/cdk/src/nuts/nut02.rs
ok300: `utoipa` needed conversion directly to/from `String`, otherwise it would show the `Id` as a struct with inner fields, not as a `String`.

I did find a way around that by manually deriving `utoipa::ToSchema`.

However it's rather verbose: https://github.com/cashubtc/cdk/compare/ed76e34707627ff076f6bf12f67f2f0f4881469f..2cab443cdc75d3833b27f66eefd78a65da356bbb

IMO the previous version was better. `TryFrom<String>` was just a wrapper around `FromStr`, so no duplication. Same for `impl From<Id> for String`.

What do you think, should I revert this last change?

thesimplekid: Makes sense if you're happier with the first version lets go with that.

ok300: Done: https://github.com/cashubtc/cdk/compare/2cab443cdc75d3833b27f66eefd78a65da356bbb..2a092669264768cc2d58d3787a6b99658b10693f

thesimplekid: Thank you

PR #421, opened 2024-10-26
feat: support custom unit

thesimplekid:


crates/cdk/src/nuts/nut00/mod.rs
```diff
@@ -373,46 +373,71 @@ pub enum CurrencyUnit {
     Usd,
     /// Euro
     Eur,
+    /// Custom currency unit (3 characters)
+    Custom([u8; 3]),
 }
```
thesimplekid, line 378: We can't assume units are 3 letter codes. So think we need to make this a string and remove Copy

thesimplekid:
Something is wrong with the CI cache on this branch closing for #438

PR #425, opened 2024-10-28
build: add rust-toolchain.toml file

tdelabro:
Let me know what you think but, especially for open source, I prefer to pin the version of the language to make sure that all contributors are working with the same one and that no bugs or changes in behaviors they may encounter are due to running a different version of the language

Run `cargo install` at the project root to download and install the language version and the components we listed.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,4 @@
+[toolchain]
+channel="1.82.0"
+components = ["rustfmt", "clippy", "rust-analyzer"]
+


tdelabro:
> Can the channel be stable instead of a specific version?

It is not a good idea, it would mean that the project toolchain is automatically updated every time a new version is released, without any human check to see if nothing has been broken during this release. One day it will just refuse to compile because some breaking change has been introduced that we don't handle

tdelabro:
> That's why I use the nix shells,

I can see the version being defined there: https://github.com/cashubtc/cdk/blob/main/flake.nix
We have to use the same in both, otherwise it loses its interest. 

It looks like you are using stable latest in it:
```
stable_toolchain = pkgs.rust-bin.stable.latest.default.override {
          targets = [ "wasm32-unknown-unknown" ]; # wasm
        };
```
but with `target = wasm32` ??? Is it the only thing you compile too? Not native, only wasm?

thesimplekid:
> but with target = wasm32 ??? Is it the only thing you compile too? Not native, only wasm?

It add the was wasm target, native is already included. 

The CI also uses the nix flake, checks both wasm and native on stable and MSRV

tdelabro:
So the only thing I would recommend would be to fix the version of the language you are using in the CLI.
And in the future make sure it's the same for both the toolchain file and the nix one

thesimplekid:
closed for https://github.com/cashubtc/cdk/pull/434

I set a specific rust-version in the flake and toolchain.toml (current stable 1.82.0)

PR #429, opened 2024-10-30
Abstract HttpClient into HttpClientMethods trait

davidcaseria:
This decouples the HTTP client implementation from the `Wallet` by allowing a CDK user to inject a custom HTTP client if desired.

This should also help with unit testing of the `Wallet` in the future.

### DIFF
--- a/crates/cdk/src/wallet/client.rs
+++ b/crates/cdk/src/wallet/client.rs
@@ -69,7 +69,8 @@ impl HttpClient {
     }
 }
 
-#[async_trait]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 impl HttpClientMethods for HttpClient {
     /// Get Active Mint Keys [NUT-01]
     #[instrument(skip(self), fields(mint_url = %mint_url))]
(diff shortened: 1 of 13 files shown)


crates/cdk/src/wallet/client.rs
```diff
@@ -367,3 +373,90 @@ impl HttpClient {
...
+        &self,
+        mint_url: MintUrl,
+        ys: Vec<PublicKey>,
+    ) -> Result<CheckStateResponse, Error>;
+
+    /// Restore request [NUT-13]
+    async fn post_restore(
+        &self,
+        mint_url: MintUrl,
+        request: RestoreRequest,
```
thesimplekid, line 460: Nice I like this PR. 

This question stems from my mistake not a change in this pr and has been in the back of my mind for awhile but think this pr is a good chance to clean it up. 

Some of the methods take the request Struct like `post_swap` but then some of the methods take the request fields and internally construct the request struct like `post_melt`. I think we should unify this and always take the request struct. What do you think does that make sense?

davidcaseria: Good question. Since this is intended to be a low-level trait that isn't required for most implementations, I prefer passing the struct to a method like `post_swap`. In case any future functions have a lot of fields, it makes the most sense to standardize them to this. But I can be convinced the other way because, in general, I prefer not having to create structs to call a function, but in this case, I think it makes the most sense.

thesimplekid: Agree think the struct makes most sense here.

davidcaseria: Updated

davidcaseria:
@thesimplekid it seems to me that the CI errors are existing errors. Are they safe to ignore?

ok300:
I think I found a way to fix at least some of the failing tests: https://github.com/davidcaseria/cdk/pull/2

ok300:
I might have a few more CI fixes: https://github.com/davidcaseria/cdk/pull/3

davidcaseria:
@ok300 I appreciate the fixes!

thesimplekid (approved):
Thanks for this @davidcaseria. And thank you @ok300 for the fixes

PR #444, opened 2024-11-08
Send current state of the subscriptions

crodas:
Improve #394

Make WebSocket subscriptions send the current status when subscribing. It would also be interesting to implement this logic into the PubSub crate. If we do this, this solution would be agnostic to the WebSocket, and any other bit of the rust code that wants to subscribe would inherit messages of current status when they subscribe. What are your thoughts @thesimplekid ?

### DIFF
--- a/crates/cdk/src/nuts/nut17/mod.rs
+++ b/crates/cdk/src/nuts/nut17/mod.rs
@@ -1,15 +1,21 @@
 //! Specific Subscription for the cdk crate
 
+use super::{BlindSignature, CurrencyUnit, PaymentMethod};
+use crate::cdk_database::{self, MintDatabase};
+pub use crate::pub_sub::SubId;
 use crate::{
-    cdk_database::{self, MintDatabase},
     nuts::{
         MeltQuoteBolt11Response, MeltQuoteState, MintQuoteBolt11Response, MintQuoteState,
         ProofState,
     },
-    pub_sub::{self, Index, Indexable, OnNewSubscription, SubscriptionGlobalId},
+    pub_sub::{self, Index, Indexable, SubscriptionGlobalId},
 };
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, ops::Deref, sync::Arc};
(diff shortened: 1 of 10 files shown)


thesimplekid:
>  It would also be interesting to implement this logic into the PubSub crate

It looks like the pubsub mod doesn't currently have access to the database so we would have to provide that? I'm happy with it in either place so if you think its worth refactoring to put it there I'm happy with that, I do see your point about making it agnostic, but if its much more complicated it may not be worht it? 

Looking at the current code it looks good my only concern is it looks like we will silently fail, granted on calls that really shouldnt fail, but in general I think its good to at least log errors if returning them is impractical as we don't want to stop the whole call failing for a partial failure.

crates/cdk/src/nuts/nut17/mod.rs
```diff
@@ -9,6 +13,20 @@ use crate::{
 };
 use serde::{Deserialize, Serialize};
 use std::ops::Deref;
+#[cfg(feature = "mint")]
+use std::sync::Arc;
```
thesimplekid, line 17: Can you reorder these imports to be consistent with the style guide. 

https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#order-of-imports

crates/cdk/src/nuts/nut17/on_subscription_no_op.rs
thesimplekid, line 1: Couldn't we remove this no op by putting the pub sub mod behind the mind feature as it is only used there?

thesimplekid (changes requested):

thesimplekid:
Rebase on main will fix the ci issue

thesimplekid (approved):

crates/cdk/src/nuts/nut17/mod.rs
thesimplekid: https://github.com/cashubtc/cdk/pull/449


fixed it here no need to hold up merging for that

Issue #443, opened 2024-11-07
Mintd fake wallet: get payment delay from config


thesimplekid:
We should allow the payment delay in fake wallet to be configurable. It maybe better to be a range, min and max and choose a random delay between those to to more closely simulate real payments.

https://github.com/cashubtc/cdk/blob/0523892b4db55cedaca59e7e14cc955251357fae/crates/cdk-mintd/src/setup.rs#L224

mubarak23:
I can take and work on this

mubarak23:
> @thesimplekid the setup should collect min and max value, then we choose a random median average value from them.

instead of this. we can generate two random unique value, and get their average

thesimplekid:
> instead of this. we can generate two random unique value, and get their average

I think we still want it in the config, For some testing we may want this to be configurable.

PR #445, opened 2024-11-08
Make delay time in fake wallet configurable

mubarak23:
PR that closed issue: #443 

added min and max delay range field to the `FakeWallet` struct, then calculate the average value inside the setup function, and pass the value in` cdk_fake_wallet::FakeWallet::new()`

### DIFF
--- a/crates/cdk-mintd/src/config.rs
+++ b/crates/cdk-mintd/src/config.rs
@@ -3,7 +3,6 @@ use std::path::PathBuf;
 use cdk::nuts::{CurrencyUnit, PublicKey};
 use cdk::Amount;
 use config::{Config, ConfigError, File};
-use rand::Rng;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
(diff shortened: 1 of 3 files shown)


crates/cdk-mintd/src/setup.rs
```diff
@@ -216,12 +216,14 @@ impl LnBackendSetup for config::FakeWallet {
             min_fee_reserve: self.reserve_fee_min,
             percent_fee_reserve: self.fee_percent,
         };
+        // calculate
+        let delay_time = (self.min_delay_time + self.max_delay_time) / 2;
```
thesimplekid, line 220: We don't want to do the average it will always be the same so there wouldn't be a point to accepting the range we could just accept what the delay should be. We want to simulate the variance between the range. If we want a specific delay 0 for example set both the min and max as 0.

crates/cdk-mintd/src/config.rs
```diff
@@ -104,14 +105,21 @@ pub struct FakeWallet {
...
     pub reserve_fee_min: Amount,
+    pub min_delay_time: u64,
+    pub max_delay_time: u64,
 }
 
 impl Default for FakeWallet {
     fn default() -> Self {
+        let mut rng = rand::thread_rng();
+        let min_rang = rng.gen_range(1..=1000);
+        let max_rang = rng.gen_range(1..=1000);
```
thesimplekid, line 116: I think 1000 seconds is too high for a default. This also doesnt check if the min is higher then the max. 

I would just set the min default to 1 and max to 3

thesimplekid (changes requested):

crates/cdk-mintd/src/setup.rs
mubarak23: so we want generate a variance between the two value

`rng.gen_range(self.min_delay_time..=self.max_delay_time)`

crates/cdk-mintd/src/config.rs
mubarak23: but setting the min at 1 and max at 3 is small, we can do 1 and 10

thesimplekid: For testing on a fake wallet waiting 10 seconds is a long time. Its a default if someone wants a it higher they can set it in the config

thesimplekid (approved):

PR #446, opened 2024-11-08
feat: nut-20 signature on mint witness

thesimplekid:
https://github.com/cashubtc/nuts/pull/188

### DIFF
--- a/crates/cdk-integration-tests/tests/fake_wallet.rs
+++ b/crates/cdk-integration-tests/tests/fake_wallet.rs
@@ -1,12 +1,12 @@
 use std::sync::Arc;
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use bip39::Mnemonic;
 use cdk::amount::SplitTarget;
 use cdk::cdk_database::WalletMemoryDatabase;
 use cdk::nuts::{
-    CurrencyUnit, MeltBolt11Request, MeltQuoteState, MintQuoteState, NotificationPayload,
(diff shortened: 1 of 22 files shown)


crates/cdk-integration-tests/tests/fake_wallet.rs
```diff
@@ -377,6 +378,87 @@ async fn test_fake_melt_change_in_quote() -> Result<()> {
...
+    let secret = SecretKey::generate();
+    let mint_quote = wallet
+        .mint_quote(100.into(), None, Some(secret.public_key()))
+        .await?;
+
+    wait_for_mint_to_be_paid(&wallet, &mint_quote.id).await?;
+
+    let mint_amount = wallet
+        .mint(&mint_quote.id, SplitTarget::default(), None, Some(secret))
+        .await?;
```
callebtc, line 399: Would it be good to generate `secret` inside `wallet.mint_quote` and also put it into the database there, so when you call `wallet.mint()` a few lines further, you don't need to provide `secret.public_key()` but could fetch it from the db again inside `wallet.mint()`?

thesimplekid: Yeah I was thinking about this I initially did it this way to avoid storing the secret in the database but I don't think there is much benefit to that. Storing it in the cdk database makes it much easier for the implementer as they do not have to store or manage secrets so I will go this route as you suggest.

PR #447, opened 2024-11-10
Update existing Id::try_from to return a u32 and impl new ones to convert to and from u64

vnprc:
These functions are very useful integrating cdk into the SRI codebase. Probably other codebases as well.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,1 +1,1 @@
-/nix/store/9bf8g8scpkrma0rwv05b4bd1qc81gihg-pre-commit-config.json
+/nix/store/v35hg96mpn8sa4i4vk9qm1f4jdyb59yj-pre-commit-config.json
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,1 +1,1 @@
-/nix/store/v35hg96mpn8sa4i4vk9qm1f4jdyb59yj-pre-commit-config.json
+/nix/store/9bf8g8scpkrma0rwv05b4bd1qc81gihg-pre-commit-config.json
--- a/crates/cdk/src/nuts/nut02.rs
+++ b/crates/cdk/src/nuts/nut02.rs
@@ -561,4 +561,15 @@ mod test {
         let id_from_uppercase = Id::from_str(&SHORT_KEYSET_ID.to_uppercase());
         assert!(id_from_uppercase.is_ok());
     }
+
+    #[test]
+    fn test_id_u64_conversion() {
+        let id = generate_random_id();
+        let u64_value = id.to_u64();
+        let converted_id = Id::from_u64(u64_value).unwrap();
+
+        assert_eq!(id, converted_id);
+        assert_eq!(id.version, converted_id.version);
+        assert_eq!(id.id, converted_id.id);
+    }
 }
--- a/crates/cdk/src/nuts/nut02.rs
+++ b/crates/cdk/src/nuts/nut02.rs
@@ -112,17 +112,23 @@ impl Id {
             id: bytes[1..].try_into()?,
         })
     }
+}
 
-    /// [`Id`] to u64
-    pub fn to_u64(&self) -> u64 {
-        let bytes = self.to_bytes();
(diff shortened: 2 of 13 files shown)


.pre-commit-config.yaml
vnprc, line 1: what's this?

thesimplekid: Yeah this is for the pre commit checks, I need to make it easier for other contributors. But for now can you just use the one from main.

Curious are you using the nix flake and it changed the file?

crates/cdk/src/nuts/nut02.rs
```diff
@@ -112,6 +112,20 @@ impl Id {
...
         })
     }
+
+    /// [`Id`] to u64
+    pub fn to_u64(&self) -> u64 {
+        let bytes = self.to_bytes();
+        let mut array = [0u8; 8];
+        array[..bytes.len()].copy_from_slice(&bytes);
+        u64::from_be_bytes(array)
+    }
```
thesimplekid, line 122: I don't think this matches the nut here https://github.com/cashubtc/nuts/blob/main/13.md#keyset-id?

Can we just reuse https://github.com/cashubtc/cdk/blob/cc5b2673674f3517d267490008f605dec3b48628/crates/cdk/src/nuts/nut02.rs#L117-L126

crates/cdk/src/nuts/nut02.rs
```diff
@@ -547,4 +561,15 @@ mod test {
...
+    #[test]
+    fn test_id_u64_conversion() {
+        let id = generate_random_id();
+        let u64_value = id.to_u64();
+        let converted_id = Id::from_u64(u64_value).unwrap();
+
+        assert_eq!(id, converted_id);
+        assert_eq!(id.version, converted_id.version);
+        assert_eq!(id.id, converted_id.id);
+    }
```
thesimplekid, line 574: Can we add a test that it matches the try from https://github.com/cashubtc/cdk/blob/cc5b2673674f3517d267490008f605dec3b48628/crates/cdk/src/nuts/nut02.rs#L117-L126

.pre-commit-config.yaml
vnprc: Yeah I used `nix develop` as usual and it auto-updated this file.

crates/cdk/src/nuts/nut02.rs
vnprc: Ah, this is a much better idea! Sorry for the garbage PR. I can take care of that test tho.

vnprc: Actually, I take it back. The existing TryFrom implementation returns a u64 with half of the least significant bytes truncated. Shouldn't it return a u32 instead of a u64?

I need functions to convert the 8 byte array to a u64 and back without data loss.

vnprc: I'm not sure what the existing TryFrom is used for, or if it is used. I think the simplest fix would be to update it to return a u32 and then write two new impls. Is this a good plan?
```
impl TryFrom<u64> for Id
impl TryFrom<Id> for u64
```

vnprc: I opened a PR in the nuts repo to clarify the language. https://github.com/cashubtc/nuts/pull/189

crates/cdk/src/nuts/nut02.rs
```diff
@@ -490,10 +509,28 @@ mod test {
     fn test_to_int() {
         let id = Id::from_str("009a1f293253e41e").unwrap();
 
-        let id_int = u64::try_from(id).unwrap();
+        let id_int = u32::try_from(id).unwrap();
         assert_eq!(864559728, id_int)
     }
 
+    #[test]
+    fn test_u64_to_id_and_back_conversion() {
```
vnprc, line 517: This function name is innaccurate and too long. Change it to something like this:
```suggestion
    fn test_to_u64_and_back() {
```

vnprc:
Crap I merged the latest from master and now the PR shows unrelated file changes. Is there a good way to fix this?

Issue #422, opened 2024-10-27
Wallet: add support for multiple active keysets per unit


ok300:
The wallet has a method `get_active_mint_keyset` which returns:

https://github.com/cashubtc/cdk/blob/09b5a55239787aa623ec6b210a7296fd8b4d8a05/crates/cdk/src/wallet/keysets.rs#L89

This assumes
- the mint has only one active keyset for this unit, or
- all of this mint's active keysets for this unit are equivalent (have the same amounts)

Neither are supported by the spec: https://cashubtc.github.io/nuts/02/#active-keysets

ISTM the wallet should track all active keysets per unit.

mubarak23:
@ok300 so essentially this should return all the keysets, not just the first one as specified above.

mubarak23:
@thesimplekid do we want to return all active keysets since a mint can have at least one  or more active keyset.

thesimplekid:
> @thesimplekid do we want to return all active keysets since a mint can have at least one or more active keyset.l

This would break other parts of cdk that expect this to return the one keyset that it should use in the method. I think what would be best would be to move this logic to a new fn `get_active_mint_keysets` that returns all the active keysets, just this funnction but remove `.first()` and returns a `Vec<KeySetInfo>`.

Then we can change this function `get_active_mint_keyset` to call that function and then select the keyset that should be used, the one with the lowest fee. This way the api is not changed in other parts of the code but we account for the possiblity of multiple active keysets in a more intelligent way.

mubarak23:
> > @thesimplekid do we want to return all active keysets since a mint can have at least one or more active keyset.l
> 
> This would break other parts of cdk that expect this to return the one keyset that it should use in the method. I think what would be best would be to move this logic to a new fn `get_active_mint_keysets` that returns all the active keysets, just this funnction but remove `.first()` and returns a `Vec<KeySetInfo>`.
> 
> Then we can change this function `get_active_mint_keyset` to call that function and then select the keyset that should be used, the one with the lowest fee. This way the api is not changed in other parts of the code but we account for the possiblity of multiple active keysets in a more intelligent way.

understood, 

the lowest fee will be check again `input_fee_ppk` in `KeySetInfo` struct

lollerfirst:
Is there in principle any reason not to have only one active keyset per unit? I know it's not in the specs, but why is my question.

thesimplekid:
[...]

PR #448, opened 2024-11-10
multiple active keysets and return active keyset with lowest fee 

mubarak23:
PR That closes issue #422 

 function `get_active_mint_keyset` to call that `get_active_mint_keysets` function and then select the keyset that should be used, the one with the lowest fee

### DIFF
--- a/crates/cdk/src/error.rs
+++ b/crates/cdk/src/error.rs
@@ -154,7 +154,7 @@ pub enum Error {
     /// No active keyset
     #[error("No active keyset")]
     NoActiveKeyset,
-     /// No active keyset
+    /// No active keyset
     #[error("No active keyset with lowest fee")]
     NoActiveKeysetWithLowestFee,
     /// Incorrect quote amount
--- a/crates/cdk/src/error.rs
+++ b/crates/cdk/src/error.rs
@@ -154,6 +154,9 @@ pub enum Error {
     /// No active keyset
     #[error("No active keyset")]
     NoActiveKeyset,
+     /// No active keyset
+    #[error("No active keyset with lowest fee")]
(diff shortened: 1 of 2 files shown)


crates/cdk/src/wallet/keysets.rs
```diff
@@ -86,6 +86,21 @@ impl Wallet {
...
+    /// Queries mint for current keysets then gets [`Keys`] for any unknown
+    /// keysets
+    #[instrument(skip(self))]
+    pub async fn get_active_mint_keyset(&self) -> Result<KeySetInfo, Error> {
+        let active_keysets = self.get_active_mint_keysets().await?;
+
+        let keyset_with_lowest_fee = active_keysets
+            .into_iter()
+            .min_by_key(|key| key.input_fee_ppk)
+            .ok_or(Error::NoActiveKeysetWithLowestFee)?;
```
thesimplekid, line 103:
```suggestion
            .ok_or(Error::NoActiveKeyset)?;
```

This will only return a `None` if there is no active keyset so we don't need a new error variant we can just use `NoActiveKeyset`

crates/cdk/src/error.rs
```diff
@@ -154,6 +154,9 @@ pub enum Error {
     /// No active keyset
     #[error("No active keyset")]
     NoActiveKeyset,
+    /// No active keyset
+    #[error("No active keyset with lowest fee")]
+    NoActiveKeysetWithLowestFee,
```
thesimplekid, line 159:
```suggestion
```

See comment below.

thesimplekid (changes requested):

thesimplekid (approved):

PR #451, opened 2024-11-12
chore: precommit yml

thesimplekid:

PR #452, opened 2024-11-13
Update Id::try_from to return a u32 and remove redundant logic in nut13::derive_path_from_keyset_id

vnprc:
Recreating [this PR ](https://github.com/cashubtc/cdk/pull/447)without including other contributors' changes in the diff.

The existing `TryFrom<Id>` trait implementation returns a u64 but can only contain values that fit within a u32 according to the [nut13 protocol definition](https://github.com/cashubtc/nuts/blob/main/13.md#keyset-id). This PR changes the existing implementation to return a u32 and defines two new TryFrom trait implementations that can be used to convert a keyset ID to and from a u64 with no data loss.

I have also opened a PR to clarify the language in NUT13 [here](https://github.com/cashubtc/nuts/pull/189).

### DIFF
--- a/crates/cdk/src/nuts/nut02.rs
+++ b/crates/cdk/src/nuts/nut02.rs
@@ -123,6 +123,7 @@ impl Id {
 }
 
 // Used to generate a compressed unique identifier as part of the NUT13 spec
+// This is a one-way function
 impl From<Id> for u32 {
     fn from(value: Id) -> Self {
         let hex_bytes: [u8; 8] = value.as_bytes();
--- a/crates/cdk/src/nuts/nut02.rs
+++ b/crates/cdk/src/nuts/nut02.rs
@@ -112,17 +112,24 @@ impl Id {
             id: bytes[1..].try_into()?,
         })
     }
+
+    /// [`Id`] as bytes
+    pub fn as_bytes(&self) -> [u8; Self::BYTELEN + 1] {
+        let mut bytes = [0u8; Self::BYTELEN + 1];
+        bytes[0] = self.version.to_byte();
+        bytes[1..].copy_from_slice(&self.id);
+        bytes
+    }
 }
 
-impl TryFrom<Id> for u32 {
-    type Error = Error;
-    fn try_from(value: Id) -> Result<Self, Self::Error> {
-        let hex_bytes: [u8; 8] = value.to_bytes().try_into().map_err(|_| Error::Length)?;
+// Used to generate a compressed unique identifier as part of the NUT13 spec
+impl From<Id> for u32 {
+    fn from(value: Id) -> Self {
+        let hex_bytes: [u8; 8] = value.as_bytes();
 
         let int = u64::from_be_bytes(hex_bytes);
 
-        let result = (int % (2_u64.pow(31) - 1)) as u32;
-        Ok(result)
+        (int % (2_u64.pow(31) - 1)) as u32
     }
 }
 
@@ -491,7 +498,7 @@ mod test {
     fn test_to_int() {
         let id = Id::from_str("009a1f293253e41e").unwrap();
 
-        let id_int = u32::try_from(id).unwrap();
+        let id_int = u32::from(id);
         assert_eq!(864559728, id_int)
     }
 
--- a/crates/cdk/src/nuts/nut02.rs
+++ b/crates/cdk/src/nuts/nut02.rs
@@ -114,14 +114,33 @@ impl Id {
     }
 }
 
-impl TryFrom<Id> for u64 {
+impl TryFrom<Id> for u32 {
     type Error = Error;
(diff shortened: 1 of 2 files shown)


crates/cdk/src/nuts/nut02.rs
```diff
@@ -114,14 +114,33 @@ impl Id {
...
+    }
+}
+
+impl TryFrom<Id> for u64 {
+    type Error = Error;
+
+    fn try_from(value: Id) -> Result<Self, Self::Error> {
+        let bytes = value.to_bytes();
+        let byte_array: [u8; 8] = bytes.try_into().map_err(|_| Error::Length)?;
+        Ok(u64::from_be_bytes(byte_array))
```
thesimplekid, line 143: I'm hesitant to put in these u64 conversion and have the u32 one that does the module as defined in the nut. I think it maybe best to change the current `TryFrom<Id> for u64` to `TryFrom<Id> for u32` as you've done but to leave the u64 conversion out. Converting to and from u64 can be done outside cdk if thats needed for something as cdk should be kept to whats defined in the nuts.

vnprc: I see now I broke some other stuff with this change to u32. It's definitely good to isolate this change.

vnprc: don't merge, nut13 unit test incoming...

vnprc:
I wrote a test for the nut13 functionality and ran it on the commit where it was broken...and it passed! Turns out we were reducing the keyset ID to 32 bits twice, once in the TryFrom impl and again in `derive_path_from_keyset_id`. Now with a unit test we can be sure that deriving the path works correctly and only reduce the keyset to u32 once.

PS I'm having to use `--no-verify` to commit. Not sure how to fix it but I can run all the checks using just.

thesimplekid:
> Turns out we were reducing the keyset ID to 32 bits twice, once in the TryFrom impl and again in derive_path_from_keyset_id

Yeah I noticed that, I'm gonna write some full integration tests deriving from seed to make sure we haven't changed that before we merge this. 

> Not sure I agree with you about leaving out the u64 TryFrom impl but you are the captain. Disagree and commit. 🫡

Happy to hear a counter argument. 

> PS I'm having to use --no-verify to commit. Not sure how to fix it but I can run all the checks using just.

I know precommit still giving issues going to fix that here https://github.com/cashubtc/cdk/pull/451

thesimplekid:
Noticed while we're here we can actually make it a `From` instead of a `TryFrom` if we add that to https://github.com/cashubtc/cdk/blob/f3ae4f4862ef42196fa24324750655175763cd7d/crates/cdk/src/nuts/nut02.rs#L107 can do this as a different pr if you don't want to include it here. 

```rust
    pub fn as_bytes(&self) -> [u8; Self::BYTELEN + 1] {
        let mut bytes = [0u8; Self::BYTELEN + 1];
        bytes[0] = self.version.to_byte();
        bytes[1..].copy_from_slice(&self.id);
        bytes
    }
```

vnprc:
Nice!

thesimplekid:
> Another potential reason not to implement From<Id> for u64 and it's inverse is that it could create confusion with From<Id> for u32. This is best solved with code documentation, IMO.

This is my issue with it - not that it's not in the spec but that it's potentially in conflict with what is. If someone has a u32 they may expect they can safely convert it to u64 to do From<u64> for Id, but that conversion path isn't actually safe since they can't know if data was lost when converting from Id to u32 via From<Id> for u32. Having these functions is too likely to lead to potential bugs, and adding docs isn't a reliable safeguard as documentation can be easily overlooked.

While I understand the desire to reduce complexity for higher-level developers and that is the goal of cdk, in this case the risk of subtle bugs outweighs the benefit. Even though implementing this trivial conversion outside cdk requires more work, it's preferable to having potentially unsafe implicit conversions in the library itself. This is especially important for a low-level library where predictability and safety should take precedence over convenience.

thesimplekid (approved):

vnprc:
Gotcha, makes sense.

> If someone has a u32 they may expect they can safely convert it to u64 to do From for Id

Yeah I thought that was where you were gonna go initially. What do you think about this?

```
impl From<u32> for Id {
    fn from(value: Id) -> Self {
        panic!("Conversion from u32 to Id is not allowed. This operation is invalid and should never be performed.");
    }
}
```

thesimplekid:
In general, we don't use panic in cdk other then tests always return an error. 

Here specifically, I don't think there is any need to add a function that will always panic or return an error the omission of a function supporting it implicitly signals it can't/shouldn't be done. This also introduces a runtime panic instead of a developer trying to do the conversion and realizing there is no function for it when developing a much preferable time for this realization. 

If we want to be more explicit about it being a one way function and why we can add to that documentation on the `From<Id> for u32`

PR #458, opened 2024-11-16
Add support for pure integration tests

ok300:
### Description

This PR embeds the `mint_url` in the `HttpClient` itself, since every of its methods was anyway always with the wallet's `mint_url` as arg.

This has a few other benefits:
- it will make it easier to encapsulate logic for mints with multiple URLs inside `HttpClient` ("if URL A fails, try URL B")
- it makes "pure integration tests" possible, which can call and test mint methods without the extra HTTP layer

-----

### Notes to the reviewers

-----

#### CHANGED

Add support for pure integration tests

----

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -780,6 +780,7 @@ name = "cdk-integration-tests"
 version = "0.5.0"
 dependencies = [
  "anyhow",
+ "async-trait",
  "axum",
  "bip39",
  "cdk",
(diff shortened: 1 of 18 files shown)


thesimplekid (approved):
ACK 479bc12f095540cb818c263530db062b4d229bf8

ok300:
I rebased to include the recent `Uuid` change.

I also had to add an explicit conversion between the `String` and `Uuid` types for the `DirectMintConnection` used in the pure integration tests : 6f0b2de99b1d896b2020dd424f3b8d5457e0d805

thesimplekid (approved):
ACK 03adfff411ad906124accbd53e64aa34e8208f97

PR #468, opened 2024-11-21
Melt token with amountless

mubarak23:
### Description
Base on the PR: [https://github.com/cashubtc/nuts/pull/173](https://github.com/cashubtc/nuts/pull/173)

mint can accept amountless invoice when melting token, 

added setting in `MeltMethodSettings` for `amount_less` also handle cases where we are trying to melt token and amount_less invoice is provvided

### Checklist

### DIFF
--- a/crates/cdk-mintd/src/main.rs
+++ b/crates/cdk-mintd/src/main.rs
@@ -148,6 +148,7 @@ async fn main() -> anyhow::Result<()> {
                 CurrencyUnit::Sat,
                 PaymentMethod::Bolt11,
                 mint_melt_limits,
+                Some(false),
                 cln.clone(),
             );
         }
@@ -167,6 +168,7 @@ async fn main() -> anyhow::Result<()> {
                     unit,
                     PaymentMethod::Bolt11,
                     mint_melt_limits,
+                    Some(false),
                     Arc::new(strike),
                 );
             }
@@ -181,6 +183,7 @@ async fn main() -> anyhow::Result<()> {
                 CurrencyUnit::Sat,
                 PaymentMethod::Bolt11,
                 mint_melt_limits,
+                Some(false),
                 Arc::new(lnbits),
             );
         }
@@ -194,6 +197,7 @@ async fn main() -> anyhow::Result<()> {
                 CurrencyUnit::Sat,
                 PaymentMethod::Bolt11,
                 mint_melt_limits,
+                Some(false),
                 Arc::new(phd),
             );
         }
@@ -207,6 +211,7 @@ async fn main() -> anyhow::Result<()> {
                 CurrencyUnit::Sat,
                 PaymentMethod::Bolt11,
                 mint_melt_limits,
+                Some(false),
                 Arc::new(lnd),
             );
         }
@@ -224,6 +229,7 @@ async fn main() -> anyhow::Result<()> {
                     unit.clone(),
                     PaymentMethod::Bolt11,
                     mint_melt_limits,
+                    Some(false),
                     fake.clone(),
                 );
             }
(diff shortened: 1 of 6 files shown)


crates/cdk/src/nuts/nut05.rs
```diff
@@ -264,6 +266,8 @@ pub struct MeltMethodSettings {
     /// Max Amount
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_amount: Option<Amount>,
+    /// Amountless
+    pub amount_less: bool,
```
thesimplekid, line 270:
```suggestion
    pub amountless: Option<bool>,
```

The NUTs PR has this without an underscore and it can be null so it much match here.

crates/cdk-mintd/src/main.rs
```diff
@@ -148,6 +148,7 @@ async fn main() -> anyhow::Result<()> {
                 CurrencyUnit::Sat,
                 PaymentMethod::Bolt11,
                 mint_melt_limits,
+                Some(false),
```
thesimplekid, line 151: CLN and some of the other backends can support paying amountless invoices, so that should be updated to actually support this change where we can and not just support the new field but disable it.

thesimplekid (changes requested):

crates/cdk/src/nuts/nut05.rs
```diff
@@ -305,6 +312,7 @@ impl Default for Settings {
             unit: CurrencyUnit::Sat,
             min_amount: Some(Amount::from(1)),
             max_amount: Some(Amount::from(1000000)),
+            amount_less: false,
```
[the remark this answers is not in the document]
thesimplekid: We shouldn't have to update any tests as this isn't a breaking change that's why in the nut PR it is an optional field. However we should add new tests to test with an amountless invoice.

crates/cdk/src/mint/melt.rs
```diff
@@ -93,6 +123,14 @@ impl Mint {
             Error::UnitUnsupported
         })?;
 
+        // check amountless (amount in meltbolt11request) is equal to payment_quote.amount
```
thesimplekid, line 126: Why do we need to check this, shouldn't we be setting the amount?

thesimplekid (changes requested):

crates/cdk-mintd/src/main.rs
```diff
@@ -148,6 +148,7 @@ async fn main() -> anyhow::Result<()> {
                 CurrencyUnit::Sat,
                 PaymentMethod::Bolt11,
                 mint_melt_limits,
+                Some(true),
```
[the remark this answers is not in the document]
mubarak23: can you explain further on "The ln backends need to be modified to support it." @thesimplekid

crates/cdk/src/mint/melt.rs
```diff
@@ -48,6 +48,32 @@ impl Mint {
         }
     }
 
+    fn check_amount_less_invoice(
```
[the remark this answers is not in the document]
mubarak23: i don't think this can be in side `check_melt_request_acceptable` because  inside `check_amount_less_invoice` we are checking to the optional amount field value is equal to the amount inside the invoice request (payment_quote)

and right now check_melt_request_acceptable is handle before  `ln.get_payment_quote()`

crates/cdk/src/mint/melt.rs
mubarak23: do you mean mint return an amount, 

if this is the case, mint does not know the amount of ecash a wallet making the melt qoute request has.

mubarak23: if mint is setting the amount coming from the option field in `MeltQuoteBolt11Request` then , we need to check to ensure only mint that support amountless payment can handle this.

thesimplekid:
close for #497

PR #470, opened 2024-11-22
fakewallet values in config

Jeezman:
## Description

when runing `cargo run --bin cdk-mintd` with fakewallet set in config.toml it gives the error
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 11.16s
     Running `target/debug/cdk-mintd`
thread 'main' panicked at crates/cdk-mintd/src/main.rs:214:60:
Fake wallet defined
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## This PR 
- adds the defaults for fakewallet to the sample config.toml file
- adds custom error output to assertion to assertion

### assertion before
```
thread 'main' panicked at crates/cdk-mintd/src/config.rs:227:38:
assertion failed: settings.fake_wallet.is_some()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### assertion after
```
thread 'main' panicked at crates/cdk-mintd/src/config.rs:219:38:
FakeWallet backend requires a valid config.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-mintd/src/config.rs
+++ b/crates/cdk-mintd/src/config.rs
@@ -108,6 +108,18 @@ pub struct FakeWallet {
     pub max_delay_time: u64,
 }
 
+impl Default for FakeWallet {
+    fn default() -> Self {
+        Self {
+            supported_units: vec![CurrencyUnit::Sat],
+            fee_percent: 0.02,
+            reserve_fee_min: 2.into(),
+            min_delay_time: 1,
+            max_delay_time: 3,
+        }
+    }
+}
+
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Default)]
 #[serde(rename_all = "lowercase")]
 pub enum DatabaseEngine {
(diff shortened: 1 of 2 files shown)


crates/cdk-mintd/src/config.rs
```diff
@@ -108,18 +108,6 @@ pub struct FakeWallet {
     pub max_delay_time: u64,
 }
 
-impl Default for FakeWallet {
```
ok300, line 111: Why remove the default implementation?

With it, it's enough to specify
```toml
ln_backend = "fakewallet"
```

Without it, everyone has to make a choice for each of those (and any future new) values.

Jeezman: @ok300 Got it! I missed that context. Revert [here](https://github.com/cashubtc/cdk/pull/470/commits/5138f409f81a1aca48d25f42129acf4daa3eb2d9)

thesimplekid (approved):

Issue #442, opened 2024-11-07
NUT17: Wallet websocket support


thesimplekid:
@crodas is working on this, doesnt seem to let me assign it

PR #473, opened 2024-11-25
Introduce subscription support in the Wallet crate.

crodas:
### Description

The last bit missing to wrap up #442

The main goal is to add a subscription to CDK Mint updates into the wallet. This feature will be particularly useful for improving the code whenever loops hit the mint server to check status changes.

The goal is to add an easy-to-use interface that will hide the fact that we're connecting to WebSocket and subscribing to events. This will also hide the fact that the CDK-mint server may not support WebSocket updates.

To be fully backward compatible, the HttpClientMethods traits have a new method, `subscribe,` which will return an object that implements `ActiveSubscription.`

In the primary implementation, there is a `SubscriptionClient` that will attempt to connect through WebSocket and will fall to the HTTP-status pull and sleep approach (the current approach), but upper stream code will receive updates as if they come from a stream of updates through WebSocket. This `SubscriptionClient` struct will also manage reconnections to WebSockets (with automatic resubscriptions) and all the low-level stuff, providing an easy-to-use interface and leaving the upper-level code with a nice interface that is hard to misuse. When `ActiveSubscription` is dropped, it will automatically unsubscribe.

-----

### Notes to the reviewers

-----

Added Websocket subscription to the Wallet.

#### FIXED

----

### Checklist


crates/cdk/src/wallet/subscription/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+                not(feature = "http_subscription"),
+                feature = "mint",
+                not(target_arch = "wasm32")
+            ))]
+            let is_ws_support = self
+                .http_client
+                .get_mint_info(mint_url.clone())
+                .await
+                .map(|info| !info.nuts.nut17.supported.is_empty())
+                .unwrap_or_default();
```
thesimplekid, line 82: If we call this on a mint that does not include NUT17 in the info response which is possible beach it is an optional nut `is_ws_support` is set to true when we need to assume false. This leads to an infinite loop of the wallet trying to connect to the ws endpoint that does not exist on the mint. 

This happens since we have the deafult of NUT17 ` Supported settings to support all the endpoints. 
https://github.com/cashubtc/cdk/blob/7d15587e3f3b5d7e992db4b12755caf289c2e31d/crates/cdk/src/nuts/nut17/mod.rs#L38-L65

I think we should change the default here to not support any endpoints and then for out mint we won't use the Default but will explicitly set that we support them. 

It may also be worth adding a time out to the ws connection attempt so the loop does not happen. 

to reproduce this you can start a cdk-mintd from the v0.4.0 tag and then use this branch as the wallet with cdk-cli.

crates/cdk/src/wallet/subscription/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+                    new_subscription_recv,
+                    on_drop_recv,
+                )
+            } else {
+                #[cfg(not(all(
+                    not(feature = "http_subscription"),
+                    feature = "mint",
+                    not(target_arch = "wasm32")
+                )))]
+                unreachable!("unreachable code");
```
thesimplekid, line 204: Would it be better to put the mint specific code behind the mint feature instead of using unreachable?

This whole file in in a wallet mod so really there should be no mint relevant code here.

crates/cdk/src/wallet/subscription/ws.rs
```diff
@@ -0,0 +1,147 @@
...
+    let mut active_subscriptions = HashMap::<SubId, mpsc::Sender<_>>::new();
+
+    loop {
+        debug!("Connecting to {}", url);
+        let (ws_stream, _) = if let Ok(result) = connect_async(&url).await {
+            result
+        } else {
+            continue;
+        };
+        debug!("Connected to {}", url);
```
thesimplekid, line 50: As mentioned above this creates an infinite loop if the wallet cannot connect to the ws endpoint, I think we should add a timeout.

crates/cdk/src/wallet/subscription/ws.rs
```diff
@@ -0,0 +1,147 @@
+use std::collections::HashMap;
+use std::sync::atomic::AtomicUsize;
+use std::sync::Arc;
+
+use futures::{SinkExt, StreamExt};
+use tokio::sync::{mpsc, RwLock};
+use tokio_tungstenite::connect_async;
+use tokio_tungstenite::tungstenite::Message;
+use tracing::{debug, error};
```
thesimplekid, line 9: NIT: Code style we don't import tracing and use the full path when we use it. 

https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#full-paths-for-logging

thesimplekid (changes requested):
Great work just a few comments.

crates/cdk/src/wallet/subscription/ws.rs
thesimplekid: Yeah I think that makes sense. I noticed it because of the above noted issue with a mint that does not signal anything for nut17 and bc of the default we try anyway. In the case of a mint signaling explicitly that it does not support ws we should not try ws at all, if it does signal support we can try and then fall back on failure.

crates/cdk/src/wallet/subscription/mod.rs
thesimplekid: It will not be false if NUT17 is not included in the info response since we use the default here https://github.com/cashubtc/cdk/blob/5ad4328a4eda7ecd7ae869db0f98799f33ab2945/crates/cdk/src/nuts/nut06.rs#L242-L245 and the default currently is that they are all supported https://github.com/cashubtc/cdk/blob/34bd97b26263b322ecfef5c92dc5d7e84c2e087c/crates/cdk/src/nuts/nut17/mod.rs#L40-L87. This is why i suggest we change the default to not supported. 

I think we should avoid the fallback in cases where we know it is not supported.

crates/cdk/src/wallet/subscription/ws.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    let mut active_subscriptions = HashMap::<SubId, mpsc::Sender<_>>::new();
+    let mut failure_count = 0;
+
+    loop {
+        tracing::debug!("Connecting to {}", url);
+        let (ws_stream, _) = if let Ok(result) = connect_async(&url).await {
+            result
+        } else {
+            failure_count += 1;
```
thesimplekid, line 72: Do we want to log each error and what the error is or too much logging?

thesimplekid:
- [ ] https://github.com/cashubtc/cdk/issues/493

thesimplekid (approved):
ACK 2462168ec99c77c61d15b8e5b63a63fd918d4473

PR #474, opened 2024-11-26
Minor improvement in the SQLite crate

crodas:
### Description

This is a minor improvement over the SQLite crate, which performs fewer SQL statements and fetches multiple results instead.

This will also remove some redundant commit() and rollback calls. Rollback already happens on drop, and there is no need for a commit if the database hasn't changed, as the transaction is used as a locking mechanism in this context.

Minor improvements to the SQLite crate to use fewer SQL statements

#### FIXED

----

### Checklist


thesimplekid (approved):
Awesome

Issue #478, opened 2024-11-27
Mint: allow caching with external Redis service


callebtc:
Currently, cached request/response pairs are written to memory (https://github.com/cashubtc/cdk/pull/361). It would be great if the mint could use a dedicated Redis caching service for this (as well or instead of RAM caching).

mubarak23:
A setup that set memory as default for cashing requests/ response, 
But allow mint to setup their own external
Redis caching service

PR #479, opened 2024-11-28
Minor improvement cache avoid clone

crodas:
### Description

Avoid cloning to serialize to JSON; instead, dereference the object, as Serde needs a reference to the object.

This minor improvement was found when researching the caching layer (to propose a fix for #478)

### Checklist


thesimplekid (approved):
Thank you. In genral I think clone is over used in the codebase so these PRs are appreciated.

Issue #455, opened 2024-11-14
docker support for cdk-mintd


aki-mizu:
It would be nice to have docker support for running a mint in the next version.

PR #480, opened 2024-11-29
add docker support

aki-mizu:
### Description

-----
This PR adds Docker support for cdk-mintd  #455

### Notes to the reviewers

-----
This would make deployment of cdk-mintd easier.

#### ADDED
Dockerfile

----

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+# Use the official Ubuntu 24.04 image as the base image
+FROM ubuntu:24.04
+
+# Set the working directory
+WORKDIR /usr/src/app
+
+# Install necessary dependencies
+RUN apt-get update && apt-get install -y \
+    curl \
+    protobuf-compiler \
+    xz-utils
+
+# Copy the source code and flake.nix into the container
+COPY . /usr/src/app
+
+# Create /nix directory and set permissions
+USER root
+RUN mkdir -m 0755 /nix && chown root /nix
+
+# Create nixbld group and users
+RUN groupadd -g 30000 nixbld && \
+    for i in $(seq 1 10); do \
+        useradd -u $((30000 + i)) -g nixbld -G nixbld -m -s /bin/false nixbld$i; \
+    done
+
+# Install Nix and set up the environment
+RUN curl -L https://nixos.org/nix/install | sh -s -- --daemon && \
+    /bin/bash -c "source /root/.nix-profile/etc/profile.d/nix.sh && \
+    exec bash -l -c 'nix-channel --update && nix-env -iA nixpkgs.nix'"
+
+# Start the Nix daemon and develop the environment
+RUN /bin/bash -c "source /root/.nix-profile/etc/profile.d/nix.sh && \
+    /root/.nix-profile/bin/nix-daemon & \
+    cd /usr/src/app && \
+    /root/.nix-profile/bin/nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command cargo build --release --bin cdk-mintd"
+
+# Set the entry point for the container
+CMD ["/usr/src/app/target/release/cdk-mintd"]


Jeezman (commented):
```
 Building 539.8s (12/13)                                                                                                                                            docker:orbstack
 => [internal] load build definition from Dockerfile                                                                                                                              0.1s
 => => transferring dockerfile: 1.34kB                                                                                                                                            0.0s
 => [internal] load metadata for docker.io/library/ubuntu:24.04                                                                                                                   2.5s
 => [auth] library/ubuntu:pull token for registry-1.docker.io                                                                                                                     0.0s
 => [internal] load .dockerignore                                                                                                                                                 0.0s
 => => transferring context: 2B                                                                                                                                                   0.0s
 => [1/8] FROM docker.io/library/ubuntu:24.04@sha256:278628f08d4979fb9af9ead44277dbc9c92c2465922310916ad0c46ec9999295                                                             0.0s
 => [internal] load build context                                                                                                                                                81.6s
 => => transferring context: 13.61GB                                                                                                                                             78.7s
 => CACHED [2/8] WORKDIR /usr/src/app                                                                                                                                             0.0s
 => [3/8] RUN apt-get update && apt-get install -y     curl     protobuf-compiler     xz-utils                                                                                   61.1s
 => [4/8] COPY . /usr/src/app                                                                                                                                                   195.8s
 => [5/8] RUN mkdir -m 0755 /nix && chown root /nix                                                                                                                               0.4s
 => [6/8] RUN groupadd -g 30000 nixbld &&     for i in $(seq 1 10); do         useradd -u $((30000 + i)) -g nixbld -G nixbld -m -s /bin/false nixbld$i;     done                  0.4s
```
Thanks for working on this. 
While testing I noticed the size of docker image is  quite on the high side, ~ 26GB.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN mkdir -m 0755 /nix && chown root /nix
+
+# Create nixbld group and users
+RUN groupadd -g 30000 nixbld && \
+    for i in $(seq 1 10); do \
+        useradd -u $((30000 + i)) -g nixbld -G nixbld -m -s /bin/false nixbld$i; \
+    done
+
+# Install Nix and set up the environment
+RUN curl -L https://nixos.org/nix/install | sh -s -- --daemon && \
```
Jeezman, line 27: Curious as to why nix is needed

Jeezman (commented):
```
2024-11-29T18:13:01.505442Z  WARN cdk_mintd::config: Error reading config file (configuration file "/root/.cdk-mintd/config.toml" not found)
Error: error returned from database: (code: 14) unable to open database file
```
Got this error while trying to run the image

aki-mizu:
> ```
> 2024-11-29T18:13:01.505442Z  WARN cdk_mintd::config: Error reading config file (configuration file "/root/.cdk-mintd/config.toml" not found)
> Error: error returned from database: (code: 14) unable to open database file
> ```
> 
> Got this error while trying to run the image

config.toml file needs to be provided externally.
If you're using docker-compose.yml, you can provide it via volumes like this:
`volumes:
      - ./config.toml:/root/.cdk-mintd/config.toml`

Jeezman:
> > ```
> > 2024-11-29T18:13:01.505442Z  WARN cdk_mintd::config: Error reading config file (configuration file "/root/.cdk-mintd/config.toml" not found)
> > Error: error returned from database: (code: 14) unable to open database file
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Got this error while trying to run the image
> 
> config.toml file needs to be provided externally. If you're using docker-compose.yml, you can provide it via volumes like this: `volumes: - ./config.toml:/root/.cdk-mintd/config.toml`

Got it. Probably needs a documentation

Dockerfile
aki-mizu: Based on description from https://github.com/aki-mizu/cdk/tree/main/crates/cdk-mintd
In order to build from source, running nix develop is necessary.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+FROM ubuntu:24.04
+
+# Set the working directory
+WORKDIR /usr/src/app
+
+# Install necessary dependencies
+RUN apt-get update && apt-get install -y \
+    curl \
+    protobuf-compiler \
+    xz-utils
```
thesimplekid, line 11: If we use the nix this would be brought in there

Dockerfile
thesimplekid: nixos has a image that we could use instead of installing it manually but we dont actually need it to build. Most of the stuff in the flake is for development

Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Use the official Ubuntu 24.04 image as the base image
+FROM ubuntu:24.04
```
thesimplekid, line 2: We don't need a full ubuntu image its larger then we need.

thesimplekid (commented):
This docker image can be made more simple have a look at this PR https://github.com/cashubtc/cdk/pull/489. We can also break the build and runtime into to containers as we don't need build deps at runtime

thesimplekid:
>  While testing I noticed the size of docker image is quite on the high side, ~ 26GB.

This is way to high something must be off the image [here](https://github.com/cashubtc/cdk/pull/489/files) is 118 mb.

thesimplekid:
>config.toml file needs to be provided externally.
> If you're using docker-compose.yml, you can provide it via volumes like this:

This can be done but I think the better way would be to add support for the mintd to be configured by env vars that can be defined in the docker-compose. This is what nutshell does. 

https://github.com/cashubtc/cdk/issues/490

Dockerfile
aki-mizu: I thought using nix is to have consistent development and deployment environment.
If not using nix, then we have to sync the rust environment to this file from time to time. 
e.g. when upgrading from rust 1.82 to 1.83.
But there might be better way to handle this if not using nix.

Dockerfile
aki-mizu: Thanks, I replaced it with minimal image.

aki-mizu:
> This docker image can be made more simple have a look at this PR #489. We can also break the build and runtime into to containers as we don't need build deps at runtime

Thank you for the suggestion.  I separated them into build and runtime stage.

aki-mizu:
> > While testing I noticed the size of docker image is quite on the high side, ~ 26GB.
> 
> This is way to high something must be off the image [here](https://github.com/cashubtc/cdk/pull/489/files) is 118 mb.

It's 128 MB now after using minimal image.

Dockerfile
aki-mizu: fixed

aki-mizu:
> nixos has a image that we could use instead of installing it manually

Thanks for the suggestion. I added nix as base image now, which makes the file much more simplified now.

> but we dont actually need it to build. Most of the stuff in the flake is for development

Let me know your thought about current approach since image size is not a concern anymore. 
Feel free to adopt #489 instead if you think that's more suitable.

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Copy workspace files and crates directory into the container
+COPY flake.nix ./flake.nix
+COPY Cargo.toml ./Cargo.toml
+COPY crates ./crates
+
+# Start the Nix daemon and develop the environment
+RUN nix-channel --update && \
+    nix-env -iA nixpkgs.nix && \
+    nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command cargo build --release --bin cdk-mintd
```
thesimplekid, line 15:
```suggestion
# Start the Nix daemon and develop the environment
RUN nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command cargo build --release --bin cdk-mintd
```

We dont need the env or to update the channel as its done by the flake

thesimplekid (changes requested):
Can we add some documentation on mapping the config file

Dockerfile
aki-mizu: fixed

aki-mizu:
> Can we add some documentation on mapping the config file

Added now, let me know if there's anything missing.

Dockerfile
Jeezman: Yeah but its really not needed for docker builds, rust version can be set in the docker file if need be. It only adds more build steps which can be avoided

aki-mizu: We want to avoid setting rust version directly in docker file since it’s already set by the flake.

thesimplekid (approved):
ACK 0d1e39769fb3b2555b997ac2f6be376707cd164b

PR #495, opened 2024-12-07
Introduce pluggable backend cache for the HTTP layer.

### LINKED ISSUE #478: Mint: allow caching with external Redis service

crodas:
### Description

Fixes #478.
-----

### Notes to the reviewers

-----

#### ADDED

Add support for custom storages for the HTTP cache layer.

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-axum/src/cache/backend/redis.rs
+++ b/crates/cdk-axum/src/cache/backend/redis.rs
@@ -51,8 +51,8 @@ impl HttpCacheStorage for HttpCacheRedis {
     }
 
     async fn get(&self, key: &HttpCacheKey) -> Option<Vec<u8>> {
-        let mut con = match self.client.get_multiplexed_tokio_connection().await {
-            Ok(con) => con,
+        let mut conn = match self.client.get_multiplexed_tokio_connection().await {
+            Ok(conn) => conn,
             Err(err) => {
                 tracing::error!("Failed to get redis connection: {:?}", err);
                 return None;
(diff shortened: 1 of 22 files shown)


thesimplekid (changes requested):
I think we just need to add docs for how to run redis or at least note it must be run externally and link to something. Maybe add it to docker compose?

Other then that LGTM

crates/cdk-axum/src/cache/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+        Some(HttpCacheKey(second_hash.into()))
+    }
+
+    /// Get a value from the cache.
+    pub async fn get<V: DeserializeOwned>(self: &Arc<Self>, key: &HttpCacheKey) -> Option<V> {
+        self.storage
+            .get(key)
+            .await
+            .map(|value| serde_json::from_slice(&value).unwrap())
+    }
```
thesimplekid, line 170: This is causing a runtime panic when run with redis. 

I added a redis container to the docker compose and the ability to read the redis config from env vars. However I'm getting a runtime panic when using redis

image

thesimplekid (changes requested):
I added a redis container to the docker-compose and the ability to read the cache config from env vars. But @crodas  can you take a look at what is causing the panic I noted.

crates/cdk-mintd/src/config.rs
```diff
@@ -210,7 +211,7 @@ pub struct MintInfo {
 
 impl Settings {
     #[must_use]
-    pub fn new(config_file_name: &Option<PathBuf>) -> Self {
+    pub fn new<P: Into<PathBuf>>(config_file_name: Option<P>) -> Self {
```
thesimplekid, line 214: NIT code style should use where https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#generics.

This applies in a few places but ill just comment here. Won't block merging for it

crates/cdk-axum/src/cache/backend/redis.rs
```diff
@@ -0,0 +1,90 @@
...
+        let mut db_key = self.prefix.clone().unwrap_or_default();
+        db_key.extend(&**key);
+
+        con.get(db_key)
+            .await
+            .map_err(|err| {
+                tracing::error!("Failed to get value from redis: {:?}", err);
+                err
+            })
+            .ok()
```
thesimplekid, line 71: It seems redis returns an empty vec if there is no value for a key. This is cause the deserlization error we've talked about offline. If we just check if the vec is empty here and return None if it is I believe this PR is good to merge.

thesimplekid (changes requested):

PR #500, opened 2024-12-12
External Logging

lollerfirst:
### Description

Defines a `ElasticsearchLayer` which implements `Layer` and can subscribe to tracing events and transmit them to ElasticSearch (specifically, another service might require another apposite Layer)

`ElastiSearccLayer` creates a multi producer single consumer channel and spawns a task that continuously attempts to receive log messages. In the `Layer` trait implementation `on_event` is called everytime a thread logs message and it write that message to the channel.

### DIFF
--- a/crates/cdk-mintd/Cargo.toml
+++ b/crates/cdk-mintd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cdk-mintd"
-version = "0.5.0"
+version = "0.6.0"
 edition = "2021"
 authors = ["CDK Developers"]
 license = "MIT"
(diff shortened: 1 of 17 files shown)


crates/cdk-mintd/src/loggers.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// - `index`: The name of the index where logs will be sent.
+    /// - `api_key`: Optional API key for authentication.
+    pub fn new(elasticsearch_url: &str, index: &str, api_key: Option<&str>) -> Self {
+        let (sender, mut receiver) = mpsc::channel(BACKLOG);
+        let client = Client::new();
+        let base_url = format!("{}/{}/_doc", elasticsearch_url.trim_end_matches('/'), index);
+        let api_key_header = api_key.map(|key| format!("ApiKey {}", key));
+
+        // Spawn an async task to process logs and send them to Elasticsearch.
+        tokio::spawn(async move {
```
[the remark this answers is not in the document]
lollerfirst: Yea sorry for the delay.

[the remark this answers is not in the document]
lollerfirst: I opted to move the handle in the `ElasticsearchLayer` object without wrapping it in a `Option` because there is always a task in this case. Wdyt?

lollerfirst:
Not quite sure how to test the API key since ElasticSearch won't mint one for me. It says some bullshit about the current license.

lollerfirst:
Ok scrap that. Managed to get one.

lollerfirst:
Should we gate this under a feature?

crates/cdk-mintd/Cargo.toml
```diff
@@ -37,6 +37,9 @@ url = "2.3"
 utoipa = { version = "4", optional = true }
 utoipa-swagger-ui = { version = "4", features = ["axum"], optional = true }
 rand = "0.8.5"
+chrono = "0.4.39"
+serde_json = "1.0.133"
+reqwest = "0.12.9"
```
thesimplekid, line 42:
```suggestion
chrono = "0.4.39"
serde_json = "1"
reqwest = "0.12"
```

Since for MSRV we need 24.3 of reqwest we should not set the patch version, cargo will use the lastest one we can. 

This change plus rebasing onto main should fix the MSRV CI errors

crates/cdk-mintd/src/loggers.rs
```diff
@@ -0,0 +1,117 @@
+use chrono::Utc;
+use reqwest::Client;
+use serde::Serialize;
+use std::fmt::Write;
+use tokio::sync::mpsc::{self, Sender};
+use tokio::task::JoinHandle;
+use tracing::field::{Field, Visit};
+use tracing::Event;
+use tracing_subscriber::layer::{Context, Layer};
+use tracing_subscriber::registry::LookupSpan;
```
thesimplekid, line 10:
```suggestion
use std::fmt::Write;

use chrono::Utc;
use reqwest::Client;
use serde::Serialize;
use tokio::sync::mpsc::{self, Sender};
use tokio::task::JoinHandle;
use tracing::field::{Field, Visit};
use tracing::Event;
use tracing_subscriber::layer::{Context, Layer};
use tracing_subscriber::registry::LookupSpan;
```

Std imports should be first, this will fix formatting ci

crates/cdk/Cargo.toml
```diff
@@ -58,6 +58,7 @@ uuid = { version = "1", features = ["v4", "serde"] }
 # -Z minimal-versions
 sync_wrapper = "0.1.2"
 bech32 = "0.9.1"
+chrono = "0.4.39"
```
thesimplekid, line 61: Imports under the -Z minimal-versions are for the case where a dep us our does not properly define their minimum deps I don't think that is the case here. 

Do we need the dep in cdk I don't see where it is used?

thesimplekid (changes requested):
Thanks for this. I have a mild preference for putting this behind a feature, but since its in mintd and not cdk and only adds deps we mostly already have from cdk not a strong one. 

Should we add elasticsearch search to the docker-compose file with a few lines of docs of how to run it? Or what is the recommended way to run with elasticsearch?

Issue #476, opened 2024-11-27
Mint RPC Server


thesimplekid:
We would like to add an RPC server to the mint so that a mint operator has both better control of mint settings with the ability to change these settings without restarting and changing the config file as currently required. This should also include methods that give the mint operator better visibility into their mint operator in order to better mange their mint and lightning liquidity, 

Below is list of methods that I think should be supported. 

#### MInt
- Update NUT06 info description etc 
- Rotate keyset
- View total issued ecash
- View total spent ecash
- View total outstanding (difference of the above two)
- Issued, spent and outstanding per keyset
- Disable/Enable mint and melt

#### Lightning
- Open channel 
- Close channel
- Balance view

cc: @davidcaseria @crodas

PR #503, opened 2024-12-16
Add ArcSwap to update Mint configuration at runtime

crodas:
### Description

The main goal is to change settings without having multiple RwLock everywhere, instead having ArcSwap to update the configuration without having access to a mutable reference to the Mint.

This will allow the RPC Server, or any other medium to update the Mint without minimum contention.

This is a pre-requisite for #476

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/config.rs
+++ b/crates/cdk/src/mint/config.rs
@@ -10,7 +10,7 @@ use super::{Id, MintInfo, MintKeySet};
 use crate::mint_url::MintUrl;
 
 /// Mint Inner configuration
-pub struct Inner {
+pub struct Config {
     /// Active Mint Keysets
     pub keysets: HashMap<Id, MintKeySet>,
     /// Mint url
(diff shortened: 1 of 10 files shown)


crates/cdk/src/mint/config.rs
```diff
@@ -0,0 +1,98 @@
...
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use arc_swap::ArcSwap;
+
+use super::{Id, MintInfo, MintKeySet};
+use crate::mint_url::MintUrl;
+
+/// Mint Inner configuration
+pub struct Inner {
```
davidcaseria, line 13: Does this need to be `pub`?

davidcaseria: That’s better. 👍

crates/cdk/src/mint/config.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Mint url
+    pub mint_info: MintInfo,
+    /// Mint config
+    pub mint_url: MintUrl,
+}
+
+/// Mint configuration
+///
+/// This struct is used to configure the mint, and it is wrapped inside a ArcSwap, so it can be
+/// updated at runtime without locking the shared config nor without requiriming a mutable reference
```
vnprc, line 25:
```suggestion
/// updated at runtime without locking the shared config nor without requiring a mutable reference
```

thesimplekid (approved):
ACK 712fa5176dbe443ae89cd3ca7e59091cca32f157

Issue #499, opened 2024-12-12
Add binary serialization


thesimplekid:
https://github.com/cashubtc/nuts/pull/199

PR #507, opened 2024-12-19
`crawB` Binary Token Serialization

lollerfirst:
### Description

* Binary tokens spec change: https://github.com/cashubtc/nuts/pull/199
* Closes this issue: #499 

-----

#### ADDED

* cdk: Ciborium transparent error for serialization
* `Token::to_raw_bytes` serializes generic token to raw bytes, choosing between a TokenV3 or TokenV4 serialization.
* `Token::try_from` for `Vec<u8>` constructs a generic token from raw bytes.
* `TokenV3::to_raw_bytes` serializes a TokenV3 to raw bytes following the spec.
* `TokenV3::from` for `Vec<u8>` constructs a TokenV3 from raw bytes.
* `TokenV4::to_raw_bytes()` serializes a TokenV4 to raw bytes following the spec.
* `TokenV4::from` for `Vec<u8>` constructs a TokenV4 from raw bytes.
* roundtrip tests with the spec test vectors.
* `Wallet::receive_raw` which receives raw binary tokens.


thesimplekid (commented):
It looks like the plan is to remove support for V3 token binary https://github.com/cashubtc/nuts/pull/202 so I think we should remove it here before merging to avoid having to take it out later.

thesimplekid (approved):
ACK 9b0cace1b9fb169536e96b3f963e0b6dbb014e99

PR #509, opened 2024-12-20
Introduce a SignatoryManager service.

crodas:
### Description

The SignatoryManager manager provides an API to interact with keysets, private keys, and all key-related operations, offering segregation between the mint and the most sensible part of the mind: the private keys.

Although the default signatory runs in memory, it is completely isolated from the rest of the system and can only be communicated through the interface offered by the signatory manager. Only messages can be sent from the mintd to the Signatory trait through the Signatory Manager.

This pull request sets the foundation for eventually being able to run the Signatory and all the key-related operations in a separate service, possibly in a foreign service, to offload risks, as described in #476.

The Signatory manager is concurrent and deferred any mechanism needed to handle concurrency to the Signatory trait.

TODO
------

- [x] Add standalone binary to spawn a GRPC server inside cdk-signatory
- [x] Add GRPC client on cdk
- [ ] Add tests
- [ ] Add cdk-signatory to the pipeline

-----

### Notes to the reviewers

Maybe a new terminology, other than Signatory and SignatoryManager can be used, but the idea I believe is solid.

-----

#### FIXED

----

### Checklist


crates/cdk/src/mint/signatory/memory.rs
```diff
@@ -0,0 +1,369 @@
...
+            // enforce only signing supported secrets as they are blinded at
+            // that point.
+            match secret.kind {
+                Kind::P2PK => {
+                    proof.verify_p2pk()?;
+                }
+                Kind::HTLC => {
+                    proof.verify_htlc()?;
+                }
+            }
```
callebtc, line 258: Functionality like this should happen in the core mint instead of here. The message should be sent to the blind signer only if all checks are successful, including scripts etc.

crates/cdk/src/mint/signatory/memory.rs
```diff
@@ -0,0 +1,369 @@
...
+                    }
+                };
+
+                let (keyset, keyset_info) = create_new_keyset(
+                    &secp_ctx,
+                    xpriv,
+                    derivation_path,
+                    Some(0),
+                    unit.clone(),
+                    max_order,
```
callebtc, line 140: Possibly out of scope but it would be nice if this parameter would accept a slice of integers (all possible amounts) instead of a single exponent (max order).

crates/cdk-signatory/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+            c,
+            keyset_id,
+            &blinded_message.blinded_secret,
+            key_pair.secret_key,
+        )?;
+
+        Ok(blinded_signature)
+    }
+
+    async fn verify_proof(&self, proof: Proof) -> Result<(), Error> {
```
thesimplekid, line 297: I don't think the full verification of the proof should be in the signatory only the verification of the signature.

crates/cdk-signatory/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+    "rand",
+    "rand-std",
+] }
+cdk-common = { path = "../cdk-common", default-features = false, features = [
+    "mint",
+] }
+tracing = "0.1.41"
+tokio = { version = "1.21", features = ["full"] }
+tonic = "0.12.3"
+prost = "0.13.4"
```
thesimplekid, line 25: These break MSRV, I'll figure that out on the management-rpc [branch](https://github.com/cashubtc/cdk/pull/543) and report back

crates/cdk-mintd/src/config.rs
```diff
@@ -211,13 +194,13 @@ pub struct MintInfo {
 
 impl Settings {
     #[must_use]
-    pub fn new<P>(config_file_name: Option<P>) -> Self
+    pub fn new<P>(config_file_name: Option<P>, check_ln: bool) -> Self
```
thesimplekid, line 197: What is check ln?

thesimplekid (commented):
> Move the signatory to their own repo. 

You mean crate right? 

This is a really good start and a structure that makes sense. I think there are still some open questions around what should be handled by the signatory and what by the mint. For example the proof verification I don't think the checking of p2pk and htlc conditions should be done by the signatory and it should only handle stuff that actually required the mint private keys. However, we do want to provide enough information that the signatory can be more then just a blind signer, doing ie rate limiting and we don't have a strong picture of what that will look like yet.

crates/cdk-signatory/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() {
+    #[cfg(feature = "grpc")]
```
ok300, line 2:
```suggestion
    println!("cargo:rerun-if-changed=src/proto/signatory.proto");
    
    #[cfg(feature = "grpc")]
```

This will make sure `cargo` only recompiles when the `proto` file changes.

crates/cdk-signatory/src/lib.rs
davidcaseria: Should the functions also take a batch of messages/proofs to limit the amount of network calls?

thesimplekid:
possibly closes https://github.com/cashubtc/cdk/issues/614

crates/cdk-signatory/src/embedded.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl Signatory for Service {
+    fn name(&self) -> String {
+        "Embedded".to_owned()
+    }
+
+    async fn blind_sign(
```
thesimplekid, line 99: Can we add tracing::instrument to these?

crates/cdk-signatory/src/proto/server.rs
```diff
@@ -0,0 +1,205 @@
...
+{
+    inner: T,
+}
+
+#[tonic::async_trait]
+impl<T> signatory_server::Signatory for CdkSignatoryServer<T>
+where
+    T: Signatory + Send + Sync + 'static,
+{
+    async fn blind_sign(
```
thesimplekid, line 23: Can we add tracing::instrument to these?

crates/cdk-sqlite/src/mint/migrations/20250415093121_drop_keystore_foreign.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE proof_new (
+    y BLOB PRIMARY KEY,
+    amount INTEGER NOT NULL,
+    keyset_id TEXT NOT NULL, -- no FK constraint here
```
thesimplekid, line 4: Why can't we use the FK?

crates/cdk/Cargo.toml
```diff
@@ -54,6 +55,7 @@ tokio = { workspace = true, features = [
     "sync",
 ] }
 getrandom = { version = "0.2" }
+cdk-signatory = { workspace = true, features = ["grpc"] }
```
thesimplekid, line 58: Should the grpc feature be optional here by adding a feature to cdk?

crates/cdk/src/mint/keysets/auth.rs
```diff
@@ -1,65 +1,50 @@
...
-        self.ensure_blind_auth_keyset_loaded(&active_keyset_id)
-            .await?;
-
-        let keysets = self.keysets.read().await;
+    pub fn auth_pubkeys(&self) -> Result<KeysResponse, Error> {
+        let key = self
+            .keysets
+            .load()
+            .iter()
+            .find(|key| key.unit == CurrencyUnit::Auth)
```
thesimplekid, line 18: Do we know that the signatory will return only active and only one keyset?

crates/cdk/src/mint/mod.rs
```diff
@@ -48,6 +53,11 @@ pub use verification::Verification;
 /// Cashu Mint
 #[derive(Clone)]
 pub struct Mint {
+    /// Signatory backend.
+    ///
+    /// It is mainly implemented in the cdk-signatory crate, and it can be embedded in the mint or
```
thesimplekid, line 58:
```suggestion
    /// It is implemented in the cdk-signatory crate, and it can be embedded in the mint or
```

crates/cdk/src/mint/mod.rs
```diff
@@ -396,82 +330,79 @@ impl Mint {
...
-            // only supported secret kinds are used as there is no way for the mint to
-            // enforce only signing supported secrets as they are blinded at
-            // that point.
-            match secret.kind {
-                Kind::P2PK => {
-                    proof.verify_p2pk()?;
-                }
-                Kind::HTLC => {
-                    proof.verify_htlc()?;
+    pub async fn verify_proofs(&self, proofs: &[Proof]) -> Result<(), Error> {
```
thesimplekid, line 378:
```suggestion
    pub async fn verify_proofs(&self, proofs: Proofs) -> Result<(), Error> {
```

Since we just call to_owned on this later should we just make this a vec?

thesimplekid (changes requested):
Really nice work

thesimplekid:
Are there any tests you added or think we can added to better cover or do you thing its well covered. Thinking mainly around rotation and generation.

crates/cdk-signatory/src/embedded.rs
```diff
@@ -0,0 +1,141 @@
...
+        Self {
+            pipeline: tx,
+            runner,
+        }
+    }
+
+    async fn runner(
+        mut receiver: mpsc::Receiver<Request>,
+        handler: Arc<dyn Signatory + Send + Sync>,
+    ) {
```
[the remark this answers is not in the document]
thesimplekid: Thats fine keep it as it is.

thesimplekid:
> I think the tests are good enough. The only bit that I removed was knowing the upcoming private key (for instance the test test_mint_keyset_gen), since that is unknown to the CDK.

Did you remove it or add it to the signatory?

crates/cdk-signatory/proto/signatory.proto
```diff
@@ -0,0 +1,173 @@
+syntax = "proto3";
```
davidcaseria, line 1: Should this be changed to match the proposed spec? https://github.com/cashubtc/nuts/pull/250

crates/cdk-signatory/src/proto/client.rs
```diff
@@ -0,0 +1,166 @@
...
+use cdk_common::error::Error;
+use cdk_common::{BlindSignature, BlindedMessage, Proof};
+use tonic::transport::{Certificate, Channel, ClientTlsConfig, Identity};
+
+use super::{blind_sign_response, boolean_response, key_rotation_response, keys_response};
+use crate::proto::signatory_client::SignatoryClient;
+use crate::signatory::{RotateKeyArguments, Signatory, SignatoryKeySet, SignatoryKeysets};
+
+/// A client for the Signatory service.
+pub struct SignatoryRpcClient {
```
davidcaseria, line 12: Should we allow callers to include a gRPC `Metadata` struct that is included in each request? This can enable a signatory to handle multiple mints.

thesimplekid (changes requested):
Looks good just noted a few things. 

Have you tested a mint upgrade. Starting a mint on the last version 0.9.2. We should expect that once upgraded the keysets are not changed and proofs generated on past versions can still be verified? 

On that note on an upgrade there isn't anything that an operator needs to do in order to run the db_signatory correct?

crates/cdk-signatory/proto/signatory.proto
```diff
@@ -0,0 +1,173 @@
...
+message BlindedMessages {
+  repeated BlindedMessage blinded_messages = 1;
+  Operation operation = 2;
+  string correlation_id = 3;
+}
+
+// Represents a blinded message
+message BlindedMessage {
+  uint64 amount = 1;
+  string keyset_id = 2;
```
davidcaseria, line 35: Should this be bytes?

crates/cdk/src/mint/mod.rs
```diff
@@ -859,28 +683,4 @@ mod tests {
             }
         }
     }
-
-    #[tokio::test]
-    async fn test_mint_keyset_gen() {
```
thesimplekid, line 864: Offline we talked about a test to ensure the signatory generates the same keyset as the current logic. This seems to be the test we want but its removed or is moved? I tested it manually with cdk-mintd and it is correct would be good to lock it in with a test.

thesimplekid (changes requested):
I think this is good to merge if we just add the test. We can resolve the other comments in a follow up.

thesimplekid:
ACK LGTM af3f585960f7caa42b78ff0d140db521ade9004a

thesimplekid (approved):

PR #518, opened 2024-12-27
Add support for African currencies to CurrencyUnit enum

Gudnessuche:
- Added Nigerian Naira (NGN), Ghanaian Cedis (GHS), Kenyan Shillings (KES), and South African Rands (ZAR) to the CurrencyUnit enum.
- Updated the derivation_index method to include these new currencies.
- Modified the FromStr and Display implementations to recognize and handle the new currency codes.
- Enhanced serialization and deserialization to support the new currencies.

This change is aimed at improving inclusivity and supporting the growing adoption of the Cashu protocol in African countries. By adding these currencies, users in Nigeria, Ghana, Kenya, and South Africa can now utilize the protocol more effectively with their local currencies.

### DIFF
--- a/crates/cdk/src/nuts/nut00/mod.rs
+++ b/crates/cdk/src/nuts/nut00/mod.rs
@@ -376,6 +376,14 @@ pub enum CurrencyUnit {
     Usd,
     /// Euro
     Eur,
+    /// Nigerian Naira
+    Ngn,
+    /// Ghanaian Cedis
+    Ghs,
+    /// Kenyan Shillings
+    Kes,
+    /// South African Rands
+    Zar,
     /// Custom currency unit
     Custom(String),
 }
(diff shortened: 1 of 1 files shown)


crates/cdk/src/nuts/nut00/mod.rs
```diff
@@ -389,6 +397,10 @@ impl CurrencyUnit {
             Self::Msat => Some(1),
             Self::Usd => Some(2),
             Self::Eur => Some(3),
+            Self::Ngn => Some(4),
+            Self::Ghs => Some(5),
+            Self::Kes => Some(6),
+            Self::Zar => Some(7),
```
thesimplekid, line 403: Thanks for this PR. This isn't an issue introduced by this PR but the current way we do derivation path by just assigning numbers does not scale well and I don't think we should add more with this method. We've discussed the idea of using a short hashing method to achieve this. https://github.com/cashubtc/cdk/pull/464#pullrequestreview-2449511528

thesimplekid:
Going to close this to track in issue #614

PR #520, opened 2025-01-02
Return proofs for wallet mint function

davidcaseria:
### Description

As it is currently written, it is impossible to see which proofs are returned for a mint since the total amount is returned. However, with the `ProofsMethods` trait, it is easy to calculate the total amount for the mint. This change returns the minted `Proofs` to make it possible to see which proofs are returned while not significantly decreasing the API's ergonomics. 

-----

### Notes to the reviewers

This is a breaking change to the API.

-----

#### CHANGED

`Wallet::mint` now returns `Proofs`.

#### FIXED

----

### Checklist

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -7,14 +7,13 @@ services:
     ports:
       - "8085:8085"
     environment:
-      - CDK_MINTD_URL=http://192.168.1.188:8085
+      - CDK_MINTD_URL=https://example.com
       - CDK_MINTD_LN_BACKEND=fakewallet
       - CDK_MINTD_LISTEN_HOST=0.0.0.0
       - CDK_MINTD_LISTEN_PORT=8085
       - CDK_MINTD_MNEMONIC=
       - CDK_MINTD_DATABASE=redb
       - CDK_MINTD_CACHE_BACKEND=memory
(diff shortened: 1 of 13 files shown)


docker-compose.yaml
```diff
@@ -7,13 +7,14 @@ services:
...
     environment:
-      - CDK_MINTD_URL=https://example.com
+      - CDK_MINTD_URL=http://192.168.1.188:8085
       - CDK_MINTD_LN_BACKEND=fakewallet
       - CDK_MINTD_LISTEN_HOST=0.0.0.0
       - CDK_MINTD_LISTEN_PORT=8085
       - CDK_MINTD_MNEMONIC=
       - CDK_MINTD_DATABASE=redb
       - CDK_MINTD_CACHE_BACKEND=memory
+      - CDK_MINTD_MNEMONIC=essence thing digital again unaware famous case usage friend claw picnic donor
```
thesimplekid, line 17: Imagine you just forgot to remove this in testing. I don't want a seed phrase in the as we've seen in other projects some people will just run it with that.

thesimplekid:
I think this make sense provides a bit more utility for the user.

thesimplekid (approved):
ACK 6c96e14bb58628186990d5fa8ee490876e45d3c8

PR #536, opened 2025-01-14
[NUT-15] LND Support for MPP Payments

lollerfirst:
### Description
Query the high probability routes for the public key in the invoice. Attempt the first route: if the failure code is 15 (TEMPORARY_CHANNEL_FAILURE), attempt the next route.

----

#### CHANGED
* `pay_invoice`: forks execution based on the presence of a partial_amount.

----

### Checklist

* [] I ran `just final-check` before committing


crates/cdk-lnd/src/lib.rs
```diff
@@ -204,47 +205,146 @@ impl MintLightning for Lnd {
...
+                    .lock()
+                    .await
+                    .router()
+                    .send_to_route_v2(fedimint_tonic_lnd::routerrpc::SendToRouteRequest {
+                        payment_hash: payment_hash.to_byte_array().to_vec(),
+                        route: Some(route),
+                        ..Default::default()
+                    })
+                    .await
+                    .map_err(|_| Error::PaymentFailed)?
```
thesimplekid, line 268: Wont this return here due to the use of `?` so if the payment fails the next route cannot be tried?

crates/cdk-lnd/src/lib.rs
```diff
@@ -204,47 +205,146 @@ impl MintLightning for Lnd {
...
+
+            // For each route:
+            // update its MPP record,
+            // attempt it and check the result
+            for mut route in routes_response.routes.into_iter() {
+                let last_hop = route.hops.last_mut().ok_or(Error::MissingLastHop)?;
+                let mpp_record = last_hop
+                    .mpp_record
+                    .as_mut()
+                    .ok_or(Error::MissingMppRecord)?;
```
thesimplekid, line 252: Do we want to return the function here or continue to next route?

crates/cdk-lnd/src/lib.rs
```diff
@@ -204,47 +205,146 @@ impl MintLightning for Lnd {
...
-            .await
-            .map_err(|_| Error::PaymentFailed)?
-            .into_inner();
-
-        let total_amount = payment_response
-            .payment_route
-            .map_or(0, |route| route.total_amt_msat / MSAT_IN_SAT as i64)
-            as u64;
+        // Detect partial payments
+        if let Some(part_amt) = partial_amount {
```
thesimplekid, line 209: NIT: Since we care about both paths the `Some` and `None` we should use a `match` and not `if let` to be consistent with code style https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#if-let

crates/cdk-lnd/src/lib.rs
lollerfirst: I think we want to return in this case. Clearly something is wrong with the response.

crates/cdk-lnd/src/lib.rs
lollerfirst: I think we only want to try more routes if we get a specific error 15 (TEMPORARY_CHANNEL_FAILURE). If we get any kind of error then something else is up.

Issue #542, opened 2025-01-19
Mint Managment RCP


thesimplekid:
- [x] Update mint info
    - [x] MOTD
    - [x] Short Desc
    - [x] Long Desc
    - [x] Contact Info
    - [x] Name
    - [x] Icon Url
    - [x] Urls
- [x] Update Mint Settings 
    - [x]  Enable/Disable
    - [x] Update Min
    - [x] Update Max
- [x] Update Melt Settings
    - [x]  Enable/Disable
    - [x] Update Min
    - [x] Update Max
- [x] Update quote ttl

thesimplekid:
#543

PR #543, opened 2025-01-19
Managment rpc

thesimplekid:
### Description

-----
closes #456 
### Notes to the reviewers

A mint management rpc can be started by enabling it in the mintd config. There is a gprpc client cli that can be run by  running `cargo r --bin cdk-mint-cli` to update the mint settings. 

If the rpc server is enabled the mint will only read settings from the config file the first time after that the rpc must be used and changes to the mint info in the config file will not be used.

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,8 @@ jobs:
             --bin cdk-mintd --no-default-features --features swagger,
             --bin cdk-mintd --no-default-features --features redis,
             --bin cdk-mintd --no-default-features --features "redis swagger",
+            --bin cdk-mintd --no-default-features --features management-rpc,
+            --bin cdk-mint-cli,
           ]
     steps:
       - name: checkout
@@ -213,6 +215,7 @@ jobs:
             -p cdk-phoenixd,
             -p cdk-fake-wallet,
             -p cdk-cln,
+            -p cdk-mint-rpc,
           ]
     steps:
       - name: checkout
--- a/crates/cdk-mint-rpc/Cargo.toml
+++ b/crates/cdk-mint-rpc/Cargo.toml
@@ -1,0 +1,41 @@
+[package]
+name = "cdk-mint-rpc"
+version = "0.6.0"
+edition = "2021"
+authors = ["CDK Developers"]
+description = "CDK mintd mint managment RPC client and server"
+license = "MIT"
+homepage = "https://github.com/cashubtc/cdk"
+repository = "https://github.com/cashubtc/cdk.git"
+rust-version = "1.63.0"                            # MSRV
+
+[[bin]]
+name = "cdk-mint-cli"
+path = "src/bin/mint_rpc_cli.rs"
+
+[dependencies]
+anyhow = "1"
+cdk = { path = "../cdk", version = "0.6.0", default-features = false, features = [
+    "mint",
+] }
+clap = { version = "~4.0.32", features = ["derive"] }
+tonic = { version = "0.9", features = [
+    "channel",
+    "tls",
+    "tls-webpki-roots",
+] }
+tracing = { version = "0.1", default-features = false, features = [
+    "attributes",
+    "log",
+] }
+tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
+tokio = { version = "1", default-features = false }
+serde_json = "1"
+serde = { version = "1", features = ["derive"] }
+thiserror = "1"
+prost = "0.11.0"
+home = "0.5.5"
+
+
+[build-dependencies]
+tonic-build = "0.9"
--- a/crates/cdk-mint-rpc/build.rs
+++ b/crates/cdk-mint-rpc/build.rs
@@ -1,0 +1,5 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo:rerun-if-changed=src/proto/cdk-mint-rpc.proto");
+    tonic_build::compile_protos("src/proto/cdk-mint-rpc.proto")?;
+    Ok(())
+}
--- a/crates/cdk-mint-rpc/src/bin/mint_rpc_cli.rs
+++ b/crates/cdk-mint-rpc/src/bin/mint_rpc_cli.rs
@@ -1,0 +1,196 @@
+use std::path::PathBuf;
+
+use anyhow::{anyhow, Result};
+use cdk_mint_rpc::cdk_mint_client::CdkMintClient;
+use cdk_mint_rpc::mint_rpc_cli::subcommands;
+use cdk_mint_rpc::GetInfoRequest;
+use clap::{Parser, Subcommand};
+use tonic::transport::{Certificate, Channel, ClientTlsConfig, Identity};
+use tonic::Request;
+use tracing::Level;
+use tracing_subscriber::EnvFilter;
+
+const DEFAULT_WORK_DIR: &str = ".cdk-mint-rpc-cli";
+
+#[derive(Parser)]
+#[command(version, about, long_about = None)]
+struct Cli {
+    /// Address of RPC server
+    #[arg(short, long, default_value = "http://127.0.0.1:8086")]
+    addr: String,
+
+    /// Logging level
+    #[arg(short, long, default_value = "debug")]
+    log_level: Level,
+
+    /// Path to working dir
+    #[arg(short, long)]
+    work_dir: Option<PathBuf>,
+
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand)]
+enum Commands {
+    /// Get info
+    GetInfo,
+    /// Update motd
+    UpdateMotd(subcommands::UpdateMotdCommand),
+    /// Update short description
+    UpdateShortDescription(subcommands::UpdateShortDescriptionCommand),
+    /// Update long description
+    UpdateLongDescription(subcommands::UpdateLongDescriptionCommand),
+    /// Update name
+    UpdateName(subcommands::UpdateNameCommand),
+    /// Update icon url
+    UpdateIconUrl(subcommands::UpdateIconUrlCommand),
+    /// Add Url
+    AddUrl(subcommands::AddUrlCommand),
+    /// Remove Url
+    RemoveUrl(subcommands::RemoveUrlCommand),
+    /// Add contact
+    AddContact(subcommands::AddContactCommand),
+    /// Remove contact
+    RemoveContact(subcommands::RemoveContactCommand),
+    /// Update nut04
+    UpdateNut04(subcommands::UpdateNut04Command),
+    /// Update nut05
+    UpdateNut05(subcommands::UpdateNut05Command),
+    /// Update quote ttl
+    UpdateQuoteTtl(subcommands::UpdateQuoteTtlCommand),
+    /// Update Nut04 quote
+    UpdateNut04QuoteState(subcommands::UpdateNut04QuoteCommand),
+    /// Rotate next keyset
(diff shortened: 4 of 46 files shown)


.github/workflows/ci.yml
```diff
@@ -82,6 +82,7 @@ jobs:
             --bin cdk-mintd --no-default-features --features swagger,
             --bin cdk-mintd --no-default-features --features redis,
             --bin cdk-mintd --no-default-features --features "redis swagger",
+            --bin mint_rpc_cli,
```
thesimplekid, line 85: add `mint_rpc_cli` lib as well as the bin

crates/cdk-mint-rpc/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+# tonic = { version = ">=0.9, <0.12", features = [
+#     "channel",
+#     "tls",
+#     "tls-webpki-roots",
+# ] }
+tonic = { version = "0.9", features = [
+    "channel",
+    "tls",
+    "tls-webpki-roots",
+] }
```
thesimplekid, line 25: We don't want to enforce such an old version of `tonic`, `prost` and `tonic-build` so we should be able to support the range to enable support of msrv and newer rust version.

thesimplekid:
cc @crodas @davidcaseria

crates/cdk-mintd/src/main.rs
```diff
@@ -359,12 +361,18 @@ async fn main() -> anyhow::Result<()> {
...
+    let mint_clone = Arc::clone(&mint);
     tokio::spawn({
         let shutdown = Arc::clone(&shutdown);
-        async move { mint.wait_for_paid_invoices(shutdown).await }
+        async move { mint_clone.wait_for_paid_invoices(shutdown).await }
     });
 
+    let _rpc_shutdown = Arc::new(Notify::new());
+
+    let mint_rpc = MintRPCServer::new("127.0.0.1", 8086, mint)?;
```
ok300, line 372: Would it make sense to place it behind a feature flag, like the swagger UI?

thesimplekid: resolved in 4ce1337a9397a6774b39a43fa779ee9d4bd8b3c2

ok300:
Fixes #542

crates/cdk-mint-rpc/src/bin/mint_rpc_cli.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    // Parse input
+    tracing_subscriber::fmt().with_env_filter(env_filter).init();
+
+    let cli = Cli::parse();
+
+    let work_dir = match &args.work_dir {
+        Some(work_dir) => work_dir.clone(),
+        None => {
+            let home_dir = home::home_dir().unwrap();
```
ok300, line 84: `unwrap()` -> `?`

crates/cdk-mint-rpc/src/bin/mint_rpc_cli.rs
```diff
@@ -0,0 +1,195 @@
...
+
+            home_dir.join(DEFAULT_WORK_DIR)
+        }
+    };
+
+    std::fs::create_dir_all(&work_dir)?;
+    tracing::debug!("Using work dir: {}", work_dir.display());
+
+    let channel = if work_dir.join("tls").is_dir() {
+        // TLS directory exists, configure TLS
```
ok300, line 94:
```suggestion
        tracing::info!("Found TLS directory, configuring TLS");
```

A log statement would be more helpful IMO.

crates/cdk-mint-rpc/src/bin/mint_rpc_cli.rs
```diff
@@ -0,0 +1,195 @@
...
+        let tls = ClientTlsConfig::new()
+            .ca_certificate(server_root_ca_cert)
+            .identity(client_identity);
+
+        Channel::from_shared(cli.addr.to_string())?
+            .tls_config(tls)?
+            .connect()
+            .await?
+    } else {
+        // No TLS directory, skip TLS configuration
```
ok300, line 109:
```suggestion
        tracing::warn!("No TLS directory found, skipping TLS configuration");
```

crates/cdk-mint-rpc/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+pub mod proto;
+
+pub mod mint_rpc_cli;
+
+pub use proto::*;
```
ok300, line 5:
```suggestion
pub mod mint_rpc_cli;
pub mod proto;
pub use proto::*;
```
NIT: alphabetic sorting

crates/cdk-mint-rpc/src/proto/cdk-mint-rpc.proto
```diff
@@ -0,0 +1,115 @@
...
+    rpc UpdateIconUrl(UpdateIconUrlRequest) returns (UpdateResponse) {}
+    rpc UpdateName(UpdateNameRequest) returns (UpdateResponse) {}
+    rpc AddUrl(UpdateUrlRequest) returns (UpdateResponse) {}
+    rpc RemoveUrl(UpdateUrlRequest) returns (UpdateResponse) {}
+    rpc AddContact(UpdateContactRequest) returns (UpdateResponse) {}
+    rpc RemoveContact(UpdateContactRequest) returns (UpdateResponse) {}
+    rpc UpdateNut04(UpdateNut04Request) returns (UpdateResponse) {}
+    rpc UpdateNut05(UpdateNut05Request) returns (UpdateResponse) {}
+    rpc UpdateQuoteTtl(UpdateQuoteTtlRequest) returns (UpdateResponse) {}
+    rpc UpdateNut04Quote(UpdateNut04QuoteRequest) returns (UpdateNut04QuoteRequest) {}
```
ok300, line 19: Is this for manual updating of quote states? If so, isn't this dangerous to allow (i.e. could interfere with the automatic quote state mgmt)?

crates/cdk-mint-rpc/src/proto/cdk-mint-rpc.proto
```diff
@@ -0,0 +1,115 @@
...
+
+message GetInfoRequest {
+}
+
+message ContactInfo {
+    string method = 1;
+    string info = 2;
+}
+
+message GetInfoResponse {
```
ok300, line 31: Maybe this should include the current settings, at least for `nut04/05`? Alternatively, maybe `UpdateNut04/05` should return the current settings?

There are methods to update them, but there is no way to get them. Since `update` will merge the settings, it would be useful to have a way to see the current ones.

crates/cdk-mint-rpc/src/proto/server.rs
```diff
@@ -0,0 +1,586 @@
...
+                let tls_config = ServerTlsConfig::new()
+                    .identity(server_identity)
+                    .client_ca_root(client_ca_cert);
+
+                Server::builder()
+                    .tls_config(tls_config)?
+                    .add_service(CdkMintServer::new(self.clone()))
+            }
+            None => {
+                tracing::info!("No valid TLS configuration found, starting insecure server");
```
ok300, line 80:
```suggestion
                tracing::warn!("No valid TLS configuration found, starting insecure server");
```
IMO `warn` is better for the insecure branch.

crates/cdk-mintd/Cargo.toml
```diff
@@ -6,9 +6,14 @@ authors = ["CDK Developers"]
 license = "MIT"
 homepage = "https://github.com/cashubtc/cdk"
 repository = "https://github.com/cashubtc/cdk.git"
-rust-version = "1.63.0"                            # MSRV
```
ok300, line 9: Removed by mistake?

crates/cdk-mintd/Cargo.toml
```diff
@@ -6,9 +6,14 @@ authors = ["CDK Developers"]
...
 description = "CDK mint binary"
 
+[features]
+default = ["management-rpc"]
```
ok300, line 12: Maybe it's better to have this disabled by default?

Keeps the attack surface smaller for the default setup. Mint operators can still enable it if they need it.

crates/cdk-mint-rpc/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "cdk-mint-rpc"
+version = "0.1.0"
```
ok300, line 3: All other CDK crates have a "unified", single version. Since there are so many inter-dependencies, maybe it's good to stick with that model and use the CDK version here (0.6.0)?

crates/cdk-mintd/src/main.rs
```diff
@@ -358,12 +360,51 @@ async fn main() -> anyhow::Result<()> {
...
+
+    if rpc_enabled {
+        if mint.mint_info().await.is_err() {
+            tracing::info!("Mint info not set on mint, setting.");
+            mint.set_mint_info(mint_builder.mint_info).await?;
+        } else {
+            tracing::info!("Mint info already set, not using config file settings.");
+        }
+    } else {
+        tracing::info!("RPC not enabled using mint info from config.");
```
ok300, line 404:
```suggestion
        tracing::info!("RPC not enabled, using mint info from config.");
```

crates/cdk-mintd/Cargo.toml
thesimplekid: mintd doesn't have an msrv or at least not 1.63 as it pulls in lnd that had a higer msrv and redb

crates/cdk-mintd/Cargo.toml
thesimplekid: I think its better to have in the default binary as I think most operator will want it, though it can be disabled at run time if they do not want to so there there is little more expose there from the excess code. The rpc is not started by default at runtime that does default to not start. 

Related if you are worried about a smaller attack surface via excess cdk code mintd should not be used as it brings in both dbs, and all the ln backends and then chooses these at runtime. I don't think this really increase the attack surface as its code that is not run, but does increase the binary.

crates/cdk-mint-rpc/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tonic_build::compile_protos("src/proto/cdk-mint-rpc.proto")?;
```
ok300, line 2:
```suggestion
    println!("cargo:rerun-if-changed=src/proto/cdk-mint-rpc.proto");
    tonic_build::compile_protos("src/proto/cdk-mint-rpc.proto")?;
```

Helps avoid unnecessary recompilations. With this, `cargo` will only recompile if the proto changed.

crates/cdk-mint-rpc/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "cdk-mint-rpc"
+version = "0.6.0"
+edition = "2021"
+authors = ["CDK Developers"]
+description = "CDK mintd mint managment rpc client"
```
ok300, line 6:
```suggestion
description = "CDK mintd mint managment RPC client and server"
```

crates/cdk-mint-rpc/build.rs
thesimplekid: Nice thank you

crates/cdk-mint-rpc/src/proto/cdk-mint-rpc.proto
thesimplekid: They are shows by curling the info endpoint. But yes more convenient to have here as well. we can add it.

PR #550, opened 2025-01-21
Introduce `cdk-database`

crodas:
### Description

This is a crate that makes it easier to create a database instance based on an enum that can be constructed from a CLI or an env variable.

This PR has been abstracted from #509, which makit es easier to create binaries that share the same settings / databases as the mintd web-server.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
I don't think this adds much benefit but it does add more overhead for us with an extra crate. Projects that use cdk as a lib will only want one database so they can just import the crate for the one they want to use (sqlite/redb). Where something like cdk where we want to support both can just import both and do his internally.

PR #575, opened 2025-02-03
Bump dependencies

ok300:
### Description

This PR bumps a few core dependencies to the latest (MSRV-compatible) version.

-----

### Checklist

### DIFF
--- a/crates/cashu/Cargo.toml
+++ b/crates/cashu/Cargo.toml
@@ -36,5 +36,5 @@ serde_with = "3"
 instant = { version = "0.1", features = ["wasm-bindgen", "inaccurate"] }
 
 [dev-dependencies]
-bip39 = "2.0"
+bip39 = "2.1"
 uuid = { version = "1", features = ["v4", "serde"] }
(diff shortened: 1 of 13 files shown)


crates/cashu/Cargo.toml
```diff
@@ -14,7 +14,7 @@ bench = []
 
 [dependencies]
 uuid = { version = "1", features = ["v4", "serde"], optional = true }
-bitcoin = { version = "0.32.2", features = [
+bitcoin = { version = "0.32.5", features = [
```
thesimplekid, line 17: cargo will automatically pull in the latest patch version so we do not want to raise this as it should be the minimum version we support.  This is true for a few of the ones updated, just going to note it here.

crates/cdk/Cargo.toml
```diff
@@ -55,16 +55,16 @@ thiserror = "1"
...
-url = "2.3"
+url = "2.5"
 utoipa = { version = "4", optional = true }
 uuid = { version = "1", features = ["v4", "serde"] }
 
 # -Z minimal-versions
-sync_wrapper = "0.1.2"
-bech32 = "0.9.1"
+sync_wrapper = "1"
+bech32 = "0.11"
```
thesimplekid, line 64: These are not direct deps of ours but transitive dependencies that do not have the minimal version defined correctly so they need to be defined here. So we cant change to a new major or minor version as we want to set the correct major/minor version for the dep. 

I haven't run the minimal check in a bit so I should redo it and add to CI and maybe we will be able to update or remove these if your direct dep has been updated. https://github.com/cashubtc/cdk/issues/505

thesimplekid (changes requested):

ok300:
Alright then I'll close this.

PR #581, opened 2025-02-05
Use Wallet TransactionDatabase; Create WalletBuilder

davidcaseria:
### Description

Adds a `TransactionDatabase` trait for wallet databases to implement to store a history of wallet transactions.

-----

### Notes to the reviewers

-----

#### CHANGED
- Renamed `WalletDatabase` to `WalletProofDatabase`.
- Wallet functions `mint`, `melt`, `send`, and `receive` now accept respective options struct.

#### ADDED
- `WalletTransactionDatabase` and `Transaction` (handled by wallet functions)
- `WalletBuilder`

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-cli/src/main.rs
+++ b/crates/cdk-cli/src/main.rs
@@ -6,10 +6,10 @@ use std::sync::Arc;
 use anyhow::{bail, Result};
 use bip39::rand::{thread_rng, Rng};
 use bip39::Mnemonic;
-use cdk::cdk_database;
-use cdk::cdk_database::WalletDatabase;
+use cdk::cdk_database::WalletProofDatabase;
+use cdk::nuts::CurrencyUnit;
 use cdk::wallet::client::HttpClient;
-use cdk::wallet::{MultiMintWallet, Wallet};
+use cdk::wallet::{MultiMintWallet, Wallet, WalletBuilder};
 use cdk_redb::WalletRedbDatabase;
 use cdk_sqlite::WalletSqliteDatabase;
 use clap::{Parser, Subcommand};
@@ -102,24 +102,6 @@ async fn main() -> Result<()> {
 
     fs::create_dir_all(&work_dir)?;
 
-    let localstore: Arc<dyn WalletDatabase<Err = cdk_database::Error> + Send + Sync> =
-        match args.engine.as_str() {
-            "sqlite" => {
-                let sql_path = work_dir.join("cdk-cli.sqlite");
-                let sql = WalletSqliteDatabase::new(&sql_path).await?;
-
-                sql.migrate().await;
-
-                Arc::new(sql)
-            }
-            "redb" => {
-                let redb_path = work_dir.join("cdk-cli.redb");
-
-                Arc::new(WalletRedbDatabase::new(&redb_path)?)
-            }
-            _ => bail!("Unknown DB engine"),
-        };
-
     let seed_path = work_dir.join("seed");
 
     let mnemonic = match fs::metadata(seed_path.clone()) {
@@ -140,24 +122,40 @@ async fn main() -> Result<()> {
         }
     };
 
+    let mut builder = WalletBuilder::new(mnemonic.to_seed_normalized("").to_vec());
+
+    let mints = match args.engine.as_str() {
+        "sqlite" => {
+            let sql_path = work_dir.join("cdk-cli.sqlite");
+            let sql = WalletSqliteDatabase::new(&sql_path).await?;
+
+            sql.migrate().await;
+
+            let db = Arc::new(sql);
+            builder = builder.proof_db(db.clone()).transaction_db(db.clone());
+            db.get_mints().await?
+        }
+        "redb" => {
+            let redb_path = work_dir.join("cdk-cli.redb");
+
+            let db = Arc::new(WalletRedbDatabase::new(&redb_path)?);
+            builder = builder.proof_db(db.clone()).transaction_db(db.clone());
+            db.get_mints().await?
+        }
+        _ => bail!("Unknown DB engine"),
+    };
+
     let mut wallets: Vec<Wallet> = Vec::new();
 
-    let mints = localstore.get_mints().await?;
-
     for (mint_url, _) in mints {
-        let mut wallet = Wallet::new(
-            &mint_url.to_string(),
-            cdk::nuts::CurrencyUnit::Sat,
-            localstore.clone(),
-            &mnemonic.to_seed_normalized(""),
-            None,
-        )?;
-        if let Some(proxy_url) = args.proxy.as_ref() {
-            let http_client = HttpClient::with_proxy(mint_url, proxy_url.clone(), None, true)?;
-            wallet.set_client(http_client);
-        }
+        let mut builder = builder.clone();
 
-        wallets.push(wallet);
+        if let Some(proxy_url) = args.proxy.as_ref() {
+            let http_client =
+                HttpClient::with_proxy(mint_url.clone(), proxy_url.clone(), None, true)?;
+            builder = builder.client(Arc::new(http_client));
+        }
+        wallets.push(builder.build(mint_url, CurrencyUnit::Sat)?);
     }
 
     let multi_mint_wallet = MultiMintWallet::new(wallets);
@@ -171,13 +169,7 @@ async fn main() -> Result<()> {
             sub_commands::melt::pay(&multi_mint_wallet, sub_command_args).await
         }
         Commands::Receive(sub_command_args) => {
-            sub_commands::receive::receive(
-                &multi_mint_wallet,
-                localstore,
-                &mnemonic.to_seed_normalized(""),
-                sub_command_args,
-            )
-            .await
+            sub_commands::receive::receive(&multi_mint_wallet, builder, sub_command_args).await
         }
         Commands::Send(sub_command_args) => {
             sub_commands::send::send(&multi_mint_wallet, sub_command_args).await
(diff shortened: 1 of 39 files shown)


crates/cdk-common/src/database/wallet.rs
```diff
@@ -14,107 +19,212 @@ use crate::nuts::{
...
+pub trait Database: ProofDatabase + TransactionDatabase {}
+
+/// Wallet Proofs Database trait
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-pub trait Database: Debug {
-    /// Wallet Database Error
-    type Err: Into<Error> + From<Error>;
-
+pub trait ProofDatabase: Debug {
```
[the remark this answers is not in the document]
davidcaseria: Can you explain more how this would work?

crates/cdk-common/src/database/wallet.rs
```diff
@@ -14,107 +19,212 @@ use crate::nuts::{
...
-    async fn get_mint_keysets(
-        &self,
-        mint_url: MintUrl,
-    ) -> Result<Option<Vec<KeySetInfo>>, Self::Err>;
+    async fn get_mint_keysets(&self, mint_url: MintUrl) -> Result<Option<Vec<KeySetInfo>>, Error>;
     /// Get mint keyset by id
-    async fn get_keyset_by_id(&self, keyset_id: &Id) -> Result<Option<KeySetInfo>, Self::Err>;
+    async fn get_keyset_by_id(&self, keyset_id: &Id) -> Result<Option<KeySetInfo>, Error>;
 
     /// Add mint quote to storage
```
davidcaseria, line 54: Should the quotes be moved to the `TransactionDatabase` trait?

crates/cdk-common/src/database/wallet.rs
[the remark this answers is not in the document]
thesimplekid: > Since this is a new trait in practica? Can we introduce a Writable trait as well?

I think this is a good idea but maybe should be a separate pr then adding the transactions. Just so we're not changing major functionality and adding features in the same one?

crates/cdk-common/src/database/wallet.rs
```diff
@@ -14,107 +19,212 @@ use crate::nuts::{
...
+    ) -> Result<Vec<ProofInfo>, Error>;
 
     /// Increment Keyset counter
-    async fn increment_keyset_counter(&self, keyset_id: &Id, count: u32) -> Result<(), Self::Err>;
+    async fn increment_keyset_counter(&self, keyset_id: &Id, count: u32) -> Result<(), Error>;
     /// Get current Keyset counter
-    async fn get_keyset_counter(&self, keyset_id: &Id) -> Result<Option<u32>, Self::Err>;
+    async fn get_keyset_counter(&self, keyset_id: &Id) -> Result<Option<u32>, Error>;
 
     /// Get when nostr key was last checked
```
davidcaseria, line 106: Are these nostr methods necessary? Should we support this some other way?

crates/cdk-common/src/database/wallet.rs
```diff
@@ -14,107 +19,212 @@ use crate::nuts::{
...
+        }
+        let hash = sha256::Hash::from_engine(hasher);
+        Self(hash.to_byte_array())
+    }
+
+    /// Get inner value
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        &self.0
+    }
+}
```
thesimplekid, line 201: I know we've brought it up before but just to flag again. Do we think there is any value in adding an it like this to the spec? I think it could be used for SIGALL on melts.

crates/cdk-common/src/database/wallet.rs
thesimplekid: https://github.com/cashubtc/cdk/blob/dcb9ab3140aa14b084593f5e2363517a691b2b82/crates/cdk-cli/src/sub_commands/receive.rs#L119-L121

Theyre used in the cli, but no they shouldn't be here. CLI or an application should have their own db for stuff like this

crates/cdk-common/src/database/wallet.rs
thesimplekid: Yeah I think it fits a little better there.

crates/cdk-common/src/database/wallet.rs
davidcaseria: I think that's a good idea. Should I add it to NUT-00? cc: @callebtc

crates/cdk-common/src/database/wallet.rs
```diff
@@ -14,107 +19,250 @@ use crate::nuts::{
...
+        Some(self.cmp(other))
+    }
+}
+
+/// Transaction direction
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum TransactionDirection {
+    Incoming,
+    Outgoing,
+}
```
ok300, line 194: How about in-place swaps? Meaning not as part of send or receive, but as a standalone wallet operation? Wallets may swap to consolidate proofs, or to change their distribution (more for this amount, less of that amount).

Since swaps can have a fee, they can affect the wallet balance. I think this should be accounted for in the tx list.

crates/cdk-common/src/database/wallet.rs
```diff
@@ -14,107 +19,250 @@ use crate::nuts::{
...
+
+impl TransactionId {
+    /// Create new [`TransactionId`]
+    pub fn new(ys: Vec<PublicKey>) -> Self {
+        let mut ys = ys;
+        ys.sort();
+        let mut hasher = sha256::Hash::engine();
+        for y in ys {
+            hasher.input(&y.to_bytes());
+        }
```
ok300, line 230: I'm not sure if only the the `Ys` should be part of this.

What if a wallet tries a tx, but it fails, then tries the same tx again with the same inputs, which succeeds? Intuitively, that should be shown as 2 txs, one failed one succeeded.

Maybe in addition to `Ys` it should hash the timestamp too?

davidcaseria: What would constitute failed transactions? I can't think of any failure reasons to persist a failed transaction.

crates/cdk-common/src/database/wallet.rs
davidcaseria: On second thought, would that scenario be an "outgoing" transaction since the `amount` was deducted from the balance? The `TransactionDirection` enum is intended to keep the `amount` field unsigned.

I will add a `fee` field to the `Transaction` struct. Internal swaps would be identified by `direction == TransactiongDirection::Outgoing && amount == fee`.

ok300: That sounds like an elegant solution.

Maybe mark it as "self-transfer" somehow (a special `metadata` field?) but otherwise outgoing tx makes sense.

crates/cdk-common/src/database/wallet.rs
ok300: > What would constitute failed transactions?

A bunch of things can lead to a tx to fail:

- send
  - if sender needs to swap, but the mint is unreachable or otherwise errors out, so no swap can occur
- mint / melt
  - if the quote reply from the mint is invalid (for example `expiry` is in the past)
  - if the quote remains `UNPAID` after the `expiry` timestamp
  - if the quote calls are fine, but calls to `/mint` or `/melt` fail (for any reason: auth, timeout, invalid response, etc)
  - if the amount to mint or melt is outside the `min/max` range defined by the mint
- receive
  - if someone received a token, but swapping the proofs failed (maybe "failed" is not the right state for such a tx)
- all types of swap (as part of send / receive, or standalone)
  - if the available inputs are not enough to cover the swap fees
  - if the available inputs match exactly the needed fees, so there can be no outputs

I'm not saying the PR should account for all possible errors. But I do think `get_transactions` should be able to reflect attempted but failed txs, with a summary for failure reason.

> I can't think of any (..) reasons to persist a failed transaction

Most (all?) mobile LN wallets show failed payments. Its a confirmation that the library actually tried to send / receive, but something went wrong. If `get_transactions` doesn't include the failed ones, app developers have to improvise and stub the failed txs, if they want to keep a good UX (on par with mobile LN wallets for example).

Another reason to store it is when contacting the support of the mint. If the app has failed txs documented in `get_transactions`, the user can get the details (quote ID, timestamp, amount, etc) that can help the mint troubleshoot the issue. For mint / melt, the tx ID could be the quote ID.

davidcaseria: I don't think I've seen an app persist a transaction for any of these cases. Usually, they are alerts at the time of the user action. I don't see any reason to keep them around forever by storing them as a transaction (especially transient network issues).

Quotes are persisted separately from transactions and can be queried independently. I don't think it makes sense to put quote data in transactions.

ok300: Of course a line has to be drawn to separate when does a transaction begin. Errors before that shouldn't reflect on the tx state, but if something goes wrong once a tx is started or in progress (for example, once the quote is pending), it seems reasonable to assume that will be reflected in a failed tx state.

On a more practical level, I don't think its possible to properly account for the proofs in the cashu wallet, without tracking for each of them, how they got in and if / how they got out. Dealing with "how they got in / out" opens the question for tx state, because the tx is often a multi-step process. For example, if a token was received, but no swap could be done, would that tx not appear in the history? What if it's a temporary mint connectivity issue and the wallet app (or CDK) retries to swap them later and it succeeds, would a history entry suddenly appear with a timestamp in the past?

IMO there is no way around dealing with failed transactions. If they're left out now, I think they'll have to be added later on.

davidcaseria: The transactions are more like add-on metadata than core components of the wallet. As we had decided earlier, the wallet is still basically a "bag of proofs." Therefore, failed transactions shouldn't need to be tracked.

The PR is now updated to add transactions after a successful mint, melt, send, or receive.

crates/cdk-cli/src/main.rs
```diff
@@ -140,24 +122,40 @@ async fn main() -> Result<()> {
...
+            let redb_path = work_dir.join("cdk-cli.redb");
+
+            let db = Arc::new(WalletRedbDatabase::new(&redb_path)?);
+            builder = builder.proof_db(db.clone()).transaction_db(db.clone());
+            db.get_mints().await?
+        }
+        _ => bail!("Unknown DB engine"),
+    };
+
+    let mut wallets: Vec<Wallet> = Vec::new();
```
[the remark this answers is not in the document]
davidcaseria: That seems like a good idea for another PR.

davidcaseria:
Closing in favor of another, more-scoped PR.

PR #596, opened 2025-02-11
Prepared Send

davidcaseria:
### Description

Prepare sends by selecting proofs and returning a fee amount before finalizing a send into a token.

-----

### Notes to the reviewers

-----

#### CHANGED
- Unifies and optimizes the proof selection algorithm to use `Wallet::select_proofs`.
- `Wallet::send` now requires a `PreparedSend`.
- `WalletDatabase` proof state update functions have been consolidated into `update_proofs_state`.

#### ADDED
- Sends should be initiated by calling `Wallet::prepare_send`.
- A `SendOptions` struct controls optional functionality for sends.
- Allow `Amount` splitting to target a fee rate amount.
- Utility functions for `Proofs`.
- Utility functions for `SendKind`.
- Completed checked arithmetic operations for `Amount` (i.e., `checked_mul` and `checked_div`). 

#### FIXED

----

### Checklist


davidcaseria:
@thesimplekid two areas I'm looking for feedback:
1. How should we support CDK users who want to bring their own `PreparedSend` logic? I made `PreparedSend` a "private" structure so that users can't modify it, causing issues with our logic.
2. How should we support memos? I have it in `SendOptions` but I think I should move it to be a parameter of the `send` function, possibly with a flag to include it in the `Token` or keep it internal with the eventual `TransactionsDatabase`.

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +29,351 @@ impl Wallet {
...
-                    true => amount + self.get_proofs_fee(&selected_proofs).await?,
-                    false => amount,
-                };
+    async fn internal_prepare_send(
+        &self,
+        amount: Amount,
+        opts: SendOptions,
+        proofs: Proofs,
+        active_keyset_id: Id,
+        force_swap: bool,
```
thesimplekid, line 136: What is the intention of `force_swap` is it just for the case where a wallet may want to rotate proofs for example from an inactive to active keyset?

crates/cdk/src/wallet/proofs.rs
```diff
@@ -153,125 +158,375 @@ impl Wallet {
...
+                // All proofs with the current amount are selected
+                if n == 0 {
+                    found = true;
+                    break;
+                }
+
+                // Try to find double the number of the next amount
+                n *= 2;
+                target_amount = curr_amount;
+            }
```
thesimplekid, line 281: Does this logic only work for powers of two? This is more of a general question and not blocking as the currently cdk mostly assumes power of 2 keysets however this isn't actually required by the nuts so we should move away from it.

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +29,351 @@ impl Wallet {
...
+            return Err(Error::InsufficientFunds);
+        }
 
-                self.swap_from_unspent(amount, conditions, include_fees)
-                    .await?
-            }
+        // Update proofs state to pending spent
+        self.localstore
+            .update_proofs_state(proofs_to_send.ys()?, State::PendingSpent)
+            .await?;
```
thesimplekid, line 249: I may have missed it but was the proof set to `Reserved` in `prepare_send`. Am I correct that the expectation is the once a proof is selected as part of a prepared send it is reserved and cannot be used in other operations unless explicitly unreserved?  This way `prepare_send` can be called multiple times without conflicting and we do not have to block other operations. 

Can we add a doc comment to the send/prepare_send fn explaining this expectation.

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +29,351 @@ impl Wallet {
...
+
+        self.localstore
+            .update_proofs_state(send.proofs().ys()?, State::Unspent)
+            .await?;
+
+        Ok(())
+    }
+}
+
+/// Prepared send
```
thesimplekid, line 276: Can we just add a few more lines explaining how the `proofs_to_swap` vs `proofs_to_send` should be uses and what they are?

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +29,351 @@ impl Wallet {
...
+pub struct SendOptions {
+    /// Memo
+    pub memo: Option<SendMemo>,
+    /// Spending conditions
+    pub conditions: Option<SpendingConditions>,
+    /// Amount split target
+    pub amount_split_target: SplitTarget,
+    /// Send kind
+    pub send_kind: SendKind,
+    /// Include fee
```
thesimplekid, line 368:
```suggestion
    /// Include fee
    ///
    /// When this is true the token created will include the amount of fees needed to redeem the token (amount + fee_to_redeem)
```

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +29,351 @@ impl Wallet {
...
+    pub include_fee: bool,
```
thesimplekid, line 369: I think the better ux is the default behavior is to include the fee always. Do we think that should be the default her in `SendOptions` or that should be done higher up the stack?

crates/cdk/src/wallet/swap.rs
```diff
@@ -188,13 +199,16 @@ impl Wallet {
...
         let active_keyset_id = self.get_active_mint_keyset().await?.id;
 
         // Desired amount is either amount passed or value of all proof
         let proofs_total = proofs.total_amount()?;
 
         let ys: Vec<PublicKey> = proofs.ys()?;
-        self.localstore.set_pending_proofs(ys).await?;
+        self.localstore
+            .update_proofs_state(ys, State::Pending)
+            .await?;
```
thesimplekid, line 211: I think we should use `Reserved` here, and keep `Pending` as a state used by the mint ie in a melt operation where the ln payment is taking some time.

thesimplekid (commented):
Great work on this PR.

> How should we support CDK users who want to bring their own PreparedSend logic? I made PreparedSend a "private" structure so that users can't modify it, causing issues with our logic.

I think we can just added a `PreparedSend::new` that way they can at least create a `PreparedSend` to pass to `send`. But I think maybe better to not worry to much about this for now to keep it simpler and if that need arises we can revisit.


>   How should we support memos? I have it in SendOptions but I think I should move it to be a parameter of the send function, possibly with a flag to include it in the Token or keep it internal with the eventual TransactionsDatabase.
I agree I think we can move it to the `send` fn maybe we can have two and internal and an external (included in token) memo and when the former is not set use the external one.

thesimplekid:
Have you done any testing around does this change the proofs that would be selected from a set from previous cdk versions? Do you expect it to? 

I would be interested to include some integrations tests on this so we don't change proofs selection accidentally.

crates/cdk-common/src/database/wallet.rs
```diff
@@ -84,14 +84,6 @@ pub trait Database: Debug {
...
         removed_ys: Vec<PublicKey>,
     ) -> Result<(), Self::Err>;
-    /// Set proofs as pending in storage. Proofs are identified by their Y
-    /// value.
-    async fn set_pending_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
-    /// Reserve proofs in storage. Proofs are identified by their Y value.
-    async fn reserve_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
-    /// Set proofs as unspent in storage. Proofs are identified by their Y
-    /// value.
-    async fn set_unspent_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err>;
```
ok300, line 94: Maybe instead of removing the 3 methods, we can keep a default implementation in the trait?

```rust
    /// Set proofs as pending in storage. Proofs are identified by their Y
    /// value.
    async fn set_pending_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err> {
        self.update_proofs_state(ys, State::Pending).await
    }
    /// Reserve proofs in storage. Proofs are identified by their Y value.
    async fn reserve_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err> {
        self.update_proofs_state(ys, State::Reserved).await
    }
    /// Set proofs as unspent in storage. Proofs are identified by their Y
    /// value.
    async fn set_unspent_proofs(&self, ys: Vec<PublicKey>) -> Result<(), Self::Err> {
        self.update_proofs_state(ys, State::Unspent).await
    }
```

This would avoid the need for wallet DBs to re-implement them (as it's done in `cdk-rexie` now) and generally preserve the simpler caller semantics from before.

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +29,351 @@ impl Wallet {
...
+    /// All proofs
+    pub fn proofs(&self) -> Proofs {
+        let mut proofs = self.proofs_to_swap.clone();
+        proofs.extend(self.proofs_to_send.clone());
+        proofs
+    }
+
+    /// Total fee
+    pub fn fee(&self) -> Amount {
+        self.swap_fee + self.send_fee
```
ok300, line 326: Shouldn't this check the options?

```suggestion
        match self.options.include_fee {
            true => self.swap_fee + self.send_fee,
            false => self.swap_fee,
        }
```

crates/cdk/src/wallet/send.rs
ok300: IMO it's better to set the `SendOptions` default here to `true`.

davidcaseria:
> Have you done any testing around does this change the proofs that would be selected from a set from previous cdk versions? Do you expect it to? 
> 
> 
> 
> I would be interested to include some integrations tests on this so we don't change proofs selection accidentally. 

Yes, it should change proof selections to be more efficient. I wrote some unit tests because I made the `select_proofs` function static.

crates/cashu/src/wallet.rs
```diff
@@ -85,3 +85,25 @@ pub enum SendKind {
...
+
+    /// Check if send kind is exact
+    pub fn is_exact(&self) -> bool {
+        matches!(self, Self::OnlineExact | Self::OfflineExact)
+    }
+
+    /// Check if send kind has tolerance
+    pub fn has_tolerance(&self) -> bool {
+        matches!(self, Self::OnlineTolerance(_) | Self::OfflineTolerance(_))
+    }
```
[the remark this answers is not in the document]
davidcaseria: I prefer to use bools instead of checking the state of Options when doing conditional logic, such as how these functions are used. I can add a tolerance function that returns an Option, but I don't see why we shouldn't provide all of these functions in the API.

crates/cdk/src/wallet/proofs.rs
```diff
@@ -153,125 +158,375 @@ impl Wallet {
...
-            .partition(|p| p.keyset_id == active_keyset_id);
+            .filter(|p| !selected_proofs.contains(p))
+            .collect::<Proofs>();
+        selected_proofs.extend(Wallet::select_proofs(
+            remaining_amount,
+            remaining_proofs,
+            active_keyset_id,
+            &HashMap::new(), // Fees are already calculated
+            false,
+        )?);
```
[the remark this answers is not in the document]
davidcaseria: I believe so. Do you have a specific edge case in mind that I can include as a test?

crates/cdk/src/wallet/send.rs
davidcaseria: This is for when we need to send proofs with conditions. Any selected proofs must be swapped.

crates/cdk/src/wallet/proofs.rs
davidcaseria: The logic should work for any keyset because of the fallback in finding the smallest amount over the target amount. However, it’s optimized for powers of two.

crates/cdk/src/wallet/send.rs
davidcaseria: Why? Typically, receivers (merchants) pay for fees.

crates/cdk/src/wallet/send.rs
davidcaseria: The send_fee should always be zero if include_fee == false. I think it's safer to keep as is.

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +29,351 @@ impl Wallet {
...
+                    .await?
+                    .into_iter()
+                    .filter_map(|p| {
+                        if p.spending_condition.is_none() {
+                            Some(p.proof)
+                        } else {
+                            None
+                        }
+                    })
+                    .collect();
```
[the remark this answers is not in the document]
thesimplekid: Sorry forgot I said I would look at this and got lost in the review. 

This pr has a fix for matching the conditions using the current api which admittedly as a bit confusing and I'm open to suggestions on. But passing in `None` means that we do not care about that condition and want all proofs no matter what that field. For example `get_proofs(None, None, None, None)` would return **ALL** proofs in the db. `get_proofs(Some("exmaple.xyz"), None, None, None` would return all proofs of that mint. So the way that spending conditions is is `None` would but proofs of any or no spending condition. So to get proofs that do not have a spending condition it would be `Some(vec![])`. Firstly, Does how I've explained how this currently works make sense? Secondly, does this api make any sense or do you have a suggestion to improve it? 

https://github.com/cashubtc/cdk/pull/654

crates/cdk/src/wallet/send.rs
thesimplekid: > Why? Typically, receivers (merchants) pay for fees.

This is true but in ecash i think this is going to lead to bad ux where someone says send me 25 sats you create a token for 25 sats without including fees and then when they receive they say why did you only send me 23 sats. 

But as i alluded to above maybe this shouldn't be pushed on the lib level and the default of false makes more sense and wallet devs can choose the ux they want.

crates/cdk-sqlite/src/wallet/migrations/20240612132920_init.sql
```diff
@@ -58,7 +58,7 @@ CREATE TABLE IF NOT EXISTS key (
 CREATE TABLE IF NOT EXISTS proof (
 y BLOB PRIMARY KEY,
 mint_url TEXT NOT NULL,
-state TEXT CHECK ( state IN ('SPENT', 'UNSPENT', 'PENDING', 'RESERVED' ) ) NOT NULL,
+state TEXT CHECK ( state IN ('SPENT', 'UNSPENT', 'PENDING', 'RESERVED', 'PENDING_SPENT' ) ) NOT NULL,
```
thesimplekid, line 61: We need to create a new migration file. We can't just modify the init as anyone with an existing wallet will get errors.

crates/cdk-sqlite/src/wallet/mod.rs
```diff
@@ -734,6 +693,29 @@ FROM proof;
...
+            .bind(y.to_bytes().to_vec())
+            .execute(&mut *transaction)
+            .await
+            .map_err(Error::from)?;
+        }
+
+        transaction.commit().await.map_err(Error::from)?;
+
+        Ok(())
+    }
```
thesimplekid, line 717: I know this is just following the bad example I set, but can we remove the for loop and do an in clause like Ive updated the mint to use. 

https://github.com/cashubtc/cdk/blob/72dff953220768db129f5ce7d4c4fb337c6f838c/crates/cdk-sqlite/src/mint/mod.rs#L1152-L1173

thesimplekid:
Not sure whats going on with the CI here it does not seem related to any of the code changes in this PR

crates/cdk-sqlite/src/wallet/migrations/20240612132920_init.sql
davidcaseria: I forgot I did this, and I couldn't figure out how to update this check in a new migration. ChatGPT had me create a new table, transfer the data, delete the old table, and rename the new table. That didn't seem right so I just did this to get it to work. Do you know how to change the check?

thesimplekid: Yeah for sqlite that is what you have to do. The ALTER TABLE in sqlite is pretty limited if it was psotgress then we could just modify the existing table. 

https://github.com/cashubtc/cdk/blob/main/crates/cdk-sqlite/src/mint/migrations/20240718203721_allow_unspent.sql

crates/cdk/src/wallet/send.rs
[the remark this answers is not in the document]
davidcaseria: My loosely held opinion is to keep the `get_proofs` API as is and add another function to return all proofs.

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +37,340 @@ impl Wallet {
...
-                let selected_proofs_amount = selected_proofs.total_amount()?;
+        self.internal_prepare_send(amount, opts, selected_proofs, force_swap)
+            .await
+    }
 
-                let amount_to_send = match include_fees {
-                    true => amount + self.get_proofs_fee(&selected_proofs).await?,
-                    false => amount,
-                };
+    async fn internal_prepare_send(
```
thesimplekid, line 122: I see there is a `cancel_send` that sets the prepared proofs back to unspent but I do not see where the proofs are reserved, maybe I'm just missing it?

My understanding is that `UNSPENT` proofs should be selected and set to `RESERVED` in the `prepare_send` and then `PreparedSend` either needs to be used and `proofs` are set `PENDINGSPENT` or `PreparedSend` needs to be canceled and proofs are moved back to `UNSPENT`. 

If we do not reserve the proofs in the prepare then if prepare is called again these proofs maybe selected again.

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +37,340 @@ impl Wallet {
...
+            return Err(Error::InsufficientFunds);
+        }
 
-                self.swap_from_unspent(amount, conditions, include_fees)
-                    .await?
-            }
+        // Update proofs state to pending spent
+        self.localstore
+            .update_proofs_state(proofs_to_send.ys()?, State::PendingSpent)
+            .await?;
```
thesimplekid, line 244: Should we check the proof state here to make sure they are Reserved or Unspent and have not already been used in another tx

crates/cdk/src/wallet/send.rs
davidcaseria: Somehow reserving the proofs got deleted. Adding a test now.

crates/cdk/src/wallet/send.rs
```diff
@@ -45,135 +37,340 @@ impl Wallet {
...
+    /// Cancel prepared send
+    pub async fn cancel_send(&self, send: PreparedSend) -> Result<(), Error> {
+        tracing::info!("Cancelling prepared send");
+
+        self.localstore
+            .update_proofs_state(send.proofs().ys()?, State::Unspent)
+            .await?;
+
+        Ok(())
+    }
```
thesimplekid, line 284: This will set all the proofs in the PreparedSend to Unspent. But we probably dont want to set proofs that are PendingSpent or Spent back. If the api is uses correctlly this wont be an issue but i think its best we check it.

crates/cdk/src/wallet/send.rs
thesimplekid: Maybe some new error ProofStateUnexpected and just add a log message.

crates/cdk/src/wallet/send.rs
davidcaseria: Is this in case a bug is introduced? I can add the check for safety but I'm not sure how that would happen.

crates/cdk/src/wallet/send.rs
davidcaseria: Yes, until `Wallet::unreserve_proofs` is called.

crates/cdk/src/wallet/send.rs
thesimplekid: > Is this in case a bug is introduced? I can add the check for safety but I'm not sure how that would happen.

My thinking to to pretect the user from the flow of prepared send -> send but then calling cancel on the prepared send. In this case where they've already created the token with send they should have to receive the token to reclaim it.

davidcaseria: This shouldn't be an issue because `PreparedSend` is not `Clone`, and both `send` and `cancel_send` move the struct, but I will add an additional check just in case.

thesimplekid: Ah good point

thesimplekid (approved):
ACK 8c7506f43c270576ea0adcef589c32c76222c87e

PR #597, opened 2025-02-11
feat: payment processor

thesimplekid:
### Description

-----

- [x] Remove strike and phd from readme

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-payment-processor/src/proto/payment_processor.proto
+++ b/crates/cdk-payment-processor/src/proto/payment_processor.proto
@@ -1,0 +1,113 @@
+syntax = "proto3";
+
+package cdk_payment_processor;
+
+service CdkPaymentProcessor {  
+    rpc GetSettings(SettingsRequest) returns (SettingsResponse) {}
+    rpc CreatePayment(CreatePaymentRequest) returns (CreatePaymentResponse) {}
+    rpc GetPaymentQuote(PaymentQuoteRequest) returns (PaymentQuoteResponse) {}
+    rpc MakePayment(MakePaymentRequest) returns (MakePaymentResponse) {}
+    rpc CheckIncomingPayment(CheckIncomingPaymentRequest) returns (CheckIncomingPaymentResponse) {}
+    rpc CheckOutgoingPayment(CheckOutgoingPaymentRequest) returns (MakePaymentResponse) {}
+    rpc WaitIncomingPayment(WaitIncomingPaymentRequest) returns (stream WaitIncomingPaymentResponse) {}
+}
+
+message SettingsRequest {}
+
+message SettingsResponse {
+  string inner = 1;
+}
+
+message CreatePaymentRequest {
+  uint64 amount = 1;
+  string unit = 2;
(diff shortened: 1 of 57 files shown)


crates/cdk-payment-processor/src/proto/payment_processor.proto
```diff
@@ -0,0 +1,112 @@
...
+
+message SettingsRequest {}
+
+message SettingsResponse {
+  string inner = 1;
+}
+
+message CreatePaymentRequest {
+  uint64 amount = 1;
+  string unit = 2; string description = 3;
```
codingpeanut157, line 23: `\n` missing

crates/cdk/src/mint/builder.rs
```diff
@@ -27,7 +29,7 @@ pub struct MintBuilder {
     /// Mint Storage backend
     localstore: Option<Arc<dyn MintDatabase<Err = database::Error> + Send + Sync>>,
     /// Ln backends for mint
-    ln: Option<HashMap<LnKey, Arc<dyn MintLightning<Err = cdk_lightning::Error> + Send + Sync>>>,
+    ln: Option<HashMap<LnKey, Arc<dyn MintPayment<Err = cdk_payment::Error> + Send + Sync>>>,
```
davidcaseria, line 32: Do you plan to make this more generic with on-chain?

thesimplekid: Yes, the plan is to use this trait for all payment backends. If we looks at the ones I created for [bolt12](https://github.com/cashubtc/cdk/blob/cc4a63bd2bdd7c9b1bd592bb4f1c9bd03d6c9d3f/crates/cdk/src/cdk_lightning/bolt12.rs#L17) and [onchain](https://github.com/cashubtc/cdk/blob/8dc537a18ea3a30b7d24c7f69d7d1de006b0111a/crates/cdk/src/cdk_onchain/mod.rs#L33). I think we can use what we have here without modify it but there would be some unused fields. The problem as I see it is to make it really general for everything we could just use `Value` but this loses typing so Ideally we don't do this though I did use it for the settings endpoint. But if we try to create types now to make it more general for things we dont support yet (onchain, bolt11) we may need to end up chaining them later anyway. So I think its best to keep the grpc api pretty close to what we know we need and currently support, bolt11, and then make changes when we know we need them and what changes we need to support bolt12, and onchain or whatever else.

That being said I'm certainly open if you have suggestions to the api at this point as I mostly just reused what we had in the previous trait.

davidcaseria: The trait is well-defined. Should the `LnKey` struct be made more generic now or later?

thesimplekid: > Should the LnKey struct be made more generic now or later? 

Good catch f46d84e631148bd4ebe5c8343c63ee405f8913e0

ok300 (approved):
Had another look at it, looks good.

PR #598, opened 2025-02-11
cdk-cln: refactor to use .call_typed

daywalker90:
### Description

Later versions of cln-rpc introduced .call_typed which are way more ergonomic so i recommend using those instead of .call

-----

### Notes to the reviewers

-----

#### None

----

### Checklist


thesimplekid (approved):
Nice thanks.

ACK 38cd88bdf9773bbbe15805a0527000bb527c5a1c

PR #605, opened 2025-02-17
BYO network transport when minting tokens

vnprc:
### Description

Hashpool operates on a slightly different model from typical ecash mints. It generates ecash tokens atomically when validating mining pool shares. In order to accomplish atomicity we need to piggyback the ecash operations on existing mining protocol messages. This means the ecash mint and wallet will not be able to make synchronous HTTP calls during the process of minting new tokens, which is the default assumption in cdk.

Here's the new flow:
1. miner finds a valid share above the pool difficulty threshold
1. proxy/wallet generates a new premint secret and saves it to the wallet
1. proxy/wallet submits a mining share and premint secret to the pool/mint
1. pool validates the mining share
1. mint creates a new quote using the block header hash of the mining share as a quote ID (is this necessary or useful?)
1. mint signs the blinded secret
1. pool returns this signature in the 'mining share accepted' message payload
1. proxy receives the signature from the mint
1. proxy retrieves the blinded secret from the cashu wallet
1. wallet validates the signature and stores the proof as an ecash token

There are two main differences with this design:
1. the wallet must store keysets and blinded secrets locally
2. the wallet must rely on local information about the mint rather than querying the mint at will

-----

### Notes to the reviewers

I have exposed some helper functions to convert between `Amount` and `AmountStr`. This helps the wallet store blinded secrets using a BTreeMap. This is kind of a code smell (Why are we converting to a string for indexing purposes? Shouldn't the data structure handle this for us?) but it was the shortest path to victory during development. I'm not sure if this is necessary. Hoping to get some clarity in this PR.

I didn't discuss it in the minting flow above but mint keysets will be propagated by the pool in a mining message payload and stored locally in the wallet. This requires a new function to add mint keysets to the wallet without making any HTTP calls.

I have opened this PR in draft state with many TODO comments and macros. I will fix these issues in due time but I wanted to get feedback on my general approach.

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/cdk_database/wallet_memory.rs
+++ b/crates/cdk/src/cdk_database/wallet_memory.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use cdk_common::database::{Error, WalletDatabase};
+use cdk_common::PreMintSecrets;
 use tokio::sync::RwLock;
 
 use crate::mint_url::MintUrl;
@@ -28,6 +29,7 @@ pub struct WalletMemoryDatabase {
     proofs: Arc<RwLock<HashMap<PublicKey, ProofInfo>>>,
     keyset_counter: Arc<RwLock<HashMap<Id, u32>>>,
     nostr_last_checked: Arc<RwLock<HashMap<PublicKey, u32>>>,
+    premint_secrets: Arc<RwLock<HashMap<String, PreMintSecrets>>>,
 }
 
 impl WalletMemoryDatabase {
@@ -38,6 +40,7 @@ impl WalletMemoryDatabase {
         mint_keys: Vec<Keys>,
         keyset_counter: HashMap<Id, u32>,
         nostr_last_checked: HashMap<PublicKey, u32>,
+        premint_secrets: HashMap<String, PreMintSecrets>,
     ) -> Self {
         Self {
             mints: Arc::new(RwLock::new(HashMap::new())),
@@ -55,6 +58,7 @@ impl WalletMemoryDatabase {
             proofs: Arc::new(RwLock::new(HashMap::new())),
             keyset_counter: Arc::new(RwLock::new(keyset_counter)),
             nostr_last_checked: Arc::new(RwLock::new(nostr_last_checked)),
+            premint_secrets: Arc::new(RwLock::new(premint_secrets)),
         }
     }
 }
@@ -352,5 +356,23 @@ impl WalletDatabase for WalletMemoryDatabase {
             .insert(verifying_key, last_checked);
 
         Ok(())
+    }
+
+    async fn add_premint_secrets(
+        &self,
+        quote_id: &str,
+        premint_secrets: &PreMintSecrets,
+    ) -> Result<(), Self::Err> {
+        self.premint_secrets
+            .write()
+            .await
+            .insert(quote_id.to_string(), premint_secrets.clone());
+        Ok(())
+    }
+    async fn get_premint_secrets(
+        &self,
+        quote_id: &str,
+    ) -> Result<Option<PreMintSecrets>, Self::Err> {
+        Ok(self.premint_secrets.read().await.get(quote_id).cloned())
     }
 }
--- a/crates/cdk/src/wallet/keysets.rs
+++ b/crates/cdk/src/wallet/keysets.rs
@@ -25,14 +25,13 @@ impl Wallet {
         Ok(keys)
     }
 
-    /// Straight up yolo those keys into the db no cap
+    /// Add a keyset to the local database and update keyset info
     pub async fn add_keyset(
         &self,
         keys: Keys,
         active: bool,
         input_fee_ppk: u64,
     ) -> Result<(), Error> {
-        // add to localstore
         self.localstore.add_keys(keys.clone()).await?;
 
         let keyset_info = KeySetInfo {
@@ -42,7 +41,6 @@ impl Wallet {
             input_fee_ppk,
         };
 
-        // somehow also add to localstore...why do I need to make two different calls?
         self.localstore
             .add_mint_keysets(self.mint_url.clone(), vec![keyset_info])
             .await?;
--- a/crates/cdk/src/wallet/keysets.rs
+++ b/crates/cdk/src/wallet/keysets.rs
@@ -25,6 +25,31 @@ impl Wallet {
         Ok(keys)
     }
 
+    /// Straight up yolo those keys into the db no cap
+    pub async fn add_keyset(
+        &self,
+        keys: Keys,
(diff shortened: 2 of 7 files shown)


crates/cdk/src/wallet/keysets.rs
```diff
@@ -97,4 +120,109 @@ impl Wallet {
             .ok_or(Error::NoActiveKeyset)?;
         Ok(keyset_with_lowest_fee)
     }
+
+    /// Get active keyset for mint from local without querying the mint
+    #[instrument(skip(self))]
+    pub async fn get_active_mint_keyset_local(&self) -> Result<KeySetInfo, Error> {
```
vnprc, line 126: Needs a better function name that indicates this function doesn't make any network calls.

crates/cdk/src/wallet/mint.rs
vnprc, line 1: I wrote these functions based on an old version of cdk and copied them over to the latest branch. Need to look back and make sure the code I copied hasn't changed in the mean time. Also, and related, I am uncomfortable with the amount of duplicated code in this module. Need to refactor the shared code between the existing mint function and these new functions.

crates/cdk/src/cdk_database/wallet_memory.rs
```diff
@@ -28,6 +29,7 @@ pub struct WalletMemoryDatabase {
     proofs: Arc<RwLock<HashMap<PublicKey, ProofInfo>>>,
     keyset_counter: Arc<RwLock<HashMap<Id, u32>>>,
     nostr_last_checked: Arc<RwLock<HashMap<PublicKey, u32>>>,
+    premint_secrets: Arc<RwLock<HashMap<String, PreMintSecrets>>>,
```
average-gary, line 32: Why String for the HashMap key here?

vnprc: Honestly, I don't know if this is necessary. It would be great if I could get rid of the string entirely.

I copied this code from somewhere else. This [code comment](https://github.com/cashubtc/cdk/pull/605/files#diff-2136a68f4fddbbe35e81ec9dff7663511b7553ef7bf7be10b7dccbbd533c2a71L219-R223) seems to indicate using a string helps with sorting these items in a BTreeMap.

```
/// String wrapper for an [Amount].
///
/// It ser-/deserializes the inner [Amount] to a string, while at the same time using the [u64]
/// value of the [Amount] for comparison and ordering. This helps automatically sort the keys of
/// a [BTreeMap] when [AmountStr] is used as key.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct AmountStr(Amount);
```

thesimplekid: We're planning to remove this memory db so i wouldn't worry about this. 

https://github.com/cashubtc/cdk/issues/607

vnprc: I use AmountStr in hashpool in the same way. Is this the recommended way to put pubkeys in a BTreeMap?

https://github.com/vnprc/hashpool/blob/master/protocols/v2/subprotocols/mining/src/cashu.rs#L270

thesimplekid: You can just use the `Amount` there no need to wrap it in the `AmoutnStr`. The reason we have to use `AmountStr` is for when the Keys response is json serialized it needs to be as a string and not an int, so that's really the only place we should use `AmountStr` everywhere else is `Amount`

crates/cdk/src/wallet/mint.rs
```diff
@@ -289,4 +290,211 @@ impl Wallet {
...
+        let active_keyset_id = self.get_active_mint_keyset_local().await?.id;
+
+        // Retrieve the keyset counter, defaulting to 0 if not found
+        // TODO do we need to update the counter if it was not returned?
+        let count = self
+            .localstore
+            .get_keyset_counter(&active_keyset_id)
+            .await?
+            .unwrap_or(0)
+            + 1;
```
vnprc, line 377: I think we have a race condition here where premint secrets could be generated twice here before incrementing the counter.

https://github.com/cashubtc/nuts/blob/main/13.md

@thesimplekid will find the exact PR and share here

vnprc: We may want to change this to atomically increment the counter here instead of when constructing proofs.

callebtc: - increment counters when the secrets are generated (not when the proofs are constructed)
- atomicity: race condition between read+write (get counter + bump counter), would require some locking mechanism

crates/cdk/src/wallet/mint.rs
```diff
@@ -289,4 +290,211 @@ impl Wallet {
...
+    ///
+    /// Returns an error if:
+    /// - The quote is unknown or has an invalid keyset (`Error::UnknownQuote` or `Error::UnknownKeySet`).
+    /// - Signature verification fails (`Error::CouldNotVerifyDleq`).
+    /// - Constructing proofs fails.
+    pub async fn verify_and_store_proofs(
+        &self,
+        signatures: [Option<BlindSignature>; 64],
+        quote_id: &str,
+    ) -> Result<Amount, Error> {
```
vnprc, line 422: Update this to return proofs instead of amount

crates/cdk/src/wallet/mint.rs
```diff
@@ -289,4 +290,211 @@ impl Wallet {
...
+            return Err(Error::PaidQuote);
+        }
+
+        // Create and store a new mint quote
+        let mint_quote = MintQuote {
+            id: quote_id.to_string(),
+            mint_url: mint_url.parse()?,
+            amount: Amount::from(amount),
+            unit: currency_unit,
+            request: "todo".to_string(), // TODO: what does request do?
```
vnprc, line 360: This doesn't really translate to our new flow because the "proof of payment" is generated before minting is even initiated. Use empty string for now.

vnprc:
We already have types for bolt11 and bolt12 quotes. Create a new type for a mining share quote.

Note to self: review the quote protocol nut

vnprc:
After team review we want to change the flow to work asynchronously within the NUT04 spec. It will work like this:
1. submit the mining share to the mint along with a pubkey (NUT20)
1. mint creates a quote and responds with the quote
1. mint can set the quote to paid if the mining share is validated
1. wallet can redeem the quote for eHash later as described in NUT04 & NUT20

This makes it easy to run a normal cashu mint and existing wallets can already support this. This decouples the mining protocol from the cashu protocol and lets us use sexy stuff like websockets without changing the mining protocol messages.

thesimplekid:
@vnprc Do you plan to still work on this PR or after out call going to go another direction?

Issue #511, opened 2024-12-23
Place storage backends behind feature flags


ok300:
Right now, to implement a wallet one needs to import the CDK and the chosen storage backend crate

```toml
cdk = { version = "v0.6.0", features = ["wallet"] }
cdk-sqlite = { version = "v0.6.0", features = ["wallet"] }
```

`mintd` has a dependency to all storage backends.

It would be simpler at least for wallet implementors to be able to choose the storage backend via a feature flag:

```toml
cdk = { version = "v0.6.0", features = ["wallet", "sqlite"] }
```

PR #613, opened 2025-02-23
Drop the in-memory database

### LINKED ISSUE #607: Remove Memory database
We've discussed removing the memory database from the mint database. In cases where an in memory database is desired an in memory sql db could be used and it removed the need for us to maintain another db within cdk. I think this is fairly obvious for the mint should be also do it for the wallet?

cc @crodas

crodas:
### Description

Fixes #607 (this PR depends on #620)

This PR drops the implementation of in-memory database traits.

They are useful for testing purposes since the tests should test our codebase and assume the database works as expected (although a follow-up PR should write a sanity test suite for all database trait implementors).

As complexity is worth with database requirements to simplify complexity and add more robustness, for instance, with the following plans to add support for transactions or buffered writes, it would become more complex and time-consuming to support a correct database trait. This PR drops the implementation and replaces it with a SQLite memory instance

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-integration-tests/Cargo.toml
```diff
@@ -20,7 +20,7 @@ rand = "0.8.5"
 bip39 = { version = "2.0", features = ["rand"] }
 anyhow = "1"
 cashu = { path = "../cashu", features = ["mint", "wallet"] }
-cdk = { path = "../cdk", features = ["mint", "wallet"] }
+cdk = { path = "../cdk", features = ["mint", "wallet", "memory_database"] }
```
thesimplekid, line 23: Do you think it makes sense to add a memory db feature? Or should we just keep it as you have to bring in one of the cdk-sqlite or redb crates. And whether the sqlite is in mem or not can be handled by that crate. The feature maybe is a little similar for the end user but it creates two para dimes to accomplish the same thing one enabling the feature one bringing in the create.  



Related: https://github.com/cashubtc/cdk/issues/511

crates/cdk/src/wallet/keysets.rs
```diff
@@ -89,6 +89,8 @@ impl Wallet {
     /// keysets
     #[instrument(skip(self))]
     pub async fn get_active_mint_keyset(&self) -> Result<KeySetInfo, Error> {
+        // Important
+        let _ = self.get_mint_info().await?;
```
thesimplekid, line 93: https://github.com/cashubtc/cdk/issues/621

thesimplekid (changes requested):
Just a small question think there was a feature flag left in from a rebase? Other then that lgtm

thesimplekid (approved):
ACK bb46e85683258f71af6e66adb61b9c14242f2509

PR #616, opened 2025-02-24
Refactor mintd used in itest

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
> There's a similar pattern in cdk_integration_tests::wallet_mint() which could probably also be replaced with wait_for_mint_to_be_paid() 

You're likely right here aec531740c49fb5f1246e342094d50d57ae9e5e7?

ok300:
I was thinking of

https://github.com/cashubtc/cdk/blob/21a2f8b7926ba139f82d7bcecc7936db67989b36/crates/cdk-integration-tests/src/lib.rs#L31-L43

thesimplekid:
Yep found it thanks 24f1d59af93ca3b7e57d02c26218d3bebc525746

ok300 (approved):

PR #618, opened 2025-02-26
Make "just format" use the nightly toolchain

ok300:
This PR enforces the nightly toolchain when calling `just format`.

### DIFF
--- a/justfile
+++ b/justfile
@@ -35,7 +35,7 @@ format:
   if [ ! -f Cargo.toml ]; then
     cd {{invocation_directory()}}
   fi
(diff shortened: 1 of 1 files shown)


thesimplekid (changes requested):
image

Cargo isn't what handles this its rustup(maybe just since I don't have rustup and it would work otherwise?), but we don't want to require rustup since we have the nix.

ok300:
You're right, this is not a good solution. It only works if `cargo` is installed through `rustup`, which may not always be the case.

PR #620, opened 2025-02-27
Do not use `INSERT OR REPLACE` in SQLite.

crodas:
### Description

Instead, use `INSERT` and `ON CONFLICT`.  The reason is that in case of conflicts, the `REPLACE` will trigger a DELETE and then perform an INSERT, as outlined in the documentation[1], and that may cause a cascade of deletion due to our FOREIGN KEYs.

Here is the official documentation:

```
When a UNIQUE or PRIMARY KEY constraint violation occurs, the REPLACE algorithm
deletes pre-existing rows that are causing the constraint violation prior to
inserting or updating the current row and the command continues executing
normally. If a NOT NULL constraint violation occurs, the REPLACE conflict
resolution replaces the NULL value with the default value for that column, or
if the column has no default value, then the ABORT algorithm is used. If
a CHECK constraint or foreign key constraint violation occurs, the REPLACE
conflict resolution algorithm works like ABORT.  When the REPLACE conflict
resolution strategy deletes rows in order to satisfy a constraint, delete
triggers fire if and only if recursive triggers are enabled.  The update
hook is not invoked for rows that are deleted by the REPLACE conflict
resolution strategy. Nor does REPLACE increment the change counter. The
exceptional behaviors defined in this paragraph might change in a future
release.
```

[1] https://www.sqlite.org/lang_conflict.html

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sqlite/src/mint/mod.rs
+++ b/crates/cdk-sqlite/src/mint/mod.rs
@@ -205,9 +205,23 @@ WHERE active = 1
 
         let res = sqlx::query(
             r#"
-INSERT OR REPLACE INTO mint_quote
+INSERT INTO mint_quote
 (id, amount, unit, request, state, expiry, request_lookup_id, pubkey)
-VALUES (?, ?, ?, ?, ?, ?, ?, ?);
+VALUES (?, ?, ?, ?, ?, ?, ?, ?)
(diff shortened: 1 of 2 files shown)


thesimplekid (approved):
Thank you. LGTM

PR #630, opened 2025-03-05
Db check on delete proofs

### LINKED ISSUE #627: `localstore.remove_proofs` should select for only pending proofs
We should add a select statement (or equivalent filtering for redb) so that remove_proofs **can not** remove spent proofs

thesimplekid:
### Description

-----

closes #627 

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ lightning-invoice = { version = "0.32.0", features = ["serde", "std"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 thiserror = { version = "1" }
-tokio = { version = "1", default-features = false }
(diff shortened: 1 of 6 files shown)


ok300:
Also a NIT: the new redb and sqlite tests look essentially the same (except for the DB instance), so maybe they should be extracted out to a common place and called with a DB instance arg?

thesimplekid:
> Also a NIT: the new redb and sqlite tests look essentially the same (except for the DB instance), so maybe they should be extracted out to a common place and called with a DB instance arg?

Thats a good point. Since this is just these two tests I'm going to leave it as is for now but we should come back to that and create more complete tests specific to the DB. I opened an issue for tracking feel free to add ideas to that https://github.com/cashubtc/cdk/issues/632.

PR #631, opened 2025-03-05
Simplify `process_swap_request`

ok300:
### Description

This PR slightly changes the `process_swap_request` implementation: it validates if the tx is balanced _before_ storing the new proofs in the DB. By doing this, we avoid having to later remove these proofs in case the tx is not balanced.

-----

### Notes to the reviewers

-----

#### FIXED

* NUT-17: Removed superfluous proof state notification

----

### Checklist


thesimplekid:
Since we now expect it to fail at the unbalanced check we should change this error 

https://github.com/cashubtc/cdk/blob/c6200331cf570f4d0c1e1fa321c719436aeb20a5/crates/cdk-integration-tests/tests/fake_wallet.rs#L1079

thesimplekid (approved):

PR #640, opened 2025-03-09
feat: Add support for sqlcipher

benthecarman:
### Description

Adds support for sqlcipher when using sqlite. This allows for encrypted database instead of just using normal sqlite 

### Notes to the reviewers

-----

#### ADDED

Added support for sqlcipher when using sqlite

#### FIXED

----

### Checklist


thesimplekid (approved):

crates/cdk-sqlite/src/common.rs
```diff
@@ -27,3 +28,29 @@ pub async fn create_sqlite_pool(path: &str) -> Result<Pool<Sqlite>, Error> {
 
     Ok(pool)
 }
+
+#[inline(always)]
+#[cfg(feature = "sqlcipher")]
+pub async fn create_sqlite_pool(path: &str, password: String) -> Result<Pool<Sqlite>, Error> {
```
ok300, line 34: This is duplicating 99% of `create_sqlite_pool`. Don't know if it's a good idea.

Why not instead update the existing `create_sqlite_pool` to handle both cases? It would only need 2 small changes:

- add an optional argument to the method
- conditionally set the `sqlcipher` password, if the feature flag is set

Like so:

```rust
#[inline(always)]
pub async fn create_sqlite_pool(
    path: &str,
    #[cfg(feature = "sqlcipher")] password: String
) -> Result<Pool<Sqlite>, Error> {
    let mut db_options = SqliteConnectOptions::from_str(path)?
        .busy_timeout(Duration::from_secs(10))
        .read_only(false)
        .pragma("busy_timeout", "5000")
        .pragma("journal_mode", "wal")
        .pragma("synchronous", "normal")
        .pragma("temp_store", "memory")
        .pragma("mmap_size", "30000000000")
        .shared_cache(true)
        .create_if_missing(true);

    #[cfg(feature = "sqlcipher")]
    {
        db_options = db_options.pragma("key", password);
    }

    let pool = SqlitePoolOptions::new()
        .min_connections(1)
        .max_connections(1)
        .idle_timeout(None)
        .max_lifetime(None)
        .connect_with(db_options)
        .await?;

    Ok(pool)
}
```

benthecarman: great idea, done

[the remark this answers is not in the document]
thesimplekid: I don't think adding the panic would add any benefit and its better to just enforce use with a password when the feature is enabled.

thesimplekid (approved):
ACK 40c53e83df9cde4ca7f957caecbfa9808fe2f78b

Thanks @benthecarman for the contribution

crates/cdk-sqlite/src/common.rs
thesimplekid: Nice, didn't realize you could do that with fn args. Could have swore I got build errors on that before maybe it was with our older msrv.

PR #647, opened 2025-03-10
Drop `nostr_last_checked` table, remove references

ok300:
### Description

#645 removed `nostr_last_checked` references from the mint storage, but the table schema and data remain.

This PR drops the table from the various storage backends:
- `cdk-sqlite`: added a migration to drop the table
- `cdk-redb`: due to a typo, the `NOSTR_LAST_CHECKED` key was pointing to the `keyset_counter` table, so no `nostr_last_checked` was ever created. Here, this PR simply removes redb references to `NOSTR_LAST_CHECKED`.
- `cdk-rexie`: bump the DB version and do not add the `NOSTR_LAST_CHECKED` object store.

-----

### Checklist


thesimplekid (approved):
Ah good catch thanks

PR #655, opened 2025-03-13
fix: verification of melt quote with empty outputs

thesimplekid:
### Description

-----

@davidcaseria  this should fix the ci issue on #596 

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


ok300 (commented):
When can a mint operation have no outputs?

thesimplekid:
> When can a mint operation have no outputs?

It can't that would fail the unbalanced check.

ok300 (approved):
I see, melt can have no outputs, not mint.

PR #657, opened 2025-03-13
Use base trait in return type of Mint's `get_settings()`

ok300:
### Description

This PR replaces `serde_json::Value` with a custom base trait `BaseMintSettings` in the mint's `get_settings()` method.

This is now possible due to the recent bump in MSRV.

----

### Checklist


ok300:
Since the generic return type has to be easily de-/serializable for the gRPC interface as well, it's simpler to stick to `serde_json::Value`.

PR #662, opened 2025-03-19
Ser/Deserialize `SecretKey` either as bytes or string

ok300:
### Description

This PR fixes an issue that prevented CDK wallets from receiving cashu tokens that include DLEQs.

-----

### Notes to the reviewers

In the NUTs, various fields that map to `SecretKey` are sometimes defined as `str` [^1] and sometimes as `bytes`[^2]. In the CBOR encoding (used for creating the cashu token), `bytes` is used. When encoding to JSON (in API responses from the mint), `str` is used.

Therefore, we need to be able to se-/de-serialize a `SecretKey` to/from either of these types, depending on the context. Until now, we only covered mappings to/from `String`, which means we could not receive a cashu token with DLEQ proofs, because deserialization would fail.

-----

#### FIXED

- Wallet: Fix ability to receive cashu tokens that include DLEQ proofs

----

### Checklist

[^1]: For example in NUT-12, `e` and `s` in a DLEQ proof: https://github.com/cashubtc/nuts/blob/fb2ad98431ca5604b7e07edb028da26bff7bf6fb/12.md?plain=1#L61-L62 . This is encoded as JSON.
[^2]: For example in NUT-00: `e`, `s` and `r` in a DLEQ proof: https://github.com/cashubtc/nuts/blob/fb2ad98431ca5604b7e07edb028da26bff7bf6fb/00.md?plain=1#L243-L245 . This is encoded using CBOR into a byte array, that is then encoded with base64 to generate a cashu token.

### DIFF
--- a/crates/cashu/src/nuts/nut01/secret_key.rs
+++ b/crates/cashu/src/nuts/nut01/secret_key.rs
@@ -8,6 +8,7 @@ use bitcoin::secp256k1;
 use bitcoin::secp256k1::rand::rngs::OsRng;
 use bitcoin::secp256k1::schnorr::Signature;
 use bitcoin::secp256k1::{Keypair, Message, Scalar};
+use serde::de::Visitor;
 use serde::{Deserialize, Deserializer, Serialize};
 
 use super::{Error, PublicKey};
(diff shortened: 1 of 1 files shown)


lollerfirst (approved):

thesimplekid (approved):
Thank you.

PR #663, opened 2025-03-20
Fix protoc build error

davidcaseria:
### Description

Depending on the version of `protoc`, the `--experimental_allow_proto3_optional` may be needed.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
Whats the error you're seeing?

thesimplekid:
> Whats the error you're seeing?

Sorry just read this really quickly and was thinking it was the version we define in Cargo.toml but its the one that on the system so we don't really have control over that

thesimplekid (approved):

davidcaseria:
Docker issues 🙄

PR #664, opened 2025-03-21
Consolidate MultiMintWallet

ok300:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/multi_mint_wallet.rs
+++ b/crates/cdk/src/wallet/multi_mint_wallet.rs
@@ -7,6 +7,9 @@ use std::collections::{BTreeMap, HashMap};
 use std::str::FromStr;
 use std::sync::Arc;
 
+use anyhow::Result;
+use cdk_common::database;
+use cdk_common::database::WalletDatabase;
 use cdk_common::wallet::WalletKey;
 use tokio::sync::Mutex;
 use tracing::instrument;
@@ -23,14 +26,20 @@ use crate::{ensure_cdk, Amount, Wallet};
 /// Multi Mint Wallet
 #[derive(Debug, Clone)]
 pub struct MultiMintWallet {
+    /// Storage backend
+    pub localstore: Arc<dyn WalletDatabase<Err = database::Error> + Send + Sync>,
     /// Wallets
     pub wallets: Arc<Mutex<BTreeMap<WalletKey, Wallet>>>,
 }
 
 impl MultiMintWallet {
-    /// New Multimint wallet
-    pub fn new(wallets: Vec<Wallet>) -> Self {
+    /// Create a new [MultiMintWallet] with initial wallets
+    pub fn new(
+        localstore: Arc<dyn WalletDatabase<Err = database::Error> + Send + Sync>,
+        wallets: Vec<Wallet>,
+    ) -> Self {
         Self {
+            localstore,
             wallets: Arc::new(Mutex::new(
                 wallets
                     .into_iter()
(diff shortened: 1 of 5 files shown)


thesimplekid:
@davidcaseria I think this may make cdk-flutter a bit easier?

crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -41,13 +53,26 @@ impl MultiMintWallet {
     }
 
     /// Add wallet to MultiMintWallet
-    #[instrument(skip(self, wallet))]
-    pub async fn add_wallet(&self, wallet: Wallet) {
+    pub async fn add_wallet(
```
thesimplekid, line 56: Why don't we keep add wallet the way it is currently done and then add a new fn `create_wallet` that does this. This way if a user wants to create a custom wallet with a different seed or something that can still be done and added to the multimint wallet.

crates/cdk/src/wallet/mod.rs
```diff
@@ -123,26 +123,26 @@ impl Wallet {
...
+    ///     let wallet = Wallet::new(mint_url, unit, Arc::new(localstore), seed, None);
     ///     Ok(())
     /// }
     /// ```
     pub fn new(
         mint_url: &str,
         unit: CurrencyUnit,
         localstore: Arc<dyn WalletDatabase<Err = database::Error> + Send + Sync>,
-        seed: &[u8],
+        seed: Arc<[u8]>,
```
thesimplekid, line 142: I think wallet can still just take a `&[u8]` and doesn't need the `Arc`

davidcaseria:
> @davidcaseria I think this may make cdk-flutter a bit easier?

Possibly. Currently, my cdk-flutter version of MultiMintWallet is a simple container of wallets. It doesn't have functions like send or receive. The user is expected to get a Wallet and then call those functions. It works well enough right now, but if we want to change it to support those functions, this will help.

crates/cdk/src/wallet/mod.rs
ok300: Unfortunately `&[u8]` needs a lifetime specifier in

https://github.com/cashubtc/cdk/blob/ad55818288b307aaf95b3f0bb5bcc7ee380cb2a6/crates/cdk/src/wallet/multi_mint_wallet.rs#L28-L34

which then also makes `MultiMintWallet::new` need one

https://github.com/cashubtc/cdk/blob/ad55818288b307aaf95b3f0bb5bcc7ee380cb2a6/crates/cdk/src/wallet/multi_mint_wallet.rs#L36-L42

and forces the caller to deal with it in every `MultiMintWallet` call, for example in the CLI:

https://github.com/cashubtc/cdk/blob/ad55818288b307aaf95b3f0bb5bcc7ee380cb2a6/crates/cdk-cli/src/sub_commands/melt.rs#L25-L28

I would much rather keep it as `Arc<[u8]>`, as it's simpler and avoids all this.

ok300: What do you think? Would you agree?

ok300: Good idea, it worked: 73ae20952aadc876fbe15efca126126da7ad8791

crates/cdk/src/wallet/multi_mint_wallet.rs
ok300: Done: e8c024ba559cf6162f12c4602b7dc78c5552e704

PR #702, opened 2025-04-02
Keysets V2

lollerfirst:
### Description

Adding support for keysets v2.

#### CHANGED

* Replaced `From` trait implementation for `Id` with functions `v1_from_keys`, `v2_from_data`.
* struct `Id`, id field:
```rust
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub enum IdBytes {
    V1([u8; 7]),
    V2([u8; 31]),
}

pub struct Id {
    version: KeySetVersion,
    id: IdBytes,
}
```
* keyset generation functions (`generate`, `generate_from_xpriv`, `create_new_keyset`) take 2 additional arguments: an optional `final_expiry` and a `version` to identify the correct version.
* creating new keysets defaults to `KeySetVersion::Version01` (creating new keysets defaults to the newer one).

#### TODO
* `ShortIdV2` for short IDs in tokens
* Fix database interactions.

----

### Checklist


crates/cashu/src/nuts/nut02.rs
```diff
@@ -103,26 +123,110 @@ impl Id {
...
+
+        let mut pubkeys_concat: Vec<u8> = keys
+            .iter()
+            .map(|(_, pubkey)| pubkey.to_bytes())
+            .collect::<Vec<[u8; 33]>>()
+            .concat();
+
+        // Add the unit
+        pubkeys_concat.extend(b"unit:");
+        pubkeys_concat.extend(unit.to_string().as_bytes());
```
ok300, line 169: Unit should be converted to lowercase before hashing: https://github.com/cashubtc/nuts/pull/182#discussion_r1939111715

crates/cashu/src/nuts/nut02.rs
```diff
@@ -103,26 +123,110 @@ impl Id {
...
     /// [`Id`] from bytes
     pub fn from_bytes(bytes: &[u8]) -> Result<Self, Error> {
-        Ok(Self {
-            version: KeySetVersion::from_byte(&bytes[0])?,
-            id: bytes[1..].try_into()?,
-        })
+        let version = KeySetVersion::from_byte(&bytes[0])?;
+        let id = match version {
+            KeySetVersion::Version00 => IdBytes::V1(bytes[1..].try_into()?),
+            KeySetVersion::Version01 => IdBytes::V2(bytes[1..].try_into()?),
```
ok300, line 129: The Keyset ID PR talks about 2 representations of a Keyset ID V2: short ID and long ID.

There is specifically a section that says

> To save space, a `Token`, as defined in [NUT-00][00], SHOULD contain an abbreviated version of the keyset `id` in the `Proof`, (the `s_id`).

This means the `Proof`s in the `Token` could have either the short or the long representation. This means `Id::from_bytes`, which is used in the deserialization of `Token`, should be able to correctly map both a short V2 ID and a long V2 ID to the new ID type.

This would also be the case when serializing (`Id::to_bytes`), as we want to prefer generating the short ID when creating the `Token` for sending.

So in short, `Id` serialization and deserialization should handle both the short and the long type of ID V2.

ok300 (commented):
Looking good.

Some tests would be nice as well, converting to and from byte arrays of different lengths.

crates/cashu/src/nuts/nut02.rs
lollerfirst: Yes you can map an IDv2 to its shorter version, but you cannot go back. The idea is when you read the short ID from the Token you should be able to map it to its complete ID just by looking at the first 7 bytes.

lollerfirst:
So @ok300 @thesimplekid I thought we could create a new `ShortId` which will be the type in the `id` field for `TokenV4Token`. In the case of v1, it will just be the keyset ID itself, while in the case of v2 it will be the first 8 bytes of IDv2.

thesimplekid:
> So @ok300 @thesimplekid I thought we could create a new ShortId enum which will be the type of the id field for TokenV4Token. In the case of v1, it will just be the keyset ID itself, while in the case of v2 it will be the first 8 bytes of IDv2.

Yes I think that makes sense. It's similar to how we handled the transition from TokenV3 to TokenV4.

ok300:
Sounds good. Maybe `ShortKeysetId` is clearer.

lollerfirst:
@ok300 @thesimplekid 
short keyset id struct:
```rust
pub struct ShortKeysetId {
    /// The version of the short keyset
    version: KeySetVersion,
    /// The improper prefix of the keyset ID bytes
    prefix: Vec<u8>, // [u8; 7]
}
```

ID from short keyset id and mint keysets information:
```
/// Selects the correct IDv2 from a list of keysets and the given short-id
    /// or returns the short-id in the case of v1.
    pub fn from_short_keyset_id(short_id: &ShortKeysetId, keysets_info: &[KeySetInfo]) -> Result<Self, Error> {
        // Check prefix length
        if short_id.prefix.len() < Self::BYTELEN_V1 || short_id.prefix.len() > Self::BYTELEN_V2 {
            return Err(Error::MalformedShortKeysetId);
        }

        match short_id.version {
            KeySetVersion::Version00 => { 
                let mut idbytes: [u8; Self::BYTELEN_V1] = [0u8; Self::BYTELEN_V1];
                idbytes.copy_from_slice(&short_id.prefix[..Self::BYTELEN_V1]);
                Ok(Self { 
                    version: short_id.version,
                    id: IdBytes::V1(idbytes)
                })
            },
            KeySetVersion::Version01 => {
                // We return the first match or error
                for keyset_info in keysets_info.iter() {
                    if keyset_info.id.id.to_vec()[..short_id.prefix.len()] == short_id.prefix {
                        return Ok(keyset_info.id);
                    }
                }
                Err(Error::UnknownShortKeysetId)
            }
        }
    }
 ```

File: crates/cashu/src/nuts/nut02.rs [hunk 3]
```diff
@@ -103,26 +136,144 @@ impl Id {
...
+        self.version
+    }
+
+    /*
     /// [`Id`] as bytes
     pub fn as_bytes(&self) -> [u8; Self::BYTELEN + 1] {
         let mut bytes = [0u8; Self::BYTELEN + 1];
         bytes[0] = self.version.to_byte();
-        bytes[1..].copy_from_slice(&self.id);
+        match self.id {
```
ok300, line 157: As the whole `as_bytes` function is commented out, maybe we can remove it?

We still have `to_bytes`.

crates/cashu/src/nuts/nut02.rs
```diff
@@ -103,26 +136,144 @@ impl Id {
...
+            IdBytes::V2(id) => bytes[1..].copy_from_slice(&id),
+        }
         bytes
+    }*/
+
+    /// *** V2 KEYSET ***
+    /// create [`Id`] v2 from keys, unit and (optionally) expiry
+    /// 1 - sort public keys by their amount in ascending order
+    /// 2 - concatenate all public keys to one byte array
+    /// 3 - concatenate the unit string to the byte array (e.g. "unit:sat")
```
ok300, line 168:
```suggestion
    /// 3 - concatenate the lowercase unit string to the byte array (e.g. "unit:sat")
```

crates/cashu/src/nuts/nut02.rs
```diff
@@ -103,26 +136,144 @@ impl Id {
...
+
+        // Add the unit
+        pubkeys_concat.extend(b"unit:");
+        pubkeys_concat.extend(unit.to_string().to_lowercase().as_bytes());
+
+        // Add the expiration
+        if let Some(expiry) = expiry {
+            pubkeys_concat.extend(b"final_expiry:");
+            pubkeys_concat.extend(expiry.to_string().as_bytes());
+        }
```
ok300, line 190: In the Keyset ID V2 PR ( https://github.com/cashubtc/nuts/pull/182 ) doesn't say to how / if to append `final_expiry:` in case the field is empty. In fact, the python example there actually appends it with no value:

https://github.com/cashubtc/nuts/blob/9c774ff0d62616c89ef280ae67ceb93b2786223f/02.md?plain=1#L93

Maybe this should keep the same behavior?

crates/cdk-rexie/src/wallet.rs
```diff
@@ -547,16 +547,19 @@ impl WalletDatabase for WalletRexieDatabase {
         Ok(())
     }
 
-    async fn add_keys(&self, keys: Keys) -> Result<(), Self::Err> {
+    async fn add_keys(&self, keyset: KeySet) -> Result<(), Self::Err> {
         let rexie = self.db.lock().await;
 
+        // Verify ID by recomputing id
+        keyset.verify_id()?;
```
ok300, line 554: I would move this before the lock.

ok300 (commented):
Looking good, only found a couple of NITs.

lollerfirst:
Please review but do not merge yet.

crates/cashu/src/nuts/nut02.rs
lollerfirst: The way I wrote it back then is I intended `final_expiry:` to be appended only if it's actually specified. Unfortunately, the example only covers the case where it's specified.

thesimplekid: We should add some test vectors to the nut on this

thesimplekid:
Sorry this took so long to get to. Looks good just a few comments. 

FYI: https://github.com/cashubtc/cdk/pull/730 should fix the ci issue

crates/cdk-redb/src/wallet/mod.rs
```diff
@@ -493,15 +494,19 @@ impl WalletDatabase for WalletRedbDatabase {
     }
 
     #[instrument(skip_all)]
-    async fn add_keys(&self, keys: Keys) -> Result<(), Self::Err> {
+    async fn add_keys(&self, keyset: KeySet) -> Result<(), Self::Err> {
```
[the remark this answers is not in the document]
lollerfirst: Uhm yea

crates/cashu/src/nuts/nut00/token.rs
```diff
@@ -684,4 +729,101 @@ mod tests {
...
+
+        let token = Token::new(mint_url.clone(), proofs, None, CurrencyUnit::Sat);
+        let token_str = token.to_string();
+
+        let token1 = Token::from_str(&token_str);
+        assert!(token1.is_ok());
+
+        let proofs1 = token1.unwrap().proofs(&keysets_info);
+        assert!(proofs1.is_err());
+    }
```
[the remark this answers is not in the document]
lollerfirst: ok

crates/cdk-redb/src/wallet/mod.rs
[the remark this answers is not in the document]
thesimplekid: That's fine was just wondering.

thesimplekid:
merging main should fix the current CI issues, I updated the stable rust version.

lollerfirst:
I think this is good.

thesimplekid:
@crodas could you advise on how to correct the migrations with rusqlite now?

thesimplekid (approved):

PR #730, opened 2025-04-19
Fix race conditions with proof state updates.

crodas:
### Description

Add a strict set of updates to prevent incorrect state changes and correct usage. Supporting the transaction at the trait level prevented some cases, but having a strict set of state change flows is better.

This bug was found while developing the signatory. The keys are read from memory, triggering race conditions at the database, and some `Pending` states are selected (instead of just selecting `Unspent`).

This PR also introduces a set of generic database tests to be executed for all database implementations, this test suite will make sure writing and maintaining new database drivers

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/state.rs
+++ b/crates/cdk-common/src/state.rs
@@ -20,10 +20,7 @@ pub enum Error {
 /// Check if the state transition is allowed
 pub fn check_state_transition(current_state: State, new_state: State) -> Result<(), Error> {
     let is_valid_transition = match current_state {
-        State::Unspent => matches!(
-            new_state,
-            State::Pending | State::Spent
-        ),
+        State::Unspent => matches!(new_state, State::Pending | State::Spent),
         State::Pending => matches!(new_state, State::Unspent | State::Spent),
         // Any other state shouldn't be updated by the mint, and the wallet does not use this
         // function
--- a/crates/cdk-common/src/database/mod.rs
+++ b/crates/cdk-common/src/database/mod.rs
@@ -16,8 +16,6 @@ pub use mint::{
 #[cfg(feature = "wallet")]
 pub use wallet::Database as WalletDatabase;
 
-use crate::state;
-
 /// CDK_database error
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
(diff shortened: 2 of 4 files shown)


crates/cdk-common/src/state.rs
```diff
@@ -0,0 +1,42 @@
...
+/// Check if the state transition is allowed
+pub fn check_state_transition(current_state: State, new_state: State) -> Result<(), Error> {
+    let is_valid_transition = match current_state {
+        State::Unspent => matches!(
+            new_state,
+            State::Pending | State::Reserved | State::PendingSpent | State::Spent
+        ),
+        State::Pending => matches!(new_state, State::Unspent | State::Spent),
+        State::Reserved => matches!(new_state, State::Pending | State::Unspent),
+        State::PendingSpent => matches!(new_state, State::Unspent | State::Spent),
```
thesimplekid, line 29: `Reserved` and `PendingSpent` are states only used by the wallet. It looks like we're only using this check on the mint. `Reserved` can also go directly to `Spent` since its a wallet state. Maybe we should just error on these two since this is intended to be used on the mint and not wallet? If we do that we should put this mod behind the mint feature.

crates/cdk-common/src/state.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[error("Token already spent")]
+    AlreadySpent,
+    /// Invalid transition
+    #[error("Invalid transition: From {0} to {1}")]
+    InvalidTransition(State, State),
+}
+
+#[inline]
+/// Check if the state transition is allowed
+pub fn check_state_transition(current_state: State, new_state: State) -> Result<(), Error> {
```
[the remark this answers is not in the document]
thesimplekid: Yes this is a good goal. Thank you.

thesimplekid (changes requested):
Thank you this looks good just the one questions on wallet/mint states

crates/cdk-common/src/state.rs
```diff
@@ -0,0 +1,42 @@
...
+    let is_valid_transition = match current_state {
+        State::Unspent => matches!(
+            new_state,
+            State::Pending | State::Reserved | State::PendingSpent | State::Spent
```
thesimplekid, line 25:
```suggestion
            State::Pending | State::Spent
```

Since this is mint only now Reserved and PendingSpent can be removed since these are only used by the wallet.

crates/cdk-common/src/lib.rs
```diff
@@ -16,6 +16,7 @@ pub mod mint;
 #[cfg(feature = "mint")]
 pub mod payment;
 pub mod pub_sub;
+pub mod state;
```
thesimplekid, line 19:
```suggestion
#[cfg(feature = "mint")]
pub mod state;
```

thesimplekid (changes requested):

thesimplekid:
I think we should do something similar for the wallet. But lets keep this PR to the mint.

thesimplekid (approved):

PR #747, opened 2025-05-08
refactor: remove mutex

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-lnd/src/lib.rs
+++ b/crates/cdk-lnd/src/lib.rs
@@ -32,7 +32,6 @@ use fedimint_tonic_lnd::lnrpc::{FeeLimit, Hop, MppRecord};
 use fedimint_tonic_lnd::tonic::Code;
 use fedimint_tonic_lnd::Client;
 use futures::{Stream, StreamExt};
-use tokio::sync::Mutex;
 use tokio_util::sync::CancellationToken;
 use tracing::instrument;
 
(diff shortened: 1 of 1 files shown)


crates/cdk-lnd/src/lib.rs
```diff
@@ -441,10 +434,9 @@ impl MintPayment for Lnd {
...
         };
 
-        let invoice = self
-            .client
-            .lock()
-            .await
+        let mut client = self.client.clone();
+
+        let invoice = client
             .lightning()
```
KnowWhoami, line 440: Instead, we should consider `&mut self` rather than `&self` in the function definition of  the trait and thus can do: 
```suggestion
let invoice = self.client
.lightning()
```

and IMO, using a client clone in the operations is causing the IT's to fail because calling an operation on the 
 client clone would not reflect changes in the `self.client`.

crates/cdk-lnd/src/lib.rs
```diff
@@ -299,15 +300,13 @@ impl MintPayment for Lnd {
...
-                    let mut routes_response: fedimint_tonic_lnd::lnrpc::QueryRoutesResponse = self
-                        .client
-                        .lock()
-                        .await
-                        .lightning()
-                        .query_routes(route_req)
-                        .await
-                        .map_err(Error::LndError)?
-                        .into_inner();
+                    let mut routes_response: fedimint_tonic_lnd::lnrpc::QueryRoutesResponse =
```
KnowWhoami, line 303: We don't require to explicitly mention the return type since rust-analyzer automatically determines the type in this case..

KnowWhoami (changes requested):

thesimplekid:
closing because of https://github.com/cashubtc/cdk/issues/748

PR #754, opened 2025-05-15
[FEAT] Golomb-encoded Set Filter

lollerfirst:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/Cargo.toml
+++ b/crates/cdk/Cargo.toml
@@ -42,10 +42,6 @@ utoipa = { workspace = true, optional = true }
 uuid.workspace = true
 jsonwebtoken = { workspace = true, optional = true }
 
-# -Z minimal-versions
-sync_wrapper = "0.1.2"
-bech32 = "0.9.1"
-
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { workspace = true, features = [
     "rt-multi-thread",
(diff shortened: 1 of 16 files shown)


crates/cdk/Cargo.toml
```diff
@@ -41,6 +41,10 @@ url.workspace = true
 utoipa = { workspace = true, optional = true }
 uuid.workspace = true
 jsonwebtoken = { workspace = true, optional = true }
+bitvec = "1.0.1"
+hex = "0.4.3"
```
thesimplekid: We shouldn't need to import hex, instead we can use out hex mod https://github.com/cashubtc/cdk/blob/main/crates/cashu/src/util/hex.rs

crates/cdk/Cargo.toml
```diff
@@ -41,6 +41,10 @@ url.workspace = true
...
+base64 = "0.22.1"
```
thesimplekid: I think bitcoin re-exports base64 maybe we can just use that?

crates/cdk/src/mint/mod.rs
```diff
@@ -301,6 +305,84 @@ impl Mint {
...
+                                    },
+                                    Ok(compressed_set) => {
+                                        let gcs_filter = cdk_common::common::GCSFilter {
+                                            num_items: compressed_set.len() as u32,
+                                            content: compressed_set,
+                                            m,
+                                            p,
+                                            time: unix_time() as i64,
+                                        };
+                                        if spent_filter.unwrap().is_some() {
```
thesimplekid: Can we remove the unwrap here. Can change to an expect since its checked above.

PR #774, opened 2025-05-29
Reclaim unspent proofs by reverting transaction

davidcaseria:
### Description

Allow a wallet user to reclaim unspent proofs by reverting a transaction. If the proofs are successfully reclaimed, the wallet will delete the transaction from the database.

-----

### Notes to the reviewers

-----

#### ADDED
- `Wallet::revert_transaction`: reclaim unspent proofs by reverting transaction

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/transactions.rs
+++ b/crates/cdk/src/wallet/transactions.rs
@@ -28,4 +28,29 @@ impl Wallet {
 
         Ok(transaction)
     }
+
+    /// Revert a transaction
+    pub async fn revert_transaction(&self, id: TransactionId) -> Result<bool, Error> {
+        let tx = self
+            .localstore
+            .get_transaction(id)
+            .await?
+            .ok_or(Error::TransactionNotFound)?;
+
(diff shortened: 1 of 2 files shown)


crates/cdk/src/wallet/transactions.rs
```diff
@@ -28,4 +28,29 @@ impl Wallet {
...
+            .get_pending_spent_proofs()
+            .await?
+            .into_iter()
+            .filter(|p| match p.y() {
+                Ok(y) => tx.ys.contains(&y),
+                Err(_) => false,
+            })
+            .collect::<Vec<_>>();
+
+        Ok(self.reclaim_unspent(pending_spent_proofs).await.is_ok())
```
thesimplekid, line 54:
```suggestion
        self.reclaim_unspent(pending_spent_proofs).await?
        Ok(())
```

Do we need the bool or would an ok be enough?

thesimplekid (approved):

PR #778, opened 2025-05-31
refactor mintd main function

asmogo:
### Description

Refactored the mintd main function. It should behave the same as before, but it’s now more readable?! 

-----

### Notes to the reviewers

I did not test every possible feature. It's currently running with a fake wallet and sqlite. 

The `use std::path::{Path}` statement was introduced to resolve warnings in `just final-check`

-----

#### FIXED

----

### Checklist


crates/cdk-mintd/src/main.rs
```diff
@@ -266,27 +383,17 @@ async fn main() -> anyhow::Result<()> {
 
             for unit in fake_wallet.clone().supported_units {
                 let fake = fake_wallet
-                    .setup(&mut ln_routers, &settings, CurrencyUnit::Sat)
-                    .await
-                    .expect("hhh");
```
asmogo, line 259: this is gone

crates/cdk-mintd/src/main.rs
```diff
@@ -137,22 +137,27 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
+trait MintCombinedDatabase: MintDatabase<cdk_database::Error> + MintKeysDatabase<Err = cdk_database::Error> {}
+
+// Implement the combined trait
+impl<T> MintCombinedDatabase for T where
+    T: MintDatabase<cdk_database::Error> + MintKeysDatabase<Err = cdk_database::Error> {}
+
```
asmogo: not sure if that's ok. This will be moved to `cdk_common::database`

crates/cdk/src/wallet/mod.rs
```diff
@@ -418,6 +418,11 @@ impl Wallet {
                     .filter(|p| response.outputs.contains(&p.blinded_message))
                     .collect();
 
+                let premint_secrets: Vec<_> = premint_secrets
+                    .iter()
+                    .filter(|p| response.outputs.contains(&p.blinded_message))
+                    .collect();
+
```
thesimplekid: Think this just came back on a merge. We do the same filter just above.
```suggestion
```

crates/cdk-mintd/src/main.rs
```diff
@@ -112,56 +181,126 @@ async fn main() -> anyhow::Result<()> {
...
+            let localstore: Arc<dyn MintDatabase<cdk_database::Error> + Send + Sync> = db.clone();
+            let keystore: Arc<dyn MintKeysDatabase<Err = cdk_database::Error> + Send + Sync> = db;
+            Ok((localstore, keystore))
         }
+    }
+}
+
+async fn setup_sqlite_database(
+    work_dir: &Path,
+    _password: String,
```
thesimplekid: Maybe its better if we use an `Option` here instead of the empty string?

thesimplekid (commented):
Thank this looks good. Just a couple small comments and once the ci runs we can merge. Much cleaner this way then the one massive fn we had before.

crates/cdk-mintd/src/main.rs
asmogo: done here 
https://github.com/cashubtc/cdk/pull/778/commits/fb5e49c6d2dbd04290e39c8f807031ae4d20d2b2

crates/cdk/src/wallet/mod.rs
asmogo: Yep sorry. This was not intentional and is now [removed](https://github.com/cashubtc/cdk/pull/778/commits/79c4a53f8ed05d0cdb4b1e34542ec1274d2847d4).

Issue #606, opened 2025-02-18
Remove `pub` properties from structs


crodas:
Follow up:

_Originally posted by @thesimplekid in https://github.com/cashubtc/cdk/issues/564#issuecomment-2665151917_
            
Remove all the `pub` from non-payload or config entries properties. Everywhere where they could be updated with consequences outside of their logic, they should be private and exposed with a getter or a Deref trait, with their log the update hidden in the struct.

thesimplekid:
Its not top priority lets prioritize wallet tx and kv store.

thesimplekid:
@crodas assigning you as we said we would do it as part of the wallet trait.

TheMhv:
I can do that but i'll need more details

PR #782, opened 2025-06-02
Secret remove pub properties 

asmogo:
### Description

ref #606

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
   self-care:
     name: Flake self-check
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
       - name: Check Nix flake inputs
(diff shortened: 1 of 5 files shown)


crates/cashu/src/nuts/nut10.rs
```diff
@@ -34,21 +34,38 @@ pub enum Kind {
...
-    pub data: String,
+    data: String,
     /// Additional data committed to and can be used for feature extensions
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub tags: Option<Vec<Vec<String>>>,
+    tags: Option<Vec<Vec<String>>>,
+}
+
+impl SecretData {
+    /// Get the nonce
```
thesimplekid:
```suggestion
impl SecretData {
    /// Create new [`SecretData`]
    pub fn new<S, V>(data: S, tags: Option<V>) -> Self
    where
        S: Into<String>,
        V: Into<Vec<Vec<String>>>,
{
    let nonce = crate::secret::Secret::generate().to_string();
    
    Self {
            nonce,
            data: data.into(),
            tags: tags.map(|v| v.into()),
    }
}
```

I think it makes sense to add a new fn on `SecretData` and then we can call that from `Secret::new` instead of generating the nonce there.

PR #783, opened 2025-06-02
Migrate from `sqlx` to rusqlite

### LINKED ISSUE #735: Remove the use of sqlx for rusqlite
We would like to have ldk-node as a first class supported backend, however currenlty we have a conflict the sql version of sqx and that used by ldk-node. To fix this conflict we should move from sqlx to rusqlite matching the version used in ldk-node

crodas:
### Description

Fixes #735

1. Add rusqlite with rusqlite with a working thread
2. Add wallet without a thread (synchronous)
3. Add custom migration

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-sqlite/src/mint/async_rusqlite.rs
```diff
@@ -0,0 +1,400 @@
+use std::marker::PhantomData;
+use std::sync::Arc;
+//use std::sync::atomic::AtomicUsize;
+//use std::sync::Arc;
```
thesimplekid:
```suggestion
```

crates/cdk-sqlite/src/mint/async_rusqlite.rs
```diff
@@ -0,0 +1,400 @@
...
+use crate::mint::Error;
+use crate::pool::Pool;
+use crate::stmt::{Column, ExpectedSqlResponse, Statement as InnerStatement, Value};
+
+const BUFFER_REQUEST_SIZE: usize = 10_000;
+
+#[derive(Debug, Clone)]
+pub struct AsyncRusqlite {
+    sender: mpsc::Sender<DbRequest>,
+    //inflight_requests: Arc<AtomicUsize>,
```
thesimplekid:
```suggestion
```

crates/cdk-sqlite/src/mint/async_rusqlite.rs
```diff
@@ -0,0 +1,400 @@
...
+                                Ok(ok) => reply_to.send(ok),
+                                Err(err) => reply_to.send(DbResponse::Error(err)),
+                            };
+                        }
+                    }
+                }
+
+                drop(conn);
+            }
+            DbRequest::Commit(reply_to) => {
```
thesimplekid, line 330: Could we just add some logging to the unexpected and error cases?

crates/cdk-sqlite/src/mint/async_rusqlite.rs
```diff
@@ -0,0 +1,400 @@
...
+impl AsyncRusqlite {
+    pub fn new(pool: Arc<Pool<SqliteConnectionManager>>) -> Self {
+        let (sender, receiver) = mpsc::channel(BUFFER_REQUEST_SIZE);
+        spawn(move || {
+            rusqlite_worker(receiver, pool);
+        });
+
+        Self {
+            sender,
+            //inflight_requests: Arc::new(0.into()),
```
thesimplekid:
```suggestion
```

crates/cdk-sqlite/src/mint/auth/mod.rs
```diff
@@ -290,70 +212,56 @@ VALUES (?, ?, ?, ?, ?);
...
 
-        transaction.commit().await.map_err(Error::from)?;
         Ok(current_state)
     }
 
     async fn add_blind_signatures(
         &self,
         blinded_messages: &[PublicKey],
-        blind_signatures: &[BlindSignature],
+        blinded_signatures: &[BlindSignature],
```
thesimplekid:
```suggestion
        blind_signatures: &[BlindSignature],
```
Signatures aren't blinded they are blind, as the wallet is the one that does the blinding so messages are blinded by the wallet and signed blind by the mint.

crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -398,634 +309,505 @@ ON CONFLICT(request_lookup_id) DO UPDATE SET
...
+                unit = excluded.unit,
+                amount = excluded.amount,
+                request = excluded.request,
+                fee_reserve = excluded.fee_reserve,
+                state = excluded.state,
+                expiry = excluded.expiry,
+                payment_preimage = excluded.payment_preimage,
+                id = excluded.id,
+                created_time = excluded.created_time,
+                paid_time = excluded.paid_time;
```
thesimplekid, line 594: i know we did this before as well and just think it never excuted so we didnt see an error but I don't think that two `ON CONFLICT` statements is valid sql. I think we should just error if there is a conflict on the id as the update state should be used?

crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -398,634 +309,505 @@ ON CONFLICT(request_lookup_id) DO UPDATE SET
...
-            r#"UPDATE melt_quote SET state = ?, paid_time = ? WHERE id = ?"#
+            r#"UPDATE melt_quote SET state = :state, paid_time = :paid_time WHERE id = :id"#
         } else {
-            r#"UPDATE melt_quote SET state = ? WHERE id = ?"#
+            r#"UPDATE melt_quote SET state = :state WHERE id = :id"#
         };
 
         let current_time = unix_time();
 
         let rec = if state == MeltQuoteState::Paid {
```
thesimplekid, line 708: We have this same if statement twice, can we combine them?

crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -1226,114 +945,57 @@ WHERE quote_id=?;
...
-                    })
-            })
-            .collect::<Result<HashMap<_, _>, _>>()?;
-
-        let states = current_states.values().collect::<HashSet<_>>();
-
-        for state in states {
-            check_state_transition(*state, proofs_state)?;
+        for state in current_states.values() {
+            check_state_transition(*state, new_state)?;
```
thesimplekid, line 982: Should we call an explicit rollback here or just rely on the transaction being dropped?

crates/cdk-sqlite/src/stmt.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// The list of arguments for the placeholders. It only supports named arguments for simplicity
+    /// sake
+    pub args: Vec<(String, Value)>,
+    /// The expected response type
+    pub expected_response: ExpectedSqlResponse,
+}
+
+impl Statement {
+    /// Creates a new statement
+    pub fn new<T: ToString>(sql: T) -> Self {
```
thesimplekid: Can we change the trait bonds to use `where`

https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#generics

crates/cdk-sqlite/Cargo.toml
```diff
@@ -16,23 +16,17 @@ default = ["mint", "wallet", "auth"]
...
 bitcoin.workspace = true
-sqlx = { version = "0.7.4", default-features = false, features = [
-    "runtime-tokio-rustls",
-    "sqlite",
-    "macros",
-    "migrate",
-    "uuid",
-] }
-libsqlite3-sys = { version = "0.27.0", features = ["bundled-sqlcipher"], optional = true }
+rusqlite = { version = "0.27", features = ["bundled"]}
```
thesimplekid:
```suggestion
rusqlite = { version = "0.31", features = ["bundled"]}
```

ldk-node is currently usening 31

crates/cdk-sqlite/src/mint/async_rusqlite.rs
```diff
@@ -0,0 +1,400 @@
+use std::marker::PhantomData;
+use std::sync::Arc;
+//use std::sync::atomic::AtomicUsize;
+//use std::sync::Arc;
+use std::thread::spawn;
```
thesimplekid, line 4: We're already using tokio, should we use a `spawn_blocking` or do you think the std thread makes more sense since its long lived?

crates/cdk-sqlite/src/mint/mod.rs
crodas: Transactions are rollback on drop, but we can be explicit as well.

crates/cdk-sqlite/src/mint/async_rusqlite.rs
crodas: I think it is better to have a dedicated OS thread, since we are not giving back the control to tokio.

crates/cdk-sqlite/Cargo.toml
crodas: Fixed in 1f831426

crates/cdk-sqlite/src/stmt.rs
crodas: Fixed in 8f1ce122

crates/cdk-sqlite/src/mint/mod.rs
crodas: Fixed in 1b644ceb

crates/cdk-sqlite/src/mint/async_rusqlite.rs
crodas: Fixed in 9cab3369

thesimplekid:
I realized currently we're using sqlx-cli to create the migration files, just black sql file with named with that datetime format. Since we're removing sqlx can we remove sqlx-cli and add a simple just command to create files in this format?

crates/cdk-sqlite/src/mint/async_rusqlite.rs
```diff
@@ -0,0 +1,519 @@
...
+        InnerStatement,
+        oneshot::Sender<DbResponse>,
+    )>();
+    let receiver = Arc::new(Mutex::new(receiver));
+
+    for _ in 0..threads {
+        let rx = receiver.clone();
+        let inflight_requests = inflight_requests.clone();
+        spawn(move || loop {
+            while let Ok((conn, sql, reply_to)) = rx.lock().unwrap().recv() {
```
thesimplekid: Can we make this an expect with why it's safe instead of an unwrap?

crates/cdk-sqlite/src/mint/auth/mod.rs
```diff
@@ -1,52 +1,48 @@
...
-            .connect_with(db_options)
-            .await?;
-
-        Ok(Self { pool })
-    }
+    pub async fn new<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
+        #[cfg(feature = "sqlcipher")]
+        let pool = create_sqlite_pool(
+            path.as_ref().to_str().ok_or(Error::InvalidDbPath)?,
+            "".to_owned(),
```
thesimplekid: Don't we need to pass a real password here?

crates/cdk-sqlite/src/mint/async_rusqlite.rs
```diff
@@ -0,0 +1,519 @@
...
+                })
+                .transpose()?;
+            DbResponse::Row(row)
+        }
+    };
+
+    let duration = start.elapsed();
+
+    if duration.as_millis() > SLOW_QUERY_THRESHOLD_MS {
+        tracing::error!("[SLOW QUERY] Took {} ms: {}", duration.as_millis(), sql.sql);
```
thesimplekid:
```suggestion
        tracing::warn!("[SLOW QUERY] Took {} ms: {}", duration.as_millis(), sql.sql);
```

thesimplekid (changes requested):
We need to support upgrading from a current mint or wallet to rusqlite. When i do this with the wallet i receive the below error: 
image
And with the mint: 
image

crates/cdk-sqlite/src/mint/async_rusqlite.rs
crodas: If it panic in any of the working thread (which it shouldn't) without releasing the lock it would be [poisoned](https://doc.rust-lang.org/nomicon/poisoning.html).

I'll add a way to error nicely

crates/cdk-sqlite/src/mint/auth/mod.rs
crodas: This is how it is currently, but I'll fix it.

thesimplekid:
@crodas Can you request my review once you have this ready working with upgrading a mint?

thesimplekid (approved):

thesimplekid:
Great LGTM

ACK 032d6e66bc5b5d3428549fce919a26c89972ee49

PR #793, opened 2025-06-06
Perform a swap before melting by default.

### LINKED ISSUE #465: Wallet melt should not over pay risking ecash stuck in pending.
When melting the wallet can over pay the melt quote and have the change returned once the payment is complete. However in an error cash this can lead to ecash being stuck in a pending state. The wallet should select proofs and swap before a melt if it does not have proofs equal to the melt quote. Swapping before increases fees and requires an extra operation but is preferable to risking stuck ecash. 

@davidcaseria just tagging you for tracking as we've talked about this before, but don't think we've fixed it.

crodas:
### Description

Fixes #465

This PR modifies the default `melt` operation, performing a swap to melt exact amounts instead of overpaying and getting the exchange token.

Although this may end up costing more in fees, it is more efficient sometimes since the overpaid amount is not unusable until the melt is finalized.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/melt.rs
+++ b/crates/cdk/src/wallet/melt.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::str::FromStr;
 
+use cdk_common::amount::SplitTarget;
 use cdk_common::wallet::{Transaction, TransactionDirection};
 use lightning_invoice::Bolt11Invoice;
 use tracing::instrument;
@@ -313,13 +314,31 @@ impl Wallet {
             .map(|k| k.id)
             .collect();
         let keyset_fees = self.get_keyset_fees().await?;
-        let input_proofs = Wallet::select_proofs(
+        let (mut input_proofs, mut exchange) = Wallet::select_exact_proofs(
             inputs_needed_amount,
             available_proofs,
             &active_keyset_ids,
             &keyset_fees,
             true,
         )?;
+
+        if let Some((proof, exact_amount)) = exchange.take() {
+            if let Ok(Some(new_proofs)) = self
+                .swap(
+                    Some(exact_amount),
+                    SplitTarget::None,
+                    vec![proof.clone()],
+                    None,
+                    false,
+                )
+                .await
+            {
+                input_proofs.extend_from_slice(&new_proofs);
+            } else {
+                // swap failed, add it back to the original set of profos
+                input_proofs.push(proof);
+            }
+        }
 
         self.melt_proofs(quote_id, input_proofs).await
     }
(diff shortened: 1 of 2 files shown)


crates/cdk/src/wallet/melt.rs
```diff
@@ -313,14 +314,32 @@ impl Wallet {
...
+        let (mut input_proofs, mut exchange) = Wallet::select_exact_proofs(
             inputs_needed_amount,
             available_proofs,
             &active_keyset_ids,
             &keyset_fees,
             true,
         )?;
 
+        if let Some((proof, exact_amount)) = exchange.take() {
+            if let Ok(Some(new_proofs)) = self
```
thesimplekid: I think we should error here and not silently fail.

crates/cdk/src/wallet/proofs.rs
```diff
@@ -176,6 +176,63 @@ impl Wallet {
         Ok(balance)
     }
 
+    /// Select exact proofs
+    ///
+    /// This function is similar to `select_proofs` but it the selected proofs will not exceed the
+    /// requested Amount, it will include a Proof and the exacto amount needed form that Proof to
```
thesimplekid, line 182:
```suggestion
    /// requested Amount, it will include a Proof and the exact amount needed for that Proof to
```

crates/cdk/src/wallet/melt.rs
crodas: We are not failing. Instead, we are doing the melt, overpaying.  But we could also error.

crates/cdk/src/wallet/proofs.rs
```diff
@@ -176,6 +176,63 @@ impl Wallet {
...
 
+    /// Select exact proofs
+    ///
+    /// This function is similar to `select_proofs` but it the selected proofs will not exceed the
+    /// requested Amount, it will include a Proof and the exact amount needed for that Proof to
+    /// perform a swap.
+    ///
+    /// The intent is to perform a swap with info, or include the Proof as part of the return if the
+    /// swap is not needed or if the swap failed.
+    pub fn select_exact_proofs(
```
crodas, line 187: I'll write unit tests for this functions later

crates/cdk/src/wallet/melt.rs
thesimplekid: the swap silently fails and then we over pay the melt, when the expected behavior is a swap before the melt to avoid overpaying.

thesimplekid:
cc @davidcaseria

thesimplekid (approved):
LGTM

PR #799, opened 2025-06-08
Updated test matrix

asmogo:
### Description

Moved the test matrix from the GitHub workflow into a shell script. This will reduce pipeline complexity and enable running the tests on your local computer

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,99 @@ jobs:
         uses: Swatinem/rust-cache@v2
       - name: Test
         run: nix develop -i -L .#stable --command just itest ${{ matrix.database }}
-          
+
+  msrv-build:
+    name: "MSRV build"
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    needs: [ pre-commit-checks, clippy, pure-itest ]
+    strategy:
+      matrix:
+        build-args:
+          [
+            -p cashu --no-default-features --features "wallet mint",
+            -p cdk-common --no-default-features --features "wallet mint",
+            -p cdk,
+            -p cdk --no-default-features --features "mint auth",
+            -p cdk --no-default-features --features "wallet auth",
+            -p cdk --no-default-features --features "http_subscription",
+            -p cdk-axum,
+            -p cdk-axum --no-default-features --features redis,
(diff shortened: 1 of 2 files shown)


thesimplekid (dismissed):

.github/workflows/ci.yml
```diff
@@ -32,314 +42,102 @@ jobs:
...
       - name: checkout
         uses: actions/checkout@v4
       - name: Install Nix
         uses: DeterminateSystems/nix-installer-action@v17
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
-      - name: Build
-        run: nix develop -i -L .#msrv --command cargo build ${{ matrix.build-args }}
+      - name: Add WASM target
+        run: nix develop -i -L .#stable --command rustup target add wasm32-unknown-unknown
```
thesimplekid: This shouldn't be needed the wasm target should be included in the nix shell by default.

asmogo: removed this here  https://github.com/cashubtc/cdk/pull/799/commits/d52fd75d354ff4a16b317d562c66510cb15c8f2a

thesimplekid (dismissed):

thesimplekid:
Just approving to trigger ci. I have not gone through this in detail.

thesimplekid:
Sorry just realized this moved from draft. Should we go over it on the call tomorrow? I've added some nix caching to the ci so looks like there is a conflict we need to resolve now.

asmogo:
> Sorry just realized this moved from draft. Should we go over it on the call tomorrow? I've added some nix caching to the ci so looks like there is a conflict we need to resolve now.  

Oh no, we're good. Just moved the ticket out of draft for testing the approval. I will not be able to join the call tomorrow. Aiming for next week.

thesimplekid (dismissed):

thesimplekid (dismissed):

asmogo:
We’re running `cargo check —example` and `cargo run —example`. I’m not sure why we have this distinction.

I’ll add more changes to avoid repeating the same workflows.

We can speed up the integration test by moving some parts back to the ci.yaml file.

PR #806, opened 2025-06-11
Wallet events

crodas:
### Description

Add support for wallet events. The idea is borrowed from [cdk_flutter](https://github.com/davidcaseria/cdk_flutter/blob/master/rust/src/api/wallet.rs#L99-L115)

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-axum/src/ws/subscribe.rs
+++ b/crates/cdk-axum/src/ws/subscribe.rs
@@ -20,7 +20,7 @@ pub(crate) async fn handle(
     let mut subscription = context
         .state
         .mint
-        .pubsub_manager
+        .pubsub_manager()
         .try_subscribe(params)
         .await
         .map_err(|_| WsError::ParseError)?;
--- a/crates/cdk-cli/src/sub_commands/cat_device_login.rs
+++ b/crates/cdk-cli/src/sub_commands/cat_device_login.rs
@@ -82,7 +82,7 @@ async fn get_device_code_token(mint_info: &MintInfo, client_id: &str) -> (String
         .expect("Nut21 defined")
         .openid_discovery;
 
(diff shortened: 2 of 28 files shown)


crates/cdk/src/wallet/events.rs
```diff
@@ -0,0 +1,220 @@
...
+                Some(unit.clone()),
+                Some(vec![State::Reserved]),
+                None,
+            );
+            let pending = db.get_proofs(
+                Some(mint_url.clone()),
+                Some(unit.clone()),
+                Some(vec![State::Pending]),
+                None,
+            );
```
thesimplekid, line 166: I wonder if querying the db like this for each balance update is the most efficient way to do this? In a fast db like we have now its probably okay but in a slow db or worst case something like a nip60 wallet this could be very inefficient to do after each call that may change the balance. I haven't gone through all the cases where this is used but my guess is we should know how much a balance is increased or decreased by, maybe we should just send that? Then if the wallet wants to it can call get_balance or just relay on the notification? 

Also as currently written there is potential for a race condition as the db is not locked across the 3 querys.

crates/cdk/src/wallet/events.rs
```diff
@@ -0,0 +1,220 @@
...
+/// Event types
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Event {
+    Balance(MintUrl, CurrencyUnit, State, Amount),
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub enum EventId {
+    Balance(MintUrl, CurrencyUnit, State),
+}
```
thesimplekid, line 33: Do we need both?

thesimplekid:
This is very out of date so going to close.

PR #820, opened 2025-06-15
Remove multiple ON CONFLICT from SQLite

### LINKED ISSUE #810: SQlite does not support multiple on conflicts
i don't believe sqlite supports multiple on conflict statements. I do not believe we need either and should error in the case of a conflict as the only change to the quote is should be via the update quote_state so we should just error when there is a conflict.

https://github.com/cashubtc/cdk/blob/5a6b28816ad791217650f3ca6c74d0e4164494e8/crates/cdk-sqlite/src/mint/mod.rs#L290-L309

crodas:
### Description

Fixes #810

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sqlite/src/mint/mod.rs
+++ b/crates/cdk-sqlite/src/mint/mod.rs
@@ -279,35 +279,15 @@ impl MintQuotesDatabase for MintSqliteDatabase {
     async fn add_mint_quote(&self, quote: MintQuote) -> Result<(), Self::Err> {
         query(
             r#"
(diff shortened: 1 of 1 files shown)


thesimplekid (dismissed):

thesimplekid:
ACK bd7f046e3196bd939ec1eb601a60b0e36d942d74 just waiting for ci to pass

thesimplekid:
ACK 5ff30d879c557a16291e4b89c86182e6e6583087

PR #822, opened 2025-06-15
Add 10 second busy timeout

### LINKED ISSUE #814: On main ci failed with sqlite db locked
Image

crodas:
### Description

Fixes #814

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sqlite/src/mint/mod.rs
+++ b/crates/cdk-sqlite/src/mint/mod.rs
@@ -816,6 +816,26 @@ impl MintProofsDatabase for MintSqliteDatabase {
 
         let current_time = unix_time();
 
+        // Check any previous proof, this query should return None in order to proceed storing
+        // Any result here would error
+        match query(r#"SELECT state FROM proof WHERE y IN (:ys) LIMIT 1"#)
+            .bind_vec(
+                ":ys",
+                proofs
+                    .iter()
+                    .map(|y| y.y().map(|y| y.to_bytes().to_vec()))
+                    .collect::<Result<_, _>>()?,
+            )
+            .pluck(&transaction)
+            .await?
+            .map(|state| Ok::<_, Error>(column_as_string!(&state, State::from_str)))
+            .transpose()?
+        {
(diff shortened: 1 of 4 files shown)


crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -852,10 +852,17 @@ impl MintProofsDatabase for MintSqliteDatabase {
...
             r#"
-            DELETE FROM proof WHERE y IN (:ys) AND state != 'SPENT'
+            DELETE FROM proof WHERE y IN (:ys) AND state NOT IN (:exclude_state)
             "#,
         )
         .bind_vec(":ys", ys.iter().map(|y| y.to_bytes().to_vec()).collect())
+        .bind_vec(
+            ":exclude_state",
+            vec![
+                QuoteState::Paid.to_string(),
```
thesimplekid:
```suggestion
                QuoteState::Spent.to_string(),
```

thesimplekid (changes requested):

crates/cdk/src/mint/melt.rs
```diff
@@ -347,9 +341,19 @@ impl Mint {
...
+        if let Some(err) = self
+            .localstore
             .add_proofs(melt_request.inputs().clone(), None)
-            .await?;
+            .await
+            .err()
+        {
+            match err {
+                cdk_common::database::Error::Duplicate => {
+                    // the proofs already exits, it will be errored by `check_ys_spendable`
```
thesimplekid: Should we just error here instead of checking spendable?

crodas: Yes, but check_ys_spendable has the correct error. Should I duplicate logic?

crodas: Is it better with 150f2596

thesimplekid: Yes thanks, I think it's better to error there then rely on an error elsewhere

PR #826, opened 2025-06-20
Split the database trait into read and transactions.

crodas:
### Description

The transaction traits will encapsulate all database changes and also expect READ-and-lock operations to read and lock records from the database for exclusive access, thereby avoiding race conditions.

The Transaction trait expects a `rollback` operation on Drop unless the transaction has been committed. 

This PR also includes work to remove redb #787 

TODO:
 - [ ] Add a set of tests for the database operations
 - [x] Move all INSERT, UPDATE and DELETE to their transaction traits
 - [x] Implement Rollback on `DROP`, although that should be already supported by  our rusqlite wrapper
-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/database/mint/mod.rs
+++ b/crates/cdk-common/src/database/mint/mod.rs
@@ -21,7 +21,7 @@ mod auth;
 pub mod test;
 
 #[cfg(feature = "auth")]
-pub use auth::MintAuthDatabase;
+pub use auth::{MintAuthDatabase, MintAuthTransaction};
 
 /// KeysDatabaseWriter
 #[async_trait]
--- a/crates/cdk-common/src/database/mint/mod.rs
+++ b/crates/cdk-common/src/database/mint/mod.rs
@@ -83,6 +83,14 @@ pub trait QuotesTransaction<'a> {
     ) -> Result<Option<mint::MeltQuote>, Self::Err>;
     /// Add [`mint::MeltQuote`]
     async fn add_melt_quote(&mut self, quote: mint::MeltQuote) -> Result<(), Self::Err>;
+
+    /// Updates the request lookup id for a melt quote
+    async fn update_melt_quote_request_lookup_id(
+        &mut self,
+        quote_id: &Uuid,
+        new_request_lookup_id: &str,
+    ) -> Result<(), Self::Err>;
+
     /// Update [`mint::MeltQuote`] state
     ///
     /// It is expected for this function to fail if the state is already set to the new state
--- a/crates/cdk-common/src/database/mint/mod.rs
+++ b/crates/cdk-common/src/database/mint/mod.rs
@@ -149,6 +149,13 @@ pub trait ProofsTransaction<'a> {
         ys: &[PublicKey],
         proofs_state: State,
     ) -> Result<Vec<Option<State>>, Self::Err>;
+
+    /// Remove [`Proofs`]
+    async fn remove_proofs(
+        &mut self,
+        ys: &[PublicKey],
+        quote_id: Option<Uuid>,
+    ) -> Result<(), Self::Err>;
 }
 
 /// Mint Proof Database trait
--- a/crates/cdk-common/src/database/mint/mod.rs
+++ b/crates/cdk-common/src/database/mint/mod.rs
@@ -67,7 +67,7 @@ pub trait QuotesTransaction<'a> {
     async fn get_mint_quote(&mut self, quote_id: &Uuid)
         -> Result<Option<MintMintQuote>, Self::Err>;
     /// Add [`MintMintQuote`]
-    async fn add_mint_quote(&mut self, quote: MintMintQuote) -> Result<(), Self::Err>;
+    async fn add_or_replace_mint_quote(&mut self, quote: MintMintQuote) -> Result<(), Self::Err>;
     /// Update state of [`MintMintQuote`]
     async fn update_mint_quote_state(
         &mut self,
--- a/crates/cdk-common/src/database/mint/mod.rs
+++ b/crates/cdk-common/src/database/mint/mod.rs
@@ -23,24 +23,81 @@ pub mod test;
 #[cfg(feature = "auth")]
 pub use auth::MintAuthDatabase;
 
+/// KeysDatabaseWriter
+#[async_trait]
+pub trait KeysDatabaseTransaction<'a, Error>: DbTransactionFinalizer<Err = Error> {
+    /// Add Active Keyset
+    async fn set_active_keyset(&mut self, unit: CurrencyUnit, id: Id) -> Result<(), Error>;
+
+    /// Add [`MintKeySetInfo`]
+    async fn add_keyset_info(&mut self, keyset: MintKeySetInfo) -> Result<(), Error>;
+}
+
 /// Mint Keys Database trait
 #[async_trait]
 pub trait KeysDatabase {
     /// Mint Keys Database Error
     type Err: Into<Error> + From<Error>;
 
-    /// Add Active Keyset
-    async fn set_active_keyset(&self, unit: CurrencyUnit, id: Id) -> Result<(), Self::Err>;
+    /// Beings a transaction
+    async fn begin_transaction<'a>(
+        &'a self,
+    ) -> Result<Box<dyn KeysDatabaseTransaction<'a, Self::Err> + Send + Sync + 'a>, Error>;
+
     /// Get Active Keyset
     async fn get_active_keyset_id(&self, unit: &CurrencyUnit) -> Result<Option<Id>, Self::Err>;
+
     /// Get all Active Keyset
     async fn get_active_keysets(&self) -> Result<HashMap<CurrencyUnit, Id>, Self::Err>;
-    /// Add [`MintKeySetInfo`]
-    async fn add_keyset_info(&self, keyset: MintKeySetInfo) -> Result<(), Self::Err>;
+
     /// Get [`MintKeySetInfo`]
     async fn get_keyset_info(&self, id: &Id) -> Result<Option<MintKeySetInfo>, Self::Err>;
+
     /// Get [`MintKeySetInfo`]s
     async fn get_keyset_infos(&self) -> Result<Vec<MintKeySetInfo>, Self::Err>;
+}
+
+/// Mint Quote Database writer trait
+#[async_trait]
+pub trait QuotesTransaction<'a> {
+    /// Mint Quotes Database Error
+    type Err: Into<Error> + From<Error>;
+
+    /// Get [`MintMintQuote`] and lock it for update in this transaction
+    async fn get_mint_quote(&mut self, quote_id: &Uuid)
+        -> Result<Option<MintMintQuote>, Self::Err>;
+    /// Add [`MintMintQuote`]
+    async fn add_mint_quote(&mut self, quote: MintMintQuote) -> Result<(), Self::Err>;
+    /// Update state of [`MintMintQuote`]
+    async fn update_mint_quote_state(
+        &mut self,
+        quote_id: &Uuid,
+        state: MintQuoteState,
+    ) -> Result<MintQuoteState, Self::Err>;
+    /// Remove [`MintMintQuote`]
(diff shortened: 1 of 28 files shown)


crates/cdk-signatory/src/common.rs
```diff
@@ -25,13 +25,14 @@ pub async fn init_keysets(
     // Get keysets info from DB
     let keysets_infos = localstore.get_keyset_infos().await?;
 
+    let mut tx = localstore.begin_transaction().await.expect("begin");
```
thesimplekid:
```suggestion
    let mut tx = localstore.begin_transaction().await?;
```

crates/cdk-signatory/src/common.rs
```diff
@@ -104,15 +105,17 @@ pub async fn init_keysets(
...
+                    tx.add_keyset_info(keyset_info).await?;
+                    tx.set_active_keyset(unit.clone(), id).await?;
                     active_keysets.insert(id, keyset);
                     active_keyset_units.push(unit.clone());
                 };
             }
         }
     }
 
+    tx.commit().await.expect("commit");
```
thesimplekid:
```suggestion
    tx.commit().await?;
```

crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -300,103 +344,170 @@ impl MintQuotesDatabase for MintSqliteDatabase {
...
+                unit = excluded.unit,
+                amount = excluded.amount,
+                request = excluded.request,
+                fee_reserve = excluded.fee_reserve,
+                state = excluded.state,
+                expiry = excluded.expiry,
+                payment_preimage = excluded.payment_preimage,
+                id = excluded.id,
+                created_time = excluded.created_time,
+                paid_time = excluded.paid_time;
```
thesimplekid: Think this just came back in on a rebase? We can't have multiple on conflicts. 

What do you think about removing the or replace. If we need to update something on the quote we can have a specific update fn?

crates/cdk-common/src/database/mint/mod.rs
```diff
@@ -23,42 +23,92 @@ pub mod test;
...
     /// Mint Keys Database Error
     type Err: Into<Error> + From<Error>;
 
-    /// Add Active Keyset
-    async fn set_active_keyset(&self, unit: CurrencyUnit, id: Id) -> Result<(), Self::Err>;
+    /// Beings a transaction
+    async fn begin_transaction<'a>(
+        &'a self,
+    ) -> Result<Box<dyn KeysDatabaseTransaction<'a, Self::Err> + Send + Sync + 'a>, Error>;
+
```
thesimplekid, line 46: Why does only the `KeysDatabase` need a begin transaction fn the others don't seem to have it?

thesimplekid (commented):
Thanks for this, I think it is a very strong improvement to our mint db. 

In terms of timing for merge this will conflict with bolt12. I think its best we do this first and then I can update the bolt12 branch on top of this. We previously discussed doing a v0.11 release middle of this week with the migration of sqlx to rusqlite and removal of redb. I think it would make sense to include this with that, what do you think?

crates/cdk-common/src/database/mint/mod.rs
crodas: We would have `begin_transactions` in 3 places (currently we have in 2 places).
1. Mint DB
2. Signatory DB (this is the signatory)
3. Auth DB (I will add soon to this PR).

crates/cdk-sqlite/src/mint/mod.rs
crodas: `INSERT OR REPLACE` did the trick, I brought it back.

crates/cdk/src/mint/melt.rs
```diff
@@ -439,40 +407,26 @@ impl Mint {
             }
         }
 
-        let quote = match self.verify_melt_request(melt_request).await {
-            Ok(quote) => quote,
-            Err(err) => {
-                tracing::debug!("Error attempting to verify melt quote: {}", err);
+        let mut tx = self.localstore.begin_transaction().await?;
```
thesimplekid, line 400: Since we start the transaction here and we do not commit until after the payment has been completed the database will be locked for the entire time it takes to make the ln payment. Most ln payments are fast but not all in the event one gets stuck the whole mint db will be blocked waiting on the one tx, I think we need to release the db lock while we wait for the ln payment.

crodas: That's a valid point, in practice a transaction should not be hold a database transaction hostage while something external is being executed.

I'll check that operation and add a locking mechanism. We still want to be able to lock resources (to avoid double spending) and a rollback mechanism.

thesimplekid: This is the intended use of the pending proof state. So the wallet can also check the status of a proof if it is stuck in a ln(or other) transaction.

crates/cdk/src/mint/check_spendable.rs
```diff
@@ -26,11 +26,12 @@ impl Mint {
             }
         }
 
+        let mut tx = self.localstore.begin_transaction().await?;
         for (state, ys) in ys_by_state {
-            self.localstore.update_proofs_states(&ys, state).await?;
+            tx.update_proofs_states(&ys, state).await?;
         }
 
-        self.localstore.remove_proofs(&unknown_proofs, None).await?;
```
asmogo, line 33: we could remove the unknown proofs Vec, since we don't use this variable anywhere. 
It seems like the `remove_proofs` function was moved to `update_proofs_states`.

crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -521,79 +779,6 @@ impl MintQuotesDatabase for MintSqliteDatabase {
...
-                unit = excluded.unit,
-                amount = excluded.amount,
-                request = excluded.request,
-                fee_reserve = excluded.fee_reserve,
-                state = excluded.state,
-                expiry = excluded.expiry,
-                payment_preimage = excluded.payment_preimage,
-                id = excluded.id,
-                created_time = excluded.created_time,
-                paid_time = excluded.paid_time;
```
asmogo, line 574: the ON CONFLICT clauses were removed. If that's intentional, we need to ensure that the REPLACE statement covers the conflict clauses. (request_lookup_id)

crates/cdk/src/mint/auth/mod.rs
```diff
@@ -205,6 +204,8 @@ impl Mint {
             }
         };
 
+        tx.commit().await?;
+
```
asmogo, line 208: consider updating the `add_proof` signature to take the state as a parameter.
Application logs would show a successful update even if the commit failed due to infrastructure or networking issues.

asmogo: or at least do the logging after transaction commit. 
```rust 
let mut tx = auth_localstore.begin_transaction().await?;
tx.add_proof(proof.clone()).await?;
let state = tx.update_proof_state(&y, State::Spent).await?;

tx.commit().await?;

tracing::debug!(
    "Successfully updated proof state to SPENT, previous state: {:?}",
    state
);
```

crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -987,16 +1059,63 @@ impl MintSignaturesDatabase for MintSqliteDatabase {
                 signature.dleq.as_ref().map(|dleq| dleq.s.to_secret_hex()),
             )
             .bind(":created_time", current_time as i64)
-            .execute(&transaction)
+            .execute(&self.transaction)
             .await
             .expect("fasdas");
```
asmogo: what's `fasdas` ?

crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -300,13 +344,208 @@ impl MintQuotesDatabase for MintSqliteDatabase {
...
 
-    async fn get_mint_quote(&self, quote_id: &Uuid) -> Result<Option<MintQuote>, Self::Err> {
+    async fn remove_mint_quote(&mut self, quote_id: &Uuid) -> Result<(), Self::Err> {
+        query(
+            r#"
+            DELETE FROM mint_quote
+            WHERE id=?
+            "#,
+        )
+        .bind(":id", quote_id.as_hyphenated().to_string())
```
asmogo: there might be an issue with this bind. The statement is using `?` positional parameter. I am not sure if this is really an issue but using `:id` would increase consistency for pos. params across the codebase.

crates/cdk/src/mint/auth/mod.rs
thesimplekid: Yeah i think it makes sense to add the state to the add_proof then we don't need the update fn here. I think this might be a bit of a carry over api model from redb, I think at one point the proof and state were different tables. So the add proof would modify the proof table and update state table the state table. Since the state is now stored in the same table, there isn't a need to add and update.

crates/cdk-sqlite/src/mint/mod.rs
crodas: Fixed in 031a05e8

crates/cdk-sqlite/src/mint/mod.rs
thesimplekid: I think I fixed this in the PR https://github.com/cashubtc/cdk/pull/845 I just merged. @crodas could you rebase and check.

crates/cdk/src/mint/melt.rs
```diff
@@ -595,19 +533,27 @@ impl Mint {
                     let mut melt_quote = quote;
                     melt_quote.request_lookup_id = payment_lookup_id;
 
-                    if let Err(err) = self.localstore.add_melt_quote(melt_quote).await {
+                    if let Err(err) = tx.add_melt_quote(melt_quote.clone()).await {
```
thesimplekid: Let's remove the use of `add_melt_quote` here what we really want to do is update the request_lookup_id lets create a specific db fn for this that uses `UPDATE` instead of insert or replace like we are currently.

crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -300,13 +344,203 @@ impl MintQuotesDatabase for MintSqliteDatabase {
...
+            .bind(":id", quote_id.as_hyphenated().to_string())
+            .execute(&self.inner)
+            .await?;
+        Ok(())
+    }
+
+    async fn add_melt_quote(&mut self, quote: mint::MeltQuote) -> Result<(), Self::Err> {
+        query(
+            r#"
+            INSERT OR REPLACE INTO melt_quote
```
thesimplekid:
```suggestion
            INSERT INTO melt_quote
```

crates/cdk/src/mint/melt.rs
```diff
@@ -595,19 +533,31 @@ impl Mint {
                     let mut melt_quote = quote;
                     melt_quote.request_lookup_id = payment_lookup_id;
 
-                    if let Err(err) = self.localstore.add_melt_quote(melt_quote).await {
+                    if let Err(err) = tx
+                        .update_melt_quote_state(&melt_quote.id, melt_quote.state)
+                        .await
+                    {
```
thesimplekid: Its the request_lookup_id that we need to update here not the state. State gets updated during `process_melt_request`

crodas: Does f091629b8a886e4905c52376d8b6e8f63a6c4f61 look better?

thesimplekid: Yes thank you

crates/cdk-sqlite/src/mint/mod.rs
thesimplekid: https://github.com/cashubtc/cdk/issues/859

thesimplekid:
ACK 5a8727d3bcdea5ead51c43a47be81a1d1cf945c7

crates/cdk/src/mint/auth/mod.rs
mubarak23: if we are adding state to `add_proof fn `, do we still need to Check previous state


Checking previous state will mean `add_proof`, interface definition will have change and return the current state after adding the state

PR #831, opened 2025-06-22
feat: remove fedimint tonic lnd

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-lnd/src/client.rs
+++ b/crates/cdk-lnd/src/client.rs
@@ -33,6 +33,10 @@ impl LndCertVerifier {
     pub(crate) async fn load(path: impl AsRef<Path>) -> Result<Self, Error> {
         let provider = default_provider();
 
+        if rustls::crypto::CryptoProvider::get_default().is_none() {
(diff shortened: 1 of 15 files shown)


crates/cdk-lnd/src/client.rs
```diff
@@ -0,0 +1,233 @@
...
+/// Custom certificate verifier for LND's self-signed certificates
+#[derive(Debug)]
+pub(crate) struct LndCertVerifier {
+    certs: Vec<Vec<u8>>,
+    provider: Arc<rustls::crypto::CryptoProvider>,
+}
+
+impl LndCertVerifier {
+    pub(crate) async fn load(path: impl AsRef<Path>) -> Result<Self, Error> {
+        let provider = default_provider();
```
a1denvalu3, line 34: Typically LND also provides a CA pem of the made up CA that signs the certificate. I'm not sure what `default_provider` does but I think this just turns off certificate signature verification?

PR #864, opened 2025-06-29
fix GLIBC compatibility issue

aki-mizu:
### Description
I'm getting `cdk-mintd: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.38 not found (required by cdk-mintd)` when running mintd 0.11.0 in docker environment
It turns out latest NixOS ships with glibc 2.38 while bookworm debian still uses glibc 2.36.
The next debian release trixie should come out soon meanwhile we may have to rely on sid version which has compatible glibc already

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

* [X] I followed the [code style guidelines](https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md)
* [X] I ran `just final-check` before committing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY crates ./crates
 RUN nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command cargo build --release --bin cdk-mintd --features redis
 
 # Create a runtime stage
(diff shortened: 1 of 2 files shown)


asmogo (approved):
LGTM. Thanks for fixing this.

PR #867, opened 2025-07-01
refactor:add proof now accept state

### LINKED ISSUE #861: Add proof accept state
https://github.com/cashubtc/cdk/pull/826#discussion_r2163074960

mubarak23:
### Description
closes #861 

This PR enhances the add_proof method by introducing an explicit state: State parameter to support more granular and intentional proof lifecycle management. Previously, the add_proof method always inserted a proof with a hardcoded i.e UNSPENT,
then make an extra call to update the proof with state of "SPENT", 

this change will mean, add_proof will accept the state, this in case where we want to save proof with the state of "UNSPENT", we can call the function and pass the state of "UNSPENT"

-----

### Notes to the reviewers

if we are adding state to add_proof , do we still need to Check previous state

Checking previous state will mean add_proof, interface definition will have change and return the current state after adding the state

-----

#### FIXED

----

### Checklist


mubarak23:
@crodas can you have a look at this

mubarak23:
@thesimplekid can have a look at this

PR #869, opened 2025-07-03
fix: install CryptoProvider when starting RPC server

gudnuf:
### Description

I was getting this error when I would run cdk-mintd with the management-rpc feature:

image
-----

### Notes to the reviewers

I'm not sure if `#[cfg(not(target_arch = "wasm32"))]` is needed. I see this same `rustls::crypto::ring::default_provider().install_default();` throughout the repo and sometimes the check for wasm32 is there and sometimes it is not. 

-----

#### FIXED
 
ensure default crypto provider is present on RPC server startup

----

### Checklist


thesimplekid:
Thanks, thought i put this everywhere it was needed looks like i missed one.

Issue #877, opened 2025-07-08
Implement postgres database


callebtc:
CDK postgres backend

PR #878, opened 2025-07-08
Add PostgreSQL support for mint and wallet

crodas:
### Description

This is a WIP branch to add PostgreSQL support, fixing #877. 

This PR depends on #890 

#### Approach

* [x] Introduce `cdk-sql-base` a generic abstraction on top of any SQL database for the mint and the wallet
    * [x] Rename `cdk-sqlite` into this crate.
    * [x] Unify the statement. 
    * [x] Add a simple placeholder parser, supporting named arguments, but let the database driver to use positional placeholder, in whatever format they need. Make it work with `bind` and `bind_vec`.
    * [x] Add custom migration for database vendor name, or migrations that applies to all. 
    * [ ] Add cache for the placeholder parsed list  
* [x] Make `cdk-sqlite` use `cdk-sql-base`
* [ ] Add `cdk-postgres`

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-mintd/src/lib.rs
+++ b/crates/cdk-mintd/src/lib.rs
@@ -3,7 +3,7 @@
 // std
 #[cfg(feature = "auth")]
 use std::collections::HashMap;
-use std::env::{self, var};
+use std::env::{self};
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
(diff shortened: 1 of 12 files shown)


crates/cdk-mintd/src/lib.rs
```diff
@@ -156,6 +157,18 @@ async fn setup_database(
             let keystore: Arc<dyn MintKeysDatabase<Err = cdk_database::Error> + Send + Sync> = db;
             Ok((localstore, keystore))
         }
+        DatabaseEngine::Postgres => {
+            let conn_url = std::env::var("DATABASE_URL").unwrap_or(
```
crodas: rename to PG_DB_URL or fail.

crates/cdk-postgres/src/db.rs
```diff
@@ -0,0 +1,155 @@
...
+
+            Ok(rows)
+        },
+        to_pgsql_error,
+    )
+    .await
+}
+
+#[inline(always)]
+pub async fn gn_pluck(conn: &Client, statement: Statement) -> Result<Option<Column>, Error> {
```
thesimplekid:
```suggestion
pub async fn pg_pluck(conn: &Client, statement: Statement) -> Result<Option<Column>, Error> {
```
typo?

PR #885, opened 2025-07-10
Increment keyset counter optimistically

davidcaseria:
### Description

Increment the keyset counter optimistically to avoid potential errors that may occur after the operation succeeds, but the counter is not incremented if the increment operation fails.

-----

### Notes to the reviewers

-----

#### FIXED
- Increment keyset counter optimistically

----

### Checklist


vnprc:
I am actually working on this same problem. I encountered this race condition in my ehash fork of cdk. My initial take was to return the counter value when calling increment but even after making this change I still encountered the race condition.

I was thinking of a resource lock on the keyset counter starting when you create the blinded secrets and ending when we increment the counter but I didn't like complexity and inefficiency it would entail.

Now that I see your PR I think an optimistic increment-and-retrieve at the time of blinded secret generation might just work. We should probably collapse the two counter functions (get and increment) into one that performs both operations atomically. Then we call that function optimistically before generating the secrets. If blinded secret generation fails we would have a gap in the deterministic secrets but I think that's fine. Gap limit issues seem less severe than race conditions to me.

I'm working on some integration tests to more clearly expose the race condition. I'll open a PR with this fix and, if you want, I'll tag you when it's ready.

vnprc:
https://github.com/cashubtc/cdk/pull/897

PR #887, opened 2025-07-11
Refactor MintBuilder

davidcaseria:
### Description

Refactor `MintBuilder` to avoid throwing any runtime errors.

Also, the OAuth Client ID is now validated when using CATs.

-----

### Notes to the reviewers

-----

#### CHANGED
- Refactored `MintBuilder` to avoid runtime errors

#### FIXED
- Verify client ID in CATs

----

### Checklist


crates/cdk-mintd/src/main.rs
```diff
@@ -237,7 +234,7 @@ async fn configure_mint_builder(
     let mint_builder = configure_lightning_backend(settings, mint_builder, &mut ln_routers).await?;
 
     // Configure signatory or seed
-    let mint_builder = configure_signing_method(settings, mint_builder).await?;
+    // let mint_builder = configure_signing_method(settings, mint_builder).await?;
```
thesimplekid: Should this be commented out. If so can we just remove it?

thesimplekid (commented):
Nice. Just one question

PR #890, opened 2025-07-15
Introduce `cdk-sql-common`

crodas:
### Description

The primary purpose of this new crate is to have a common and shared codebase for all SQL storage systems. It would force us to write standard SQL using best practices for all databases.

This crate has been extracted from #878

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sql-common/src/mint/mod.rs
+++ b/crates/cdk-sql-common/src/mint/mod.rs
@@ -1124,12 +1124,12 @@ where
                 request_lookup_id_kind
             FROM
                 mint_quote
-            WHERE request_lookup_id = :request_lookup_id"#,
+            WHERE request_lookup_id = :request_lookup_id
+            AND request_lookup_id_kind = :request_lookup_id_kind
+            "#,
         )?
-        .bind(
-            "request_lookup_id",
-            serde_json::to_string(request_lookup_id)?,
-        )
+        .bind("request_lookup_id", request_lookup_id.to_string())
+        .bind("request_lookup_id_kind", request_lookup_id.kind())
         .fetch_one(&self.db)
         .await?
         .map(|row| sql_row_to_mint_quote(row, vec![], vec![]))


crates/cdk-sql-common/src/mint/migrations/postgres/20250710212391_init.sql
thesimplekid, line 1: Should postgres be in this pr?

crates/cdk-sql-common/src/mint/migrations/sqlite/1_fix_sqlx_migration.sql
thesimplekid, line 1: Have you tested going from a v0.11 mint to this?

crates/cdk-sql-common/src/mint/migrations/postgres/20250710212391_init.sql
crodas: No. I will remove it.

crates/cdk-sql-common/src/mint/migrations/sqlite/1_fix_sqlx_migration.sql
crodas: I did, it is working fine.

crates/cdk-sql-common/src/common.rs
```diff
@@ -0,0 +1,46 @@
...
+    )?
+    .execute(conn)
+    .await?;
+
+    // Apply each migration if it hasn’t been applied yet
+    for (name, sql) in migrations {
+        if let Some((prefix, _)) = name.split_once(['/', '\\']) {
+            if prefix != db_prefix {
+                continue;
+            }
```
asmogo: If the names match, we should remove the prefix from the names. Otherwise, this will cause all migrations to be rerun because they were moved to a new subdirectory.
Alternatively, we could remove the path from all `pub static MIGRATIONS`.

crates/cdk-sql-common/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[dependencies]
+async-trait.workspace = true
+cdk-common = { workspace = true, features = ["test"] }
+bitcoin.workspace = true
+thiserror.workspace = true
+tokio.workspace = true
+tracing.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+lightning-invoice.workspace = true
```
asmogo, line 28: this is unused

asmogo (changes requested):

crates/cdk-sql-common/Cargo.toml
crodas: great catch. I'll fix it.

crodas: Fixed in e1cc72ecff76d801e2299d548002e5a4a955d9c7

crates/cdk-sql-common/src/common.rs
crodas: Isn't e1cc72ecff76d801e2299d548002e5a4a955d9c7 a better solution?

asmogo: That should do the job. Thanks 👍  I had something similar in mind like using the `file_name()` function on the path. 

```
diff --git a/crates/cdk-sql-common/build.rs b/crates/cdk-sql-common/build.rs
--- a/crates/cdk-sql-common/build.rs	(revision 316538beb9aee2bf89637bcd1744e92f6d8e1c50)
+++ b/crates/cdk-sql-common/build.rs	(date 1753447116890)
@@ -17,15 +17,14 @@
         let skip_path = parent.to_str().unwrap_or_default().len();
         let dest_path = parent.join("migrations.rs");
         let mut out_file = File::create(&dest_path).expect("Failed to create migrations.rs");
-
-        let skip_name = migration_path.to_str().unwrap_or_default().len();
-
+        
+        
         writeln!(out_file, "/// @generated").unwrap();
         writeln!(out_file, "/// Auto-generated by build.rs").unwrap();
         writeln!(out_file, "pub static MIGRATIONS: &[(&str, &str)] = &[").unwrap();
 
         for path in &files {
-            let rel_name = &path.to_str().unwrap().replace("\\", "/")[skip_name + 1..]; // for Windows
+            let rel_name = &path.file_name().unwrap().to_str().unwrap();
             let rel_path = &path.to_str().unwrap().replace("\\", "/")[skip_path..]; // for Windows
             writeln!(
                 out_file,
```

crates/cdk-sql-common/src/pool.rs
```diff
@@ -30,13 +30,17 @@ pub trait ResourceManager: Debug {
...
+    type Config: Clone + Debug;
 
     /// The error the resource may return when creating a new instance
     type Error: Debug;
 
     /// Creates a new resource with a given config
-    fn new_resource(config: &Self::Config) -> Result<Self::Resource, Error<Self::Error>>;
+    fn new_resource(
+        config: &Self::Config,
+        still_valid: Arc<AtomicBool>,
```
asmogo: Im not sure why we need that `still_valid` here. It seems like we always create the resource in valid state true.
The timeout should be moved to the configuration.

crodas: `still_valid` is used to signal the resource manager if the current resource is stale and it should be dropped (and a be eventually replaced with a new one).

This makes no sense for SQLite but it does for pgsql, to reconnect stale connections.

crodas: I acknowledge that the name is not very descriptive; a better name could be used. Any suggestions?

asmogo: OK gotcha. That makes sense. I haven’t looked into the psql implementation yet.

Perhaps we should simply call it `stale` and negate the drop condition? 

Do we need this as a parameter to the `new_resource` function? 
New resources might always be valid (or not stale)

crodas: I love it. I will rename it and document the behaviour, thanks for the feedback

crodas: is it better in cd066415c53ef2e0e9ac0c4eef3421bbeaa6c6b0 /cc @asmogo ?

crates/cdk-sql-common/src/pool.rs
```diff
@@ -30,13 +30,20 @@ pub trait ResourceManager: Debug {
...
+    type Config: Clone + Debug;
 
     /// The error the resource may return when creating a new instance
     type Error: Debug;
 
-    /// Creates a new resource with a given config
-    fn new_resource(config: &Self::Config) -> Result<Self::Resource, Error<Self::Error>>;
+    /// Creates a new resource with a given config.
+    ///
+    /// If `stale` is every set to TRUE it is assumed the resource is no longer valid and it will be
```
thesimplekid:
```suggestion
    /// If `stale` is ever set to TRUE it is assumed the resource is no longer valid and it will be
```

crates/cdk-sql-common/src/stmt.rs
```diff
@@ -0,0 +1,357 @@
...
+                parts,
+                cached_sql,
+                sql: None,
+                cache,
+                ..Default::default()
+            })
+        } else {
+            let parts = split_sql_parts(sql)?;
+
+            let _ = cache.write().map(|mut cache| {
```
thesimplekid:
```suggestion
if let Ok(mut cache) = self.cache.write() {
    if let Some((_, cached_sql)) = cache.get_mut(&original_sql) {
        *cached_sql = Some(sql.clone().into());
    }
} else {
    tracing::warn!("Failed to acquire write lock for SQL statement cache");
}
```

Just add some logging instead of ignoring the error

crates/cdk-sql-common/src/common.rs
```diff
@@ -0,0 +1,49 @@
...
+    // Apply each migration if it hasn’t been applied yet
+    for (name, sql) in migrations {
+        let basename = if let Some((prefix, basename)) = name.split_once(['/', '\\']) {
+            if prefix != db_prefix {
+                continue;
+            }
+            basename
+        } else {
+            name
+        };
```
thesimplekid:
```suggestion
let basename = match name.split_once(['/', '\\']) {
    Some((prefix, basename)) => {
        if prefix != db_prefix {
            continue;
        }
        basename
    }
    None => name,
};
```

Our code style says we should use match here https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#if-let.

thesimplekid (changes requested):
Overall I think this looks good I am just getting a runtime error that needs to be addressed. I am getting the Error: Duplicate entry when i restart the mint. This happened when I did not have a mint db created on by starting the mint, then stopping the mint and attempting to restart. 

Image

crates/cdk-sql-common/src/common.rs
```diff
@@ -0,0 +1,50 @@
...
+                basename
+            }
+            None => name,
+        };
+
+        let is_missing = query("SELECT name FROM migrations WHERE name = :name")?
+            .bind("name", basename)
+            .pluck(conn)
+            .await?
+            .is_none();
```
asmogo: removing the base name and using the name variable as bind parameter solves the issue mentioned in https://github.com/cashubtc/cdk/pull/890#pullrequestreview-3058149066. This was introduced due to a previous change request. 
Maybe we should consider removing the path from the name while creating the MIGRATIONS in build.rs. 
The path (database engine) is still referenced by file location.

crates/cdk-sql-common/src/pool.rs
asmogo: Yep. Thanks!

crates/cdk-sql-common/src/common.rs
crodas: Is it better in fc7adba0?

asmogo (approved):
ACK. Didn’t test https://github.com/cashubtc/cdk/pull/890/commits/0041c135f7872cb9dda7e68c1adfad8d013672c8 explicitly but it seems ok.

Issue #894, opened 2025-07-16
Wallet: Race Condition in Keyset Counter Operations


vnprc:
Summary

  Multiple wallet functions contain a race condition in their keyset counter operations that could lead to duplicate cryptographic secrets being generated under concurrent access.

  Problem Description

  Several wallet functions use a get-then-increment pattern for keyset counters:

  1. Read the current counter value (get_keyset_counter)
  2. Generate cryptographic secrets using that counter value
  3. Increment the counter separately (increment_keyset_counter)

  When two concurrent operations execute this sequence, they could read the same counter value and generate identical secrets, potentially causing the mint to reject duplicate
  blinded messages.

  Affected Functions

  1. issue_bolt11() - VULNERABLE

  File: crates/cdk/src/wallet/issue/issue_bolt11.rs
  Lines: 218-223 (get counter) → 232-238 (generate secrets) → 285-287 (increment counter)

  // VULNERABLE PATTERN
  let count = self.localstore.get_keyset_counter(&active_keyset_id).await?;
  let count = count.map_or(0, |c| c + 1);

  let premint_secrets = PreMintSecrets::from_xpriv(
      active_keyset_id,
      count,  // Uses the read counter value
      self.xpriv,
      amount_mintable,
      &amount_split_target,
  )?;

  // Later...
  self.localstore.increment_keyset_counter(&active_keyset_id, proofs.len() as u32).await?;

  2. melt_bolt11() - VULNERABLE

  File: crates/cdk/src/wallet/melt/melt_bolt11.rs
  Lines: 144-149 (get counter) → 151-156 (generate secrets) → 225-227 (increment counter)

  // VULNERABLE PATTERN
  let count = self.localstore.get_keyset_counter(&active_keyset_id).await?;
  let count = count.map_or(0, |c| c + 1);

  let premint_secrets = PreMintSecrets::from_xpriv_blank(
      active_keyset_id,
      count,  // Uses the read counter value
      self.xpriv,
      proofs_total - quote_info.amount,
  )?;

  // Later...
  self.localstore.increment_keyset_counter(&active_keyset_id, change_proofs.len() as u32).await?;

  Root Cause

  - No application-level locking around the get-generate-increment sequence
  - Separate database calls with timing gaps between operations
  - No transaction boundaries wrapping the entire operation
  - Database connection pooling only protects connection access, not data consistency

  Race Condition Scenario

  // Thread A and Thread B both execute concurrently:

  // 1. Both read counter = 5
[...]

PR #897, opened 2025-07-17
fix: atomically increment keyset counter

vnprc:
this prevents race conditions causing duplicate blind messages

remove get_keyset_counter() function
update increment_keyset_counter() to return the new count value
  - sqlite and redb implementations change all wallet operations to use the new atomic increment pattern
  - issue_bolt11.rs - Bolt11 invoice minting
  - issue_bolt12.rs - Bolt12 offer minting
  - melt_bolt11.rs - Melting with change proofs
  - swap.rs - Token swapping operations

### Description

https://github.com/cashubtc/cdk/issues/894

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sqlite/src/wallet/mod.rs
+++ b/crates/cdk-sqlite/src/wallet/mod.rs
@@ -751,21 +751,36 @@ ON CONFLICT(id) DO UPDATE SET
 
     #[instrument(skip(self), fields(keyset_id = %keyset_id))]
     async fn increment_keyset_counter(&self, keyset_id: &Id, count: u32) -> Result<u32, Self::Err> {
-        let new_counter_value = Statement::new(
-            r#"
-            UPDATE keyset
-            SET counter=counter+:count
-            WHERE id=:id
-            RETURNING counter
-            "#,
-        )
-        .bind(":count", count)
-        .bind(":id", keyset_id.to_string())
-        .plunk(&self.pool.get().map_err(Error::Pool)?)
-        .map_err(Error::Sqlite)?
-        .ok_or_else(|| Error::Sqlite(rusqlite::Error::QueryReturnedNoRows))?;
+        let query = format!(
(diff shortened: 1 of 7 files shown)


crates/cdk-sqlite/src/wallet/mod.rs
```diff
@@ -750,39 +750,24 @@ ON CONFLICT(id) DO UPDATE SET
...
     #[instrument(skip(self), fields(keyset_id = %keyset_id))]
-    async fn increment_keyset_counter(&self, keyset_id: &Id, count: u32) -> Result<(), Self::Err> {
-        Statement::new(
+    async fn increment_keyset_counter(&self, keyset_id: &Id, count: u32) -> Result<u32, Self::Err> {
+        let new_counter_value = Statement::new(
             r#"
             UPDATE keyset
             SET counter=counter+:count
             WHERE id=:id
+            RETURNING counter
```
crodas: Is this standard? Remember that I'm trying to have a generic SQL layer #890

crodas: Here is some really old [prior art](https://code.flickr.net/2010/02/08/ticket-servers-distributed-unique-primary-keys-on-the-cheap/)

vnprc: lol "the dumbest possible thing that will work" i couldn't have said it better myself

i rewrote this function to use a transaction to atomically increment and get the new count value since apparently standard SQL doesn't support returning values from an update. (really SQL? ¯\\\_(ツ)_/¯)

thesimplekid:
> Are we planning to add wallet db transaction support?

Yes, I think we should. It especially makes sense in the the case of larger wallet, maybe some sort of services operating a wallet. For the use can now of single user/process wallet it doesn't really add much. But since we have much of the code we can share with the mint it makes sense to do them the same way.

vnprc:
superceded by #944

PR #898, opened 2025-07-17
feat: prepared send via confim

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


davidcaseria:
What is the benefit of passing the `Wallet` to the `PreparedSend` over the current reverse process? If we want people to confirm the `PreparedSend`, should we store the `Wallet` instance in a field so it doesn't have to be passed later?

davidcaseria:
See #925 for the alternative

PR #900, opened 2025-07-22
refactor Nut10SecretRequest and add tests with provided test vectors

lollerfirst:
### Description

Makes the NUT10 secret inside a payment request conformant to [the latest changes](https://github.com/cashubtc/nuts/pull/274)

### Checklist


thesimplekid:
In the nuts test vectors there are cbor payment requests. I think we should include those in this as well to make sure we can decode them. https://github.com/lollerfirst/nuts/blob/a91268498346d63f367b38d0ae73374927746917/tests/18-tests.md?plain=1#L26-L31

thesimplekid (approved):
LGTM thanks

PR #912, opened 2025-07-26
Add NUT-19 support in the wallet

### LINKED ISSUE #870: Wallet support fot NUT19 cached requests
Currently the cdk mint supports cached responses however the wallet does not take advantage of them. We should add support in our http client, where if there is an http failure the wallet will reattempt the request. 

https://github.com/cashubtc/cdk/blob/cd63760324ef64928f496a5103cf5d6b1985a8be/crates/cdk/src/wallet/mint_connector/http_client.rs#L196

https://github.com/cashubtc/nuts/blob/main/19.md

@crodas Any interest in this one?

crodas:
### Description

Fixes #870

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/mint_connector/http_client.rs
+++ b/crates/cdk/src/wallet/mint_connector/http_client.rs
@@ -1,7 +1,5 @@
 use std::collections::HashSet;
-#[cfg(feature = "auth")]
-use std::sync::Arc;
-use std::sync::OnceLock;
+use std::sync::{Arc, RwLock as StdRwLock};
 use std::time::{Duration, Instant};
 
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


crates/cdk/src/wallet/mint_connector/http_client.rs
```diff
@@ -190,8 +196,67 @@ impl HttpClient {
...
+
+            if result.is_ok() {
+                return result;
+            }
+
+            match result.as_ref() {
+                Err(
+                    Error::Database(_) | Error::HttpError(_) | Error::Custom(_) | Error::Payment(_),
+                ) => {
+                    // retry request, if possible
```
thesimplekid, line 253: Could we just log that we're retrying and what error we got.

crates/cdk/src/wallet/mint_connector/http_client.rs
```diff
@@ -329,33 +399,60 @@ impl MintConnector for HttpClient {
...
     }
 
     /// Helper to get mint info
     async fn get_mint_info(&self) -> Result<MintInfo, Error> {
         let url = self.mint_url.join_paths(&["v1", "info"])?;
-        self.core.http_get(url, None).await
+        let info: MintInfo = self.core.http_get(url, None).await?;
+
+        if self.cache_support.get().is_none() {
+            let _ = self.cache_support.set((
```
thesimplekid: In theory the cached endpoints could change and maybe should be updated each time. However, it very unlikely to change so its probably fine to leave as a once cell for cheaper reads over a RwLock but thought I would mention it.

PR #918, opened 2025-07-29
fix(sqlite): fix bolt12 db migration errors

vnprc:
- add missing columns to the wallet get_mint_quotes query
- fix migration column name: amount_minted → amount_issued
- fix migration default value: 'BOLT11' → 'bolt11'
- add test coverage for mint quote db operations
- add PaymentMethod string parsing tests

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

* [ ] I followed the [code style guidelines](https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md)
* [ ] I ran `just final-check` before committing

PR #920, opened 2025-07-30
test: add PaymentMethod parsing and db read/write tests

vnprc:
### Description

the sql column error was fixed in the sql abstraction layer refactor but these are decent tests so i saved from #918

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid (approved):
LGTM thanks

PR #922, opened 2025-07-30
feat: enhance auth config

gudnuf:
### Description

Nuts 21 and 22 allow for any endpoint to be protected with clear or blind auth. This PR adds more fine grained configurations to allow users to enable/disable auth on every endpoint and configure the auth type as clear or blind. Minting blind auth tokens always requires clear auth.

-----

#### Defaults 

I've chosen the defaults based on my experience with implementing auth in my wallet and trying to provide a smooth UX. Only minting, swapping, and restoring require blind auth. Everything else is unauthenticated. Of course, wallets should not assume these defaults, but it makes the wallet experience better if they are set in this way.

- **checkstate**: by default I do not require auth so that any wallet can easily input a token and check if its spent before prompting the user to authenticate
- **get/check mint quotes**: by default creating and checking mint quotes does not require auth which allows a server (ie. lnurl server) to generate mint quotes on behalf of a user. This one introduces the risk that a wallet which does not check auth requirements could create a mint quote but never be able to actually mint the ecash.
- **melting**: melting does not require auth by default because there is no good solution for sending ecash from an auth mint to a non-authenticated user. By disabling auth on melts anyone is able to withdraw from the mint

There is a reasonable argument to enable blind auth on everything by default.

#### Config names
I renamed some config options and changed their type from boolean to an enum of 'clear', 'blind', or 'none'. Is this acceptable? Or do we need to make this change more backwards compatible? 

I renamed them in this way because the other option I saw was to keep the old configs (ie `enabled_mint=true`) and introduce new configs to configure the auth type (ie. `mint_auth_type='clear'`). This is verbose and makes the config harder to read.

-----

#### CHANGED

changed auth configs to allow for fine grained control of protected endpoints

#### ADDED

added the following config options:

```toml
#Authentication settings for endpoints
# Options: "clear", "blind", "none" (none = disabled)

mint = "blind"
get_mint_quote = "none"  
check_mint_quote = "none"

melt = "none"
get_melt_quote = "none"
check_melt_quote = "none"

swap = "blind"
restore = "blind"
check_proof_state = "none"
```

#### REMOVED

removed the following config options:

```toml
enabled_mint=true
enabled_melt=true
enabled_swap=true
enabled_check_mint_quote=true
enabled_check_melt_quote=true
enabled_restore=true
enabled_check_proof_state=true
```

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-mintd/src/env_vars/mod.rs
+++ b/crates/cdk-mintd/src/env_vars/mod.rs
@@ -60,6 +60,8 @@ impl Settings {
 
         #[cfg(feature = "auth")]
         {
+            use crate::config::AuthType;
+
             // Check env vars for auth config even if None
             let auth = self.auth.clone().unwrap_or_default().from_env();
 
@@ -67,9 +69,15 @@ impl Settings {
             if auth.openid_discovery != String::default()
                 || auth.openid_client_id != String::default()
                 || auth.mint_max_bat != 0
-                || auth.enabled_mint
-                || auth.enabled_melt
-                || auth.enabled_swap
+                || auth.mint != AuthType::Blind
+                || auth.get_mint_quote != AuthType::None
+                || auth.check_mint_quote != AuthType::None
+                || auth.melt != AuthType::None
+                || auth.get_melt_quote != AuthType::None
+                || auth.check_melt_quote != AuthType::None
+                || auth.swap != AuthType::Blind
+                || auth.restore != AuthType::Blind
+                || auth.check_proof_state != AuthType::None
             {
                 self.auth = Some(auth);
             } else {
--- a/crates/cdk-mintd/src/env_vars/mod.rs
+++ b/crates/cdk-mintd/src/env_vars/mod.rs
@@ -60,25 +60,11 @@ impl Settings {
 
         #[cfg(feature = "auth")]
         {
-            use crate::config::AuthType;
-
             // Check env vars for auth config even if None
             let auth = self.auth.clone().unwrap_or_default().from_env();
 
-            // Only set auth if env vars are present and have non-default values
-            if auth.openid_discovery != String::default()
-                || auth.openid_client_id != String::default()
-                || auth.mint_max_bat != 0
-                || auth.mint != AuthType::Blind
-                || auth.get_mint_quote != AuthType::None
-                || auth.check_mint_quote != AuthType::None
-                || auth.melt != AuthType::None
-                || auth.get_melt_quote != AuthType::None
-                || auth.check_melt_quote != AuthType::None
-                || auth.swap != AuthType::Blind
-                || auth.restore != AuthType::Blind
-                || auth.check_proof_state != AuthType::None
-            {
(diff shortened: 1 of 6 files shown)


crates/cdk-mintd/src/main.rs
```diff
@@ -42,6 +42,8 @@ use cdk::nuts::{ContactInfo, MintVersion, PaymentMethod};
 use cdk::types::QuoteTTL;
 use cdk_axum::cache::HttpCache;
 use cdk_mintd::cli::CLIArgs;
+#[cfg(feature = "auth")]
```
gudnuf: @thesimplekid I ran `just final-check` and `cargo fmt` locally and everything passed, but then when I pushed the "Cargo fmt" CI test was failing until I added this feature flag. How do I run that check locally?

thesimplekid: It's due to what features are enabled or disabled. just final-check it just with all features enabled I think. You can see what features are enabled in the ci cargo clippy --bin cdk-mintd --no-default-features --features fakewallet. We don't have a good script to run them all locally. @asmogo has a pr that would help this https://github.com/cashubtc/cdk/pull/799

crates/cdk-mintd/src/env_vars/mod.rs
```diff
@@ -60,16 +60,24 @@ impl Settings {
...
-                || auth.enabled_swap
+                || auth.mint != AuthType::Blind
+                || auth.get_mint_quote != AuthType::None
+                || auth.check_mint_quote != AuthType::None
+                || auth.melt != AuthType::None
+                || auth.get_melt_quote != AuthType::None
+                || auth.check_melt_quote != AuthType::None
+                || auth.swap != AuthType::Blind
+                || auth.restore != AuthType::Blind
+                || auth.check_proof_state != AuthType::None
```
thesimplekid: Instead of doing this here could we add a fn to the `Auth` struct the just returns a bool enabled/disabled. Then here we can just call `auth.auth_enabled()`. Just to make it more clear what this does.

thesimplekid (changes requested):
Just one small thing we can clean up I think and then LGTM. Thanks for doing this.

crates/cdk-mintd/src/env_vars/mod.rs
gudnuf: Okay what do you think about this instead [499b43c](https://github.com/cashubtc/cdk/pull/922/commits/499b43c1d6741c4256738bc6c8d14df27709d8a0) ?

The previous checks where a bit strange because it would treat auth as disabled if all the values were default, but you could easily want all the defaults, so then the only thing that would determine if auth is enabled is if the mint_max_bat was zero or not.

thesimplekid: Looks good

thesimplekid (approved):
ACK 499b43c1d6741c4256738bc6c8d14df27709d8a0

PR #928, opened 2025-07-31
New Secret Derivation

lollerfirst:
### Description

The new keyset derivation uses HMAC-SHA512 instead of BIP32.

----

### Checklist


lollerfirst:
@thesimplekid  now enforces [u8; 64] everywhere.

crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -29,7 +29,7 @@ use crate::{ensure_cdk, Amount, Wallet};
 pub struct MultiMintWallet {
     /// Storage backend
     pub localstore: Arc<dyn WalletDatabase<Err = database::Error> + Send + Sync>,
-    seed: Arc<[u8]>,
+    seed: Arc<[u8; 64]>,
```
thesimplekid:
```suggestion
    seed: [u8; 64],
```

Since its a fixed size array now we should be able to remove the `Arc`.

thesimplekid (approved):

PR #931, opened 2025-08-01
Working on a better database abstraction

crodas:
After [this question in the chat](https://matrix.to/#/!oJFtttFHGfnTGrIjvD:matrix.cashu.space/$oJFtttFHGfnTGrIjvD:matrix.cashu.space/$I5ZtjJtBM0ctltThDYpoCwClZFlM6PHzf8q2Rjqmso8) regarding a database transaction within the same function, I realized a few design flaws in our SQL database abstraction, particularly regarding transactions.

1. Our upper abstraction got it right, where a transaction is bound with `&mut self`, so Rust knows how to handle its lifetime with' async/await'.
2. The raw database does not; instead, it returns &self, and beginning a transaction takes &self as well, which is problematic for Rust, but that's not all. It is fundamentally wrong. A transaction should take &mut self when beginning a transaction, as that connection is bound to a transaction and should not be returned to the pool. Currently, that responsibility lies with the implementor. If a mistake is made, a transaction could be executed in two or more connections.
3. The way a database is bound to our store layer is through a single struct, which may or may not internally utilize our connection pool. This is also another design flow, in this PR, a connection pool is owned, and to use a connection, it should be requested, and that connection is reference with mutable when beginning a transaction

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-sqlite/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![warn(missing_docs)]
 #![warn(rustdoc::bare_urls)]
 
+mod async_sqlite;
```
thesimplekid, line 6: I think you forgot to add this file to git?

crates/cdk-sql-common/src/database.rs
```diff
@@ -32,22 +34,170 @@ pub trait DatabaseExecutor: Debug + Sync + Send {
...
+{
+    /// Consumes the current transaction committing the changes
+    async fn commit(conn: &mut W) -> Result<(), Error> {
+        query("COMMIT")?.execute(conn).await?;
+        Ok(())
+    }
+
+    /// Begin a transaction
+    async fn begin(conn: &mut W) -> Result<(), Error> {
+        query("BEGIN")?.execute(conn).await?;
```
thesimplekid, line 185:
```suggestion
        query("BEGIN IMMEDIATE")?.execute(conn).await?;
```

Should we be using BEGIN IMMEDIATE?

crates/cdk-sql-common/src/mint/auth/mod.rs
```diff
@@ -15,36 +16,43 @@ use tracing::instrument;
...
 
 /// Mint SQL Database
 #[derive(Debug, Clone)]
-pub struct SQLMintAuthDatabase<DB>
+pub struct SQLMintAuthDatabase<DB, RM, C, E>
 where
     DB: DatabaseConnector,
+    RM: ResourceManager<Resource = DB, Config = C, Error = E>,
+    C: Debug + Clone + Send + Sync,
+    E: Debug + std::error::Error + Send + Sync + 'static,
```
thesimplekid: This PR looks good my only concern is the complexity with the generics here. I think it would be best if we could simplify it a bit.

crodas: Is it better in 7607695fcce92bd8097edc3997473fb0f4165dcf ?

crates/cdk-sql-common/src/database.rs
crodas: Fixed in b40cf8fb

thesimplekid (approved):

PR #937, opened 2025-08-06
feat: add NUT-20 mint quote lookup API

vnprc:
- add quote lookup endpoint to get quotes by locking keys
- update auth tests for new MintQuoteLookup route path
- add integration tests for quote lookup functionality

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nutXX.rs
+++ b/crates/cashu/src/nuts/nutXX.rs
@@ -2,7 +2,7 @@
 
 use serde::{Deserialize, Serialize};
 
-use super::PublicKey;
+use super::{PaymentMethod, PublicKey};
 
 /// NUT-XX Settings
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Default)]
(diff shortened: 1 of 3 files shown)


thesimplekid:
> @thesimplekid you mentioned in the last cdk meeting that this would be better implemented as a custom API and added to the cdk router in my instantiation of the mint

That depends on if we think we should add this to the nuts or not, I think that's still open for discussion. If we do add it to the nuts this approach is best, if not then its not a cashu thing and needs to be handled separately. To do it sepratlly you would need to keep your own DB of the pubkey to quote mappings and then create an axum router with an endpoint to expose it. Currently you would need to fork or have your own mintd as there isn't a way to give mintd an extra router. The router gets created [here](https://github.com/cashubtc/cdk/blob/51b26eae621828d8dc5853b2e1ce111ab28efd55/crates/cdk-mintd/src/lib.rs#L693-L702) but I think we could add this to mintd as it might be useful to other were mintd accepts any router and just adds it to its own.

vnprc:
concept nack on the NUT PR

PR #943, opened 2025-08-08
Fixed bug to avoid empty calls `get_proofs_states`

crodas:
### Description

This bug was found while adding support for postgres and writing tests #878 

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid (approved):

thesimplekid:
I think we need to start adding tests for things like this.

PR #944, opened 2025-08-08
fix: implement atomic keyset counter

vnprc:
- remove get_keyset_counter
- update increment_keyset_counter to atomically increment and return counter value
- replace get+increment pattern with atomic increment everywhere

### Description

updated PR to replace #897
uses the new wallet DB transactions to atomically increment and get keyset counter using `FOR UPDATE`

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


vnprc:
@crodas done

PR #946, opened 2025-08-11
Introduce `run_db_operation_sync` and `run_db_operation`

crodas:
### Description

These functions are designed as a single funnel to talk to the database, whether it is synchronous or asynchronous.

This single funnel will log SQL queries and slow operations, providing a clear and unified debug message for the problematic query, so it can be optimized accordingly (for instance, missing indexes or unbound SQL requests).

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid (approved):
ACK 937d9ac43b68bda67aa685bd4531bfb0df8fee33

PR #947, opened 2025-08-11
Add explicit rollback.

crodas:
### Description

The proof_writer implicit rollback on Drop is too slow for our tests causing race conditions with Postgres.

This commit enhances the rollback logic and makes it explicit

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/proof_writer.rs
+++ b/crates/cdk/src/mint/proof_writer.rs
@@ -157,8 +157,8 @@ impl ProofWriter {
         };
 
         tracing::info!(
-            "Rollback proofs to their original states {:?} {:?}",
-            ys,
+            "Rollback {} proofs to their original states {:?}",
+            ys.len(),
             original_states
         );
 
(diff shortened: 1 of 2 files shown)


crates/cdk/src/mint/proof_writer.rs
```diff
@@ -143,13 +143,28 @@ impl ProofWriter {
...
         } else {
             return Ok(());
         };
-        reset_proofs_to_original_state(tx, &ys, original_states).await?;
+
+        tracing::info!(
+            "Rollback proofs to their original states {:?} {:?}",
+            ys,
+            original_states
+        );
```
thesimplekid: I think this logging is a bit too verbose, maybe debug on the number of proofs rolled back?

thesimplekid (commented):
This looks good. Can we add a doc comment to the drop that we prefer the explicit roll back due to unpredictable timing of using the drop with the spawn task and an warn loggins statement that it was used?

crates/cdk/src/mint/proof_writer.rs
crodas: Fixed in 30a7556c

thesimplekid (approved):

PR #954, opened 2025-08-12
Do not perform external calls during a database transaction.

crodas:
### Description

The codebase was used to correctly perform signatory calls during a database transaction, as the signatory was previously exclusively in process. However, a few months ago, it was changed to be a trait that can be either local or remote. Making external calls to services, adding latency, during an ongoing database transaction is a bad idea because it will lock the rows until the service call is finalized, which is unpredictable.

The issue is even worse in our pipeline where the SQLite storage driver is used with the ":memory:" path, which forces the Database pool to have a size of 1. Since our tests run in parallel, they would randomly fail.

This issue was failing in the CI, but the error was not making the pipeline fail. This bug was fixed as well.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-sql-common/src/mint/migrations.rs
```diff
@@ -1,6 +1,7 @@
 /// @generated
 /// Auto-generated by build.rs
 pub static MIGRATIONS: &[(&str, &str, &str)] = &[
+    ("", "20250812132015_drop_melt_request.sql", include_str!(r#"./migrations/20250812132015_drop_melt_request.sql"#)),
```
thesimplekid: Is this right? I think it should only be sqlite and why is it included in this pr?

thesimplekid: Looks like https://github.com/cashubtc/cdk/pull/953 was wrong

thesimplekid: should be fixed by https://github.com/cashubtc/cdk/pull/956

thesimplekid: I think you'll just need to manually delete this change and force push

crodas: Done

PR #963, opened 2025-08-13
Add `wait_for_payment` function in the wallet

### LINKED ISSUE #941: Wait for invoice to be paid
Currently we don't have a nice fn on the wallet that will wait for an invoice to be paid. If you look at the examples we look checking for state. We should add a wallet fn that takes a mint quote and a time out and just waits until it is paid and mints once it is, erroring if it can't mint or times out

crodas:
### Description

Fixes #941

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/wait.rs
+++ b/crates/cdk/src/wallet/wait.rs
@@ -42,6 +42,7 @@ impl From<WaitableEvent> for WalletSubscription {
 }
 
 impl Wallet {
+    #[inline(always)]
     async fn wait_and_mint_quote(
         &self,
         quote: MintQuote,
@@ -57,6 +58,23 @@ impl Wallet {
     /// Mints an amount and returns the invoice to be paid, and a BoxFuture that will finalize the
     /// mint once the invoice has been paid
     pub async fn mint_once_paid(
+        &self,
+        amount: Amount,
+        description: Option<String>,
+        timeout_duration: Duration,
+    ) -> Result<(String, impl Future<Output = Result<Proofs, Error>> + '_), Error> {
+        self.mint_once_paid_ex(
+            amount,
+            description,
+            Default::default(),
+            None,
+            timeout_duration,
+        )
+        .await
+    }
+
+    /// Similar function to mint_once_paid but with no default options
+    pub async fn mint_once_paid_ex(
         &self,
         amount: Amount,
         description: Option<String>,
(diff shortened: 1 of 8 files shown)


crates/cdk-integration-tests/tests/test_fees.rs
```diff
@@ -29,9 +28,10 @@ async fn test_swap() {
     let invoice = Bolt11Invoice::from_str(&mint_quote.request).unwrap();
     pay_if_regtest(&get_temp_dir(), &invoice).await.unwrap();
 
-    wait_for_mint_to_be_paid(&wallet, &mint_quote.id, 10)
+    wallet
+        .wait_for_payment(&mint_quote, Duration::from_secs(60))
         .await
-        .unwrap();
+        .expect("Mint to be paid");
```
thesimplekid: Don't change this now but I don't mind unwrap in tests whatever you prefer

crates/cdk/src/wallet/wait.rs
```diff
@@ -0,0 +1,80 @@
...
+            WaitableEvent::MeltQuote(quote_id) => {
+                WalletSubscription::Bolt11MeltQuoteState(vec![quote_id])
+            }
+            WaitableEvent::MintQuote(quote_id) => {
+                WalletSubscription::Bolt11MintQuoteState(vec![quote_id])
+            }
+        }
+    }
+}
+
```
thesimplekid, line 43: I think we should have another fn `mint_once_paid` to allow this flow: 

```
let quote = wallet.mint_quote(amount, None).await?;

let waiter = wallet.wait_for_payment(&quote.id, Duration::from_secs(10));
let proofs = waiter.await?;
```

With `wait_payment` it does not mint once it is paid so another fn call is needed.

crates/cdk/src/wallet/wait.rs
```diff
@@ -0,0 +1,80 @@
...
+        event: T,
+        timeout_duration: Duration,
+    ) -> BoxFuture<'_, Result<(), Error>>
+    where
+        T: Into<WaitableEvent>,
+    {
+        let subs = self.subscribe::<WalletSubscription>(event.into().into());
+
+        Box::pin(async move {
+            timeout(timeout_duration, async {
```
thesimplekid, line 111: This will fall back to polling right if the mint does not support ws?

thesimplekid (changes requested):

crates/cdk/src/wallet/wait.rs
crodas: Yes, the subscription function already does that. There is a ws.rs client and an http.rs client

thesimplekid: Yeah that's what I thought just wanted to make sure. Thanks

PR #965, opened 2025-08-14
chore(flake): add NIX_PATH for flake

ubbabeck:
### Description

This pr set the `NIX_PATH` as an environment variable since the `rust analyzer` needs it to function properly. 

removed 
`input 'crane' has an override for a non-existent input 'nixpkgs'` 

-----

### Notes to the reviewers

Use nix develop .#\<dev-shell\>

-----

### FIXED

Rust analyzer
----

### Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,37 +21,48 @@
 
     crane = {
       url = "github:ipetkov/crane";
-      inputs.nixpkgs.follows = "nixpkgs";
     };
 
     pre-commit-hooks.url = "github:cachix/pre-commit-hooks.nix";
   };
 
-  outputs = { self, nixpkgs, rust-overlay, flake-utils, pre-commit-hooks, ... }:
-    flake-utils.lib.eachDefaultSystem (system:
+  outputs =
+    { self
+    , nixpkgs
+    , rust-overlay
+    , flake-utils
+    , pre-commit-hooks
+    , ...
+    }@inputs:
+    flake-utils.lib.eachDefaultSystem (
+      system:
       let
         overlays = [ (import rust-overlay) ];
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -60,38 +71,53 @@
...
+            mprocs
+
+            # Needed for github ci
+            libz
+          ]
+          ++ libsDarwin;
 
         # Common arguments can be set here to avoid repeating them later
         nativeBuildInputs = [
+          pkgs.rust-analyzer
```
thesimplekid, line 117: Is this the only change? Rest is fmt?

flake.nix
```diff
@@ -151,53 +183,70 @@
               cargo update -p pest_meta --precise 2.8.0
               cargo update -p pest --precise 2.8.0
```
thesimplekid: These are removed now that we updated msrv. Can you rebase?

flake.nix
ubbabeck: No I also removed the [`inputs.nixpkgs.follows`](https://github.com/cashubtc/cdk/pull/965/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0L24)  for crane as it is was an unnecessary override. And set [`NIX_PATH`](https://github.com/cashubtc/cdk/pull/965/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R90)  as the rust analyzer needed it.

flake.nix
ubbabeck: Rebased

PR #967, opened 2025-08-15
feat: log to file

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-mintd/src/lib.rs
```diff
@@ -780,6 +806,35 @@ pub async fn run_mintd(
     run_mintd_with_shutdown(work_dir, settings, shutdown_signal(), db_password).await
 }
 
+/// Run mintd with proper logging setup and cleanup
+/// This is the recommended way to run mintd as it handles the logging guard properly
+pub async fn run_mintd_with_logging(
```
asmogo:
```suggestion
pub async fn run_mintd(
```

This got me confused. I would prefer calling the function `run_mintd` with `enable_logging` as an arg, or remove the arg / else statement.

thesimplekid: Yeah thats better applied in https://github.com/cashubtc/cdk/pull/967/commits/3cd4669e5a03a241baeb200382c4b9810f80b930

PR #975, opened 2025-08-18
Http retry should only happen at Http Error (which are transport errors)

### LINKED ISSUE #939: Mint connector should not retry on known error
In the case we get a known error from the mint we should not attempt to retry the request. The caching and retry should only be used for possible http errors.

https://github.com/cashubtc/cdk/blob/5521a377e8ba67e220f3d25722519f1837aad664/crates/cdk/src/wallet/mint_connector/http_client.rs#L207

Image

crodas:
### Description

Fixes #939

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/mint_connector/http_client.rs
+++ b/crates/cdk/src/wallet/mint_connector/http_client.rs
@@ -270,7 +270,7 @@ impl HttpClient {
             match result.as_ref() {
                 Err(Error::HttpError(status_code, _)) => {
                     let status_code = status_code.to_owned().unwrap_or_default();
-                    if status_code >= 400 && status_code <= 499 {
+                    if (400..=499).contains(&status_code) {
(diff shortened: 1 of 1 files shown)


crates/cdk/src/wallet/mint_connector/http_client.rs
```diff
@@ -249,7 +268,13 @@ impl HttpClient {
             }
 
             match result.as_ref() {
-                Err(Error::Database(_) | Error::HttpError(_) | Error::Custom(_)) => {
+                Err(Error::HttpError(status_code, _)) => {
+                    let status_code = status_code.to_owned().unwrap_or_default();
+                    if status_code <= 400 && status_code >= 499 {
```
thesimplekid:
```suggestion
                    if status_code >= 400 && status_code <= 499 {
```

thesimplekid (approved):

PR #981, opened 2025-08-20
fix: lnbits payment check and units

thesimplekid:
---------

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-lnbits/src/lib.rs
+++ b/crates/cdk-lnbits/src/lib.rs
@@ -368,13 +368,7 @@ impl MintPayment for LNbits {
                 unit: CurrencyUnit::Msat,
                 payment_id: payment.details.payment_hash,
             }]),
-
-            false => Ok(vec![WaitPaymentResponse {
-                payment_identifier: payment_identifier.clone(),
-                payment_amount: Amount::ZERO,
-                unit: CurrencyUnit::Msat,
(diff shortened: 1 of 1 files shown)


crates/cdk-lnbits/src/lib.rs
```diff
@@ -352,12 +355,27 @@ impl MintPayment for LNbits {
...
+                unit: CurrencyUnit::Msat,
+                payment_id: payment.details.payment_hash,
+            }]),
+
+            false => Ok(vec![WaitPaymentResponse {
+                payment_identifier: payment_identifier.clone(),
+                payment_amount: Amount::ZERO,
+                unit: CurrencyUnit::Msat,
+                payment_id: payment.details.payment_hash,
+            }]),
```
thesimplekid: Should we just return an empty list here?

asmogo: This would allow removing the additional [ZERO check](https://github.com/cashubtc/cdk/pull/981/files#diff-3f34620819b76e5d79a77f6133983f0b0afe923226d9bf8b81d135647da77412R50)

thesimplekid: Yeah looking at cln unpaid are filtered out so think its best to do the same here.

PR #984, opened 2025-08-20
compatibility for migrating Nutshell Mints

lollerfirst:
### Description

This PR aims to provide the most basic compatibility for Mints migrating from Nutshell

-----

### CHANGED

 - Change cdk-mintd to optionally accept a variable-length seed directly. NOT explicitly stated in the `config.toml.example`, so new mint operators don't use it.
 - Change all UUID instances to a new `QuoteId` enum that is either a base64 string or a UUID. When generating a quote id, it's always UUID. When reading from DB or from a request, then it can also accomodate for base64.


lollerfirst:
@thesimplekid

crates/cashu/src/nuts/nut04.rs
```diff
@@ -25,6 +25,92 @@ pub enum Error {
...
+    Base64,
+    /// Invalid quote ID
+    #[error("neither a valid UUID nor a valid base64 string")]
+    InvalidQuoteId,
+}
+
+/// Mint Quote ID
+#[derive(Serialize, Debug, Clone, PartialEq, PartialOrd, Ord, Eq, Hash)]
+#[serde(untagged)]
+pub enum QuoteId {
```
gudnuf: seems wrong to define QouoteId in Nut04 because it's not specific to minting. imo it should be owned equally by both NUT 04, 05, and specific payment methods. Maybe we should define it in the spec in NUT00, then in CDK also define it in NUT00. 

When you use it in other nuts and define a NUT04 error that's strange to me, again because it's not a really a mint operation specific error

thesimplekid: Spec defines quote id as just a string we shouldn't change that uuid is a cdk restriction. But yes thinks that arent spec genrally shouldn't be in the `cashu` crate, I think we should be able to just move this to `cdk-common` without too much trouble.

a1denvalu3: @gudnuf : we cannot make the cashu crate depend on cdk_common. Instead me and @thesimplekid discussed it and opted to move `QuoteId` to lib.rs, always in the cashu crate.

PR #985, opened 2025-08-20
Introduce Future Streams for Payments and Minting Proofs

crodas:
### Description

Introduce Future Streams (`ProofStream`, `PaymentStream`) for Payments and Proofs, an easier to use interface, async friendly, to interact for the mint waiting for payments of mints for Bolt11 and Bolt12.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/bolt12.rs
+++ b/crates/cdk-integration-tests/tests/bolt12.rs
@@ -337,8 +337,14 @@ async fn test_regtest_bolt12_mint_extra() -> Result<()> {
         .pay_bolt12_offer(Some(pay_amount_msats), mint_quote.request.clone())
         .await?;
 
+    let payment = wallet
+        .wait_for_payment(&mint_quote, tokio::time::Duration::from_secs(15))
+        .await?
+        .unwrap();
+
     let state = wallet.mint_bolt12_quote_state(&mint_quote.id).await?;
 
+    assert_eq!(payment, state.amount_paid);
     assert_eq!(state.amount_paid, (pay_amount_msats / 1_000).into());
     assert_eq!(state.amount_issued, Amount::ZERO);
 
--- a/crates/cdk-integration-tests/tests/regtest.rs
+++ b/crates/cdk-integration-tests/tests/regtest.rs
@@ -310,12 +310,7 @@ async fn test_cached_mint() {
         .expect("failed to pay invoice");
 
     let _proofs = wallet
-        .wait_and_mint_quote(
-            quote.clone(),
-            SplitTarget::default(),
-            None,
-            tokio::time::Duration::from_secs(15),
-        )
+        .wait_for_payment(&quote, tokio::time::Duration::from_secs(15))
         .await
         .expect("payment");
 


crates/cdk/src/wallet/streams/proof.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+impl<'a> ProofStream<'a> {
+    /// Create a new Stream
+    pub fn new(
+        wallet: &'a Wallet,
+        mint_quote: MintQuote,
+        amount_split_target: SplitTarget,
+        spending_conditions: Option<SpendingConditions>,
+        timeout: Duration,
```
thesimplekid: Lets make this a cancel token instead of a duration. This gives the user more control of when they want to stop waiting on the stream.

crates/cdk/src/mint/issue/mod.rs
```diff
@@ -456,7 +456,7 @@ impl Mint {
                         self.pubsub_manager.mint_quote_bolt12_status(
                             mint_quote.clone(),
                             wait_payment_response.payment_amount,
-                            Amount::ZERO,
+                            mint_quote.amount_issued(),
```
thesimplekid: We want this to be to be Amount::ZERO there is no issuing happening here we only got a payment.

crates/cdk/src/mint/issue/mod.rs
```diff
@@ -626,9 +626,17 @@ impl Mint {
...
 
-        tx.increment_mint_quote_amount_issued(&mint_request.quote, mint_request.total_amount()?)
+        let amount_issued = mint_request.total_amount()?;
+
+        tx.increment_mint_quote_amount_issued(&mint_request.quote, amount_issued)
             .await?;
 
+        self.pubsub_manager.mint_quote_bolt12_status(
+            mint_quote.clone(),
+            mint_quote.amount_paid(),
```
thesimplekid:
```suggestion
            Amount::ZERO,
```

We did not get a payment here we only issued so we do not need to update the amount payed.

crates/cdk/examples/mint-token.rs
```diff
@@ -36,24 +36,32 @@ async fn main() -> Result<(), Error> {
...
+    );
+
+    while let Some(proofs) = proof_streams.next().await {
+        let proofs = match proofs {
+            Ok(proofs) => proofs,
+            Err(err) => {
+                tracing::error!("Proof stream ended with {:?}", err);
+                break;
+            }
+        };
```
thesimplekid: For the examples where using bolt12 we probably don't need the stream the stream is really for bolt12 keeping it open and re minting.

crates/cdk-integration-tests/tests/test_fees.rs
```diff
@@ -96,15 +100,18 @@ async fn test_fake_melt_change_in_quote() {
...
+    );
 
-    let _mint_amount = wallet
-        .mint(&mint_quote.id, SplitTarget::default(), None)
+    let _proofs = proof_streams
+        .next()
         .await
-        .unwrap();
+        .expect("payment")
+        .expect("no error");
```
thesimplekid: Will just just keep the stream open for 60 seconds or because it bolt11 the stream ends after one? Either way I don't think we need the stream when we're just minting once.

crates/cdk/src/wallet/streams/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl Wallet {
+    #[inline(always)]
+    /// Mints a mint quote once it is paid
+    pub fn proof_stream(
+        &self,
+        quote: MintQuote,
+        amount_split_target: SplitTarget,
+        spending_conditions: Option<SpendingConditions>,
+    ) -> ProofStream<'_> {
```
davidcaseria: Would streaming the mint quote to get status updates be beneficial?

davidcaseria (commented):
This will be very helpful!

crates/cdk/src/wallet/streams/mod.rs
crodas: This would return the proofs already, meaning the quote would already be paid. Maybe it should also be useful to have a status update stream? So the UI/UX can be changed accordingly?

davidcaseria: I don't have a strong opinion, but I wanted to raise the issue to consider whether returning the status is useful.

thesimplekid: I think here since you're subscribing to one mint quote proofs is good. I think another fn to subscribe to many mint quotes you would want to return the proofs and the quote id then to know what proof has been paid. I think @crodas started on a wallet subscription method in another pr.

crates/cdk/examples/auth_wallet.rs
thesimplekid, line 1: In the examples where we use bolt11 I don't think it makes sense to use the stream instead we should leave it how it was with the wait and mint fn. Then we should add an example that is just for streaming and mints multiple times. fake.thesimplekid.dev will repay bolt12 quotes with any amounts so we should use this to show how it is used. The stream really only makes sense for quotes that can get multiple updates for now just bolt12.

Issue #979, opened 2025-08-20
Secret Memory Leakage Due to Extensive Cloning


vnprc:
# Problem Description

The CDK codebase has a fundamental security issue where cryptographic secrets are copied throughout memory without proper cleanup. The `Secret` type implements `Clone`, which creates copies that persist in memory even when the original secret goes out of scope.

# Current State Analysis

## Secret Cloning Hotspots

The `Secret` type is cloned extensively throughout the codebase:

    // cashu/src/nuts/nut00/mod.rs:433
    secret: self.secret.clone(),

    // cashu/src/nuts/nut00/mod.rs:506  
    secret: self.secret.clone(),

    // cashu/src/nuts/nut00/mod.rs:894
    self.iter().map(|pm| pm.secret.clone()).collect()

## The Fundamental Issue

1. Secret generation: `Secret::generate()` creates a secret
2. Immediate cloning: Secret gets moved into `Proof` structs via `.clone()`
3. Multiple copies: Proof conversions (`ProofV3 ↔ ProofV4 ↔ Proof`) clone secrets again
4. Collection operations: Methods like `secrets()` clone all secrets in a collection
5. No cleanup: Cloned copies remain in memory until overwritten by other data

Result: Secret data scattered throughout memory in multiple uncleared copies.

---

# Security Impact

- Secrets remain in memory longer than necessary
- Memory dumps could expose secret values
- Swap files may contain uncleared secrets
- Process memory inspection reveals sensitive data

---

# Proposed Solution: Reference-Counted Secure Secrets

Implement an `Arc<SecretBox>` approach where all "clones" share the same underlying secret data.

## Implementation

    use std::sync::Arc;
    use zeroize::{Zeroize, ZeroizeOnDrop};

    #[derive(ZeroizeOnDrop)]
    struct SecretBox {
        data: String,
    }

    #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
    #[serde(transparent)]
    pub struct Secret(Arc<SecretBox>);

    impl Secret {
        pub fn generate() -> Self {
            let mut random_bytes = [0u8; 32];
            rand::thread_rng().fill_bytes(&mut random_bytes);
            let secret_data = hex::encode(random_bytes);
            random_bytes.zeroize();

            Secret(Arc::new(SecretBox { data: secret_data }))
        }

        pub fn new<S: Into<String>>(secret: S) -> Self {
[...]

PR #988, opened 2025-08-21
feat: zeroize cryptographic secrets on drop

vnprc:
implement zeroize on Drop for Secret, Wallet, and MultiMintWallet
this erases sensitive memory addresses before deallocation

### Description

Minor security improvement that prevents a memory dump, swap file, or process memory inspection attack from exposing secrets left in deallocated memory addresses.

This is good hygiene and could be useful as part of a defense in depth strategy. Currently, it doesn't accomplish much because the master seed is kept in Wallet memory for the duration of the process.

A more meaningful security improvement would be to encrypt the seed on disk and load it into memory only when actively deriving keys. This would have a high performance impact due to the high frequency of ecash operations. Perhaps a better approach would be to focus on device architectures with a secure computing environment such as a secure enclave on a mobile phone and using something like [VLS](https://github.com/lightning-signer) on mint servers.

In any case this PR has very low performance impact and it's better than nothing.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/Cargo.toml
+++ b/crates/cashu/Cargo.toml
@@ -36,6 +36,7 @@ serde_with.workspace = true
 regex = { workspace = true, optional = true }
 strum = { workspace = true, optional = true }
 strum_macros = { workspace = true, optional = true }
+zeroize = "1"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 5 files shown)


vnprc:
#979

PR #993, opened 2025-08-23
fix(db): make add_keys idempotent across db impls

vnprc:
- redb: return Ok() instead of Err(database::Error::Duplicate)
- cdk-sql-common: add ON CONFLICT(id) DO NOTHING clause
- unit test idempotency for all db impls

### Description

Make add_keys function idempotent across all database implementations to prevent duplicate key insertion errors.

  This change ensures that calling add_keys multiple times with the same keyset will not result in an error, making the operation safe
  for retry scenarios and more robust in concurrent environments.

  Changes Made

  - cdk-redb: Modified add_keys to return Ok(()) instead of Err(database::Error::Duplicate) when keys already exist, with debug logging
  instead of warning
  - cdk-sql-common: Added ON CONFLICT(id) DO NOTHING clause to the keyset insertion query to handle duplicate key scenarios gracefully
  - All database implementations: Added comprehensive test cases test_add_keys_idempotent to verify the idempotent behavior works
  correctly

  Test Coverage

  Added tests for all database backends (Postgres, SQLite, ReDB) that verify:
  - First add_keys call succeeds
  - Second add_keys call with same keyset succeeds (idempotent behavior)
  - Keys can be retrieved correctly after multiple insertion attempts
  - No data corruption occurs from duplicate calls

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

`just final-check` crashed my computer and i didn't have time to rerun it, maybe we should reevaluate the resource cost of this just recipe

### DIFF
--- a/crates/cdk-sqlite/src/wallet/mod.rs
+++ b/crates/cdk-sqlite/src/wallet/mod.rs
@@ -188,17 +188,7 @@ mod tests {
         use std::collections::BTreeMap;
 
         // Create in-memory SQLite database for testing
-        let path = std::env::temp_dir()
-            .to_path_buf()
-            .join(format!("cdk-test-add-keys-{}.sqlite", uuid::Uuid::new_v4()));
-
-        #[cfg(feature = "sqlcipher")]
-        let db = WalletSqliteDatabase::new((path, "password".to_string()))
-            .await
-            .unwrap();
-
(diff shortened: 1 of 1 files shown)


thesimplekid:
We discussed this on a call going to close for reasoned mentioned there.

PR #995, opened 2025-08-25
onchain

gudnuf:
### Description

Adds onchain as NUT-26 and implements onchain in fake-wallet.

-----

### Notes to the reviewers

There's still more to do, here's a todo list of things I can think of that still need to be done:

- [ ] Onchain melt quotes use onchain txids as payment proof. Right now I just store the txid on the `MeltQuote`'s `payment_preimage` column. We should consider either adding a transaction_id column or something changing the column name to `payment_proof`.
- [ ] handle confirmed vs unconfirmed payments. We need to create a payment for the onchain mint quote and then update the quote's `amount_paid` and `amount_unconfirmed` based on the payment status. If we receive a payment notification that is unconfirmed, then we increment `amount_unconfirmed`. This same payment will then get updated to confirmed and we should decrement `amount_unconfirmed`, then increment `amount_paid`
- [ ] add number of blocks required for confirmation to mint info

-----

#### CHANGED

- Created `MeltQuoteResponse` enum to handle new `MeltQuoteOnchainResponse`.

#### ADDED

- NUT-26 onchain

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/melt/melt_onchain.rs
+++ b/crates/cdk/src/wallet/melt/melt_onchain.rs
@@ -1,0 +1,73 @@
+//! Melt Onchain
+//!
+//! Implementation of melt functionality for onchain Bitcoin transactions
+
+use cdk_common::nut26::MeltQuoteOnchainRequest;
+use cdk_common::wallet::MeltQuote;
+use tracing::instrument;
+
+use crate::nuts::MeltQuoteOnchainResponse;
+use crate::{Amount, Error, Wallet};
+
+impl Wallet {
+    /// Melt Quote for onchain Bitcoin transaction
+    #[instrument(skip(self, request))]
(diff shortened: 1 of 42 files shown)


crates/cdk/src/wallet/melt/melt_onchain.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        let quote = MeltQuote {
+            id: quote_res.quote,
+            amount: quote_res.amount,
+            request,
+            unit: self.unit.clone(),
+            fee_reserve: quote_res.fee_reserve,
+            state: quote_res.state,
+            expiry: quote_res.expiry,
+            payment_preimage: None, // Onchain transactions don't have preimages like Lightning
```
gudnuf, line 36: this should be the transaction id that the mint send bitcoin to

PR #1004, opened 2025-08-26
deterministic currency unit derivation path

vnprc:
### Description

Move the CurrencyUnit enum from `nut00` where it was originally implemented before being added to the spec to `nut01` to match the spec definition.

Implement deterministic currency unit derivation paths.

- use SHA-256 so results are identical across platforms and rust versions
- NFC normalization, lowercasing, and trim to ensure logically equivalent strings map to the same index
- reserve a low integer range for known variants

-----

### Notes to the reviewers

See also this impl from the portal team: https://github.com/PortalTechnologiesInc/cdk/commit/0051f57cf29bcf8d8b815a8cb38b24cd6e3fa6c7
and this impl from a previous hashpool cdk branch: https://github.com/vnprc/cdk/commit/0d11a5bd75dffc31496be2d5fbac72182c753529

The implementation in this PR improves on both of the above impls.
-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut01/mod.rs
+++ b/crates/cashu/src/nuts/nut01/mod.rs
@@ -235,33 +235,36 @@ impl CurrencyUnit {
             Self::Usd => Some(2),
             Self::Eur => Some(3),
             Self::Auth => Some(4),
-            Self::Custom(s) => {
-                // 1) NFC normalization: composes equivalent Unicode sequences (e.g., "e" + U+0301) into a single
-                //    canonical code point (e.g., "é") so visually/equivalently identical strings hash the same
-                // 2) lowercase: avoids case-induced divergence ("USD" vs "usd")
-                // 3) trim: removes accidental leading/trailing whitespace (" usd " vs "usd")
-                let norm = s.nfc().collect::<String>().to_lowercase();
-                let norm = norm.trim();
-
-                // use SHA-256 so that the same normalized string always yields the same digest
-                let digest = Sha256::hash(norm.as_bytes());
-
-                // take 4 bytes in a fixed endianness to get a u32
-                let x = u32::from_be_bytes([digest[0], digest[1], digest[2], digest[3]]) as u64;
-
-                // map x into the inclusive interval [RESERVED, u32::MAX].
-                // compute the size of that interval:
-                //   size = (u32::MAX - RESERVED + 1)
-                // use u64 math to avoid overflow.
-                let interval_size = (u32::MAX as u64) - (RESERVED as u64) + 1;
-
-                // Fold x uniformly into [0, interval_size - 1].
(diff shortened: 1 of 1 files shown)


crates/cashu/src/nuts/nut01/mod.rs
```diff
@@ -192,6 +196,127 @@ impl MintKeyPair {
...
+    /// Deterministic derivation index for each currency unit
+    ///
+    /// - use SHA-256 so results are identical across platforms
+    ///   and rust versions
+    /// - NFC normalization, lowercasing, and trim to ensure logically
+    ///   equivalent strings map to the same index.
+    /// - reserve a low integer range for known variants
+    pub fn derivation_index(&self) -> Option<u32> {
+        // reserved band for well-known units
+        const RESERVED: u32 = 10_000;
```
thesimplekid: I don't see why we need to reserve anything beyond what we already have hard-coded we should just hash everything else.

vnprc: so you think we'll never hard code another currency unit?

vnprc: I guess the only reason to hard code one is to get a unique derivation path which this PR solves in a better way.

thesimplekid:
> Suppose we want to add a new enum value, how does that work? Do we implement it as a custom variant?

What do you mean enum value?

vnprc:
a new CurrencyUnit enum variant, like how `Auth` was recently added.
```
pub enum CurrencyUnit {
    /// Sat
    #[default]
    Sat,
    /// Msat
    Msat,
    /// Usd
    Usd,
    /// Euro
    Eur,
    /// Auth
    Auth,
    /// Custom currency unit
    Custom(String),
}
```

If we reserve a small range of hard coded derivation path values for the future it becomes trivial to add a new enum variant that is more ergonomic for developers to use. The downside is minor IMO: you get an infinitesimally larger chance of collision. We don't need to reserve 10,000 spaces but I think a small reserved range is a good idea.

thesimplekid:
> I think a small reserved range is a good idea.

for what? Even the current ones only need to stay fixed for backwards compatibility. The only point of the reserved is to map the unit to the derivation index, we can do that now by hashing.

vnprc:
I was about to suggest we take a page from the secp256k1 library and implement different hash functions but I just saw that apoelstra removed this functionality last week. https://github.com/rust-bitcoin/rust-secp256k1/commit/54828aceb9c3fb4e299095d8eb73c669acb8bb93

We could standardize on the rust default hash function for simplicity. Would this make derivation paths different in some library upgrade scenarios?

vnprc:
https://github.com/cashubtc/nuts/pull/292

PR #1005, opened 2025-08-26
feat: redact secrets from Debug and Display impls

vnprc:
redact Debug impls for SecretKey, Secret, Wallet, and MultiMintWallet redact Display impls for SecretKey and Secret
wallet structs don't implement Display, so no issue there

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut01/secret_key.rs
+++ b/crates/cashu/src/nuts/nut01/secret_key.rs
@@ -15,7 +15,7 @@ use super::{Error, PublicKey};
 use crate::SECP256K1;
 
 /// SecretKey
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "swagger", derive(utoipa::ToSchema))]
 pub struct SecretKey {
     #[cfg_attr(feature = "swagger", schema(value_type = String))]
@@ -38,7 +38,15 @@ impl From<secp256k1::SecretKey> for SecretKey {
 
 impl fmt::Display for SecretKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.to_secret_hex())
+        write!(f, "[REDACTED]")
+    }
+}
+
+impl fmt::Debug for SecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SecretKey")
+            .field("inner", &"[REDACTED]")
+            .finish()
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/cashu/src/nuts/nut01/secret_key.rs
```diff
@@ -38,7 +38,15 @@ impl From<secp256k1::SecretKey> for SecretKey {
...
-        write!(f, "{}", self.to_secret_hex())
+        write!(f, "[REDACTED]")
+    }
+}
+
+impl fmt::Debug for SecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SecretKey")
+            .field("inner", &"[REDACTED]")
+            .finish()
```
thesimplekid: I don't think we need to do this because secp256k1 already does: https://docs.rs/secp256k1/0.31.1/src/secp256k1/key.rs.html#59

crates/cashu/src/nuts/nut01/secret_key.rs
```diff
@@ -38,7 +38,15 @@ impl From<secp256k1::SecretKey> for SecretKey {
 
 impl fmt::Display for SecretKey {
```
thesimplekid, line 39: We currently use this in the Serialize impl so we need to check if we use that anywhere it doesn't make sense to just have redacted for that.

crates/cashu/src/secret.rs
```diff
@@ -35,6 +35,18 @@ impl Default for Secret {
     }
 }
 
+impl fmt::Debug for Secret {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("Secret").field(&"[REDACTED]").finish()
+    }
+}
+
+impl fmt::Display for Secret {
```
thesimplekid: We need to go to string on the secret to store it in the db. This would not save any secrets and store everything as redacted. 

https://github.com/cashubtc/cdk/blob/44cef59fe567dff2302f63385535444f740adf07/crates/cdk-sql-common/src/wallet/mod.rs#L750

crates/cashu/src/nuts/nut01/secret_key.rs
vnprc: oh cool, that macro is way better

crates/cashu/src/nuts/nut01/secret_key.rs
vnprc: Oops. I think we can leave this change because SecretKey uses a custom Serialize impl that calls `self.to_secret_hex()` or `self.as_secret_bytes()` and not Display or `to_string()`.

I'll put it back.

PR #1006, opened 2025-08-27
Minor file organization

crodas:
### Description

Minor file organization to group similar crates together, to make it more obvious at first sight what each crate does, and to be less overwhelming for newcomers.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
I don't have a very string opinion on it, kinda default to keeping it as is but if others like it I don't mind. I don't want to do it before promethus though.

PR #1007, opened 2025-08-27
Add suport for Bolt12 notifications for HTTP subscription

### LINKED ISSUE #992: Emulate `NotificationPayload::MintQuoteBolt12Response` for http subscription
As a bonus, also include a http/ws testsuite that connects to a fake mint, ideally internal.

crodas:
### Description

This commit adds support for Mint Bolt12 Notifications for HTTP when Mint does not support WebSocket or the wallet decides not to use it, and falls back to HTTP.

This PR fixes #992

TODO: Add a test compiling the wallet without HTTP support to test it

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/builder.rs
+++ b/crates/cdk/src/wallet/builder.rs
@@ -26,7 +26,7 @@ pub struct WalletBuilder {
     #[cfg(feature = "auth")]
     auth_wallet: Option<AuthWallet>,
     seed: Option<[u8; 64]>,
-    prefer_http_subscription: bool,
+    use_http_subscription: bool,
     client: Option<Arc<dyn MintConnector + Send + Sync>>,
 }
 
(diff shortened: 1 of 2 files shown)


crates/cdk/src/wallet/builder.rs
```diff
@@ -50,6 +52,19 @@ impl WalletBuilder {
         Self::default()
     }
 
+    /// Use HTTP for wallet subscriptions to mint events
+    pub fn prefer_http_subscription(mut self) -> Self {
+        self.prefer_http_subscription = true;
+        self
+    }
```
thesimplekid: Is this a prefer or a require? WS makes sense to be prefer since we fallback to http if mint does not support ws, but here I think its setting wallet to never use ws?

thesimplekid: @crodas we should still change this

crodas: Yeah, sorry!

crodas: Fixed in f261b105

thesimplekid (approved):

PR #1010, opened 2025-08-27
adding more LDK configuration settings

asmogo:
### Description
added bip39 mnemonic seed and announcement addresses to LDK configuration.

-----

### Notes to the reviewers

-----

#### ADDED
* [CDK-LDK] bip39 mnemonic seed 
* [CDK-LDK] configurable announcement addresses 
* [CDK-LDK] configurable log
* 
#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-ldk-node/src/lib.rs
+++ b/crates/cdk-ldk-node/src/lib.rs
@@ -160,35 +160,16 @@ impl CdkLdkNodeBuilder {
     }
 
     /// Builds the [`CdkLdkNode`] instance
-    pub fn build(self) -> Result<CdkLdkNode, Error> {
-        CdkLdkNode::new(self)
-    }
-}
-
-impl CdkLdkNode {
-    /// Create a new CDK LDK Node instance
-    ///
-    /// # Arguments
-    /// * `network` - Bitcoin network (mainnet, testnet, regtest, signet)
-    /// * `chain_source` - Source of blockchain data (Esplora or Bitcoin RPC)
-    /// * `gossip_source` - Source of Lightning network gossip data
-    /// * `storage_dir_path` - Directory path for node data storage
-    /// * `fee_reserve` - Fee reserve configuration for payments
-    /// * `listening_address` - Socket addresses for peer connections
-    /// * `runtime` - Optional Tokio runtime to use for starting the node
-    ///
-    /// # Returns
-    /// A new `CdkLdkNode` instance ready to be started
     ///
     /// # Errors
     /// Returns an error if the LDK node builder fails to create the node
-    pub fn new(builder: CdkLdkNodeBuilder) -> Result<Self, Error> {
+    pub fn build(self) -> Result<CdkLdkNode, Error> {
         let mut ldk = Builder::new();
-        ldk.set_network(builder.network);
-        tracing::info!("Storage dir of node is {}", builder.storage_dir_path);
-        ldk.set_storage_dir_path(builder.storage_dir_path);
+        ldk.set_network(self.network);
+        tracing::info!("Storage dir of node is {}", self.storage_dir_path);
+        ldk.set_storage_dir_path(self.storage_dir_path);
 
-        match builder.chain_source {
+        match self.chain_source {
             ChainSource::Esplora(esplora_url) => {
                 ldk.set_chain_source_esplora(esplora_url, None);
             }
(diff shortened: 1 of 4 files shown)


thesimplekid:
@asmogo Should we reopen this? If i remember it was in draft so I never really looked at it.

asmogo:
yes. I'll merge main and retest everything before reopening.

crates/cdk-mintd/src/setup.rs
```diff
@@ -329,17 +330,31 @@ impl LnBackendSetup for config::LdkNode {
         // We need to get the actual socket address struct from ldk_node
         // For now, let's construct it manually based on the cdk-ldk-node implementation
         let listen_address = vec![socket_addr.into()];
-
-        let mut ldk_node = cdk_ldk_node::CdkLdkNode::new(
+        let mem: Mnemonic = settings.clone().info.mnemonic.unwrap().parse().unwrap();
```
thesimplekid: We should not make the seed required as its not backwards compatible, existing nodes that use the current method of ldk-node random bytes stored on disk would break. 

If we want to force new nodes to use a mnemonic one option would be if no mnemonic is given we check if the node dir is there with a secret `keys_seed` is the file name. If that is there we allow the node to start with that if not we error with a message a mnemonic must be set.

asmogo: done in [c7bafcf](https://github.com/cashubtc/cdk/pull/1010/commits/c7bafcfab437ce2e750ff37854eeae5600af30e5)

crates/cdk-ldk-node/src/lib.rs
```diff
@@ -220,10 +220,14 @@ impl CdkLdkNode {
 
         ldk.set_node_alias("cdk-ldk-node".to_string())?;
         if builder.seed.is_some() {
-            ldk.set_entropy_bip39_mnemonic(builder.seed.unwrap(), None);
+            ldk.set_entropy_bip39_mnemonic(builder.seed.expect("valid seed"), None);
```
thesimplekid: we can use an `if let Some(seed) = builder.seed {}` here and it removes the need to check is some and then unwrap it.

crates/cdk-mintd/src/setup.rs
```diff
@@ -330,7 +331,38 @@ impl LnBackendSetup for config::LdkNode {
...
+            // Check if this is a new node or an existing node
+            let storage_dir = PathBuf::from(&storage_dir_path);
+            let keys_seed_file = storage_dir.join("keys_seed");
+
+            if !keys_seed_file.exists() && !storage_dir.exists() {
+                // This is a new node and no mnemonic is provided
+                tracing::warn!(
+                    "ldk_node_mnemonic should be set in the [ldk_node] configuration section."
+                );
+            }
```
thesimplekid: Should we error here? If we want to enforce the use of mnemonics on new node I think we need to error and not allow start up not just warn. If we want to allow both with and without mnemonics then we shouldn't warn.

asmogo: My initial implementation was erroring. I thought it might be a better UX to start the node and warn the user.
Erroring might be better since this warning could be missed by the user.

asmogo: fixed in https://github.com/cashubtc/cdk/pull/1010/commits/a8b8e619be646282c4906dd0585b380ce4da0135

crates/cdk-ldk-node/src/lib.rs
asmogo: fixed in https://github.com/cashubtc/cdk/pull/1010/commits/f8e15f32e72030f910b08aa39d020a9f63adf080

crates/cdk-mintd/src/setup.rs
thesimplekid: > My initial implementation was erroring. I thought it might be a better UX to start the node and warn the user.
> Erroring might be better since this warning could be missed by the user.

I think it depends on what we want to allow. If we allow the node be created without the mnemonic its likly the error could be missed and  the node will start to be used with the randomly generated seed and itll be too late to switch to a mnemonic. There for we should accept this and not log an error. 

However, the other option is we do not allow this path and require a mnemonic unless there is an existing seed file then we do allow this path for backwards compatibility. I think this is the best path going forward nodes will use a mnemonic and we do not break current nodes.  Looks like thats what youve done in https://github.com/cashubtc/cdk/commit/a8b8e619be646282c4906dd0585b380ce4da0135

thesimplekid:
ACK a8b8e619be646282c4906dd0585b380ce4da0135

asmogo:
Thanks for the updates. 
ACK https://github.com/cashubtc/cdk/pull/1010/commits/74afea3423d8fe53bc41121ece977ca8af284bbd

PR #1012, opened 2025-08-28
Abstract the HTTP Transport

crodas:
### Description

This PR allows replacing the HTTP transport layer with another library, allowing wallet ffi to provide a better-suited HTTP library that would be used instead of Reqwest.

TODO:
-  ~[ ] Implement a reqwest sync client~ Maybe another client? I think that having it as optional is good enough, the wrappers can implement their own HTTP.
     
      

> Conversely, the functionality in reqwest::blocking must not be executed within an async runtime, or it will panic when attempting to block. If calling directly from an async function, consider using an async [reqwest::Client](https://docs.rs/reqwest/latest/reqwest/struct.Client.html) instead. If the immediate context is only synchronous, but a transitive caller is async, consider changing that caller to use [tokio::task::spawn_blocking](https://docs.rs/tokio/1.47.1/x86_64-unknown-linux-gnu/tokio/task/blocking/fn.spawn_blocking.html) around the calls that need to block.
>
> Source: https://docs.rs/reqwest/latest/reqwest/blocking/index.html

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/test_fees.rs
+++ b/crates/cdk-integration-tests/tests/test_fees.rs
@@ -32,7 +32,7 @@ async fn test_swap() {
             mint_quote.clone(),
             SplitTarget::default(),
             None,
-            tokio::time::Duration::from_secs(15),
+            tokio::time::Duration::from_secs(60),
         )
         .await
         .expect("payment");
@@ -92,7 +92,7 @@ async fn test_fake_melt_change_in_quote() {
             mint_quote.clone(),
             SplitTarget::default(),
             None,
-            tokio::time::Duration::from_secs(15),
+            tokio::time::Duration::from_secs(60),
         )
         .await
         .expect("payment");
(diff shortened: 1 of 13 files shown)


thesimplekid:
> [ ] Implement a reqwest sync client Maybe another client? I think that having it as optional is good enough, the wrappers can implement their own HTTP.

minreq is a decent lib for blocking if we need one. We don't want to use reqwests blocking as its not really blocking.

crates/cdk-integration-tests/tests/test_fees.rs
```diff
@@ -92,7 +92,7 @@ async fn test_fake_melt_change_in_quote() {
             mint_quote.clone(),
             SplitTarget::default(),
             None,
-            tokio::time::Duration::from_secs(15),
+            tokio::time::Duration::from_secs(60),
```
thesimplekid, line 95: I don't mind setting these higher if they solve our ci issues but I think its likely covering up an issue more then solving it, so we should try to solve it for real in another pr.

thesimplekid (commented):
Do we need to allow setting this on the `Wallet::builder` or how does this get used? Maybe we can add an example. But trait and rest of this LGTM

crates/cdk-integration-tests/tests/test_fees.rs
crodas: I'll focus on this later

PR #1022, opened 2025-09-01
feat(cdk): add generic key-value store functionality for mint databases

thesimplekid:
Implements a comprehensive KV store system with transaction support, namespace-based organization, and validation for mint databases.

- Add KVStoreDatabase and KVStoreTransaction traits for generic storage
- Include namespace and key validation with ASCII character restrictions
- Add database migrations for kv_store table in SQLite and PostgreSQL
- Implement comprehensive test suite for KV store functionality
- Integrate KV store traits into existing Database and Transaction bounds

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-sqlite/src/mint/mod.rs
```diff
@@ -31,6 +31,93 @@ mod test {
 
     mint_db_test!(provide_db);
 
+    #[tokio::test]
+    async fn test_kvstore_functionality() {
```
crodas: This test should be moved into the cdk-sql-common crate using generic

crodas: Even better to put this tests inside [cdk-common](https://github.com/cashubtc/cdk/blob/main/crates/cdk-common/src/database/mint/test.rs), have [mint_db_test](https://github.com/cashubtc/cdk/blob/main/crates/cdk-common/src/database/mint/test.rs#L127C1-L135C7) call your tests passing the database implementation through the generic, as in [here](https://github.com/cashubtc/cdk/blob/main/crates/cdk-common/src/database/mint/test.rs#L84-L87)

thesimplekid: cfec30a376cb70ddb2306c84a23ff39e0c8ee023?

crodas: Yes! Perfect

thesimplekid:
> Can we have a wrapper type for the KVStore? Something that makes it easier to use, for instance, where the value is a generic that implements serde serialize to store, and deserialize to return. Other than that, this looks OK and I'll approve it.

We can do this in a follow up I think.

PR #1028, opened 2025-09-02
chore: move `pay_request` logic into cdk lib

lollerfirst:
### Description

move `pay_request` logic into cdk lib

### DIFF
--- a/crates/cdk/Cargo.toml
+++ b/crates/cdk/Cargo.toml
@@ -12,7 +12,7 @@ license.workspace = true
 
 [features]
 default = ["mint", "wallet", "auth"]
-wallet = ["dep:futures", "dep:reqwest", "cdk-common/wallet", "dep:rustls"]
+wallet = ["dep:futures", "dep:reqwest", "cdk-common/wallet", "dep:rustls", "dep:nostr-sdk"]
 mint = ["dep:futures", "dep:reqwest", "cdk-common/mint", "cdk-signatory"]
 auth = ["dep:jsonwebtoken", "cdk-common/auth", "cdk-common/auth"]
 bip353 = ["dep:trust-dns-resolver"]
@@ -44,6 +44,11 @@ utoipa = { workspace = true, optional = true }
 uuid.workspace = true
 jsonwebtoken = { workspace = true, optional = true }
 trust-dns-resolver = { version = "0.23.2", optional = true }
+nostr-sdk = { optional = true, version = "0.41.0", default-features = false, features = [
+    "nip04",
+    "nip44",
+    "nip59"
+]}
 
 # -Z minimal-versions
 sync_wrapper = "0.1.2"
--- a/crates/cdk/src/wallet/payment_request.rs
+++ b/crates/cdk/src/wallet/payment_request.rs
@@ -1,0 +1,160 @@
+//! Utilities for paying NUT-18 Payment Requests.
+//!
+//! This module prepares and broadcasts payments for Cashu NUT-18 payment requests using either
+//! Nostr or HTTP transports when available. If no transport is present in the request, an error
+//! is returned so callers can handle alternative delivery mechanisms explicitly.
+
+use cdk_common::{Amount, PaymentRequest, PaymentRequestPayload, TransportType};
+use nostr_sdk::nips::nip19::Nip19Profile;
+use nostr_sdk::{Client as NostrClient, EventBuilder, FromBech32, Keys};
+use reqwest::Client;
+
+use crate::wallet::SendOptions;
+use crate::Wallet;
+
+/// Pay a NUT-18 PaymentRequest using a specific wallet.
+///
+/// - If the request contains a Nostr or HttpPost transport, it will try those (preferring Nostr).
+/// - If no usable transport is present, this returns an error.
+/// - If the request has no amount, a `custom_amount` must be provided.
+pub async fn pay_request(
+    payment_request: PaymentRequest,
+    matching_wallet: &Wallet,
+    custom_amount: Option<Amount>,
+) -> Result<(), crate::error::Error> {
+    use crate::error::Error;
+
+    let amount = match payment_request.amount {
+        Some(amount) => amount,
+        None => match custom_amount {
+            Some(a) => a,
+            None => return Err(Error::AmountUndefined),
+        },
+    };
+
+    let transports = payment_request.transports.clone();
+
+    // Prefer Nostr to avoid revealing IP, fall back to HTTP POST.
+    let transport = transports
+        .iter()
+        .find(|t| t._type == TransportType::Nostr)
+        .or_else(|| {
+            transports
+                .iter()
+                .find(|t| t._type == TransportType::HttpPost)
+        });
+
+    let prepared_send = matching_wallet
+        .prepare_send(
+            amount,
+            SendOptions {
+                include_fee: true,
+                ..Default::default()
+            },
+        )
+        .await?;
+
+    let token = prepared_send.confirm(None).await?;
+
+    // We need the keysets information to properly convert from token proof to proof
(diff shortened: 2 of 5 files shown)


crates/cdk/src/wallet/payment_request.rs
```diff
@@ -0,0 +1,160 @@
...
+
+use crate::wallet::SendOptions;
+use crate::Wallet;
+
+/// Pay a NUT-18 PaymentRequest using a specific wallet.
+///
+/// - If the request contains a Nostr or HttpPost transport, it will try those (preferring Nostr).
+/// - If no usable transport is present, this returns an error.
+/// - If the request has no amount, a `custom_amount` must be provided.
+pub async fn pay_request(
```
thesimplekid: I think we should impl this on the wallet and have it take &self like most of out other fns.

so the caller is `self.pay_request(payment_request, amount).await?`

crates/cdk/Cargo.toml
```diff
@@ -44,6 +44,11 @@ utoipa = { workspace = true, optional = true }
 uuid.workspace = true
 jsonwebtoken = { workspace = true, optional = true }
 trust-dns-resolver = { version = "0.23.2", optional = true }
+nostr-sdk = { optional = true, version = "0.41.0", default-features = false, features = [
```
thesimplekid: Lets bump nostr-sdk while we're here `0.43`

crates/cdk/Cargo.toml
```diff
@@ -12,7 +12,7 @@ license.workspace = true
 
 [features]
 default = ["mint", "wallet", "auth"]
-wallet = ["dep:futures", "dep:reqwest", "cdk-common/wallet", "dep:rustls"]
+wallet = ["dep:futures", "dep:reqwest", "cdk-common/wallet", "dep:rustls", "dep:nostr-sdk"]
```
thesimplekid: I think nostr should be an optional feature of the wallet what do you think?

a1denvalu3: That would require conditional code compilation within the functions (whether we support the nostr feature or not) and considering that the only two transports for payment requests are either POST or NOSTR i don't think that's worth it.

thesimplekid: But as it is it forces the nostr dep on every cdk wallet even those not using nostr

a1denvalu3: @thesimplekid so is it possible for a feature to depend on another feature? because obviously we'd want the nostr feature only be usable for wallets.

thesimplekid: Yes we do it here for auth in mintd.

https://github.com/cashubtc/cdk/blob/734e62b04acb366e5740cf6844c80efe6570a1f2/crates/cdk-mintd/Cargo.toml#L30

a1denvalu3: @thesimplekid can this be resolved, now?

crates/cdk/src/wallet/payment_request.rs
```diff
@@ -0,0 +1,573 @@
...
+    /// Optional HTLC hash condition (mutually exclusive with `preimage`)
+    pub hash: Option<String>, // HTLC hash
+    /// Optional HTLC preimage (mutually exclusive with `hash`)
+    pub preimage: Option<String>, // HTLC preimage
+    /// Transport type for the request: "nostr", "http", or "none"
+    pub transport: String, // "nostr", "http", or "none"
+    /// Target URL for HTTP transport (required if `transport == http`)
+    pub http_url: Option<String>, // when transport == http
+    /// List of Nostr relay URLs to include in the nprofile (used if `transport == nostr`)
+    pub nostr_relays: Option<Vec<String>>, // when transport == nostr
```
thesimplekid, line 205: We should be able to remove this in favor of the Payment request builder. 

https://github.com/cashubtc/cdk/blob/ce3d25dfb9cc90e0729f22a8eeaba5966034f029/crates/cashu/src/nuts/nut18/payment_request.rs#L90

crates/cdk/src/wallet/payment_request.rs
```diff
@@ -0,0 +1,573 @@
...
+    /// - `(PaymentRequest, None)` otherwise.
+    ///
+    /// Errors when:
+    /// - `unit` cannot be parsed, relay URLs are invalid, or P2PK/HTLC parameters are malformed.
+    ///
+    /// Notes:
+    /// - Sets `single_use = true` to discourage replays.
+    /// - Ephemeral Nostr keys are intentional; keep `NostrWaitInfo` only as long as needed for reception.
+    #[cfg(feature = "nostr")]
+    pub async fn create_request(
```
thesimplekid, line 353: Think this can be removed using the builder as well?

crates/cdk/src/wallet/payment_request.rs
```diff
@@ -0,0 +1,573 @@
...
+            match transport_type.as_str() {
+                "nostr" => {
+                    let keys = Keys::generate();
+                    let relays = if let Some(custom_relays) = &params.nostr_relays {
+                        if !custom_relays.is_empty() {
+                            custom_relays.clone()
+                        } else {
+                            vec![
+                                "wss://relay.nos.social".to_string(),
+                                "wss://relay.damus.io".to_string(),
```
thesimplekid: In the lib I'm not sure we should provide default relays force the add (including cdk-cli) to provide those?

crates/cdk/src/wallet/payment_request.rs
```diff
@@ -0,0 +1,573 @@
...
+            transports,
+            nut10,
+        };
+
+        Ok(req)
+    }
+
+    /// Wait for a Nostr payment for the previously constructed PaymentRequest and receive it into the wallet.
+    #[cfg(feature = "nostr")]
+    pub async fn wait_for_nostr_payment(&self, info: NostrWaitInfo) -> Result<Amount> {
```
thesimplekid, line 511: Should we make this a stream like we do for minting?

thesimplekid (commented):
Few comments. Thanks for working on this

crates/cdk/src/wallet/payment_request.rs
a1denvalu3: All that `PaymentRequestBuilder` does for you is set stuff and then instantiate a `PaymentRequest` with the stuff you've provided. So it's pretty useless in this case. I'm not sure if what you're suggesting here is to use it as the type for the params, use it inside `create_request` or in place of `create_request` (in which case all of the logic goes back to cdk-cli anyway).

thesimplekid: Okay this is fine then. It just seems there is too much over lap between CreateRequestParams and The payment request builder?

crates/cdk/src/wallet/payment_request.rs
a1denvalu3: I've made this into a stream

thesimplekid:
@crodas Can you review the streams?

crates/cdk/src/wallet/streams/nostr.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        if this.rx_future.is_none() {
+            let mut rx = this.rx.take().expect("receiver");
+            this.rx_future = Some(Box::pin(async move {
+                let item = rx.recv().await;
+                (item, rx)
+            }));
+        }
+
+        let mut fut = this.rx_future.take().expect("future present");
```
crodas: Instead of expect() / unwrap(), I would advise to instead do `ok_or(Error::Internal)?`, I think that having code that panics in prod is not good.

a1denvalu3: I don't mind changing this but I think this should be an unreachable state, so we are "expecting" not to reach it.

Issue #767, opened 2025-05-27
Should this be bytes?


crodas:
Should this be bytes?

_Originally posted by @davidcaseria in https://github.com/cashubtc/cdk/pull/509#discussion_r2098855406_

PR #1032, opened 2025-09-03
Update the signatory.proto file to match NUT-XXX

### LINKED ISSUE #762: Possibly out of scope but it would be nice if this parameter would accept a slice of integers (all possible amounts) instead of a single exponent (max order).
Possibly out of scope but it would be nice if this parameter would accept a slice of integers (all possible amounts) instead of a single exponent (max order).

_Originally posted by @callebtc in https://github.com/cashubtc/cdk/pull/509#discussion_r1895506004_

crodas:
### Description

Fixes #762, #767

Source: https://github.com/cashubtc/nuts/pull/250/files

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-signatory/src/proto/signatory.proto
+++ b/crates/cdk-signatory/src/proto/signatory.proto
@@ -131,28 +131,6 @@ message BlindSignatureDLEQ {
   bytes s = 2;
 }
 
-// Witness type
-message Witness {
-  oneof witness_type {
-    P2PKWitness p2pk_witness = 1;
-    HTLCWitness htlc_witness = 2;
-  }
-}
-
-// P2PKWitness type
-message P2PKWitness {
-  // List of signatures
-  repeated string signatures = 1;
-}
-
-// HTLCWitness type
-message HTLCWitness {
-  // Preimage
-  string preimage = 1;
-  // List of signatures
-  repeated string signatures = 2;
-}
-
 enum ErrorCode {
   ERROR_CODE_UNSPECIFIED = 0;
   ERROR_CODE_AMOUNT_OUTSIDE_LIMIT = 1;


crates/cdk-sql-common/src/mint/migrations/postgres/3_add_signatory_amounts.sql
thesimplekid, line 1: can we name these like we do for sqlite with the date.

crodas: Added tests in 7c86efe0887c2ebc91f42c54389586d947a038e4

crates/cdk-signatory/src/proto/signatory.proto
```diff
@@ -120,6 +131,28 @@ message BlindSignatureDLEQ {
...
+  repeated string signatures = 1;
+}
+
+// HTLCWitness type
+message HTLCWitness {
+  // Preimage
+  string preimage = 1;
+  // List of signatures
+  repeated string signatures = 2;
+}
```
thesimplekid: Can't these be removed? The signatory does not verify HTLC or P2PK the mint does.

thesimplekid (commented):
Other then that LGTM

crates/cdk-signatory/src/proto/signatory.proto
crodas: I will remove it and update the NUT document

crodas: Fixed in d74a3a3b61e73b4fd18a2db44b98bad182b929f1

thesimplekid (approved):

PR #1050, opened 2025-09-08
Close websocket connections sooner

crodas:
### Description

Close websocket connections sooner

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-axum/src/ws/mod.rs
+++ b/crates/cdk-axum/src/ws/mod.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 
-use axum::extract::ws::{Message, WebSocket};
+use axum::extract::ws::{CloseFrame, Message, WebSocket};
 use cdk::mint::QuoteId;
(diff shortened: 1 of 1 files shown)


crates/cdk-axum/src/ws/mod.rs
```diff
@@ -74,12 +74,20 @@ pub async fn main_websocket(mut socket: WebSocket, state: MintState) {
...
             }
-            Some(Ok(Message::Text(text))) = socket.next() => {
+
+            from_ws = socket.next() => {
+                let text = if let Some(Ok(Message::Text(message))) = from_ws {
+                    message
+                } else {
+                    // Client disconnected or sent something that is not a text
+                    return;
+                };
```
thesimplekid: I think we should handle close explicitly. Also this will disconnect on a Ping which is not correct. 

https://docs.rs/axum/latest/axum/extract/ws/enum.Message.html

Issue #1029, opened 2025-09-02
Feature request cdk-wallet: store P2PK key and lookup automatically on token receive


callebtc:
creating a P2PK key should store the private key in the cdk wallet db. when receiving a token that's locked to one of our keys, we can lookup the private key and sign the token automatically.

lescuer97:
> [@lescuer97](https://github.com/lescuer97) are you planning to take this or [@crodas](https://github.com/crodas)?

I need to finish some stuff first but I would like to take this one.

thesimplekid:
closed by https://github.com/cashubtc/cdk/pull/1466

PR #1053, opened 2025-09-08
feat: P2PK key storage and auto-sign on receive

yashbhutwala:
Implements #1029

- Add wallet DB trait for P2PK key storage (add/get/list/remove)
- SQL-backed storage + migrations (sqlite/postgres)
- Auto-merge stored keys with options to sign proofs on receive
- Wallet helpers to add/generate/list P2PK signing keys

Followed development guide: formatted, typos, clippy; validated pure tests.

### DIFF
--- a/crates/cdk/src/wallet/receive.rs
+++ b/crates/cdk/src/wallet/receive.rs
@@ -104,9 +104,7 @@ impl Wallet {
                         if !p2pk_signing_keys.contains_key(&x_only)
                             && !missing_db_keys.contains(&x_only)
                         {
-                            if let Some(stored) =
-                                self.localstore.get_p2pk_key(pubkey.clone()).await?
-                            {
+                            if let Some(stored) = self.localstore.get_p2pk_key(pubkey).await? {
                                 p2pk_signing_keys.insert(x_only, stored);
                             } else {
                                 missing_db_keys.insert(x_only);
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,6 +111,7 @@ strum = "0.27.1"
 strum_macros = "0.27.1"
 rustls = { version = "0.23.27", default-features = false, features = ["ring"] }
 prometheus = { version = "0.13.4", features = ["process"], default-features = false }
+zeroize = "1"
 
 
 
--- a/crates/cdk-ffi/Cargo.toml
+++ b/crates/cdk-ffi/Cargo.toml
@@ -30,7 +30,6 @@ uniffi = { version = "0.29", features = ["cli", "tokio"] }
 url = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
 
-
 [features]
 default = ["postgres"]
 # Enable Postgres-backed wallet database support in FFI
--- a/crates/cdk-ffi/src/database.rs
+++ b/crates/cdk-ffi/src/database.rs
@@ -94,11 +94,11 @@ pub trait WalletDatabase: Send + Sync {
 
     // P2PK signing key storage
     /// Store a P2PK signing key
-    async fn add_p2pk_key(&self, secret_key: SecretKey) -> Result<(), FfiError>;
+    async fn add_p2pk_key(&self, secret_key: Arc<SecretKey>) -> Result<(), FfiError>;
     /// Fetch a P2PK signing key by public key
-    async fn get_p2pk_key(&self, pubkey: PublicKey) -> Result<Option<SecretKey>, FfiError>;
+    async fn get_p2pk_key(&self, pubkey: PublicKey) -> Result<Option<Arc<SecretKey>>, FfiError>;
     /// List stored P2PK signing keys
-    async fn list_p2pk_keys(&self) -> Result<Vec<P2pkSigningKey>, FfiError>;
+    async fn list_p2pk_keys(&self) -> Result<Vec<Arc<P2pkSigningKey>>, FfiError>;
     /// Remove a stored P2PK signing key by public key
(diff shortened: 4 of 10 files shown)


yashbhutwala:
> Should the wallet use the seed to make deterministic secrets instead of randomly generated secret keys (e.g., using BIP32)?

Great question — and I think we’re talking about two different classes of secrets:

- Ephemeral blinding secrets (used when minting proofs): These are intentionally one-time, unlinkable values. Making them deterministic from a seed doesn’t buy us recoverability (Cashu is bearer; you need the proofs, not a seed, to restore funds) and introduces coordination/duplication risks across devices. For these, we should keep using high-entropy randomness.
- Long‑lived keys (e.g., P2PK lock keys, auth keys, or anything we must be able to re-derive): These should indeed come from the wallet seed. Using hardened derivation via BIP32/SLIP‑0010 is appropriate here so users can restore keys across devices.

In this PR, the “secret keys” in question are of the first type, so I kept them random by design.

crates/cdk-common/src/database/wallet.rs
```diff
@@ -118,4 +118,23 @@ pub trait Database: Debug {
...
+    async fn get_p2pk_key(&self, _pubkey: PublicKey) -> Result<Option<SecretKey>, Self::Err> {
+        Ok(None)
+    }
+    /// List all stored P2PK signing keys. Default returns empty list.
+    async fn list_p2pk_keys(&self) -> Result<Vec<SecretKey>, Self::Err> {
+        Ok(vec![])
+    }
+    /// Remove a stored P2PK signing key by pubkey. Default no-op.
+    async fn remove_p2pk_key(&self, _pubkey: PublicKey) -> Result<(), Self::Err> {
+        Ok(())
```
thesimplekid: I don't think these should be no ops especially ones that do not return errors we should force db backends to support this.

crates/cdk-integration-tests/tests/p2pk_autosign.rs
thesimplekid, line 1: I don't think these tests get run as part of ci? We can just move them to the pure tests.

crates/cdk/src/wallet/mod.rs
```diff
@@ -662,6 +662,28 @@ impl Wallet {
...
+        let sk = crate::nuts::SecretKey::generate();
+        let pk = sk.public_key();
+        self.localstore.add_p2pk_key(sk).await?;
+        Ok(pk)
+    }
+
+    /// List stored P2PK signing keys
+    pub async fn list_p2pk_signing_keys(&self) -> Result<Vec<crate::nuts::SecretKey>, Error> {
+        self.localstore.list_p2pk_keys().await.map_err(Into::into)
+    }
```
thesimplekid: We store the pubkey and secret key, why don't we return both?

crates/cdk/src/wallet/receive.rs
```diff
@@ -51,9 +51,29 @@ impl Wallet {
...
+        }
+
+        // Merge in any keys stored in the wallet database
+        let stored_keys = self.localstore.list_p2pk_keys().await.unwrap_or_default();
+        for sk in stored_keys {
+            let x = sk.x_only_public_key(&SECP256K1).0;
+            if seen.insert(x) {
+                merged_keys.push(sk);
+            }
+        }
```
thesimplekid, line 63: It seems to me the effect of this is for each receive even ones where no proof needs a signature we get all secret keys from the db and create a list of them, this seem unnecessary. Why don't we iter over the proofs get all pubkeys for those and then get the secret keys for only those pubkeys.

thesimplekid (changes requested):

crates/cdk-common/src/database/wallet.rs
yashbhutwala: Updated the trait so these methods are mandatory (no default no-ops) and wired them through every wallet backend. The SQL store keeps its existing table, Redb has a v5 migration to create p2pk_signing_keys, and the FFI bridge now exposes the concrete impls, so all backends persist and surface P2PK keys consistently.

crates/cdk-integration-tests/tests/p2pk_autosign.rs
yashbhutwala: Moved both autosign scenarios into integration_tests_pure.rs so they run with the rest of the pure wallet/mint CI suite, and deleted the standalone file

crates/cdk/src/wallet/mod.rs
yashbhutwala: Switched list_p2pk_signing_keys (and the underlying db APIs) to return (pubkey, secret) pairs. Callers now get the stored public key alongside the secret so they can correlate entries without re-deriving.

crates/cdk/src/wallet/receive.rs
yashbhutwala: Reworked the merge logic: we seed the map with any keys provided in ReceiveOptions, then, as we inspect each proof, we only call get_p2pk_key for the pubkeys actually referenced (track misses to avoid repeat queries). No more list_p2pk_keys() scan on every receive.

davidcaseria:
Should there be a wallet function to create a p2pk payment request that generates and saves the key as a convenience method so users don't have to remember to save it themselves?

thesimplekid:
> Should there be a wallet function to create a p2pk payment request that generates and saves the key as a convenience method so users don't have to remember to save it themselves?

Yes that's why I haven't merged. I don't think user should have to generate a secret key like they do now; should call a fn as you suggest.

Issue #342, opened 2024-09-13
Wallet: determine valid amounts from mint's `keys` response


prusnak:
Same as https://github.com/cashubtc/nutshell/issues/618

ISTM that CDK also hardcodes/assumes amounts of `2^n` and not the ones provided by the mint:

https://github.com/cashubtc/cdk/blob/f978b080d5d77455f23fdcc7dc44036277e792f2/crates/cdk/src/wallet/mod.rs#L756-L758

thesimplekid:
Yes I noticed this a bit ago but haven't fixed it yet.

https://github.com/cashubtc/cdk/issues/124

PR #1055, opened 2025-09-09
Include supported amounts instead of assuming the power of 2

### LINKED ISSUE #955: Wallet assumes the mint have the amounts of power of two
Although that may be the case, the mint shares all the supported amount, and that last should be used instead.

crodas:
### Description

The mint's signatory defines the amounts and the wallet, and the mint, when paying, should use them instead of assuming the supported amounts are 2^(0..32), which is not part of the spec.

Fixes #955, #342

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/integration_tests_pure.rs
+++ b/crates/cdk-integration-tests/tests/integration_tests_pure.rs
@@ -339,6 +339,7 @@ async fn test_attempt_to_swap_by_overflowing() {
             cdk::Error::NUT03(cdk::nuts::nut03::Error::Amount(_)) => (),
             cdk::Error::AmountOverflow => (),
             cdk::Error::AmountError(_) => (),
+            cdk::Error::TransactionUnbalanced(_, _, _) => (),
             _ => {
                 panic!("Wrong error returned in swap overflow {:?}", err);
             }


crates/cdk/src/wallet/keysets.rs
```diff
@@ -153,19 +153,32 @@ impl Wallet {
...
     ///
     /// Returns the input fee rate (per-proof-per-thousand) for a specific keyset ID from
     /// cached keysets in the local database. This is an offline operation that does not
     /// contact the mint. If the keyset is not found locally, returns an error.
-    pub async fn get_keyset_fees_by_id(&self, keyset_id: Id) -> Result<u64, Error> {
-        self.get_keyset_fees()
+    pub async fn get_keyset_fees_and_amounts_by_id(
+        &self,
+        keyset_id: Id,
+    ) -> Result<(u64, Vec<u64>), Error> {
```
thesimplekid: I have a preference for a struct with named fields over the tuple

crates/cdk/src/wallet/proofs.rs
```diff
@@ -13,6 +13,12 @@ use crate::nuts::{
 use crate::types::ProofInfo;
 use crate::{ensure_cdk, Amount, Error, Wallet};
 
+/// Fees and Amounts supported
+pub type FeesAndAmounts = (u64, Vec<u64>);
```
thesimplekid: Could just define a struct here instead of type for the tuple.

crates/cdk/src/wallet/mod.rs
```diff
@@ -325,34 +325,32 @@ impl Wallet {
 
     /// Get amounts needed to refill proof state
     #[instrument(skip(self))]
-    pub async fn amounts_needed_for_state_target(&self) -> Result<Vec<Amount>, Error> {
+    pub async fn amounts_needed_for_state_target(
+        &self,
+        amounts_ppk: &[u64],
```
thesimplekid: I don't think the amounts are ppk? The is ppk but the amounts are just the amounts the keyset supports?

crates/cdk/src/wallet/keysets.rs
crodas: That's a fair ask, I'll make it happen.

crates/cdk/src/wallet/proofs.rs
crodas: Yep, I'll do that.

crates/cdk/src/wallet/mod.rs
crodas: A better name would be keyset_amounts. I'll rename them

crodas: Fixed in 3d198e87b94490de028031c70058fd59561cee87

crates/cdk/src/wallet/proofs.rs
crodas: Fixed in 3d198e87b94490de028031c70058fd59561cee87

thesimplekid (approved):

PR #1060, opened 2025-09-10
fix: replace std::time with instant for wasm

gudnuf:
### Description

`std::time` is unusable in wasm environments. I replaced all usage of `std::time` relevant to a wasm wallet with the `instant` crate. For native targets instant is just a type alias for std::time

-----

### Notes to the reviewers

Side note: the [instant](https://crates.io/crates/instant) documentation says its no longer maintained, but cdk already used instant so I decided to keep it. 

-----

#### FIXED

----

### Checklist


thesimplekid:
Since you pointed out instant is not maintained can we remove it? Seems like it should be pretty easy to just switch out for https://docs.rs/web-time/latest/web_time/

gudnuf:
> Since you pointed out instant is not maintained can we remove it? Seems like it should be pretty easy to just switch out for https://docs.rs/web-time/latest/web_time/

I removed instant and replaced with web_time, but I left the rest of the usage of std::time that doesn't seem to be used by a wasm wallet

PR #1062, opened 2025-09-10
fix: make http wallet subscriptions wasm compatible

gudnuf:
### Description

Makes the wallet subscriptions compatible with wasm. 

For wasm, we now use gloo timers instead of tokio intervals. 
Also, when creating the http worker for wasm we don't return a join handle because that's not a thing in wasm. Instead we break out of the loop manually.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/subscription/mod.rs
+++ b/crates/cdk/src/wallet/subscription/mod.rs
@@ -13,6 +13,8 @@ use cdk_common::subscription::Params;
 use tokio::sync::{mpsc, RwLock};
 use tokio::task::JoinHandle;
 use tracing::error;
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen_futures;
 
 use super::Wallet;
 use crate::mint_url::MintUrl;
@@ -207,7 +209,7 @@ impl SubscriptionClient {
             new_subscription_notif,
             on_drop_notif,
             subscriptions: subscriptions.clone(),
-            worker: Some(Self::start_worker(
+            worker: Self::start_worker(
                 prefer_ws_method,
                 http_client,
                 url,
@@ -215,7 +217,7 @@ impl SubscriptionClient {
                 new_subscription_recv,
                 on_drop_recv,
                 wallet,
-            )),
+            ),
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/cdk/src/wallet/subscription/mod.rs
```diff
@@ -228,7 +230,7 @@ impl SubscriptionClient {
         new_subscription_recv: mpsc::Receiver<SubId>,
         on_drop_recv: mpsc::Receiver<SubId>,
         wallet: Arc<Wallet>,
-    ) -> JoinHandle<()> {
+    ) -> Option<JoinHandle<()>> {
```
gudnuf, line 233: The other option I could think of was to implement separate wam functions that don't return a join handle, but this seemed better as the only change we need to make is to have `http_worker` return `None` is wasm

crates/cdk/src/wallet/subscription/mod.rs
```diff
@@ -304,12 +306,15 @@ impl SubscriptionClient {
         );
 
         #[cfg(target_arch = "wasm32")]
-        let ret = tokio::task::spawn_local(http_worker);
+        {
+            wasm_bindgen_futures::spawn_local(http_worker);
+            None
```
crodas, line 311: Could you please elaborate on why returning None is better here, rather than returning the JoinHandle?

Other than this, everything else looks good, thanks @gudnuf

gudnuf: From what I could figure out join handles aren't a thing in wasm because it's single threaded. 

[See](https://github.com/cashubtc/cdk/pull/1062/files#diff-b0ad00b24f33a6357d2d27de3efdbccf92d366f0b93bf242b3ef938aa5e439eaR148) how I had to break out of the loop manually for the wasm version of `http_main` when the SubscriptionClient is dropped. 

I'm not going to pretend to understand everything that's going on here so if you think there's a better way I can try to figure it. 

Update:

Another option would be to try to return a dummy join handle

or

`wasm_bindgen_futures::spawn_local` returns `()`, so we could make the `SubscriptionClient` struct like:

```rs
pub struct SubscriptionClient {
    new_subscription_notif: mpsc::Sender<SubId>,
    on_drop_notif: mpsc::Sender<SubId>,
    subscriptions: Arc<RwLock<HashMap<SubId, WsSubscriptionBody>>>,
    #[cfg(not(target_arch = "wasm32"))]
    worker: Option<JoinHandle<()>>,
    #[cfg(target_arch = "wasm32")]
    worker: Option<()>,
}
```

However, I didn't like this because then we need seperate wasm implementations for everything that currently returns a JoinHandle. The `worker` join handle was already optional, so I decided to take advantage of that.

PR #1064, opened 2025-09-10
feat: per-request tor circuits with arti

lollerfirst:
# Tor transport for CDK (arti + hyper integration)

## Summary

This PR introduces a Tor-backed HTTP transport implementation for the `cdk` crate under the `tor` feature.
It enables the wallet/mint connector to issue HTTP(S) requests over the Tor network using the Tor Project’s
Arti client and a Hyper 0.14-based connector.

- New module: `wallet::mint_connector::transport::tor_transport`
- New type: `TorAsync` that implements the existing `Transport` trait when the `tor` feature is enabled.
- Uses `arti-client` and `arti-hyper` (both v0.19.0), with `hyper` 0.14 and `http` 0.2 for the Tor code path.
- Bootstraps a Tor client once (lazily), reusing it across requests.
- Supports isolation tokens via Arti’s `IsolationToken`/`StreamPrefs` primitives (per-request isolation follow‑up noted below).

## Motivation

- Provide an optional, native Tor transport for wallet/mint HTTP interactions without relying on a local SOCKS proxy.
- Keep the implementation self-contained, feature-gated, and compatible with Arti 0.19.

## Implementation details

- `TorAsync` stores:
  - `Arc<tokio::sync::OnceCell<TorClient<PreferredRuntime>>>` to lazily create and bootstrap the Tor client once.
  - Optional `IsolationToken` used for per-request isolation.
  - Optional host matcher regex to restrict which hosts may use the Tor transport (aligned with the existing `with_proxy` hook semantics in other transports).

- On first use, the OnceCell runs `TorClient::create_bootstrapped(TorClientConfig::default())` to bootstrap.

- HTTPS across Tor is performed by `arti_hyper::ArtiHttpConnector` with a `tls-api-native-tls` connector.
  - We build a `hyper` 0.14 client using the Arti connector to perform requests.

- Headers and request building use `http` 0.2 types to avoid conflicts with other `http` versions.

### Isolation

- The `Transport::new_isolated()` method returns a clone with a fresh `IsolationToken`.
- `StreamPrefs` is used to carry the token.
- Per-request isolation is now enforced: when an isolation token is set, the transport creates a `StreamPrefs` with that token and calls `tor_client.clone_with_prefs(prefs)`, then builds the `ArtiHttpConnector` with this cloned client for the request. This ensures the request uses its own isolated circuit preferences.

## Dependency changes (tor path only)

- Added (optional under `tor`):
  - `tor-rtcompat = { version = "0.19.0", features = ["tokio", "rustls"] }`
  - `tls-api = "0.9"`
  - `tls-api-native-tls = "0.9"`
- Adjusted (tor path):
  - `hyper = "0.14"`
  - `http = "0.2"`
  - Kept `arti-client = "0.19.0"` and `arti-hyper = "0.19.0"`.
- Removed for tor path: `http-body-util` (to avoid pulling `http` 1.x and mixing header types across code paths).

These changes localize the Tor transport code to the `hyper 0.14` + `http 0.2` ecosystem required by `arti-hyper 0.19`, avoiding header type mismatches.

## Backwards compatibility

- Tor functionality is guarded behind the `tor` feature; default features remain unchanged.
- The existing non-Tor HTTP transport is unaffected.

## Usage

- Build with Tor support:

  ```sh
  cargo build -p cdk --features tor
  ```

- Create a transport and use as usual. The `TorAsync` transport is selected automatically when building with the `tor` feature (based on how you wire it in your wallet/mint connector).

- To create an isolated transport (fresh circuits):

  ```rust
  // Given a TorAsync transport instance `t`, get a new isolated clone:
  let isolated = t.new_isolated();
  ```

### DIFF
--- a/crates/cdk/Cargo.toml
+++ b/crates/cdk/Cargo.toml
@@ -21,6 +21,17 @@ bip353 = ["dep:hickory-resolver"]
 swagger = ["mint", "dep:utoipa", "cdk-common/swagger"]
 bench = []
 http_subscription = []
+tor = [
+    "wallet",
+    "dep:arti-client",
+    "dep:arti-hyper",
+    "dep:hyper",
+    "dep:http",
+    "dep:rustls",
+    "dep:tor-rtcompat",
+    "dep:tls-api",
+    "dep:tls-api-native-tls",
+]
 prometheus = ["dep:cdk-prometheus"]
 
 [dependencies]
@@ -39,6 +50,7 @@ serde_json.workspace = true
 serde_with.workspace = true
 tracing.workspace = true
 thiserror.workspace = true
+
 futures = { workspace = true, optional = true, features = ["alloc"] }
 url.workspace = true
 utoipa = { workspace = true, optional = true }
@@ -73,13 +85,23 @@ tokio-tungstenite = { workspace = true, features = [
     "rustls-tls-native-roots",
     "connect"
 ] }
+# Tor dependencies (optional; enabled by feature "tor")
+hyper = { version = "0.14", optional = true, features = ["client", "http1", "http2"] }
+http = { version = "0.2", optional = true }
+arti-client = { version = "0.19.0", optional = true, default-features = false, features = ["tokio", "rustls"] }
+arti-hyper = { version = "0.19.0", optional = true }
 rustls = { workspace = true, optional = true }
+tor-rtcompat = { version = "0.19.0", optional = true, features = ["tokio", "rustls"] }
+tls-api = { version = "0.9", optional = true }
+tls-api-native-tls = { version = "0.9", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 tokio = { workspace = true, features = ["rt", "macros", "sync", "time"] }
 cdk-signatory = { workspace = true, default-features = false }
 getrandom = { version = "0.2", features = ["js"] }
 ring = { version = "0.17.14", features = ["wasm32_unknown_unknown_js"] }
+rustls = { workspace = true, optional = true }
+
 uuid = { workspace = true, features = ["js"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
--- a/crates/cdk/src/wallet/mint_connector/http_client.rs
+++ b/crates/cdk/src/wallet/mint_connector/http_client.rs
@@ -47,6 +47,31 @@ impl<T> HttpClient<T>
 where
     T: Transport + Send + Sync + 'static,
 {
+    /// Create new [`HttpClient`] with a provided transport implementation.
+    #[cfg(feature = "auth")]
+    pub fn with_transport(
+        mint_url: MintUrl,
+        transport: T,
+        auth_wallet: Option<AuthWallet>,
+    ) -> Self {
+        Self {
+            transport: transport.into(),
+            mint_url,
+            auth_wallet: Arc::new(RwLock::new(auth_wallet)),
+            cache_support: Default::default(),
+        }
+    }
+
+    /// Create new [`HttpClient`] with a provided transport implementation.
+    #[cfg(not(feature = "auth"))]
+    pub fn with_transport(mint_url: MintUrl, transport: T) -> Self {
+        Self {
+            transport: transport.into(),
+            mint_url,
+            cache_support: Default::default(),
+        }
+    }
+
     /// Create new [`HttpClient`]
     #[cfg(feature = "auth")]
     pub fn new(mint_url: MintUrl, auth_wallet: Option<AuthWallet>) -> Self {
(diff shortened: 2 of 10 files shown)


crates/cdk/src/wallet/mint_connector/transport.rs
```diff
@@ -17,26 +17,36 @@ pub trait Transport: Default + Send + Sync + Debug + Clone {
...
-        proxy: Url,
+        proxy: url::Url,
         host_matcher: Option<&str>,
         accept_invalid_certs: bool,
-    ) -> Result<(), Error>;
+    ) -> Result<(), super::Error>;
+
+    /// Create a new isolated instance of this transport. For Tor-backed transports,
+    /// this returns a clone bound to a fresh isolation token; for others this is a clone.
+    fn new_isolated(&self) -> Self {
```
crodas: The HTTP client should be stateless, so I see no value in cloning to use once, aside from it not being a generic need. Therefore, I think `new_isolated` is not generic enough to be included here; instead, a new IsolationToken should be passed to the request function (inside the function, the cloning of the Tor client is already happening).

Other than this little feedback, everything looks great.  Thank you @lollerfirst

a1denvalu3: @crodas  I wonder if I should make it that `request` always draws a fresh isolation token and clones with preferences with that isolation token. I don't really see a point in allowing for a choice here.

crodas: Yes, that would be even better I think, great idea.

crates/cdk/src/wallet/mint_connector/transport.rs
```diff
@@ -178,3 +182,184 @@ impl Transport for Async {
         })
     }
 }
+
+#[cfg(feature = "tor")]
+pub mod tor_transport {
```
thesimplekid: Can you put this tor_transport as its own mod https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#sub-modules

crates/cdk/Cargo.toml
```diff
@@ -66,13 +79,33 @@ tokio-tungstenite = { workspace = true, features = [
...
 ring = { version = "0.17.14", features = ["wasm32_unknown_unknown_js"] }
+hyper = { version = "0.14", optional = true, features = ["client", "http1", "http2"] }
+http = { version = "0.2", optional = true }
+arti-client = { version = "0.19.0", optional = true, default-features = false, features = ["tokio", "rustls"] }
+arti-hyper = { version = "0.19.0", optional = true }
+once_cell = { version = "1", optional = true }
+rustls = { workspace = true, optional = true }
+tor-rtcompat = { version = "0.19.0", optional = true, features = ["tokio", "rustls"] }
+tls-api = { version = "0.9", optional = true }
+tls-api-native-tls = { version = "0.9", optional = true }
```
thesimplekid: It seems to be the same deps are in both the wasm32 and not wasm32? Can't we just move them out of the target flags?

crates/cdk/src/wallet/mint_connector/mod.rs
```diff
@@ -21,7 +21,11 @@ pub mod transport;
 /// Auth HTTP Client with async transport
 #[cfg(feature = "auth")]
 pub type AuthHttpClient = http_client::AuthHttpClient<transport::Async>;
+/// Tor Http Client with async transport
+#[cfg(feature = "tor")]
+pub type HttpClient = http_client::HttpClient<transport::tor_transport::TorAsync>;
 /// Http Client with async transport
+#[cfg(not(feature = "tor"))]
```
thesimplekid: I think we talked about this but to refresh. I am correct in that if the tor feature is enabled it will always be a tor client and there is not way to configure at runtime what client to use? Is there a reason we have to do this or its a choice?

I'm thinking there is going to wallets that want to choose this at runtime not compile time for example giving users a setting to turn on or off tor depending on their needs, this is what mutiny wallet does.

crates/cdk/src/wallet/mint_connector/http_client.rs
```diff
@@ -215,7 +236,8 @@ where
     #[instrument(skip(self), fields(mint_url = %self.mint_url))]
     async fn get_mint_keysets(&self) -> Result<KeysetResponse, Error> {
         let url = self.mint_url.join_paths(&["v1", "keysets"])?;
-        self.transport.http_get(url, None).await
+        let transport = self.transport.clone();
```
crodas, line 245: Why are cloning the transport before using it?

crates/cdk/src/wallet/mint_connector/transport/tor_transport.rs
```diff
@@ -0,0 +1,176 @@
...
+pub struct TorAsync {
+    tor: Arc<OnceCell<arti_client::TorClient<tor_rtcompat::PreferredRuntime>>>,
+    isolation: IsolationToken,
+}
+
+impl Clone for TorAsync {
+    fn clone(&self) -> Self {
+        Self {
+            tor: self.tor.clone(),
+            isolation: IsolationToken::new(),
```
crodas: This cloning should occur _before_ calling any GET/POST operation, inside those functions, to avoid cloning outside of this transport. This requirement is unique to this transport and not global; therefore, this behavior should be contained within this context. 

Related to https://github.com/cashubtc/cdk/pull/1064/files#r2349830048

lollerfirst:
@thesimplekid 

- Reworked Tor transport to use a pool of isolated arti Tor clients.
  - Bootstrap a single TorClient once, then build N isolated clients via TorClient::isolated_client().
  - Each request selects the next client in round-robin using an AtomicUsize (i+1 % L).
- Added async constructors:
  - TorAsync::new().await (default pool size = 10)
  - TorAsync::with_pool_size(n).await
- Updated MultiMintWallet::new_with_tor to initialize TorAsync with TorAsync::new().await (instead of Default).
- Requests now use ArtiHttpConnector with the selected isolated client; HTTP behavior unchanged.

Issue #1034, opened 2025-09-04
trust dns trust dns has been rebranded to Hickory DNS


thesimplekid:
We use this for bip353 so we need to update out dep. 

https://docs.rs/trust-dns-resolver/latest/trust_dns_resolver/

PR #1068, opened 2025-09-12
Add `resolve_dns_txt` to HttpTransport and MintConnector

### LINKED ISSUE #1036: BIP353 DNS query does not use mint connector
Currently, the get_bip353_quote function internally uses trust_dns to perform DNS queries. This means that if a wallet is using a custom mint connector—for example, one designed for Tor—the DNS query will not be routed through Tor, potentially exposing the user's IP address.

One solution is to add a get_bip353 method to the MintConnector. This method would be responsible for resolving and verifying the BIP-353 address. In our default MintConnector, we can continue using the existing library (trust_dns), but move the logic from the mint_quote function into the connector itself.

This approach opens the door for other transport mechanisms, to implement their own logic for resolving addresses privately.

crodas:
### Description

Fixes #1036, #1034

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/mint_connector/mod.rs
+++ b/crates/cdk/src/wallet/mint_connector/mod.rs
@@ -28,6 +28,9 @@ pub type HttpClient = http_client::HttpClient<transport::Async>;
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 pub trait MintConnector: Debug {
+    /// Resolve the DNS record getting the TXT value
+    async fn resolve_dns_txt(&self, domain: &str) -> Result<Vec<String>, Error>;
+
     /// Get Active Mint Keys [NUT-01]
     async fn get_mint_keys(&self) -> Result<Vec<KeySet>, Error>;
     /// Get Keyset Keys [NUT-01]
--- a/crates/cdk/src/wallet/mint_connector/mod.rs
+++ b/crates/cdk/src/wallet/mint_connector/mod.rs
@@ -30,9 +30,7 @@ pub type HttpClient = http_client::HttpClient<transport::Async>;
 pub trait MintConnector: Debug {
     #[cfg(feature = "bip353")]
     /// Resolve the DNS record getting the TXT value
-    async fn resolve_dns_txt(&self, _domain: &str) -> Result<Vec<String>, Error> {
-        todo!()
-    }
+    async fn resolve_dns_txt(&self, _domain: &str) -> Result<Vec<String>, Error>;
 
     /// Get Active Mint Keys [NUT-01]
     async fn get_mint_keys(&self) -> Result<Vec<KeySet>, Error>;
--- a/crates/cdk/src/wallet/mint_connector/mod.rs
+++ b/crates/cdk/src/wallet/mint_connector/mod.rs
@@ -28,7 +28,7 @@ pub type HttpClient = http_client::HttpClient<transport::Async>;
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 pub trait MintConnector: Debug {
-    #[cfg(feature = "bip353")]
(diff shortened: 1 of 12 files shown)


crates/cdk/src/wallet/mint_connector/transport.rs
```diff
@@ -5,6 +5,10 @@ use cdk_common::AuthToken;
 use reqwest::Client;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
+#[cfg(feature = "bip353")]
+use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
+#[cfg(feature = "bip353")]
+use trust_dns_resolver::TokioAsyncResolver;
```
thesimplekid: Can we change this to https://docs.rs/hickory-resolver/latest/hickory_resolver/

crates/cdk/src/wallet/mint_connector/transport.rs
```diff
@@ -14,6 +18,9 @@ use crate::error::ErrorResponse;
 #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 pub trait Transport: Default + Send + Sync + Debug + Clone {
+    /// DNS resolver to get a TXT record from a domain name
+    async fn resolve_dns_txt(&self, domain: &str) -> Result<Vec<String>, Error>;
```
thesimplekid: Should we feature flag it here for bip353 instead of having the fn that just returns an error?

Also option to just remove the feature and just always have it?

crodas: I like to always have it, but returning an error, so using is easier to use

thesimplekid: Why? It seems having a fn that will just always error is more confusing.

crates/cdk/src/wallet/mint_connector/transport.rs
```diff
@@ -14,6 +20,12 @@ use crate::error::ErrorResponse;
 #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 pub trait Transport: Default + Send + Sync + Debug + Clone {
+    #[cfg(feature = "bip353")]
+    /// DNS resolver to get a TXT record from a domain name
+    async fn resolve_dns_txt(&self, _domain: &str) -> Result<Vec<String>, Error> {
+        todo!()
+    }
```
thesimplekid:
```suggestion
    async fn resolve_dns_txt(&self, _domain: &str) -> Result<Vec<String>, Error>;
```

crates/cdk/src/wallet/mint_connector/mod.rs
```diff
@@ -28,6 +28,12 @@ pub type HttpClient = http_client::HttpClient<transport::Async>;
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 pub trait MintConnector: Debug {
+    #[cfg(feature = "bip353")]
+    /// Resolve the DNS record getting the TXT value
+    async fn resolve_dns_txt(&self, _domain: &str) -> Result<Vec<String>, Error> {
+        todo!()
+    }
```
thesimplekid:
```suggestion
    async fn resolve_dns_txt(&self, _domain: &str) -> Result<Vec<String>, Error>;
```

thesimplekid (changes requested):

crates/cdk/src/wallet/mint_connector/mod.rs
crodas: Fixed in c2bf0b1f

crates/cdk/src/wallet/mint_connector/transport.rs
crodas: Fixed in c2bf0b1f

PR #1076, opened 2025-09-15
feat: balance for units cdk-cli

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


vnprc:
I think it would be better to query the local db for all proofs with all currency units and print out a summary. If you feel strongly that the cli should make an API call to the mint, that should probably be an optional param because it will always fail when the cli can't connect to the mint.

thesimplekid:
I don't think a connection to the mint is required for the balance. When we start the cli we try and refresh keysets but it should just be logging an error if it can't; the balance check should work? https://github.com/cashubtc/cdk/blob/4f65441c0de3e3c05bbe5dfb5cb5a957a26aa898/crates/cdk-cli/src/main.rs#L211-L220.

For example here, I see the logging output of the mints I can't connect to but can still check the balances:
image

PR #1088, opened 2025-09-17
fix(build): use OUT_DIR for migration files

vnprc:
- write generated migration files to OUT_DIR instead of source directory
- copy migration SQL files to OUT_DIR for inclusion in build artifacts
- use absolute paths from OUT_DIR in include_str! macros
- update consumer modules to include from OUT_DIR using concat! macro

these changes enable cdk to support nix sandbox builds

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sql-common/build.rs
+++ b/crates/cdk-sql-common/build.rs
@@ -1,4 +1,5 @@
 use std::cmp::Ordering;
+use std::env;
 use std::fs::{self, File};
 use std::io::Write;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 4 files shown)


thesimplekid:
@crodas can you review?

thesimplekid:
> @vnprc @thesimplekid We can safely delete src/mint/auth/migrations.rs, src/mint/migrations.rs and src/wallet/migrations.rs

Do you mean remove from source control?

thesimplekid (approved):

PR #1109, opened 2025-09-23
fix: handle fiat melt amount conversions

gudnuf:
### Description

Allows the payment processor to do currency conversion when getting the melt quote. Before, if a melt request was in a fiat unit, the `Amount::to_unit` would fail. This change moves `check_melt_request_acceptable` to after the the payment processor can tell us the amount in the requested unit.

I also added currency conversion to fake wallet using mempool's price api (https://mempool.space/api/v1/prices)

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/melt.rs
+++ b/crates/cdk/src/mint/melt.rs
@@ -142,19 +142,6 @@ impl Mint {
             ..
         } = melt_request;
 
-        let amount_msats = melt_request.amount_msat()?;
-
-        let amount_quote_unit = to_unit(amount_msats, &CurrencyUnit::Msat, unit)?;
-
-        self.check_melt_request_acceptable(
-            amount_quote_unit,
-            unit.clone(),
-            PaymentMethod::Bolt11,
-            request.to_string(),
-            *options,
-        )
-        .await?;
-
         let ln = self
             .payment_processors
             .get(&PaymentProcessorKey::new(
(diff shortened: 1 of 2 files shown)


crates/cdk/src/mint/melt.rs
```diff
@@ -297,6 +285,16 @@ impl Mint {
...
+        // Validate using processor quote amount for currency conversion
+        self.check_melt_request_acceptable(
+            payment_quote.amount,
+            unit.clone(),
+            PaymentMethod::Bolt12,
+            request.clone(),
+            *options,
+        )
+        .await?;
+
```
thesimplekid, line 305: I think we need to move this back where it was before getting the payment quote. The unit of the payment quote might not be that of the melt quote. So here were using the payment quote amount but the unit from the melt quote.

gudnuf: It needs to be after the payment processor gets the quote so that we know the unit amount of the invoice.

With regards to the unit, using `payment_quote.unit` vs `unit` should always be the same because the payment processor is obtained based on the request unit. Using the melt request unit here made more sense to me because `check_melt_request_acceptable` uses the unit to check the melt settings; however using the payment quote unit might be better because then if a processor returns a unit other than the specified processor unit that would fail (although seems like an implementation error, not runtime)

gudnuf:
> I think in the long run we should just remove it from the response and amount in the response should be the quote unit

Yea I think that makes sense. Returning the unit that the processor converted the invoice amount to is a nice sanity check, but it should be expected that the processor always converts to it's own unit.

Either way, the check you added looks good

Side note: it could be nice to make the `Amount` struct include the unit that the amount is in

thesimplekid:
> Side note: it could be nice to make the Amount struct include the unit that the amount is in

Yes it would. @davidcaseria and I have both tried this at separate times and its truly a mess of a change.

PR #1117, opened 2025-09-24
feat: added postgres to ffi

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-ffi/src/database.rs
+++ b/crates/cdk-ffi/src/database.rs
@@ -4,10 +4,30 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use cdk::cdk_database::WalletDatabase as CdkWalletDatabase;
+#[cfg(feature = "postgres")]
+use cdk_postgres::WalletPgDatabase as CdkWalletPgDatabase;
 use cdk_sqlite::wallet::WalletSqliteDatabase as CdkWalletSqliteDatabase;
 
 use crate::error::FfiError;
 use crate::types::*;
+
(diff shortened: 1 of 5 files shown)


crates/cdk-ffi/src/database.rs
thesimplekid, line 1: Can we split this up into multiple files just for organizaion? At least a psgl and sqlite make sense.

justfile
```diff
@@ -429,7 +429,7 @@ _ffi-lib-ext:
 
 # Build the FFI library
 ffi-build *ARGS="--release":
-  cargo build {{ARGS}} --package cdk-ffi
+  cargo build {{ARGS}} --package cdk-ffi --features postgres
```
thesimplekid, line 432: Technically we don't need the feature flag since its a default feature but if we want to leave it to be explicit that's fine just wanted to mention it.

crates/cdk-ffi/src/database.rs
asmogo: yes. 
I think there is also still some redundancy in the WalletDatabaseBridge. 
Going to continue working on that.

asmogo: fixed in https://github.com/cashubtc/cdk/pull/1117/commits/28c77fbb5b93deffe13fd2b9afbf6ad09917156f

thesimplekid (approved):

davidcaseria:
There are some `println`s left in the code, but otherwise LGTM

PR #1118, opened 2025-09-24
feat: uniffi bindings for golang

asmogo:
### Description

Still in draft because of these prerequisites
- [ ] https://github.com/NordSecurity/uniffi-bindgen-go/pull/82
- [ ] https://github.com/NordSecurity/uniffi-bindgen-go/pull/77 (to prevent uniffi downgrade)

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/justfile
+++ b/justfile
@@ -477,7 +477,7 @@ ffi-generate LANGUAGE *ARGS="--release": ffi-build
   if [[ "$LANG" == "go" ]]; then
     if ! command -v uniffi-bindgen-go >/dev/null 2>&1; then
       echo "⬇️  Installing uniffi-bindgen-go..."
-      cargo install uniffi-bindgen-go --git https://github.com/NordSecurity/uniffi-bindgen-go --tag v0.4.0+v0.28.3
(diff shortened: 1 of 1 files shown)


asmogo:
I think we should move forward and merge this one as well. The only downside is that we are forced to downgrade the uniffi crate to 0.28.3

thesimplekid:
> I think we should move forward and merge this one as well. The only downside is that we are forced to downgrade the uniffi crate to 0.28.3

I don't mind downgrading now as I don't think we lose anything we're using. I do think there are a few things in recent versions we may want to take advantage of, but nothing we need right now that blocks downgrading. My only concern would be we get stuck there if the go generator never upgrades.

It seems ldk-node is on 0.28.3 as well for this reason https://github.com/lightningdevkit/ldk-node/pull/733#discussion_r2690852332.

Another potential issue is if the dart generator gets to a state where we can use it but requires lets say 0.30.0 but were stuck on 0.28.3 to support go, we're in a position of having to choose to drop the go support in order to upgrade. But since that is not ready yet I believe, hopefully go upgrades by then. 

cc @davidcaseria do you think downgrading is okay?

PR #1121, opened 2025-09-25
MultiMintWallet add mints even if they are offline

thesimplekid:
Remove unnecessary mint info fetching and keyset refresh during mint addition, and improve error propagation in token restoration flow

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/multi_mint_wallet.rs
+++ b/crates/cdk/src/wallet/multi_mint_wallet.rs
@@ -205,9 +205,6 @@ impl MultiMintWallet {
             )?
         };
 
-        wallet.fetch_mint_info().await?;
-        wallet.refresh_keysets().await?;
-
         let mut wallets = self.wallets.write().await;
         wallets.insert(mint_url, wallet);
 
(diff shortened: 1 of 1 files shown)


crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -205,9 +205,6 @@ impl MultiMintWallet {
             )?
         };
 
-        wallet.fetch_mint_info().await?;
-        wallet.refresh_keysets().await?;
```
davidcaseria, line 209: Does the MultiMintWallet need to expose functions to maintain this functionality?

thesimplekid: Maybe. Refresh keysets actually also gets the mint info. So even with this removed we do get the lasted mint info and keysets with any online operations. Though if you mean expose fns so a user can manually get lasted mint info and keysets I don't see a down side.

davidcaseria: If it's already handled within the Wallet, then we can merge as-is and add it later if needed.

davidcaseria (approved):

crates/cdk/src/wallet/multi_mint_wallet.rs
thesimplekid: Yeah I've been thinking if there is a better way to handle this maybe adding a cache of some kind to the mint connector. Currently for every online request the wallet will get the latest mint info and keyset info. I think this is probably too aggressive and adds unnecessary requests and latency.

PR #1124, opened 2025-09-25
token: add spending-condition inspection helpers and token_secrets()

lollerfirst:
### Description

This PR adds a lightweight API to inspect `SpendingConditions` directly from a `Token` and without requiring Mint keysets:

- New helpers on Token:
  - `spending_conditions`
  - `p2pk_pubkeys`
  - `p2pk_refund_pubkeys`
  - `htlc_hashes`
  - `locktimes`
- New token_secrets() unifies V3/V4 proof traversal and removes code duplication.
- Bypasses short→long keyset-id mapping since only Secret is needed for condition parsing.
- Improves DX: external users can extract P2PK and HTLC metadata without mint context.

### DIFF
--- a/crates/cashu/src/nuts/nut00/token.rs
+++ b/crates/cashu/src/nuts/nut00/token.rs
@@ -2,23 +2,21 @@
 //!
 //! <https://github.com/cashubtc/nuts/blob/main/00.md>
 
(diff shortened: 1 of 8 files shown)


lollerfirst:
@thesimplekid addressed

PR #1127, opened 2025-09-25
rename ln settings in toml configuration

asmogo:
### Description
- Replace Ln/LnBackend with Backend/BackendType across mintd
- Move settings.ln.ln_backend to settings.backend.name
- Update example config: [ln] -> [backend], ln_backend -> name
- Adjust env var mapping and internal references accordingly
- Add serde aliases for fake_wallet and grpc_processor
- Update integration tests to the new settings schema

BREAKING CHANGE: rename config table and key:
- [ln] -> [payment_backend]
- ln_backend -> kind

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-mintd/src/env_vars/mod.rs
+++ b/crates/cdk-mintd/src/env_vars/mod.rs
@@ -93,7 +93,7 @@ impl Settings {
         self.info = self.info.clone().from_env();
         self.mint_info = self.mint_info.clone().from_env();
         self.payment_backend = self.payment_backend.clone().from_env();
-        // The following section is providing backwards compatability #1127
+        // The following section is providing backwards compatibility #1127
         // todo -- remove this section
         if self.payment_backend.kind == PaymentBackendKind::None
             && self.ln.ln_backend != PaymentBackendKind::None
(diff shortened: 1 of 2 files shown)


crates/cdk-mintd/example.config.toml
```diff
@@ -84,9 +84,9 @@ max_connections = 20
 # Connection timeout in seconds (optional, defaults to 10)
 connection_timeout_seconds = 10
 
-[ln]
+[backend]
```
thesimplekid: Should it be payment_backend. Just backend could mean database or something?

crates/cdk-mintd/example.config.toml
```diff
@@ -84,9 +84,9 @@ max_connections = 20
...
 # Required ln backend `cln`, `lnd`, `fakewallet`, 'lnbits', 'ldknode'
-ln_backend = "fakewallet"
+name = "fakewallet"
```
thesimplekid:
```suggestion
kind = "fakewallet"
```

Is kind better?

crates/cdk-mintd/example.config.toml
asmogo: yes. I like that.

asmogo:
I have updated this to be backwards compatible with the old config https://github.com/cashubtc/cdk/pull/1127/commits/60037e7f2e5e53349f49505df871fef5d237b5cd

We print deprecation warnings to std out if we see the old [ln] configuration is still active.

PR #1153, opened 2025-10-03
Add cdk-mintd module and package

jurraca:
### Description

Add a NixOS module for the `cdk-mintd` binary. This enables pulling in a mint as a systemd service. In order to do this, we also implement a `cdk-mintd` package which is added to the flake.

-----

### Notes to the reviewers

Looking for conceptual feedback, and how hard you think maintaining this would be--it's a very large module.

In order to get a package working, we need a `Cargo.lock` file. I added one in the root of the project, but maybe this should be in the crate directory itself?
Was trying to get a test working but might not be worth it.
Haven't tested all the options, but the module seems to build without Nix errors.

-----

#### ADDED
Nix module and package.
Cargo.lock file

#### REMOVED
Cargo.lock in .gitignore

#### FIXED

----

### Checklist


crates/cdk-mintd/package.nix
```diff
@@ -0,0 +1,44 @@
...
+, pkg-config
+, openssl
+, sqlite
+, postgresql
+, protobuf
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "cdk-mintd";
+  version = "0.12.0";
```
notjustanyusername, line 12: Version should be read from Cargo.toml.

crates/cdk-mintd/test.nix
```diff
@@ -0,0 +1,102 @@
...
+            pass
+
+    server = HTTPServer(("127.0.0.1", 8080), Handler)
+    print("Mock cdk-mintd listening on 127.0.0.1:8080")
+    server.serve_forever()
+    '
+  '';
+  module = self.nixosModules.default;
+in
+pkgs.nixosTest {
```
notjustanyusername, line 37: I tried running the test but got an error "Ln backend must be set".

I think if there is going to be a module for cdk-mint in this project then there has to be a test as well. Yes its quite a bit of work to get it set up at first but it helps so much with maintenance/reliability/documentation.

There's no lightning module in nixpkgs so maybe you can just use ldk embedded or pull in [nix-bitcoin](https://github.com/fort-nix/nix-bitcoin/tree/master).

crates/cdk-mintd/module.nix
```diff
@@ -0,0 +1,841 @@
...
+        ExecStart = "${cfg.package}/bin/cdk-mintd --config ${cfg.configFile}";
+
+        # Capability restrictions
+        CapabilityBoundingSet = "";
+        AmbientCapabilities = "";
+      };
+    };
+
+    # Open firewall port if listening on all interfaces
+    networking.firewall.allowedTCPPorts = mkIf (cfg.settings.info.listen_host == "0.0.0.0") [ cfg.settings.info.listen_port ];
```
notjustanyusername, line 796: Usually there is an option to automatically open firewall or not. Some people may be using nftables for example.

crates/cdk-mintd/package.nix
```diff
@@ -0,0 +1,44 @@
+{ lib
+, rustPlatform
+, pkg-config
+, openssl
+, sqlite
+, postgresql
+, protobuf
+}:
+
+rustPlatform.buildRustPackage rec {
```
notjustanyusername, line 10: I notice crane is in the flake inputs (but not actually used). Did you consider using it to build the package? If developers on the project want to make use of more nix features then it might be better to use crane. It allows for more flexibility and artifact reuse when building the whole workspace.

PR #1181, opened 2025-10-10
Deterministic Currency Unit Derivation Paths

vnprc:
### Description

Hard coding currency units and their corresponding derivation paths for deterministic secrets is an unsustainable strategy because it forces us to update all cashu implementations for each new currency unit. We've gotten away with it this far because cashu only supports a limited set of currencies, with sats being the most popular by far.

Ehash will have very different requirements because each mining epoch (time in between mining rewards) will use a different currency unit. So we need a way to dynamically map currency units to derivation paths.

This PR introduces an algorithm to deterministically generate a derivation path based on the currency unit name.

First, some code cleanup. This PR moves the currency unit code from the nut00 crate to nut01 where it belongs. (The spec was updated after the initial cdk CurrencyUnit implementation.)

Now to business: we force all currency unit names to uppercase and calculate the derivation path in this way:
- normalize unicode characters to NFC format
- lowercase (probably change to uppercase)
- trim
- SHA256 hash
- modulo 2^31 - 1

Opening this draft PR to get initial feedback on the implementation. I will also open a NUT PR to propose this new specification.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut01/mod.rs
+++ b/crates/cashu/src/nuts/nut01/mod.rs
@@ -252,15 +252,15 @@ impl CurrencyUnit {
 
     /// Generate deterministic derivation index for custom currency units
     fn custom_derivation_index(s: &str, reserved: u32) -> Option<u32> {
-        // 1) NFC normalization: composes equivalent Unicode sequences (e.g., "e" + U+0301) into a single
-        //    canonical code point (e.g., "é") so visually/equivalently identical strings hash the same
-        // 2) lowercase: avoids case-induced divergence ("USD" vs "usd")
-        // 3) trim: removes accidental leading/trailing whitespace (" usd " vs "usd")
-        let norm = s.nfc().collect::<String>().to_lowercase();
-        let norm = norm.trim();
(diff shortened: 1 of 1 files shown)


crates/cashu/src/nuts/nut01/mod.rs
```diff
@@ -192,6 +196,147 @@ impl MintKeyPair {
...
+        }
+    }
+
+    /// Generate deterministic derivation index for custom currency units
+    fn custom_derivation_index(s: &str, reserved: u32) -> Option<u32> {
+        // 1) NFC normalization: composes equivalent Unicode sequences (e.g., "e" + U+0301) into a single
+        //    canonical code point (e.g., "é") so visually/equivalently identical strings hash the same
+        // 2) lowercase: avoids case-induced divergence ("USD" vs "usd")
+        // 3) trim: removes accidental leading/trailing whitespace (" usd " vs "usd")
+        let norm = s.nfc().collect::<String>().to_lowercase();
```
vnprc: change to uppercase

PR #1200, opened 2025-10-21
feat: optimize pending mint quotes query performance

vnprc:
- add get_pending_mint_quotes trait method to wallet database interface
- implement optimized SQL query filtering by payment method and balance
- create composite index on (payment_method, amount_paid, amount_issued)
- add created_time column to mint_quote table for ordering
- implement get_pending_mint_quotes for SQLite, ReDB, and FFI backends
- add wallet method to retrieve pending quotes filtered by mint URL and expiry

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-ffi/.cargo/config.toml
+++ b/crates/cdk-ffi/.cargo/config.toml
@@ -1,0 +1,7 @@
+[target.'cfg(target_os = "android")']
+rustflags = [
+    "-C", "link-arg=-z",
+    "-C", "link-arg=max-page-size=16384",
+    "-C", "link-arg=-z",
+    "-C", "link-arg=common-page-size=16384",
+]


crates/cdk/src/wallet/issue/issue_bolt11.rs
```diff
@@ -157,6 +155,17 @@ impl Wallet {
...
     }
 
+    /// Get pending mint quotes
+    /// Returns mint quotes that have mintable balance or are bolt12 quotes (reusable).
+    /// Filters out expired quotes and quotes from other mints.
+    #[instrument(skip(self))]
+    pub async fn get_pending_mint_quotes(&self) -> Result<Vec<MintQuote>, Error> {
+        let mut pending_quotes = self.localstore.get_pending_mint_quotes().await?;
+        let unix_time = unix_time();
+        pending_quotes.retain(|quote| quote.mint_url == self.mint_url && quote.expiry > unix_time);
```
thesimplekid: We shouldn't filter by expiry time here even if it is after expiry time we should still check with the mint if we can mint.

crates/cdk-redb/src/wallet/mod.rs
```diff
@@ -511,6 +511,25 @@ impl WalletDatabase for WalletRedbDatabase {
...
+
+        Ok(table
+            .iter()
+            .map_err(Error::from)?
+            .flatten()
+            .flat_map(|(_id, quote)| serde_json::from_str::<MintQuote>(quote.value()).ok())
+            .filter(|quote| {
+                // Return quotes with pending balance OR bolt12 quotes
+                (quote.amount_paid > quote.amount_issued && u64::from(quote.amount_paid) > 0)
+                    || quote.payment_method.to_string() == "bolt12"
```
thesimplekid: Why do we compare the strings here can't we use the enum?

crates/cdk-common/src/database/wallet.rs
```diff
@@ -62,6 +62,8 @@ pub trait Database: Debug {
     async fn get_mint_quote(&self, quote_id: &str) -> Result<Option<WalletMintQuote>, Self::Err>;
     /// Get mint quotes from storage
     async fn get_mint_quotes(&self) -> Result<Vec<WalletMintQuote>, Self::Err>;
+    /// Get pending mint quotes from storage (quotes with mintable balance or bolt12 quotes)
+    async fn get_pending_mint_quotes(&self) -> Result<Vec<WalletMintQuote>, Self::Err>;
```
thesimplekid: Can we rename to unpaid over pending.

thesimplekid (dismissed):

thesimplekid:
Not really sure what happened to the merges here so https://github.com/cashubtc/cdk/pull/1303

PR #1204, opened 2025-10-24
Add database transaction trait for cdk wallet

### LINKED ISSUE #1203: Update the Database trait for wallet and support for transactions

crodas:
### Description

Fixes #1203

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-ffi/src/database.rs
+++ b/crates/cdk-ffi/src/database.rs
@@ -1,9 +1,17 @@
 //! FFI Database bindings
 
 use std::collections::HashMap;
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 
-use cdk::cdk_database::WalletDatabase as CdkWalletDatabase;
+use cdk_common::database::{
+    DbTransactionFinalizer, DynWalletDatabaseTransaction, WalletDatabase as CdkWalletDatabase,
+    WalletDatabaseTransaction as CdkWalletDatabaseTransaction,
+};
+use cdk_common::task::spawn;
+use cdk_sql_common::pool::DatabasePool;
+use cdk_sql_common::SQLWalletDatabase;
+use tokio::sync::Mutex;
 
 use crate::error::FfiError;
 #[cfg(feature = "postgres")]
@@ -11,42 +19,20 @@ use crate::postgres::WalletPostgresDatabase;
 use crate::sqlite::WalletSqliteDatabase;
 use crate::types::*;
 
-/// FFI-compatible trait for wallet database operations
-/// This trait mirrors the CDK WalletDatabase trait but uses FFI-compatible types
-#[uniffi::export(with_foreign)]
+/// FFI-compatible wallet database trait (read-only operations + begin_db_transaction)
+/// This trait mirrors the CDK WalletDatabase trait structure
+#[uniffi::export]
 #[async_trait::async_trait]
 pub trait WalletDatabase: Send + Sync {
-    // Mint Management
-    /// Add Mint to storage
-    async fn add_mint(
-        &self,
-        mint_url: MintUrl,
-        mint_info: Option<MintInfo>,
-    ) -> Result<(), FfiError>;
-
-    /// Remove Mint from storage
-    async fn remove_mint(&self, mint_url: MintUrl) -> Result<(), FfiError>;
+    /// Begin a database transaction
+    async fn begin_db_transaction(&self)
+        -> Result<Arc<WalletDatabaseTransactionWrapper>, FfiError>;
 
     /// Get mint from storage
     async fn get_mint(&self, mint_url: MintUrl) -> Result<Option<MintInfo>, FfiError>;
 
     /// Get all mints from storage
     async fn get_mints(&self) -> Result<HashMap<MintUrl, Option<MintInfo>>, FfiError>;
-
-    /// Update mint url
-    async fn update_mint_url(
-        &self,
-        old_mint_url: MintUrl,
-        new_mint_url: MintUrl,
-    ) -> Result<(), FfiError>;
-
-    // Keyset Management
-    /// Add mint keyset to storage
-    async fn add_mint_keysets(
-        &self,
-        mint_url: MintUrl,
-        keysets: Vec<KeySetInfo>,
-    ) -> Result<(), FfiError>;
 
     /// Get mint keysets for mint url
     async fn get_mint_keysets(
@@ -57,22 +43,11 @@ pub trait WalletDatabase: Send + Sync {
     /// Get mint keyset by id
     async fn get_keyset_by_id(&self, keyset_id: Id) -> Result<Option<KeySetInfo>, FfiError>;
 
-    // Mint Quote Management
-    /// Add mint quote to storage
-    async fn add_mint_quote(&self, quote: MintQuote) -> Result<(), FfiError>;
-
     /// Get mint quote from storage
     async fn get_mint_quote(&self, quote_id: String) -> Result<Option<MintQuote>, FfiError>;
 
     /// Get mint quotes from storage
     async fn get_mint_quotes(&self) -> Result<Vec<MintQuote>, FfiError>;
-
-    /// Remove mint quote from storage
-    async fn remove_mint_quote(&self, quote_id: String) -> Result<(), FfiError>;
-
-    // Melt Quote Management
-    /// Add melt quote to storage
-    async fn add_melt_quote(&self, quote: MeltQuote) -> Result<(), FfiError>;
 
     /// Get melt quote from storage
     async fn get_melt_quote(&self, quote_id: String) -> Result<Option<MeltQuote>, FfiError>;
@@ -80,26 +55,8 @@ pub trait WalletDatabase: Send + Sync {
     /// Get melt quotes from storage
     async fn get_melt_quotes(&self) -> Result<Vec<MeltQuote>, FfiError>;
 
-    /// Remove melt quote from storage
-    async fn remove_melt_quote(&self, quote_id: String) -> Result<(), FfiError>;
-
-    // Keys Management
-    /// Add Keys to storage
-    async fn add_keys(&self, keyset: KeySet) -> Result<(), FfiError>;
-
     /// Get Keys from storage
     async fn get_keys(&self, id: Id) -> Result<Option<Keys>, FfiError>;
-
-    /// Remove Keys from storage
-    async fn remove_keys(&self, id: Id) -> Result<(), FfiError>;
-
-    // Proof Management
-    /// Update the proofs in storage by adding new proofs or removing proofs by their Y value
-    async fn update_proofs(
-        &self,
-        added: Vec<ProofInfo>,
-        removed_ys: Vec<PublicKey>,
-    ) -> Result<(), FfiError>;
 
     /// Get proofs from storage
     async fn get_proofs(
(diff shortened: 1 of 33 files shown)


crodas:
> What part of the wallet needs database transaction support? Thereby, what is the most significant risk for FFI wallets, given that they don't support database transactions?


I will open a follow up PR, basically all changes should be done in a transaction. The ffi database trait and the cdk-common database trait should match.

crates/cdk/src/wallet/receive.rs
```diff
@@ -22,17 +23,18 @@ impl Wallet {
     #[instrument(skip_all)]
     pub async fn receive_proofs(
         &self,
+        tx: &mut Tx<'_, '_>,
```
davidcaseria: Exposing the database transaction in the public API does not seem right. Can we have the Wallet keep a mutex for the transaction in its state?

crodas: I agree, I will move all the functions that receive a tx private, and the public function will not receive anything from the outside. 

```
    pub async fn receive_proofs(...) -> will call receive_proofs_with_tx
```

A mutex will not work because the point of a transaction begin/end in a call-flow.  I was thinking to wrap the wallet of a WalletInTx privage struct that would interact with a tx instead of a localstorage.  But this would be private.

What do you suggest @thesimplekid ?

crodas: I meant this

```
diff --git a/crates/cdk/src/wallet/proofs.rs b/crates/cdk/src/wallet/proofs.rs
index 94252d01..298442b1 100644
--- a/crates/cdk/src/wallet/proofs.rs
+++ b/crates/cdk/src/wallet/proofs.rs
@@ -92,8 +92,18 @@ impl Wallet {
     /// Reclaim unspent proofs
     ///
     /// Checks the stats of [`Proofs`] swapping for a new [`Proof`] if unspent
+    pub async fn reclaim_unspent(&self, proofs: Proofs) -> Result<(), Error> {
+        let mut tx = self.localstore.begin_db_transaction().await?;
+        self.reclaim_unspent_with_tx(&mut tx, proofs).await?;
+        tx.commit().await?;
+        Ok(())
+    }
+
+    /// Reclaim unspent proofs with transaction
+    ///
+    /// Checks the stats of [`Proofs`] swapping for a new [`Proof`] if unspent
     #[instrument(skip(self, tx, proofs))]
-    pub async fn reclaim_unspent(&self, tx: &mut Tx<'_, '_>, proofs: Proofs) -> Result<(), Error> {
+    pub async fn reclaim_unspent_with_tx(&self, tx: &mut Tx<'_, '_>, proofs: Proofs) -> Result<(), Error> {
         let proof_ys = proofs.ys()?;

         let transaction_id = TransactionId::new(proof_ys.clone());
(END)
```

thesimplekid: What is the motivation for having the receive proofs with tx I looked through where its being used and didn't see why it is needed?

crodas: To avoid deadlocks,  you can't access localstore at the same time with an open transaction with SQLite, which is excellent for development, because you should have a consistent point of view of the database (localstore vs Tx) in other databases.

thesimplekid: Yes, I understand that but I don't see anywhere where we are receive proofs where we want to use an existing tx.

thesimplekid: I think its fine lets keep it

crates/cdk/src/wallet/swap.rs
thesimplekid, line 1: We're holding db txs's across network calls here but also in other places can we avoid this?

crates/cdk/src/wallet/proofs.rs
```diff
@@ -61,20 +77,55 @@ impl Wallet {
             .collect())
     }
 
+    /// Get this wallet's [Proofs] that match the args with transaction
+    pub(super) async fn get_proofs_with_tx(
```
thesimplekid: We have two apis for the same thing. Some times we have two fns one _with_tx. But then in other fns we just have one fn that accepts an Option<tx> we should be consistent with how we handle this.

crodas: We will drop all options, any function that accepts a transaction is private / pub super

crodas:
@davidcaseria @thesimplekid My main doubt this afternoon was about the `cdk_ffi::database::Database` trait and whether we would like it to be 100% compatible with the proposed changes to the `cdk_common::database::Wallet*`

I later discovered that `cdk_ffi::database::Database`  provides raw database access to the target language. For that reason, I think the approach proposed here is enough. Any other attempt to share the transaction behaviour would be overkill. 

Ideally, in the future, we will drop raw access and provide any needed APIs in the wallet for accessing its data.

What do you guys think? If you agree with what I wrote earlier, this PR will be final and actually require no follow-up.

crates/cdk-sql-common/src/wallet/mod.rs
thesimplekid, line 1: I noticed now that we have some fns that are both in txs and outside we duplicate sql in for some fns. I think we should refactor so we have the sql in a common fn that can be used in both, this would make it less likely we update one and forget to update the other. The ones I noticed are:

  - get_keyset_by_id() (lines 548 & 952)
  - get_keys() (lines 571 & 1061)
  - get_mint_quote() (lines 591 & 975)
  - get_melt_quote() (lines 621 & 1033)
  - get_proofs() (lines 649 & 1082)

I think this might be true in some places on the mint as well. I am okay if we do this as a follow up PR thought and not hold this up.

crates/cdk/src/wallet/swap.rs
thesimplekid: This can make a network call but we have a db tx open. Could we refactor this so we pass in the active keyset id so we can make the network call before the tx is opened?

crates/cdk/src/wallet/swap.rs
```diff
@@ -210,13 +215,13 @@ impl Wallet {
 
         // Desired amount is either amount passed or value of all proof
         let proofs_total = proofs.total_amount()?;
-
-        let ys: Vec<PublicKey> = proofs.ys()?;
-        self.localstore
-            .update_proofs_state(ys, State::Reserved)
+        let fee = self.get_proofs_fee(&proofs).await?;
+        let fee_and_amounts = self
+            .get_keyset_fees_and_amounts_by_id(active_keyset_id)
```
thesimplekid: This can also make a network call so should be done out of the tx.

crates/cdk/src/wallet/receive.rs
```diff
@@ -141,8 +141,8 @@ impl Wallet {
             &keys,
         )?;
 
-        self.localstore
-            .increment_keyset_counter(&active_keyset_id, recv_proofs.len() as u32)
+        let mut tx = self.localstore.begin_db_transaction().await?;
+        tx.increment_keyset_counter(&active_keyset_id, recv_proofs.len() as u32)
             .await?;
```
thesimplekid, line 162: I think we don't need this increment there is an increment inside the create_swap?

crates/cdk-sql-common/src/wallet/mod.rs
crodas: That's true. We should have the SQL in a single place so they never get out of sync.

crodas:
@thesimplekid I now saw the missing bit, it was `WalletDatabaseBridge`. I'll work on it.

crates/cdk/src/wallet/receive.rs
crodas: This code is older than this PR, so I don't know but I could check later

crates/cdk/src/wallet/swap.rs
crodas: Fixed in bdfdb053dd9093aaa8f312672039c635e881607f

crates/cdk/src/wallet/swap.rs
crodas: Fixed in bdfdb053dd9093aaa8f312672039c635e881607f

crates/cdk-sql-common/src/wallet/mod.rs
crodas: Fixed in bdfdb053dd9093aaa8f312672039c635e881607f

crates/cdk-ffi/src/database.rs
```diff
@@ -571,16 +439,833 @@ impl CdkWalletDatabase for WalletDatabaseBridge {
...
+
+pub(crate) struct FfiWalletSQLDatabase<RM>
+where
+    RM: DatabasePool + 'static,
+{
+    inner: SQLWalletDatabase<RM>,
+    // SAFETY: The transaction borrows from `inner`, but we guarantee through
+    // our API that `inner` outlives the transaction. The transaction is always
+    // explicitly committed or rolled back, and we never drop `inner` while a
+    // transaction is active. Automatic rollback on drop ensures no dangling state.
```
thesimplekid: Think this is left over from the unsafe code?

crates/cdk-ffi/src/database.rs
```diff
@@ -571,16 +439,833 @@ impl CdkWalletDatabase for WalletDatabaseBridge {
...
+where
+    RM: DatabasePool + 'static,
+{
+    async fn begin(&self) -> Result<(), FfiError> {
+        let mut inner_tx = self.tx.lock().await;
+        if inner_tx.is_some() {
+            return Err(FfiError::Database {
+                msg: "Nested transactions not supported".to_owned(),
+            });
+        }
```
thesimplekid: I notice that this will fail fast if there is already a tx in progress. This is different from the api for non-ffi where we will wait for a connection to become available from the pool. This is maybe okay in the ffi context or we could add waiting logic here. If we think the fail-fast is best for ffi we should document it.

crodas: We can push towards the clean separation of the Database layer and Transaction, or keep it all together as we have now. I think a clear separation would be best, as we have in Rust. 
Basically, it will fail in cases like this, where an ongoing transaction is expected. 
This model although simpler is limited to one ongoing transaction per model. I'll push the improvements now that I got rid of the lifetimes associated to a tx.

crates/cdk-ffi/src/multi_mint_wallet.rs
```diff
@@ -23,20 +23,20 @@ pub struct MultiMintWallet {
...
 #[uniffi::export(async_runtime = "tokio")]
 impl MultiMintWallet {
-    /// Create a new MultiMintWallet from mnemonic using WalletDatabase trait
+    /// Create a new MultiMintWallet from mnemonic using WalletDatabaseFfi trait
     #[uniffi::constructor]
     pub fn new(
         unit: CurrencyUnit,
         mnemonic: String,
-        db: Arc<dyn crate::database::WalletDatabase>,
+        db: crate::database::WalletDatabaseType,
```
thesimplekid: Doesn't making this a closed enum and not a trait make it impossible for someone to implement the trait in another lang and use it?

crodas: Fixed in fab1983aab8d90278fb6597cb2352fb40972c7bb

crates/cdk-ffi/src/database.rs
```diff
@@ -492,121 +449,956 @@ impl CdkWalletDatabase for WalletDatabaseBridge {
...
+}
+
+/// Transaction wrapper for FFI
+pub(crate) struct FfiWalletTransaction {
+    tx: Arc<Mutex<Option<DynWalletDatabaseTransaction>>>,
+}
+
+impl Drop for FfiWalletTransaction {
+    fn drop(&mut self) {
+        let tx = self.tx.clone();
```
thesimplekid:
```suggestion
    fn drop(&mut self) {
        if let Ok(guard) = self.tx.try_lock() {
            if guard.is_none() {
                return;
            }
        }
        let tx = self.tx.clone();
```

We can probally avoid the spawn in most cases.

crates/cdk-ffi/src/database.rs
```diff
@@ -492,121 +449,956 @@ impl CdkWalletDatabase for WalletDatabaseBridge {
...
+        let cdk_id = transaction_id.try_into()?;
+        tx.remove_transaction(cdk_id)
+            .await
+            .map_err(|e| FfiError::Database { msg: e.to_string() })
+    }
+}
+
+/// FFI-safe database type enum
+#[derive(uniffi::Enum, Clone)]
+pub enum WalletDatabaseType {
```
thesimplekid: I don't think this is used anymore?

crates/cdk-ffi/src/database.rs
crodas: > I think this looks good. But will have to test how it actually works in bindings, specifically with the async spawn in the drop. Have you looked into how this would be handled?

Drop is something that happens in Rust, I guess we should have a runtime when that happens. But I'll test shortly.

crates/cdk-ffi/src/database.rs
crodas: I think it is exposed to the wrapped language so they can easily instantiate the native database connectors, right @davidcaseria ?

crodas: Fixed in b9f0e63e

crates/cdk-ffi/src/database.rs
crodas: You were right, I'm debugging now

```
thread '<unnamed>' (91996174) panicked at /Users/cesar/projects/rust/cdk/crates/cdk-common/src/task.rs:14:5:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread '<unnamed>' (91996174) panicked at /Users/cesar/projects/rust/cdk/crates/cdk-sql-common/src/database.rs:106:13:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

crodas: It is working as expected; we have Python tests that confirm this behaviour.

thesimplekid (approved):

PR #1206, opened 2025-10-24
fix: sig_all_msg_to_sign

asmogo:
### Description
https://github.com/cashubtc/nuts/pull/302

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


Cargo.toml
thesimplekid, line 1: taking this commit should remove the need for this 

https://github.com/cashubtc/cdk/pull/1207

PR #1212, opened 2025-10-26
Various mint bugfixes for swap and melt. SIG_INPUTS+SIG_ALL, locktimes, P2PK+HTLC. Also updates the SIG_ALL message for amount-switching

SatsAndSports:
### Description

_This includes updates for the protocol (https://github.com/cashubtc/nuts/pull/302) around how the message for SIG_ALL is constructed (with amount and secret, not keyset_id). However, even before that protocol update was proposed, there were a number of bugs in the mint, and the protocol update depends on those fixes. This PR includes these bugfixes too_

- The mint bugs are listed below. For example, all SIG_ALL swaps were rejected in swap. Also, locktime wasn't handled correctly in either P2PK or HTLC. And melt was accepting many invalid melts.
- This PR introduces a new trait called `SpendingConditionsVerification`, which is shared by swap and melt. This centralizes all checking of NUT-10 spending conditions (P2PK, HTLC, SIG_ALL/SIG_INPUTS, locktimes, refunds) into one place. Code re-use is a priority in this design, in order to fix the bugs.
- This PR reorganizes `process_swap_request` and `Mint.melt` to use this trait with a single simple call to `.verify_spending_conditions`. 
- This PR includes 35 new tests for all the combinations in swap and melt. Also, this PR also includes the updated test vectors for SIG_ALL that are in `nuts/tests/11-test.md`

Some issues fixed by this:
- SIG_ALL swaps were always rejected. This is because the SIG_INPUTS code (`verify_p2pk` in particular) would reject the SIG_ALL, as it didn't understand SIG_ALL.
- melt was accepting all SIG_INPUTS transactions without verification, it was verifying only the SIG_ALL transactions
- locktime wasn't handled consistently. Some of the code treated everything as pre-locktime, and therefore it didn't the correct post-locktime behaviour

#### Testing

```
# To run the 29 new swap tests:
cargo test -p cdk --lib mint::swap::spending_conditions::
# To run the 6 new melt tests:
cargo test -p cdk --lib mint::melt::spending_conditions::
# To run the updated test vectors:
cargo test -p cashu nuts::nut11::tests::
```

#### Notes

_Some smaller details:_

- A new function `get_pubkeys_and_required_sigs_for_p2pk` is implemented, to ease the identification of the relevant keys and the number of signatures required. It now also returns a boolean to specifiy if a preimage is required; this is to support HTLC. It compares the current time to the locktime and identifies the relevant keys. This PR uses this function for SIG_ALL and SIG_INPUTS cases, to encourage code reuse.
- `verify_p2pk`, which is a method on `Proof`, is rewritten to remove the bug mentioned above and to use the new shared code in order to ensure consistency between SIG_INPUTS and SIG_ALL. `verify_p2pk` also asserts that it's NOT called with SIG_ALL; as SIG_ALL verification shouldn't call `verify_p2pk`
- `can_sign_sig_all` is removed. It's unnecessary to have a wallet-side check of whether a SIG_ALL is valid. If we duplicate the code in the wallet and mint, there is a risk of a discrepancy. _(Option: when the wallet creates a swap_request, maybe the wallet could verify it using `verify_spending_conditions_for_swap` if we want?)_

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/test_helpers/mod.rs
+++ b/crates/cdk/src/test_helpers/mod.rs
@@ -7,4 +7,5 @@
 
 #[cfg(feature = "mint")]
 pub mod mint;
+#[cfg(feature = "mint")]
 pub mod nut10;


crates/cashu/src/nuts/nut11/mod.rs
```diff
@@ -134,87 +190,43 @@ impl Proof {
...
-        //   refund pubkeys (n_sigs_refund-of-refund). This supersedes normal pubkey enforcement
-        //   after expiry.
-        // - If locktime has passed and no refund keys are present, proof becomes spendable
-        //   without further key checks (anyone-can-spend behavior).
-        // - Otherwise (before locktime), enforce normal multisig on the set of authorized
-        //   pubkeys: Secret.data plus optional `pubkeys` tag, requiring n_sigs unique signers.
+        debug_assert!(
+            secret.kind() == Kind::P2PK,
+            "verify_p2pk called with non-P2PK secret - this is a bug"
+        );
```
thesimplekid: I think these should be more than debug asserts and should return errors at runtime.

Debug asserts should be used for checking our developer assumptions and only in places where we can only reach them in private functions, for example. For things that a user could do, like calling sign_p2pk on the wrong proof type, we should return an error.

crates/cdk/src/mint/swap/tests/htlc_sigall_spending_conditions_tests.rs
thesimplekid, line 1: Can we move these test files and helpers into a test folder mod within the swap directory.

crates/cashu/src/nuts/nut00/mod.rs
```diff
@@ -305,13 +305,17 @@ impl Witness {
     /// Add signatures to [`Witness`]
     pub fn add_signatures(&mut self, signatues: Vec<String>) {
         match self {
+            // P2PKWitness.signatures is Vec<String> (not Option), so extend() works directly.
+            // Note: P2PKWitness::default() creates an empty Vec, ensuring signatures is never uninitialized.
```
thesimplekid: I think we can remove these comments. In general, comments are really helpful and we should use more of them to describe *why* we're doing something rather than *how*. Since `extend()` works directly on the Vec here and the initialization flow is pretty clear from the code itself, these comments might be a bit redundant.

That said, leaning towards more comments is a good instinct

crates/cashu/src/nuts/nut11/mod.rs
```diff
@@ -1038,30 +862,12 @@ impl SwapRequest {
     }
 
     /// Validate SIG_ALL conditions and signatures for the swap request
+    ///
+    /// DEPRECATED: Use `verify_spending_conditions()` instead.
+    /// The original implementation didn't take locktime or HTLC into account.
+    #[deprecated(since = "0.4.0", note = "Use verify_spending_conditions() instead. The original implementation didn't take locktime or HTLC into account.")]
```
thesimplekid: I think this version number is wrong. But we don't need to mark as deprecated we can just remove it if its not needed or wrong. We will include it in a minor release since we're pre 1.0 its okay

crates/cashu/src/nuts/nut11/mod.rs
```diff
@@ -1113,6 +919,10 @@ impl<Q: std::fmt::Display + Serialize + DeserializeOwned> MeltRequest<Q> {
     }
 
     /// Verify all inputs have matching secrets and tags
+    ///
+    /// WARNING: This function may be incomplete. According to NUT-11, when SIG_ALL is present,
+    /// all inputs must have: (1) same kind, (2) SIG_ALL flag, (3) same data, (4) same tags.
+    /// This function only checks (3) and (4), but not (1) and (2).
```
thesimplekid: Lets make sure its right before merging.

crates/cashu/src/nuts/nut14/mod.rs
```diff
@@ -72,83 +65,83 @@ impl Proof {
...
+            spending_conditions.sig_flag != super::SigFlag::SigAll,
+            "verify_htlc called with SIG_ALL proof - this is a bug"
+        );
+
+        debug_assert!(
+            secret.kind() == super::Kind::HTLC,
+            "verify_htlc called with non-HTLC secret - this is a bug"
+        );
+
+        // Get the appropriate spending conditions based on locktime
```
thesimplekid, line 114: Should return errors here as well not debug asserts

crates/cashu/src/nuts/nut14/mod.rs
```diff
@@ -72,83 +65,83 @@ impl Proof {
...
+        let now = unix_time();
+        let (preimage_needed, relevant_pubkeys, relevant_num_sigs_required) =
+            super::nut10::get_pubkeys_and_required_sigs(&secret, now)
+                .map_err(Error::NUT11)?;
+
+        // don't extract the witness until it's needed. Remember a post-locktime
+        // zero-refunds proof is acceptable here, and therefore a Witness isn't always
+        // needed
+
+        // If preimage is needed (before locktime), verify it
```
thesimplekid: Can you clarify this comment.

crates/cashu/src/nuts/nut10.rs
```diff
@@ -105,6 +108,393 @@ impl Secret {
     }
 }
 
+/// Get the relevant public keys and required signature count for P2PK or HTLC verification
+///
+/// Takes into account locktime - if locktime has passed, returns refund keys,
+/// otherwise returns primary pubkeys/hash path. Returns (preimage_needed, pubkeys, required_sigs).
+/// For P2PK, preimage_needed is always false. For HTLC, preimage_needed is true before locktime.
+pub fn get_pubkeys_and_required_sigs(
```
thesimplekid: Does this need to be public? Maybe can be just public to the crate seems like an internal fn.

crates/cashu/src/nuts/nut10.rs
```diff
@@ -105,6 +108,393 @@ impl Secret {
...
+        if let Some(refund_keys) = &conditions.refund_keys {
+            // Locktime has passed and refund keys exist - use refund keys
+            let refund_sigs = conditions.num_sigs_refund.unwrap_or(1);
+            Ok((false, refund_keys.clone(), refund_sigs))
+        } else {
+            // Locktime has passed with no refund keys - anyone can spend
+            Ok((false, vec![], 0))
+        }
+    } else {
+        // Before locktime: logic differs between P2PK and HTLC
```
thesimplekid, line 158: These are helpful comments. May make sense to even link to parts of spec.

crates/cashu/src/nuts/nut10.rs
```diff
@@ -105,6 +108,393 @@ impl Secret {
...
+                Ok((true, pubkeys, required_sigs))
+            }
+        }
+    }
+}
+
+use super::Proofs;
+
+/// Trait for requests that spend proofs (SwapRequest, MeltRequest)
+pub trait VerificationForSpendingConditions {
```
thesimplekid:
```suggestion
pub trait SpendingConditionVerification {
```
Save a word?

crates/cashu/src/nuts/nut10.rs
```diff
@@ -105,6 +108,393 @@ impl Secret {
...
+        }
+
+        Ok(())
+    }
+
+    /// Verify spending conditions for each input individually
+    ///
+    /// Handles SIG_INPUTS mode, non-NUT-10 secrets, and any other case where inputs
+    /// are verified independently rather than as a group.
+    /// This function will NOT be called if any input has SIG_ALL.
```
thesimplekid, line 369: Do we check this?

crates/cashu/src/nuts/nut11/mod.rs
SatsAndSports: Agreed. I'll return an error to the user instead of the `debug_assert`. I'll add new errors in some cases:

image


> For things that a user could do, like calling sign_p2pk on the wrong proof type, we should return an error.

I assumed (without thinking really) that `verify_p2pk` was private and wouldn't be called directly by the user. But I was wrong, it's public and users might therefore call it directly.

====

Thanks for all the comments yesterday, I'll try to work through them today

crates/cdk/src/mint/swap/tests/htlc_sigall_spending_conditions_tests.rs
SatsAndSports: Done. Those test files are now in a folder called `crates/cdk/src/mint/spending_conditions`

I'm not very familiar with how Rust projects are organized, so don't hesitate to suggest other changes

crates/cashu/src/nuts/nut00/mod.rs
SatsAndSports: Agreed. I've removed those comments. I'll push it later today

crates/cashu/src/nuts/nut11/mod.rs
SatsAndSports: Agreed. I've removed it (`verify_sig_all`), and will push it later today

There is another option, if anybody wants to keep a public function with the same name and which is about SIG_ALL checks. The trait gives this class a function called `verify_full_sig_all_check`. We could rename that function to `verify_sig_all` and make it public

crates/cashu/src/nuts/nut11/mod.rs
SatsAndSports: I've now removed that, including that warning.

I added that warning earlier, to remind myself to look at that function. Anyway, it's removed now. It wasn't handling locktime

crates/cashu/src/nuts/nut14/mod.rs
SatsAndSports: Done. Errors instead of `debug_assert`

crates/cashu/src/nuts/nut14/mod.rs
SatsAndSports: I've rewritten that comment in the code. Copied here also:

```
        // While a Witness is usually needed in a P2PK or HTLC proof, it's not                                                                    
        // always needed. If we are past the locktime, and there are no refund                                                                    
        // keys, then the proofs are anyone-can-spend:                                                                                            
        //     NUT-11: "If the tag locktime is the unix time and the mint's local                                                                 
        //              clock is greater than locktime, the Proof becomes spendable                                                               
        //              by anyone, except if [there are no refund keys]"                                                                          
        // Therefore, this function should not extract any Witness unless it                                                                      
        // is needed to get a preimage or signatures.
```

crates/cashu/src/nuts/nut10.rs
SatsAndSports: You're right. Changed from `pub` to `pub(crate)`

crates/cashu/src/nuts/nut10.rs
SatsAndSports: Agreed. Changing from `VerificationForSpendingConditions` to `SpendingConditionVerification`

crates/cashu/src/nuts/nut10.rs
SatsAndSports: That function  is private and is only called from one place, where `verify_spending_conditions` has already verified that there is no SIG_ALL in any input. So the checking is done there

thesimplekid: We should add a debug assert then since its something in our control we're committing not to do.

SatsAndSports:
TODOs (today hopefully):

- [x] sync with 'origin/main', with tests still passing
- [ ] update melt to use the verify_conditions, with some tests
- [ ] `just format`, and anything else related to linting
- [ ] are there some `#[ignore]` tests? Maybe can be activated again the with new test vectors
- [ ] deal with the 'unused' warnings
- [ ] some comments just refer to 'swap', but this works for 'melt' too. Update the comments

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -24,8 +24,11 @@ use crate::types::PaymentProcessorKey;
 use crate::util::unix_time;
 use crate::{ensure_cdk, Amount, Error};
 
-mod melt_saga;
-pub(super) mod shared;
+pub(crate) mod melt_saga;
+pub(crate) mod shared;
+
+#[cfg(test)]
+mod spending_conditions;
```
thesimplekid: We should rename this test mod  to just test and not spending conditions.

crates/cdk/src/mint/swap/mod.rs
```diff
@@ -3,12 +3,15 @@ use cdk_prometheus::METRICS;
...
-use super::nut11::{enforce_sig_flag, EnforceSigFlag};
-use super::{Mint, SigFlag, SwapRequest, SwapResponse};
+use super::{Mint, SwapRequest, SwapResponse};
 use crate::Error;
+use cdk_common::SpendingConditionVerification;
 
 pub mod swap_saga;
 
+#[cfg(test)]
+pub(crate) mod spending_conditions;
```
thesimplekid: We should rename this test mod to just test and not spending conditions.

crates/cashu/src/nuts/nut03.rs
```diff
@@ -91,6 +91,32 @@ impl SwapRequest {
...
+        let mut msg = String::new();
+
+        // Add all input secrets and C values in order
+        // msg = secret_0 || C_0 || ... || secret_n || C_n
+        for proof in &self.inputs {
+            msg.push_str(&proof.secret.to_string());
+            msg.push_str(&proof.c.to_hex());
+        }
+
+        // Add all output amount, keyset_id, and B_ values in order
```
thesimplekid: We don't include the keyset_id anymore right?

crates/cashu/src/nuts/nut05.rs
```diff
@@ -151,6 +151,38 @@ impl<Q: Serialize + DeserializeOwned> MeltRequest<Q> {
...
+        let mut msg = String::new();
+
+        // Add all input secrets and C values in order
+        // msg = secret_0 || C_0 || ... || secret_n || C_n
+        for proof in &self.inputs {
+            msg.push_str(&proof.secret.to_string());
+            msg.push_str(&proof.c.to_hex());
+        }
+
+        // Add all output amount, keyset_id, and B_ values in order (if any)
```
thesimplekid: Same here we don't include keyset_id

crates/cashu/src/nuts/nut05.rs
```diff
@@ -151,6 +151,38 @@ impl<Q: Serialize + DeserializeOwned> MeltRequest<Q> {
     }
 }
 
+impl<Q: std::fmt::Display> super::nut10::SpendingConditionVerification for MeltRequest<Q> {
```
thesimplekid: Can you use where here https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#generics

crates/cdk/src/mint/melt/spending_conditions/htlc_sigall_spending_conditions_tests.rs
```diff
@@ -0,0 +1,181 @@
+#![cfg(test)]
```
thesimplekid: I don't think we should need the cfg test here since the whole mod is a test.

crates/cashu/src/nuts/nut10.rs
```diff
@@ -105,6 +108,432 @@ impl Secret {
...
+                // (data contains hash, not pubkey)
+                let pubkeys = conditions.pubkeys.clone().unwrap_or_default();
+                let required_sigs = conditions.num_sigs.unwrap_or(1);
+                Ok((true, pubkeys, required_sigs))
+            }
+        }
+    }
+}
+
+use super::Proofs;
```
thesimplekid, line 187: Can we move this to the top with other imports

crates/cdk/src/test_helpers/nut10.rs
thesimplekid, line 1: Can we move this file to a test helpers mod. Like we do for cdk https://github.com/cashubtc/cdk/tree/main/crates/cdk/src/test_helpers

crates/cashu/src/nuts/nut05.rs
SatsAndSports: Correct. This comment is out of date

crates/cashu/src/nuts/nut03.rs
SatsAndSports: Correct. This comment is out of date

crates/cashu/src/nuts/nut10.rs
SatsAndSports: I added a few more words, including quotes from the NUTS, to the comment at the top of this code

Issue #621, opened 2025-03-02
Need to get mint info before we can do other operations


thesimplekid:
We should clean up the api so that this is more clear and the error does not happen at the database level and it is clear that the mint info needs to be fetched.

https://github.com/cashubtc/cdk/blob/2827cc4a7a0764882149878f8af45287f290df09/crates/cdk/src/wallet/keysets.rs#L92-L93

thesimplekid:
All yours thanks.

thesimplekid:
This led to this bug where the auth wallet wasn't created:

https://github.com/cashubtc/cdk/pull/704

PR #1240, opened 2025-10-29
Introduce MintMetadataCache for efficient key and metadata management

### LINKED ISSUE #1126: improve checking of keyset and mint info
Currently, for every online request  (mint, melt, or swap) we first query the mint for its latest mint info and keysets. This ensures we always have up‑to‑date data, but it can be overly aggressive and adds many extra round‑trips for a busy wallet.

Instead, we could move this logic into the mint connector (or maybe just elsewhere in the wallet). When the wallet initiates an operation, it would first check its local cache: if the stored keysets and mint info are recent enough (within a defined threshold), it would use them; if they are stale, it would then query the mint for the latest version.

(https://github.com/cashubtc/cdk/pull/1121#discussion_r2379529982)

crodas:
# Description

Fixes #1126 and #621

## Summary

This PR introduces a new `MintMetadataCache` component that provides efficient, on-demand caching of mint metadata (mint info, keysets, and cryptographic keys) with atomic updates and deferred database persistence. This replaces the previous approach of querying the database repeatedly for keyset and key information.

### Problems

1. **Repeated Database Queries**
   - Every key lookup required a database query (`get_keyset_by_id()`)
   - Keyset information fetched from database on every access
   - No in-memory caching of frequently accessed data
   - High I/O overhead for common operations

2. **Scattered Metadata Management**
   - Keys, keysets, and mint info managed separately
   - No unified view of mint state
   - Difficult to ensure consistency across related data

3. **Synchronous Operations**
   - Database writes blocked operations
   - No way to defer expensive persistence operations
   - All operations had to wait for database I/O

4. **No Multi-Wallet Support**
   - Each wallet instance would query database independently
   - No way to share cached metadata across wallet instances
   - Duplicated memory and network fetches for same mint

#### 1. Pull-Based Loading
- Data fetched on-demand from mint HTTP API
- No background threads or workers
- Simple, predictable behavior

```rust
// Load from cache if available, fetch if not
let metadata = cache.load(&storage, &client).await?;

// Force fresh fetch from mint
let fresh = cache.load_from_mint(&storage, &client).await?;
```

#### 2. Atomic Cache Updates
- Uses `ArcSwap` for lock-free reads
- Entire metadata snapshot updated atomically
- Readers always see consistent state
- No partial updates or race conditions

#### 3. Deferred Database Persistence
- Cache updates return immediately
- Database writes happen in background tasks via `spawn()`
- Version tracking prevents redundant writes
- Supports multiple storage backends simultaneously

#### 4. Multi-Database Support
- Tracks which databases have been synced via pointer identity
- Handles wallets using different storage backends
- Only writes to databases that need updating

#### 5. Separate Auth Support
- Dual freshness tracking for regular and auth keysets
- `load()` for regular keysets
- `load_auth()` for blind auth keysets
- Both can be cached independently

### Performance Improvements

1. **Eliminated Database Queries**
   - Keys and keysets accessed from in-memory cache
   - O(1) HashMap lookups instead of database queries
   - Significantly reduced I/O operations

2. **Lock-Free Reads**
   - `ArcSwap` provides lock-free concurrent access
   - No contention between readers
   - Scales well with concurrent operations

3. **Deferred Writes**
   - Operations return immediately without waiting for database
   - Database persistence happens asynchronously
   - No blocking on I/O operations

4. **Shared Caching**
   - Multiple wallets for same mint can share cache instance
   - Reduces memory usage
   - Eliminates redundant network fetches

### Code Quality Improvements

1. **Clear Responsibilities**
   - Metadata caching separated from wallet logic
   - Single source of truth for mint metadata
   - Well-defined module boundaries

2. **Better Documentation**
   - Comprehensive doc comments on all public APIs
   - Clear examples of usage patterns
   - Explicit about cache vs. fresh data behavior

3. **Simplified Logic**
   - No background workers or message passing
   - Synchronous reasoning about cache state
   - Easier to test and debug

4. **WASM Compatible**
   - No reliance on background threads
   - Compatible with single-threaded WASM runtime
   - Uses `task::spawn()` which adapts to platform

### Data Flow

```
┌─────────────────┐
│   Wallet API    │
│ (load_keyset)   │
└────────┬────────┘
         │
         ▼
┌────────────────────┐
│ MintMetadataCache  │
│    .load()         │
└────────┬───────────┘
         │
    ┌────┴────┐
    │ Cached? │
    └────┬────┘
         │
    ┌────┴────┐
    │   Yes   │───► Return cached data (O(1))
    │         │
    │   No    │───► fetch_from_http()
    └─────────┘            │
                          ▼
                  ┌───────────────┐
                  │ Update cache  │ (atomic swap)
                  └───────┬───────┘
                          │
                          ▼
                  ┌───────────────┐
                  │ spawn() task  │ (background)
                  │ persist_to_db │
                  └───────────────┘
```

### Version Tracking

Each database is tracked by pointer identity:
```rust
db_sync_versions: HashMap<usize, usize>
                         ^^^^    ^^^^
                         │       └─ Cache version synced to this DB
                         └───────── Pointer ID of Arc<dyn WalletDatabase>
```

This allows:
- Multiple wallets with same mint but different databases
- Automatic sync detection (only write if versions differ)
- Efficient deferred persistence

### Integration Points

1. **Wallet Struct**
   ```rust
   pub struct Wallet {
       // ... other fields
       pub metadata_cache: Arc<MintMetadataCache>,
   }
   ```

2. **WalletBuilder**
   ```rust
   builder
       .metadata_cache(shared_cache)  // Optional: share across wallets
       .metadata_caches(cache_map)    // Optional: pre-populated cache map
       .build()
   ```

3. **Wallet Methods**
   - `load_keyset_keys()` - Uses cache for key lookups
   - `load_mint_keysets()` - Returns cached keysets
   - `refresh_keysets()` - Forces fresh fetch
   - `get_active_keyset()` - Gets from cache
   - `get_keyset_fees_and_amounts()` - All from cache

### API Changes

**Before:**
```rust
// Direct database queries
let keyset = self.localstore.get_keyset_by_id(&id).await?;
let keys = self.localstore.get_keys(&id).await?;
```

**After:**
```rust
// Cache-first with automatic fetch
let metadata = self.metadata_cache.load(&self.localstore, &self.client).await?;
let keyset = metadata.keysets.get(&id)?;
let keys = metadata.keys.get(&id)?;
```

### Backward Compatibility

- All existing wallet APIs remain unchanged
- Internal implementation uses cache instead of direct DB
- Database still serves as persistent store
- No breaking changes to public APIs

## Performance Metrics

Expected improvements:
- **Keyset lookups**: Database query → O(1) HashMap lookup
- **Key access**: Database query → O(1) HashMap lookup
- **Write latency**: Synchronous DB write → Async spawn
- **Memory sharing**: Per-wallet storage → Shared cache (when using same mint)

### DIFF
--- a/crates/cdk/src/wallet/builder.rs
+++ b/crates/cdk/src/wallet/builder.rs
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::time::Duration;
 
 use cdk_common::database;
+use cdk_common::parking_lot::Mutex;
 #[cfg(feature = "auth")]
 use cdk_common::AuthToken;
 #[cfg(feature = "auth")]
@@ -27,6 +29,7 @@ pub struct WalletBuilder {
     seed: Option<[u8; 64]>,
     use_http_subscription: bool,
     client: Option<Arc<dyn MintConnector + Send + Sync>>,
+    metadata_cache_ttl: Option<Duration>,
     metadata_cache: Option<Arc<MintMetadataCache>>,
     metadata_caches: HashMap<MintUrl, Arc<MintMetadataCache>>,
 }
@@ -42,6 +45,7 @@ impl Default for WalletBuilder {
             auth_wallet: None,
             seed: None,
             client: None,
+            metadata_cache_ttl: None,
             use_http_subscription: false,
             metadata_cache: None,
             metadata_caches: HashMap::new(),
@@ -58,6 +62,12 @@ impl WalletBuilder {
     /// Use HTTP for wallet subscriptions to mint events
     pub fn use_http_subscription(mut self) -> Self {
         self.use_http_subscription = true;
+        self
+    }
+
+    /// Set metadata_cache_ttl
+    pub fn set_metadata_cache_ttl(mut self, metadata_cache_ttl: Option<Duration>) -> Self {
+        self.metadata_cache_ttl = metadata_cache_ttl;
         self
     }
 
@@ -201,6 +211,8 @@ impl WalletBuilder {
             }
         };
 
+        let metadata_cache_ttl = self.metadata_cache_ttl;
+
         let metadata_cache = self.metadata_cache.unwrap_or_else(|| {
             // Check if we already have a cache for this mint in the HashMap
             if let Some(cache) = self.metadata_caches.get(&mint_url) {
@@ -216,7 +228,7 @@ impl WalletBuilder {
             unit,
             localstore,
             metadata_cache,
-            metadata_cache_ttl: None.into(),
+            metadata_cache_ttl: Arc::new(Mutex::new(metadata_cache_ttl)),
             target_proof_count: self.target_proof_count.unwrap_or(3),
             #[cfg(feature = "auth")]
             auth_wallet: Arc::new(RwLock::new(self.auth_wallet)),
--- a/crates/cdk/src/wallet/mint_metadata_cache.rs
+++ b/crates/cdk/src/wallet/mint_metadata_cache.rs
@@ -44,9 +44,6 @@ use crate::wallet::MintConnector;
 use crate::wallet::{AuthMintConnector, AuthWallet};
 use crate::{Error, Wallet};
 
-/// Default TTL
-pub const DEFAULT_TTL: Duration = Duration::from_secs(300);
-
 /// Metadata freshness and versioning information
 ///
 /// Tracks when data was last fetched and which version is currently cached.
@@ -214,6 +211,7 @@ impl MintMetadataCache {
     ///
     /// * `storage` - Database to persist metadata to (async background write)
     /// * `client` - HTTP client for fetching from mint server
+    /// * `ttl` - Optional TTL, if not provided it is assumed that any cached data is good enough
     ///
     /// # Returns
     ///
@@ -233,14 +231,15 @@ impl MintMetadataCache {
         ttl: Option<Duration>,
     ) -> Result<Arc<MintMetadata>, Error> {
         // Acquire lock to ensure only one fetch at a time
-        let ttl = ttl.unwrap_or(DEFAULT_TTL);
         let current_version = self.metadata.load().status.version;
         let _guard = self.fetch_lock.lock().await;
 
         // Check if another caller already updated the cache while we waited
         let current_metadata = self.metadata.load().clone();
         if current_metadata.status.is_populated
-            && current_metadata.status.updated_at + ttl > Instant::now()
+            && ttl
+                .map(|ttl| current_metadata.status.updated_at + ttl > Instant::now())
+                .unwrap_or(true)
             && current_metadata.status.version > current_version
         {
             // Cache was just updated by another caller - return it
@@ -275,6 +274,7 @@ impl MintMetadataCache {
     ///
     /// * `storage` - Database to persist metadata to (if fetched or stale)
     /// * `client` - HTTP client for fetching from mint (only if cache empty)
+    /// * `ttl` - Optional TTL, if not provided it is assumed that any cached data is good enough
     ///
     /// # Returns
     ///
@@ -295,7 +295,6 @@ impl MintMetadataCache {
     ) -> Result<Arc<MintMetadata>, Error> {
         let cached_metadata = self.metadata.load().clone();
         let storage_id = Self::arc_pointer_id(storage);
-        let ttl = ttl.unwrap_or(DEFAULT_TTL);
 
         // Check what version of cache this database has seen
         let db_synced_version = self
@@ -306,7 +305,9 @@ impl MintMetadataCache {
             .unwrap_or_default();
 
         if cached_metadata.status.is_populated
-            && cached_metadata.status.updated_at + ttl > Instant::now()
+            && ttl
+                .map(|ttl| cached_metadata.status.updated_at + ttl > Instant::now())
+                .unwrap_or(true)
         {
             // Cache is ready - check if database needs updating
             if db_synced_version != cached_metadata.status.version {
@@ -319,7 +320,7 @@ impl MintMetadataCache {
         }
 
         // Cache not populated - fetch from mint
-        self.load_from_mint(storage, client, Some(ttl)).await
+        self.load_from_mint(storage, client, ttl).await
     }
 
     /// Load auth keysets and keys (auth feature only)
--- a/crates/cdk/src/wallet/mod.rs
+++ b/crates/cdk/src/wallet/mod.rs
@@ -2,7 +2,6 @@
 
 use std::collections::HashMap;
 use std::fmt::Debug;
-use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
@@ -79,53 +78,6 @@ pub use types::{MeltQuote, MintQuote, SendKind};
 
 use crate::nuts::nut00::ProofsMethods;
 
-/// Cloneable mutex
-///
-/// Each instance is independent from each other, that's why an Arc is not an option
-#[derive(Debug)]
-pub struct CloneableMutex<T>(Mutex<T>)
-where
-    T: Clone + Debug;
-
-impl<T> From<T> for CloneableMutex<T>
-where
-    T: Clone + Debug,
-{
-    fn from(value: T) -> Self {
-        Self(Mutex::new(value))
-    }
-}
-
-impl<T> CloneableMutex<T>
-where
-    T: Clone + Debug,
-{
-    /// Clone the inner guarded object
-    pub fn clone_inner(&self) -> T {
-        (*self.0.lock().deref()).clone()
-    }
-}
-
-impl<T> Clone for CloneableMutex<T>
-where
-    T: Clone + Debug,
-{
-    fn clone(&self) -> Self {
-        Self(Mutex::new(self.0.lock().clone()))
-    }
-}
-
-impl<T> Deref for CloneableMutex<T>
-where
-    T: Clone + Debug,
-{
-    type Target = Mutex<T>;
-
-    fn deref(&self) -> &Self::Target {
-        &self.0
-    }
-}
-
 /// CDK Wallet
 ///
 /// The CDK [`Wallet`] is a high level cashu wallet.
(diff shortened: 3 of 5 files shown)


thesimplekid:
Closes https://github.com/cashubtc/cdk/issues/1126

crates/cdk-common/src/lib.rs
```diff
@@ -8,6 +8,33 @@
 #![warn(missing_docs)]
 #![warn(rustdoc::bare_urls)]
 
+pub mod task {
```
thesimplekid: Can we make this its own file.

https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#sub-modules

crates/cdk/src/wallet/auth/auth_wallet.rs
```diff
@@ -170,89 +174,61 @@ impl AuthWallet {
...
+            .clone())
     }
 
-    /// Get blind auth keysets from local database or go online if missing
+    /// Get blind auth keysets from KeyManager cache or trigger refresh if missing
     ///
-    /// First checks the local database for cached blind auth keysets. If keysets are not found locally,
-    /// goes online to refresh keysets from the mint and updates the local database.
+    /// First checks the KeyManager cache for keysets. If keysets are not cached,
+    /// triggers a refresh from the mint and waits briefly before checking again.
```
thesimplekid: Where is the wait mentioned in this comment I don't see it here if get_keysets does that internally we should remove the comment here

crates/cdk/src/wallet/key_manager.rs
```diff
@@ -0,0 +1,1064 @@
...
+
+    /// Refresh interval
+    refresh_interval: Duration,
+
+    /// Internal counter for each registered/mint wallet
+    counter: AtomicUsize,
+
+    /// Semaphore to limit concurrent HTTP requests to mint servers
+    /// This is stored to keep it alive and is passed to the refresh loop
+    #[allow(dead_code)]
```
thesimplekid: Why is this dead?

crates/cdk/src/wallet/key_manager.rs
```diff
@@ -0,0 +1,1064 @@
...
+            .next()
+            .ok_or(Error::IncorrectMint)?
+            .client
+            .clone();
+
+        let keyset = http_client.get_mint_keyset(*keyset_id).await?;
+
+        // Persist to all databases
+        for storage in &storages {
+            let _ = storage.add_keys(keyset.clone()).await.inspect_err(|e| {
```
thesimplekid: We should verify the keyset id before adding it to storage.

crates/cdk/src/wallet/key_manager.rs
```diff
@@ -0,0 +1,1064 @@
...
+    }
+
+    /// Load mint info and keys from all registered databases
+    ///
+    /// Iterates through all storage backends and loads keysets and keys into cache.
+    /// This is called on first access when cache is empty.
+    async fn fetch_mint_info_and_keys_from_db(
+        registration: &MintRegistration,
+    ) -> Result<(), Error> {
+        debug!(
```
thesimplekid: Can we use the full path here. https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#full-paths-for-logging

crates/cdk/src/wallet/key_manager.rs
```diff
@@ -0,0 +1,1064 @@
...
+                        )
+                    });
+            }
+
+            if !registration.cache.load().is_ready {
+                let mint_url = registration.mint_url.clone();
+                let _ = tokio::time::timeout(
+                    Duration::from_secs(60),
+                    Self::fetch_from_http(registration, refresh_scheduler),
+                )
```
thesimplekid: Doesn't this bypass the semaphore so more then the limit of http requests can occur?

crates/cdk/src/wallet/key_manager.rs
```diff
@@ -0,0 +1,1064 @@
...
+                .inspect_err(|e| warn!("Failed to fetch keys for {} with error {}", mint_url, e));
+            }
+        });
+    }
+
+    /// Refresh keys from mint server
+    ///
+    /// Spawns an async task with 60s timeout. HTTP requests are limited to
+    /// MAX_CONCURRENT_HTTP_REQUESTS concurrent requests via semaphore.
+    fn fetch_and_sync_mint_task(
```
thesimplekid: I'm not sure why we have this fn and sync_mint_task do we need both can they be combined?

crates/cdk/src/wallet/key_manager.rs
crodas: We only throttle refreshes, if that code is reached, we need to ping the mint right away since we have zero keysets in our db.

crates/cdk/src/wallet/key_manager.rs
crodas: Yeah, it's confusing, I'll name them better.

1. Sync: Will wall all available datastores and try to get info from any of them. If found, it will replicate that info to all stores. Otherwise it will work as fetch without the semaphore
2. Fetch: Tries to sync, then fetches from the mint, always, using the semphore.

I'll combine them both

crates/cdk/src/wallet/key_manager/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// for the cache to be updated with a newer version.
+    pub async fn refresh(&self) -> Result<Vec<KeySetInfo>, Error> {
+        let last_version = self.cache.load().refresh_version;
+
+        self.send_message(MessageToWorker::FetchMint);
+
+        for _ in 0..MAX_RETRY {
+            if let Some(keysets) = {
+                let cache = self.cache.load();
+                if last_version < cache.refresh_version {
```
thesimplekid: This send_message will silently fail in the admittedly unlikely event the message cannot be sent on the channel but then we will still go into the polling loop. If we know the channel message failed we should return early with an error I think.

crodas:
> Nice work on this, moving to single mint makes this simpler. However, I'm wondering if the background worker is the best fit for us. I don't think we should add a background task to the wallet. This could have privacy implications and leak metadata about wallet activity. I think using a simple cache with TTL that can be configured is better, since it will only go online when the wallet is going online anyway. I also think this method would be simpler and reduce the code complexity since there is no background task or message passing.

The alternative is to have the key_manager accept a MintConnector in the constructor, which it may use right there to fetch the keys, or drop it.

When refresh_keys is called, a MintConnector should be passed. Also, the Wallet would call refresh_keys less frequently internally.

How do you like the outlined plan @thesimplekid  ? If you agree, I can make it happen quickly with the code that I already wrote.

crates/cdk/Cargo.toml
```diff
@@ -40,6 +40,7 @@ cdk-common.workspace = true
 cbor-diag.workspace = true
 async-trait.workspace = true
 anyhow.workspace = true
+once_cell = "1.20.2"
```
thesimplekid: I don't think this is actually used

thesimplekid (commented):
What is the api for a user that whats to check how old a cache is and if they should fresh or not? It seems like were still calling refresh on every online operation and always going to the mint and not taking advantage of the cache?

I'm thinking we should provide an api on the wallet refresh if expired or something that accepts a Duration and if the cache is older then that we get from the mint?

crates/cdk/Cargo.toml
crodas: It was used, I'll remove it now.

crates/cdk/src/wallet/builder.rs
```diff
@@ -162,10 +201,22 @@ impl WalletBuilder {
...
+                Arc::new(MintMetadataCache::new(mint_url.clone()))
+            }
+        });
+
         Ok(Wallet {
             mint_url,
             unit,
             localstore,
+            metadata_cache,
+            metadata_cache_ttl: None.into(),
```
thesimplekid: Why is this none? shouldn't we allow this to be set in the builder?

crates/cdk/src/wallet/mint_metadata_cache.rs
```diff
@@ -271,9 +291,11 @@ impl MintMetadataCache {
         &self,
         storage: &Arc<dyn WalletDatabase<Err = database::Error> + Send + Sync>,
         client: &Arc<dyn MintConnector + Send + Sync>,
+        ttl: Option<Duration>,
     ) -> Result<Arc<MintMetadata>, Error> {
         let cached_metadata = self.metadata.load().clone();
         let storage_id = Self::arc_pointer_id(storage);
+        let ttl = ttl.unwrap_or(DEFAULT_TTL);
```
thesimplekid: Doesn't having this unwrap defeat the purpose of the option? If it is None then we should not auto refresh and should only refresh when the user explicitly calls it.

crates/cdk/src/wallet/mod.rs
```diff
@@ -75,6 +79,53 @@ pub use types::{MeltQuote, MintQuote, SendKind};
...
+where
+    T: Clone + Debug,
+{
+    type Target = Mutex<T>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
```
thesimplekid: The CloneableMutex implementation seems weird to me. I think we should just remove Clone from the Wallet (and mint) and force the use of an Arc if multiple copies are needed, but that can be another pr. I think for this using Arc<RwLock<Option<Duration>>> instead. The current CloneableMutex design suggests each clone should have independent TTL, but in practice, if you're cloning wallets that share the same cache, they should probably share TTL settings too. This is more efficient and likely the desired behavior.

crodas: I'm in favour of removing the Clone of the wallet, in another PR. I'll apply the changes you suggested here.

crodas: Fixed in 11adbf9f3675f25557047876a6f090b7e3388a0f

crates/cdk/src/wallet/builder.rs
crodas: Fixed in 11adbf9f3675f25557047876a6f090b7e3388a0f

crates/cdk/src/wallet/mint_metadata_cache.rs
crodas: Fixed in 11adbf9f3675f25557047876a6f090b7e3388a0f

thesimplekid:
ACK https://github.com/cashubtc/cdk/commit/11adbf9f3675f25557047876a6f090b7e3388a0f

PR #1244, opened 2025-10-31
PreMintSecrets: fix `into_iter()`

codingpeanut157:
### Description
In the `impl Iterator for PremintSecrets` the function `next` uses `Vec::pop` which pops an element from the end.
The result is an iterator that iterates over the elements in the reverse order respect to the iterator returned by `PremintSecrets::iter()`.

This commits fixes the bug by using `Vec::remove(0)`

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid (approved):

thesimplekid:
This might not be the most performant solution since for every remove we need to shift all elements. However, since these are probably small enough lists its likely fine. But we could consider using a VecDeque so we have pop front if this becomes a concern.

Issue #1104, opened 2025-09-22
New websocket subscriptions should check mint quote states


gudnuf:
Mint quotes are no longer checked on startup since https://github.com/cashubtc/cdk/pull/873. Currently the http endpoints check mint quote state, but if a wallet is only using websockets the `on_new_subscription` method will only check the quote state in the database rather than asking the LN backend. New subscriptions should check mint quote states to make sure the wallet gets the latest actual state.

For reference, I tried to implement this [here](https://github.com/gudnuf/cdk/commit/05309c2c36e32f82839cbb3775299d9b1050c1bd);  however, we can't pass the `Mint` instance to the `PubSubManager` because that creates a circular dependency, so I passed the `payment_processors` to the `PubSubManager`. This does not seem like the best approach, so I'm not opening a PR.

crodas:
Fixed on #1246

PR #1246, opened 2025-11-02
Fix websocket issues and mint quotes

### LINKED ISSUE #1209: Fix issue with websocket
When checking the status of a mint quote form the websocket the check only happens against the database and not the processor,  it should be similar to `check_mint_quote` and not only the database.

Also address these logings

Image

crodas:
### Description

Fixes #1209 and #1104

When checking with HTTP pulling, mint quotes would ping the ln backend if needed to get any payment that may not be included, so the mint uses this chance to correct its database.

This check was missing for database subscriptions
-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/subscription.rs
+++ b/crates/cdk/src/mint/subscription.rs
@@ -27,7 +27,7 @@ pub struct MintPubSubSpec {
 }
 
 impl MintPubSubSpec {
-    /// Similar to Mint::check_mint_quote_paid, but copied to avoid
+    /// Call Mint::check_mint_quote_payments to update the quote pinging the payment backend
     async fn get_mint_quote(
         &self,
         quote_id: &QuoteId,


crates/cdk/src/mint/subscription.rs
```diff
@@ -1,27 +1,54 @@
...
 
 /// Mint subtopics
 #[derive(Clone)]
 pub struct MintPubSubSpec {
     db: DynMintDatabase,
+    payment_processors: Arc<HashMap<PaymentProcessorKey, DynMintPayment>>,
 }
 
 impl MintPubSubSpec {
+    /// Similar to Mint::check_mint_quote_paid, but copied to avoid
```
thesimplekid: I think this comment is cut off?

thesimplekid (commented):
Nice looks good just think one doc comment needs to be fixed.

crates/cdk/src/mint/subscription.rs
crodas: Yeah, that was the old doc-doc from a previous approach. Lemme update it.

PR #1247, opened 2025-11-02
feat: add keyset_amounts table to track issued and redeemed amounts

### LINKED ISSUE #1130: Store total issued and total burned per keyset
Curently for the management rpc total issued response (https://github.com/cashubtc/cdk/blob/6d0003a4fceb7738b0e4d0f8d76b8925bbf5fd6e/crates/cdk/src/mint/mod.rs#L997) and the total redeemed (https://github.com/cashubtc/cdk/blob/6d0003a4fceb7738b0e4d0f8d76b8925bbf5fd6e/crates/cdk/src/mint/mod.rs#L1032) we compute it at the time of the query by summing the blind_signature table and proofs table responsibility. This is obviously inefficient and will not scale well. Instead we should just store a counter in the keysets table.

crodas:
### Description

Fixes #1130

This commit introduces a new `keyset_amounts` table to efficiently track the total issued and redeemed amounts per keyset, replacing the need to calculate these values on-demand from the proof and blind_signature tables.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/test_swap_flow.rs
+++ b/crates/cdk-integration-tests/tests/test_swap_flow.rs
@@ -18,7 +18,6 @@ use cashu::{CurrencyUnit, Id, PreMintSecrets, SecretKey, SpendingConditions, Sta
 use cdk::mint::Mint;
 use cdk::nuts::nut00::ProofsMethods;
 use cdk::Amount;
-use cdk_common::database::mint::{ProofsDatabase, SignaturesDatabase};
 use cdk_integration_tests::init_pure_tests::*;
 
 /// Helper to get the active keyset ID from a mint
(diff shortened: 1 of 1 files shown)


thesimplekid (commented):
What is the benefit of doing this in a new table? I was just thinking we could add two new columns to the `keyset` table one `total_issued` and the other `total_redeemed`.  This keeps all the keyset data in table.

crates/cdk-integration-tests/tests/test_swap_flow.rs
```diff
@@ -58,6 +59,29 @@ async fn test_swap_happy_path() {
         .expect("Could not get proofs");
 
     let keyset_id = get_keyset_id(&mint).await;
+
+    // Check initial amounts after minting
+    let total_issued = mint.total_issued().await.unwrap();
+    let total_redeemed = mint.total_redeemed().await.unwrap();
```
thesimplekid, line 64: I think we should add specific tests for this and not just include it in other tests.

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -1698,6 +1736,19 @@ where
                     .bind("signed_time", current_time as i64)
                     .execute(&self.inner)
                     .await?;
+
+                    query(
+                        r#"
+                        INSERT INTO keyset_amounts (keyset_id, type, amount)
+                        VALUES (:keyset_id, 'issued', :amount)
+                        ON CONFLICT (keyset_id, type)
+                        DO UPDATE SET amount = keyset_amounts.amount + :amount
```
thesimplekid: I wonder if it would be better if create the first entry with a zero amount and then update that each time instead of doing the insert with the on conflict.

crates/cdk-sql-common/src/mint/migrations/sqlite/20251102000000_create_keyset_amounts.sql
```diff
@@ -0,0 +1,24 @@
+-- Create keyset_amounts table
+CREATE TABLE IF NOT EXISTS keyset_amounts (
+    keyset_id TEXT NOT NULL,
+    type TEXT NOT NULL,
+    amount INTEGER NOT NULL DEFAULT 0,
+    PRIMARY KEY (keyset_id, type)
+);
```
thesimplekid, line 6: What if we changed the schema to have one row per keyset with `total_issued` and `total_redeemed` columns, instead of two rows per keyset with a `type` column?

crodas: I like this, will make this happen

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -219,6 +219,43 @@ where
...
+                r#"
+                UPDATE keyset_amounts
+                SET total_redeemed = total_redeemed + amounts.total
+                FROM (
+                    SELECT keyset_id, COALESCE(SUM(amount), 0) as total
+                    FROM proof
+                    WHERE y IN (:ys)
+                    GROUP BY keyset_id
+                ) amounts
+                WHERE keyset_amounts.keyset_id = amounts.keyset_id
```
thesimplekid: What if we update the `get_current_states` to return the `amount` and `keyset_id` with the y this way we can avoid the second query on the proofs table here.

thesimplekid: I think if we do this we need to make sure the get current state locks the rows affected to avoid a race condition.

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -219,6 +219,43 @@ where
...
+                "#,
+            )?
+            .bind_vec("ys", ys.iter().map(|y| y.to_bytes().to_vec()).collect())
+            .execute(&self.inner)
+            .await?;
+
+            // If no rows were updated, INSERT new rows (rare case - first time seeing this keyset)
+            if updated == 0 {
+                query(
```
thesimplekid: I think I meant something different then adding an update then an insert if nothing is affected. I meant can we create the entry with a zero amount in the `keysets_amounts` table when we create the keyset, then on a `proof` state update we know it exists. But maybe this isn't possible since the keyset table is the signatory? If thats the case the upsert pattern with `INSERT INTO` and an `ON CONFLICT` is probably better. As i think this `UPDATE` then `INSERT` pattern is a race condition.

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -1698,6 +1754,33 @@ where
...
+                    .bind("amount", u64::from(signature.amount) as i64)
+                    .bind("keyset_id", signature.keyset_id.to_string())
+                    .execute(&self.inner)
+                    .await?;
+
+                    // If no rows updated, INSERT (rare case - first issuance for this keyset)
+                    if updated == 0 {
+                        query(
+                            r#"
+                            INSERT INTO keyset_amounts (keyset_id, total_issued, total_redeemed)
```
thesimplekid: Same applies here https://github.com/cashubtc/cdk/pull/1247/files#r2492850899

crates/cdk-integration-tests/tests/integration_tests_pure.rs
thesimplekid, line 1: We should add a test for if a swap included multiple keysets that they are all updated correctly.

crates/cdk-common/src/database/mint/mod.rs
```diff
@@ -343,16 +347,38 @@ pub trait SignaturesDatabase {
...
+pub trait MintKeysTransaction<'a> {
+    /// Database Error
+    type Err: Into<Error> + From<Error>;
+
+    /// Function to call everytime the signatory pushes information towards the mint
+    async fn set_signatory_keysets(
+        &mut self,
+        pubkey: &PublicKey,
+        keysets: &[MintKeySetInfo],
+    ) -> Result<(), Self::Err>;
```
thesimplekid: Wouldn't this need to be added to the signatory spec? I'm wondering if the upsert pattern is better it would remove the need for this.

crodas: This is for the mint only; the spec is unaware, and this is a chance for the mint to react to new keys, but no private key is given, so no need for the spec. The idea is to have a shared place to react when keys are read from the signatory. Does this make sense?

What is missing from the spec, I just realized, is the notification when keys are rotated by another process (all mints should be notified). But that's a topic for another day.

thesimplekid: > What is missing from the spec, I just realized, is the notification when keys are rotated by another process (all mints should be notified). But that's a topic for another day.

Yeah guess thats what I was thinking this was for. However, isnt the upsert pattern still easier and we can just remove this.

crodas: Done already

PR #1250, opened 2025-11-03
fix: add proof recovery mechanism for failed wallet operations

### LINKED ISSUE #1180: Stored proofs remain in pending state when melt fails due to mint's maximum proof limit
## Description

When attempting to melt a large number of proofs and the mint rejects the operation due to exceeding the maximum number of proofs that can be burned in a single operation, the proofs become stuck in a "pending" state in the CDK wallet database. These proofs are then not returned by `get_unspent_proofs()` in subsequent redemption attempts, effectively making them unavailable for spending.

## Expected Behavior

When a melt operation fails (regardless of the reason), proofs that were marked as "pending" should be reverted back to "unspent" state so they can be:
1. Retrieved by subsequent `get_unspent_proofs()` calls
2. Available for retry in the next redemption cycle
3. Not permanently locked in an intermediate state

## Actual Behavior

When `wallet.melt()` or `wallet.melt_proofs()` fails due to mint rejection (e.g., exceeding proof count limit), the proofs remain in "pending" state and are effectively lost from the wallet's available balance. They are neither spendable nor visible to the wallet's standard queries for unspent proofs.

## Steps to Reproduce

1. Store a large number of proofs in the CDK wallet (e.g., >1000 proofs)
2. Attempt to melt all proofs using `wallet.melt(quote_id)` or `wallet.melt_proofs(quote_id, proofs)`
3. Mint rejects the operation with an error like "too many proofs" or "maximum proof limit exceeded"
4. Subsequent calls to `wallet.get_unspent_proofs()` return empty or incomplete results
5. `wallet.total_balance()` might still show the balance, but proofs are inaccessible

## Environment

- **CDK Version**: 0.13.1
- **CDK-SQLite Version**: 0.13.1
- **Wallet Database**: SQLite with WAL mode
- **Scenario**: Automatic redemption task attempting to melt accumulated proofs to Lightning
- **Mint Example**: mint.coinos.io (has 1000 proof limit per melt operation)

## Code Context

In our implementation, we have a redemption loop that processes stored tokens:

```rust
// Get all spendable proofs
let proofs = match wallet_clone.get_unspent_proofs().await {
    Ok(p) => p,
    Err(e) => {
        error!("Failed to get spendable proofs: {}", e);
        continue;
    }
};

// ... proof selection logic ...

// Attempt to melt proofs
let melt_result = wallet_clone.melt(&quote.id).await;

match melt_result {
    Ok(result) => {
        // Success - proofs are properly burned
        info!("Successfully melted proofs");
    },
    Err(e) => {
        // ISSUE: Proofs remain in pending state
        error!("Failed to melt proofs: {}", e);
        // No automatic state recovery happens here
    }
}
```

## Workaround

We've implemented `CASHU_MAX_PROOFS_PER_MELT` configuration to limit the number of proofs sent per melt operation:

```rust
// Select proofs to melt based on configured limit
let (proofs_to_melt, selected_total_msat) = if max_proofs_per_melt > 0 && proofs.len() > max_proofs_per_melt {
    // Limit exceeded - select only the configured number of proofs
    let selected_proofs: Vec<_> = proofs.iter().take(max_proofs_per_melt).cloned().collect();
    // ... calculate selected total ...
    (selected_proofs, selected_total)
} else {
    // No limit or under limit - use all proofs
    (proofs.clone(), total_amount_msat)
};
```

However, this doesn't solve the underlying issue - if a melt operation fails for **any reason**, the proofs should be returned to unspent state.

## Impact

- **Loss of funds**: Proofs stuck in pending state are effectively inaccessible
- **Manual intervention required**: Database must be manually inspected and proof states corrected
- **Poor user experience**: Failed redemptions result in "lost" tokens from the user's perspective
- **Silent failure**: Wallet continues to show balance but proofs can't be spent

## Proposed Solution

The CDK wallet should implement automatic state rollback when melt operations fail:

1. **Before melt**: Mark proofs as pending
2. **During melt**: Execute the melt operation
3. **On success**: Mark proofs as spent/burned
4. **On failure**: **Automatically revert proofs back to unspent state**

This could be implemented with:
- Transaction-like semantics for proof state changes
- Catch block in melt operations that reverts state on error
- State recovery mechanism that checks for stale "pending" proofs on wallet initialization

## Related Issues

This affects any scenario where melt operations can fail, including:
- Mint proof limits (this issue)
- Network failures during melt
- Mint downtime
- Invalid quote IDs
- Fee reserve calculation errors
- Timeout errors

## Additional Context

We're building an L402 nginx module that accepts Cashu tokens and automatically redeems them to Lightning. This issue surfaced when users started sending many small-denomination tokens, causing the accumulated proof count to exceed mint limits.

The workaround of limiting proofs per melt works for the specific proof-count-limit issue, but the underlying state management problem remains a concern for production reliability.

**Repository**: https://github.com/DhananjayPurohit/ngx_l402
**Relevant Code**: [src/cashu.rs#L625-L794](https://github.com/DhananjayPurohit/ngx_l402/blob/main/src/cashu.rs#L625-L794)

crodas:
### Description

This commit introduces a new `try_proof_operation` helper that wraps wallet operations (swap, melt) with automatic proof recovery in case of network or mint failures. When an operation fails, the wallet now attempts to recover by marking proofs as unspent and swapping them to prevent loss of funds.

Fixes #1180

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk/src/wallet/try_proof_operation.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[cfg(not(target_arch = "wasm32"))]
+pub trait MaybeSend: Send {}
+#[cfg(target_arch = "wasm32")]
+pub trait MaybeSend {}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl<T: ?Sized + Send> MaybeSend for T {}
+#[cfg(target_arch = "wasm32")]
+impl<T: ?Sized> MaybeSend for T {}
```
thesimplekid: Can we add some comments about why we need this. I know its for wasm but can we document it.

crodas: Yeah, sure. That's because async tasks are not Send in wasm.

crates/cdk/src/wallet/reclaim.rs
```diff
@@ -0,0 +1,105 @@
...
+                        )
+                        .is_ok();
+
+                    if swap_reverted_proofs {
+                        tracing::error!(
+                            "Attempting to swap exposed {} proofs to new proofs",
+                            inputs.len()
+                        );
+                        for proofs in inputs.chunks(BATCH_PROOF_SIZE) {
+                            if let Err(inner_err) = self.reclaim_unspent(proofs.to_owned()).await {
```
crodas, line 72: The reclaim may be too much, maybe hide it behind a feature flag or just remove it for now, the `unspent` should be enough.

PR #1251, opened 2025-11-03
feat: custom axum router

asmogo:
### Description

Add support for custom payment methods and custom Axum routes. Custom payment methods can be provided by gRPC payment processors using the [SettingsResponse](https://github.com/cashubtc/cdk/pull/1251/files#diff-a8498671bfd5722cd1d754011e6b07a93045a7d4da100d9ad7c145ecdf05d582R17) proto message.

- [x] Rename `NutCustom` to `CustomPaymentMethod`
- [x] Update mintd to NOT load MintInfo from storage on start

-----

### Notes to the reviewers

-----

#### CHANGED
- Payment Processor: Extended protobuf definitions and client/server logic for custom payment processing. 
#### ADDED
- Custom Router Implementation: Added custom_router.rs and custom_handlers.rs in cdk-axum to support custom Axum routing and handlers.
#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-cli/src/sub_commands/mint.rs
+++ b/crates/cdk-cli/src/sub_commands/mint.rs
@@ -9,7 +9,6 @@ use cdk::wallet::MultiMintWallet;
 use cdk::{Amount, StreamExt};
 use cdk_common::nut00::KnownMethod;
 use clap::Args;
-use lightning::types::payment;
 use serde::{Deserialize, Serialize};
 
 use crate::utils::get_or_create_wallet;


crates/cdk/src/wallet/receive.rs
```diff
@@ -207,10 +207,23 @@ impl Wallet {
...
+                "Unit mismatch detected - Token unit: {:?}, Wallet unit: {:?}",
+                unit,
+                self.unit
+            );
+            tracing::debug!("Token data: {:?}", token);
+            tracing::debug!("Encoded token: {}", encoded_token);
+        }
 
         ensure_cdk!(unit == self.unit, Error::UnsupportedUnit);
+        println!("units passed");
```
asmogo: remove this. This was used for debugging.

crates/cdk-payment-processor/src/proto/payment_processor.proto
```diff
@@ -128,16 +135,24 @@ message Bolt12OutgoingPaymentOptions {
...
+  string offer = 1;
+  optional uint64 max_fee_amount = 2;
+  optional uint64 timeout_secs = 3;
+  optional MeltOptions melt_options = 5;
+}
 
 enum OutgoingPaymentOptionsType {
   OUTGOING_BOLT11 = 0;
   OUTGOING_BOLT12 = 1;
+  OUTGOING_PAYMENT_OPTIONS_TYPE_CUSTOM = 2;
```
asmogo: This was prefixed to avoid collisions. 
https://buf.build/docs/best-practices/style-guide/#enums
https://buf.build/docs/lint/rules/#enum_value_prefix

We might want to prefix all other enums as well.

crates/cdk-lnd/src/lib.rs
```diff
@@ -116,12 +116,13 @@ impl Lnd {
...
             wait_invoice_cancel_token: CancellationToken::new(),
             wait_invoice_is_active: Arc::new(AtomicBool::new(false)),
-            settings: Bolt11Settings {
+            settings: PaymentProcessorSettings {
                 mpp: true,
                 unit: CurrencyUnit::Msat,
                 invoice_description: true,
                 amountless: true,
                 bolt12: false,
+                custom: "".to_string(),
```
asmogo: I don't like defaulting like this.

crates/cdk-common/src/payment.rs
```diff
@@ -398,7 +448,7 @@ pub struct PaymentQuoteResponse {
 
 /// Ln backend settings
 #[derive(Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize)]
-pub struct Bolt11Settings {
+pub struct PaymentProcessorSettings {
```
asmogo: This was renamed. 
We might want to move this to the .proto file

crates/cashu/src/nuts/auth/nut21.rs
```diff
@@ -121,70 +119,134 @@ pub enum Method {
...
         .filter(|path| regex.is_match(&path.to_string()))
         .collect())
 }
+fn transform_custom(input: &str) -> String {
+    // Replace '_' with '/' and then ensure the string ends with "v1/"
+    let mut s = input.replace('_', "/");
+
+    if !s.starts_with("/v1/") {
+        s.insert_str(0, "/v1/");
+    }
```
asmogo: I think this function fixes this 
https://github.com/cashubtc/cdk/pull/1251/files#diff-02d320f8bc111746b6a271a1e5fd84c2002aafe839a4fab3bf5e62006287aa3fR822-R825

crates/cdk-mintd/example.config.toml
```diff
@@ -170,6 +167,12 @@ max_delay_time = 3
 # port = 50051
 # tls_dir = "/path/to/tls"
 
+# [custom_payment_methods]
+# Configure custom payment methods (requires grpc_processor)
+# This creates dedicated routes like /v1/mint/quote/paypal, /v1/mint/paypal, etc.
+# enabled = ["paypal", "venmo", "cashapp"]
+# Note: Your gRPC payment processor must implement support for these custom methods
+
```
asmogo: r we ok with this configuration?

crates/cdk-payment-processor/src/proto/client.rs
```diff
@@ -157,6 +167,9 @@ impl MintPayment for PaymentProcessorClient {
         let mut inner = self.inner.clone();
 
         let request_type = match &options {
+            cdk_common::payment::OutgoingPaymentOptions::Custom(_) => {
+                OutgoingPaymentRequestType::Bolt11Invoice
```
asmogo: this is wrong

crates/cdk-mintd/example.config.toml
asmogo: this has been updated in [feat: refactor custom payment method handling](https://github.com/cashubtc/cdk/pull/1251/commits/988fb296f5479bf0c266d9660e211f4da41f7558)

crates/cdk-common/src/payment.rs
asmogo: done in https://github.com/cashubtc/cdk/pull/1251/commits/e57c47bfe64e70417ee1db6246904e538902025f

crates/cdk/src/wallet/receive.rs
asmogo: done https://github.com/cashubtc/cdk/pull/1251/commits/f197f6cdf28513d66bc88b7bb68e544ff61c822d

crates/cdk-payment-processor/src/proto/client.rs
asmogo: done https://github.com/cashubtc/cdk/pull/1251/commits/df7292a874b184f8c74bddb76729c652cf43beb2

crates/cashu/src/nuts/nut_custom.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct MeltQuoteCustomRequest {
+    /// Custom payment method name
+    pub method: String,
+    /// Payment request string (method-specific format)
+    pub request: String,
+    /// Currency unit
+    pub unit: CurrencyUnit,
+    /// Method-specific data (opaque JSON)
+    #[serde(default)]
+    pub data: Value,
```
thesimplekid: I think we should add #[serde(flatten)]. This way a new payment methods could define named fields like we state in nut04/05 and we would preserve the fields names. This way if a new payment method did get merged into cdk after starting as a custom one client implementations wouldn't have to change we would just define the full type in cdk.

We could make this a hashmap where the key is the field name. 
```
  #[serde(flatten)]
    data: HashMap<String, serde_json::Value>,
```

crates/cdk-payment-processor/src/proto/client.rs
```diff
@@ -99,7 +98,15 @@ impl MintPayment for PaymentProcessorClient {
 
         let settings = response.into_inner();
 
-        Ok(serde_json::from_str(&settings.inner)?)
+        Ok(cdk_common::payment::SettingsResponse {
+            bolt11: settings.bolt11,
+            custom: settings.custom,
+            mpp: settings.mpp,
+            unit: settings.unit,
+            invoice_description: settings.invoice_description,
```
thesimplekid: Should this be bolt11 specific?

crates/cdk-payment-processor/src/proto/client.rs
```diff
@@ -99,7 +98,15 @@ impl MintPayment for PaymentProcessorClient {
 
         let settings = response.into_inner();
 
-        Ok(serde_json::from_str(&settings.inner)?)
+        Ok(cdk_common::payment::SettingsResponse {
+            bolt11: settings.bolt11,
+            custom: settings.custom,
+            mpp: settings.mpp,
```
thesimplekid: Should this be bolt11 specific?

crates/cdk-mintd/example.config.toml
thesimplekid: Do we need to define them here? Or can we just get them from the grpc at runtime?

crates/cashu/src/nuts/auth/nut21.rs
```diff
@@ -5,8 +5,6 @@ use std::str::FromStr;
 
 use regex::Regex;
 use serde::{Deserialize, Serialize};
-use strum::IntoEnumIterator;
-use strum_macros::EnumIter;
```
thesimplekid, line 9: We should remove this dep if we don't use it anymore

crates/cashu/src/nuts/nut04.rs
```diff
@@ -249,6 +249,8 @@ pub enum MintMethodOptions {
         /// Mint supports setting bolt11 description
         description: bool,
     },
+    /// Custom Options
+    Custom {},
```
thesimplekid, line 255: Should this have a value flattened as I described here?

https://github.com/cashubtc/cdk/pull/1251/files#r2492890368

crates/cashu/src/nuts/nut_custom.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Amount to mint
+    pub amount: Amount,
+    /// Currency unit
+    pub unit: CurrencyUnit,
+    /// Method-specific data (opaque JSON)
+    #[serde(default)]
+    pub data: Value,
+    /// Optional description
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
```
thesimplekid: Think this is bolt11 specific and we shouldnt have it always defined if a payment type wants it they can  put it on data.

crates/cdk/src/mint/issue/mod.rs
```diff
@@ -30,6 +31,13 @@ pub enum MintQuoteRequest {
     Bolt11(MintQuoteBolt11Request),
     /// Lightning Network BOLT12 offer request
     Bolt12(MintQuoteBolt12Request),
+    /// Custom payment method request
+    Custom {
+        /// Payment method name (e.g., "paypal", "venmo")
+        method: String,
```
thesimplekid, line 39: Our `PaymentMethod` enum has a custom variant should we use that?

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -336,6 +338,117 @@ impl Mint {
...
+                method: method.to_string(),
+                request: request.clone(),
+                data: data.clone(),
+            },
+            unit.clone(),
+            payment_quote.amount,
+            payment_quote.fee,
+            unix_time() + melt_ttl,
+            payment_quote.request_lookup_id.clone(),
+            None, // Custom methods don't use options
```
thesimplekid, line 438: Should they maybe there should be a custom variant on that?

thesimplekid (commented):
Nice work. Initial comments

crates/cdk-mintd/example.config.toml
asmogo: No configuration required. These will be fetched from the payment_processor

crates/cashu/src/nuts/auth/nut21.rs
asmogo: done

crates/cashu/src/nuts/nut_custom.rs
asmogo: i will update this to be a flatten hashmap

asmogo: this was fully removed

crates/cdk-payment-processor/src/proto/payment_processor.proto
```diff
@@ -15,29 +15,49 @@ service CdkPaymentProcessor {
 message EmptyRequest {}
 
 message SettingsResponse {
-  string inner = 1;
+  string unit = 1;
+  Bolt11Settings bolt11 = 2;
+  Bolt12Settings bolt12 = 3;
+  map<string, string> custom = 4;
+}
```
asmogo, line 22: we need to keep in mind that will break compatibility with currently existing gRPC payment processors.

crates/cashu/src/nuts/nut04.rs
asmogo: is this still relevant?

crates/cashu/src/nuts/auth/nut22.rs
```diff
@@ -334,10 +332,10 @@ mod tests {
...
         let expected_protected: HashSet<ProtectedEndpoint> = HashSet::from_iter(vec![
             ProtectedEndpoint::new(Method::Post, RoutePath::Swap),
-            ProtectedEndpoint::new(Method::Get, RoutePath::MintBolt11),
-            ProtectedEndpoint::new(Method::Get, RoutePath::MintQuoteBolt11),
-            ProtectedEndpoint::new(Method::Get, RoutePath::MintQuoteBolt12),
-            ProtectedEndpoint::new(Method::Get, RoutePath::MintBolt12),
+            ProtectedEndpoint::new(Method::Get, RoutePath::Mint("bolt11".to_string())),
+            ProtectedEndpoint::new(Method::Get, RoutePath::MintQuote("bolt11".to_string())),
+            ProtectedEndpoint::new(Method::Get, RoutePath::MintQuote("bolt12".to_string())),
+            ProtectedEndpoint::new(Method::Get, RoutePath::Mint("bolt12".to_string())),
```
thesimplekid: I think we should pass in the `PaymentMethod` enum here and not just a string.

crates/cashu/src/nuts/nut04.rs
```diff
@@ -305,6 +308,79 @@ impl Settings {
...
+    pub pubkey: Option<PublicKey>,
+}
+
+/// Custom payment method mint quote response
+///
+/// This is a generic response type for custom payment methods.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(feature = "swagger", derive(utoipa::ToSchema))]
+#[serde(bound = "Q: Serialize + for<'a> Deserialize<'a>")]
+pub struct MintQuoteCustomResponse<Q> {
```
thesimplekid, line 369: How do we know these will work for any payment method? I think for ehash they don't as the share can be included in the mint quote request. 

What about something like: 
```rust
 pub struct MintQuoteCustomResponse<Q> {
      pub quote: Q,
      pub state: QuoteState,
      pub expiry: Option<u64>,
      pub amount: Option<Amount>,
      pub unit: Option<CurrencyUnit>,
      #[serde(flatten, default, skip_serializing_if = "Value::is_null")]
      pub extra: serde_json::Value,
  }
```
  With #[serde(flatten)], the extra fields merge into the parent JSON rather than being nested:
```json
  {
    "quote": "abc123",
    "state": "UNPAID",
    "amount": 1000,
    "paypal_link": "https://paypal.me/merchant",
    "paypal_email": "merchant@example.com"
  }
```
This separation also enables proper validation layering: the mint can verify the well-defined fields (amount, unit, state, etc.) while passing extra through to the gRPC payment processor for method-specific validation.

It also provides a clean upgrade path: when a payment method becomes speced, its fields can be promoted from extra to well-defined struct fields without breaking existing clients. For example, bolt12 added amount_paid and amount_issued fields that bolt11 doesn't have—with this pattern, new methods can start with fields in extra, and once standardized, those fields move to the struct proper. The JSON shape stays the same either way.

  We should apply the same pattern for melt responses (which currently reuse MeltQuoteBolt11Response for all methods).

crates/cashu/src/nuts/nut06.rs
```diff
@@ -697,7 +700,7 @@ mod tests {
         let mint_info_with_nut15 = MintInfo {
             name: Some("Test Mint".to_string()),
             nuts: Nuts::default().nut15(vec![MppMethodSettings {
-                method: crate::PaymentMethod::Bolt11,
+                method: crate::PaymentMethod::from("bolt11"),
```
thesimplekid: I think we should add constructors for the known types to avoid having to pass the string. `new_bol11`

crates/cashu/src/nuts/nut04.rs
asmogo: I think these fields will work for any payment method. I was even planing to reduce this down to quote, unit, request and probably the state, even if this is not specified in nut 4. 

gRPC payment processors with custom payment methods like "paypal" or "stripe" can use the "request" field in the `MintQuoteCustomResponse` and `MeltQuoteCustomRequest` to forward extra data to the payment processor. 

The most important argument for adding this extra field is the clean upgrade path for wallets imho.

Cargo.lock
thesimplekid, line 1: Do you know why this changed? I don't see an change to any Cargo.toml.

crates/cashu/src/nuts/nut00/mod.rs
```diff
@@ -639,46 +639,172 @@ impl<'de> Deserialize<'de> for CurrencyUnit {
...
+    pub fn is_bolt12(&self) -> bool {
+        matches!(self, Self::Known(KnownMethod::Bolt12))
+    }
+}
+
+impl Default for PaymentMethod {
+    fn default() -> Self {
+        Self::Known(KnownMethod::Bolt11)
+    }
+}
```
thesimplekid: Should we have a default for this? I think maybe not.

crates/cdk-mintd/src/lib.rs
```diff
@@ -803,6 +809,11 @@ async fn setup_authentication(
             add_endpoint(ws_protected_endpoint, &auth_settings.websocket_auth);
         }
 
+        // Custom payment method endpoints will be added dynamically after the mint is built
```
thesimplekid: Is this correct and in the right place? Does it refer to auth endpoints?

crates/cdk-payment-processor/src/proto/payment_processor.proto
thesimplekid: We should maybe start versioning the proto file.

crates/cdk/src/mint/issue/mod.rs
```diff
@@ -586,13 +631,13 @@ impl Mint {
...
-        if mint_quote.payment_method == PaymentMethod::Bolt12 && mint_quote.pubkey.is_none() {
+        if mint_quote.payment_method == "bolt12" && mint_quote.pubkey.is_none() {
             tracing::warn!("Bolt12 mint quote created without pubkey");
             return Err(Error::SignatureMissingOrInvalid);
         }
 
-        let mint_amount = match mint_quote.payment_method {
-            PaymentMethod::Bolt11 => {
+        let mint_amount = match mint_quote.payment_method.as_str() {
+            "bolt11" => {
```
thesimplekid: If we use the Custom on the payment method could we keep the enum matching over the string? https://github.com/cashubtc/cdk/pull/1251/changes#r2492948059

crates/cdk/src/mint/builder.rs
```diff
@@ -249,49 +249,105 @@ impl MintBuilder {
...
-
-            let mut mpp = self.mint_info.nuts.nut15.clone();
-
-            mpp.methods.push(mpp_settings);
-
-            self.mint_info.nuts.nut15 = mpp;
+        let method_str = method.to_string();
+
+        // Handle bolt11 methods
+        if method_str == "bolt11" {
```
thesimplekid: I would prefer to use match on the enum then else if on the string if we can. It ensures we match all patterns including ones we can add in the future.

asmogo: fixed in https://github.com/cashubtc/cdk/pull/1251/commits/ba8ca1f9a42561e039acfa39a1e0f435df0478c0

crates/cdk/src/mint/issue/mod.rs
asmogo: fixed in https://github.com/cashubtc/cdk/pull/1251/commits/ba8ca1f9a42561e039acfa39a1e0f435df0478c0

crates/cdk-mintd/src/lib.rs
asmogo: Yes, it does refer to the protected_endpoints in auth mint. These also have to be dynamically generated based on the provided payment processor methods [here](https://github.com/cashubtc/cdk/pull/1251/changes#diff-02d320f8bc111746b6a271a1e5fd84c2002aafe839a4fab3bf5e62006287aa3fR1001). 

I'll update the comment to use the correct function name.

crates/cdk-payment-processor/src/proto/payment_processor.proto
asmogo: > This was prefixed to avoid collisions. https://buf.build/docs/best-practices/style-guide/#enums https://buf.build/docs/lint/rules/#enum_value_prefix
> 
> We might want to prefix all other enums as well.

https://github.com/cashubtc/cdk/pull/1251/changes/50a77aa120f3800b87ea1fab5d8567abd5660b77 was necessary for this.

crates/cdk/src/mint/builder.rs
```diff
@@ -250,48 +250,103 @@ impl MintBuilder {
...
-            };
-
-            let mut mpp = self.mint_info.nuts.nut15.clone();
-
-            mpp.methods.push(mpp_settings);
-
-            self.mint_info.nuts.nut15 = mpp;
+        // Handle bolt11 methods
+        if method.is_bolt11() {
+            if let Some(ref bolt11_settings) = settings.bolt11 {
```
thesimplekid: A match would be better here then the if, else if else.

crates/cashu/src/nuts/auth/nut21.rs
thesimplekid, line 1: I think we need to update the matching route paths to account for the custom path otherwise they wont be projected by auth. 


Have you tried custom paths with auth? 


```
diff --git a/crates/cashu/src/nuts/auth/nut21.rs b/crates/cashu/src/nuts/auth/nut21.rs
index 15557bdd..579d4b78 100644
--- a/crates/cashu/src/nuts/auth/nut21.rs
+++ b/crates/cashu/src/nuts/auth/nut21.rs
@@ -387,6 +387,16 @@ mod tests {
         assert!(matches!(result.unwrap_err(), Error::InvalidRegex(_)));
     }

+    #[test]
+    // #[ignore = "regex matching does not include custom payment methods"]
+    fn test_matching_route_paths_custom_payment_methods() {
+        let paths = matching_route_paths("^/v1/mint/.*").unwrap();
+
+        assert!(paths.contains(&RoutePath::Mint("bolt11".to_string())));
+        assert!(paths.contains(&RoutePath::Mint("bolt12".to_string())));
+        assert!(paths.contains(&RoutePath::Mint("paypal".to_string())));
+    }
+
     #[test]
     fn test_route_path_to_string() {
         // Test that to_string() returns the correct path strings
```

asmogo: I'm assuming that there are custom routes with auth in my deployment. Will check an verify that the routes are actually authenticated. 
Looking at the code, it seems like they are not authenticated. Thanks

crates/cdk/src/mint/builder.rs
asmogo: done

crates/cashu/src/nuts/auth/nut21.rs
asmogo: This route matcher is used for serialization. 

Custom paths are checked in the axum router using the `verify_auth` function. We should move this to a middleware.

thesimplekid:
> I noticed this log message but that's probably related to the payment processor implementation
> 
> ```
> cache_post_melt_custom{method=Path("bolt11")}:post_melt_custom{method="bolt11"}:melt:finalize: cdk::mint::melt::shared: Payment lookup id changed post payment from Some(CustomId("0f86ad86-2090-448d-87ee-78d33d7f850e")) to tr_1SnRedJHYTtcuc3YHUTgv5jA
> ```

Yes, this happens if the processors returns a different lookup id after the make_payment then it did after the get quote. For some payment backends this is needed and out of our control but its better if this doesnt happen.

thesimplekid (approved):

PR #1253, opened 2025-11-04
feat: P2BK

a1denvalu3:
# Pay-to-Blinded-Key (P2BK) Implementation for Enhanced Privacy

## Overview
This PR implements [NUT-28: Pay-to-Blinded-Key](https://github.com/cashubtc/nuts/blob/main/28.md), a privacy enhancement for P2PK (NUT-11) transactions that enables "silent payments" in Cashu. P2BK allows tokens to be locked to a public key without exposing which public key they're locked to, even to the mint itself.

## Key Features

- **Enhanced Privacy**: Tokens can be sent to recipients without the mint learning the recipient's public key
- **Canonical Slot Mapping**: Supports multi-key proofs with proper slot organization
- **Backward Compatible**: Works with existing mints without requiring protocol changes on the mint side

## Technical Implementation

- **ECDH Key Derivation**: Uses Elliptic Curve Diffie-Hellman to derive shared secrets between sender and receiver
- **Blinding Operations**: Implements cryptographic blinding of public keys using random scalar multiplication
- **Key Recovery**: Enables receivers to recover the correct signing key through the shared ECDH secret
- **Validation**: Includes comprehensive tests for key derivation, blinding, and signing operations

## User-Facing Changes

- Added `use_p2bk` option to the wallet's send operations to toggle P2BK functionality
- When enabled, all P2PK transactions will use blinded keys for enhanced privacy
----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/integration_tests_pure.rs
+++ b/crates/cdk-integration-tests/tests/integration_tests_pure.rs
@@ -1599,14 +1599,22 @@ async fn test_p2bk_multi_key_receive() {
 
     let secret1 = SecretKey::generate();
     let secret2 = SecretKey::generate();
-    
+
     // Multisig 1-of-2 (data key + 1 pubkey in tags)
-    let conds = cashu::nuts::Conditions::new(None, Some(vec![secret2.public_key()]), None, Some(1), None, None).unwrap();
+    let conds = cashu::nuts::Conditions::new(
+        None,
+        Some(vec![secret2.public_key()]),
+        None,
+        Some(1),
+        None,
+        None,
+    )
+    .unwrap();
     let spending_conditions = SpendingConditions::P2PKConditions {
         data: secret1.public_key(),
         conditions: Some(conds),
     };
-    
+
     let send_amount = Amount::from(10);
 
     let prepared = wallet_sender
@@ -1638,6 +1646,6 @@ async fn test_p2bk_multi_key_receive() {
         )
         .await
         .expect("Receiver should be able to redeem P2PK token with second key");
-        
+
     assert_eq!(send_amount, received_amount);
 }


crates/cdk/src/wallet/send.rs
```diff
@@ -255,6 +260,160 @@ impl PreparedSend {
...
+    /// * `pubkey` - The primary pubkey to blind (from `data` field)
+    /// * `conditions` - The P2PK conditions containing additional pubkeys and refund keys
+    ///
+    /// # Returns
+    /// * `Result<(), Error>` - Success or error during blinding operation
+    #[instrument(skip(proofs, pubkey, conditions), level = "debug")]
+    fn apply_p2bk_blinding(
+        proofs: &mut Proofs,
+        pubkey: &cdk_common::PublicKey,
+        conditions: &Option<cdk_common::nuts::nut11::Conditions>,
```
crodas: Is there a way to do `Option<&dk_common::nuts::nut11::Conditions>`

crates/cashu/src/nuts/nut28/test_vectors.rs
thesimplekid, line 1: Nice, I like how you separated the test vectors that are spec'd from other tests we should do that for the other nuts as well.

crates/cashu/src/nuts/nut28/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+use bitcoin::hashes::sha256::Hash as Sha256;
+use bitcoin::hashes::{Hash, HashEngine};
+use bitcoin::secp256k1::Secp256k1;
+use thiserror::Error;
+
+use crate::nuts::nut01::{PublicKey, SecretKey};
+use crate::Id;
+
+// Create a static SECP256K1 context that we'll use for operations
+static SECP: LazyLock<Secp256k1<bitcoin::secp256k1::All>> = LazyLock::new(Secp256k1::new);
```
thesimplekid, line 40: Not really a comment on this PR but more note to sell. Currently we use secp from bitcoin whitch is an older version then available  if we had it as a direct dep, in newer version they removed the context so we should not have to do this. I'm not sure if we use anything else from bitcoin (besides hashes) and should keep doing it this way or we just use it to get secp and should use use that directly.

crates/cashu/src/nuts/nut00/mod.rs
```diff
@@ -827,20 +915,182 @@ impl PreMintSecrets {
...
+                        let add_blinding_scalar =
+                            ecdh_kdf(&ephemeral_key, pubkey, keyset_id, slot)?;
+
+                        // Blind the additional pubkey
+                        let blinded_pubkey = blind_public_key(pubkey, &add_blinding_scalar)?;
+                        blinded_pubkeys.push(blinded_pubkey);
+                    }
+
+                    blinded_conditions.pubkeys = Some(blinded_pubkeys);
+                }
```
thesimplekid:
```suggestion
                if let Some(pubkeys) = conditions.pubkeys {
                    let mut blinded_pubkeys: Vec<PublicKey> = Vec::with_capacity(pubkeys.len());

                    // Blind each additional pubkey using slots 1 through N
                    for (idx, pubkey) in pubkeys.iter().enumerate() {
                        let slot = (idx + current_idx) as u8;
                        if slot > 10 {
                            tracing::warn!(
                                "Too many pubkeys to blind (max 10 slots), skipping rest"
                            );
                            break;
                        }



                        // Derive blinding scalar for this pubkey
                        let add_blinding_scalar =
                            ecdh_kdf(&ephemeral_key, pubkey, keyset_id, slot)?;

                        // Blind the additional pubkey
                        let blinded_pubkey = blind_public_key(pubkey, &add_blinding_scalar)?;
                        blinded_pubkeys.push(blinded_pubkey);
                    }
                    
                    current_idx += blinded_pubkeys.len(); 
                    blinded_conditions.pubkeys = Some(blinded_pubkeys);
                }
```

I think we're updating the index in the wrong spot here and it should only be done after the loop so the refund can see it. The pubkeys gets increment by idx incrementing it in the loop causes a slot to get skipped. See the below test that fails with the current implementation but passes with the suggested fix. 

```rust
    #[test]
    fn test_slot_numbers_are_consecutive() {
        let keyset_id = Id::from_str("009a1f293253e41e").unwrap();

        // Create 3 different keys
        let key1 = SecretKey::generate().public_key();
        let key2 = SecretKey::generate().public_key();
        let key3 = SecretKey::generate().public_key();

        let ephemeral_sk = SecretKey::generate();

        let conditions = SpendingConditions::P2PKConditions {
            data: key1,
            conditions: Some(Conditions {
                pubkeys: Some(vec![key2, key3]),
                refund_keys: None,
                num_sigs: Some(1),
                sig_flag: SigFlag::SigInputs,
                locktime: None,
                num_sigs_refund: None,
            }),
        };

        let (_, blinded) =
            PreMintSecrets::apply_p2bk(conditions, keyset_id, Some(ephemeral_sk.clone())).unwrap();

        // Extract blinded keys
        let (blinded_key1, blinded_others) = match blinded {
            SpendingConditions::P2PKConditions { data, conditions } => {
                (data, conditions.unwrap().pubkeys.unwrap())
            }
            _ => panic!("Wrong type"),
        };

        // For each slot, try to derive and see if it matches
        // Slot 0 should match key1
        let r0 = ecdh_kdf(&ephemeral_sk, &key1, keyset_id, 0).unwrap();
        let test0 = blind_public_key(&key1, &r0).unwrap();
        assert_eq!(blinded_key1, test0, "Slot 0 should be key1");

        // Slot 1 should match key2
        let r1 = ecdh_kdf(&ephemeral_sk, &key2, keyset_id, 1).unwrap();
        let test1 = blind_public_key(&key2, &r1).unwrap();
        assert_eq!(blinded_others[0], test1, "Slot 1 should be key2");

        // Slot 2 should match key3 (FAILS with buggy code - it uses slot 3!)
        let r2 = ecdh_kdf(&ephemeral_sk, &key3, keyset_id, 2).unwrap();
        let test2 = blind_public_key(&key3, &r2).unwrap();
        assert_eq!(blinded_others[1], test2, "Slot 2 should be key3");
    }
```

a1denvalu3: I've fixed this issue and included the test

crates/cashu/src/nuts/nut18/payment_request.rs
```diff
@@ -104,6 +96,7 @@ pub struct PaymentRequestBuilder {
     description: Option<String>,
     transports: Vec<Transport>,
     nut10: Option<Nut10SecretRequest>,
+    nut28: Option<bool>,
```
robwoodgate: Payment request support has been removed from the spec for now

crates/cashu/src/nuts/nut28/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+    let shared = pubkey.mul_tweak(&SECP, &secret_key.as_scalar())?;
+
+    // SharedSecret exposes 32 bytes (x-coordinate)
+    let z_x: [u8; 32] = shared.x_only_public_key().0.serialize();
+
+    // Build KDF input: domain tag || x-only(Z) || keyset_id (raw bytes) || canonical_slot (1 byte) || counter (1 byte)
+    let mut engine = Sha256::engine();
+    engine.input(b"Cashu_P2BK_v1");
+    engine.input(&z_x);
+    engine.input(&keyset_id.to_bytes());
```
robwoodgate: keyset_id was removed in latest spec

crates/cdk/src/wallet/receive/saga/mod.rs
```diff
@@ -150,23 +150,65 @@ impl<'a> ReceiveSaga<'a, Initial> {
...
                     match secret.kind() {
                         Kind::P2PK => {
                             let data_key = PublicKey::from_str(secret.secret_data().data())?;
                             pubkeys.push(data_key);
                         }
                         Kind::HTLC => {
+                            // HTLC data is a hash, so it's not a pubkey.
+                            // But wait, the spec says: By implication, it also extends NUT-14 (HTLC).
+                            // But HTLC slot 0 is the hash? No! HTLC data is the hash. HTLC does not have a pubkey in `data`.
+                            // So we just add the pre-image and skip slot 0 pubkey.
```
thesimplekid: Some ai comments we should remove

crates/cdk/src/wallet/receive/saga/mod.rs
```diff
@@ -150,23 +150,65 @@ impl<'a> ReceiveSaga<'a, Initial> {
...
+                        pubkeys.append(&mut refund_keys);
+                    }
+
+                    for (i, pubkey) in pubkeys.iter().enumerate() {
+                        let slot = match secret.kind() {
+                            Kind::P2PK => i as u8,
+                            _ => (i + 1) as u8, // HTLC skips slot 0 since it's a hash, not a pubkey
+                        };
+                        if let Some(ephemeral_key) = proof.p2pk_e {
+                            #[cfg(feature = "wallet")]
```
thesimplekid: We're in the wallet mod so I don't think we need the feature flag here.

crates/cdk/src/wallet/receive/saga/mod.rs
```diff
@@ -297,6 +340,10 @@ impl<'a> ReceiveSaga<'a, Prepared> {
 
             for blinded_message in pre_swap.swap_request.outputs_mut() {
                 for signing_key in p2pk_signing_keys.values() {
+                    // Blinded messages don't have p2pk_e because they are outputs of the swap.
+                    // Wait, we are signing the outputs of the swap using p2pk_signing_keys!
+                    // If the user provided p2pk keys to be signed on output messages,
+                    // this assumes standard P2PK since output P2BK requires ephemeral keys which is handled at creation.
```
thesimplekid:
```suggestion
```

crates/cdk/src/wallet/swap/mod.rs
```diff
@@ -27,6 +27,7 @@ impl Wallet {
         amount_split_target: SplitTarget,
         input_proofs: Proofs,
         spending_conditions: Option<SpendingConditions>,
+        use_p2bk: bool,
```
thesimplekid: We should probably put the new argument at the end. It maybe worth creating a SwapOptions struct to reduce the number or augments on the swap fn, but that can be a separate pr.

crates/cdk/src/wallet/swap/mod.rs
```diff
@@ -57,9 +59,10 @@ impl Wallet {
         amount_split_target: SplitTarget,
         proofs: Proofs,
         spending_conditions: Option<SpendingConditions>,
+        use_p2bk: bool,
         include_fees: bool,
         proofs_fee_breakdown: &ProofsFeeBreakdown,
-    ) -> Result<PreSwap, Error> {
+    ) -> Result<(PreSwap, Option<Vec<crate::nuts::nut01::SecretKey>>), Error> {
```
thesimplekid: Should we add the SecretKeys to the PreSwap struct instead of a tuple?

crates/cdk/src/wallet/receive/saga/mod.rs
a1denvalu3: yes

PR #1256, opened 2025-11-06
refactor: replace proof swap with state check in error recovery

### LINKED ISSUE #1255: The reclaim may be too much, maybe hide it behind a feature flag or just remove it for now, the `unspent` should be enough.
The reclaim may be too much, maybe hide it behind a feature flag or just remove it for now, the `unspent` should be enough.

_Originally posted by @crodas in https://github.com/cashubtc/cdk/pull/1250#discussion_r2495044351_

crodas:
### Description

This commit improves the proof recovery mechanism by replacing the swap operation with a more efficient state check approach when recovering from failed operations.

Fixes #1255

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/reclaim.rs
+++ b/crates/cdk/src/wallet/reclaim.rs
@@ -1,7 +1,6 @@
 use std::collections::HashMap;
 use std::future::Future;
 
-use cdk_common::wallet::TransactionId;
 use cdk_common::{CheckStateRequest, ProofsMethods};
 use tracing::instrument;
 


crates/cdk/src/wallet/reclaim.rs
```diff
@@ -47,6 +94,11 @@ impl Wallet {
             match f.await {
                 Ok(r) => Ok(r),
                 Err(err) => {
+                    println!(
+                        "Http operation failed with \"{}\", revering  {} proofs states to UNSPENT",
```
a1denvalu3:
```suggestion
                        "Http operation failed with \"{}\", reverting  {} proofs states to UNSPENT",
```

crates/cdk/src/wallet/reclaim.rs
```diff
@@ -30,9 +35,51 @@ impl<T: ?Sized> MaybeSend for T {}
...
+                    unspent
+                        .iter()
+                        .map(|x| x.y())
+                        .collect::<Result<Vec<_>, _>>()?,
+                    state,
+                )
+                .await?;
+        }
+
+        match self.localstore.remove_transaction(transaction_id).await {
```
a1denvalu3: Why are we removing the transaction

crodas: This is a standard practice that I copied from other places, @thesimplekid. I'm fine with dropping this line. Wdyt?

thesimplekid: Where else do we remove it? My intuition is that we shouldn't have to remove it and it should never be added if the tx fails. Since I don't believe we have a failed tx state and we only add successful ones.

crates/cdk/src/wallet/reclaim.rs
```diff
@@ -47,6 +94,11 @@ impl Wallet {
             match f.await {
                 Ok(r) => Ok(r),
                 Err(err) => {
+                    println!(
+                        "Http operation failed with \"{}\", reverting  {} proofs states to UNSPENT",
+                        err,
+                        inputs.len()
+                    );
```
thesimplekid: These should be logs not prints.

thesimplekid (approved):
LGTM. ACK 91e89a542dbbbf76c02b03c327e72fab5bebf827

PR #1257, opened 2025-11-08
bring signatory up to date with the remote signer spec

lescuer97:
### Description

Changes the signatory keyset generation and updates the proto spec to follow this [pr](https://github.com/cashubtc/nuts/pull/250)

1- Generates new keysets with the correct derivations paths. 
2- returns the derivation_index as version
3- add test vectors from remote signer

-----

### Notes to the reviewers

This should not affect any already created keyset. 

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut02.rs
+++ b/crates/cashu/src/nuts/nut02.rs
@@ -80,6 +80,23 @@ impl KeySetVersion {
             _ => Err(Error::UnknownVersion),
         }
     }
+
+    /// [`KeySetVersion`] from proto value
+    pub fn from_proto_i32(value: i32) -> Result<Self, Error> {
+        match value {
+            1 => Ok(Self::Version00),
+            2 => Ok(Self::Version01),
+            _ => Err(Error::UnknownVersion),
+        }
+    }
+
+    /// [`KeySetVersion`] to proto value
+    pub fn to_proto_i32(&self) -> i32 {
+        match self {
+            Self::Version00 => 1,
+            Self::Version01 => 2,
+        }
+    }
 }
 
 impl fmt::Display for KeySetVersion {
@@ -918,6 +935,25 @@ mod test {
         let id_from_bytes = Id::from_bytes(&id_bytes).unwrap();
 
         assert_eq!(id_from_bytes, id);
+    }
+
+    #[test]
+    fn test_keyset_version_proto_mapping() {
+        assert_eq!(
+            KeySetVersion::from_proto_i32(1).unwrap(),
+            KeySetVersion::Version00
+        );
+        assert_eq!(
+            KeySetVersion::from_proto_i32(2).unwrap(),
+            KeySetVersion::Version01
+        );
+        assert!(matches!(
+            KeySetVersion::from_proto_i32(0),
+            Err(Error::UnknownVersion)
+        ));
+
+        assert_eq!(KeySetVersion::Version00.to_proto_i32(), 1);
+        assert_eq!(KeySetVersion::Version01.to_proto_i32(), 2);
     }
 
     #[test]
--- a/crates/cdk-signatory/src/proto/convert.rs
+++ b/crates/cdk-signatory/src/proto/convert.rs
@@ -345,8 +345,7 @@ impl From<crate::signatory::RotateKeyArguments> for RotationRequest {
             unit: Some(value.unit.into()),
             amounts: value.amounts,
             input_fee_ppk: value.input_fee_ppk,
-            // added a plus one to be correct with the keyset enum
-            keyset_id_type: (value.keyset_id_type.to_byte() as i32 + 1),
+            keyset_id_type: value.keyset_id_type.to_proto_i32(),
             final_expiry: value.final_expiry,
         }
     }
@@ -356,7 +355,6 @@ impl TryInto<crate::signatory::RotateKeyArguments> for RotationRequest {
     type Error = Status;
 
     fn try_into(self) -> Result<crate::signatory::RotateKeyArguments, Self::Error> {
-        let keyset_id_type: u8 = self.keyset_id_type as u8;
         Ok(crate::signatory::RotateKeyArguments {
             unit: self
                 .unit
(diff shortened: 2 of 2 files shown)


crates/cashu/src/nuts/nut00/mod.rs
```diff
@@ -577,16 +579,14 @@ pub enum CurrencyUnit {
...
-            _ => None,
-        }
+    ///  Big endian encoded integer of the first 4 bytes of the sha256 hash of the unit string.
+    pub fn derivation_index(&self) -> u32 {
+        use bitcoin::hashes::sha256;
+
+        let unit_str = self.to_string().to_lowercase();
+        let bytes = <sha256::Hash as BitcoinHash>::hash(unit_str.as_bytes());
+        // Take the first 4 bytes and convert to u32 (big endian) make sure the integer
+        u32::from_be_bytes([bytes[0], bytes[1], bytes[2], bytes[3]]) & !(1 << 31)
```
thesimplekid, line 620: Is this spec'd? @vnprc has been working on something similar. https://github.com/cashubtc/cdk/pull/1004

lescuer97: yeah, it has always been part of the remote signer spec: https://github.com/cashubtc/nuts/pull/250/files. 

The only new addition is the modulo on the sha256sum. @lollerfirst  had a look at that and gave me some thumbs up on it

crodas: @lescuer97 This will be a breaking change for the known types, how are planning to address that?

vnprc: yes I have open PRs to cdk and the NUTs repo to implement deterministic derivation paths
cdk PR https://github.com/cashubtc/cdk/pull/1181
NUTs PR https://github.com/cashubtc/nuts/pull/292

lescuer97: > yes I have open PRs to cdk and the NUTs repo to implement deterministic derivation paths cdk PR #1181 NUTs PR [cashubtc/nuts#292](https://github.com/cashubtc/nuts/pull/292)

I was not aware of this PRs. @vnprc  would you like to use mine? they seems to be equally as effective.

lescuer97: > @lescuer97 This will be a breaking change for the known types, how are planning to address that?

you are right, it would break it for library users. I could make a new version and make a deprecation warning for the older function. What do you think?

lescuer97: @crodas  I added back and depracted the original index function.

crates/cashu/src/nuts/nut00/mod.rs
```diff
@@ -578,6 +580,10 @@ pub enum CurrencyUnit {
 #[cfg(feature = "mint")]
 impl CurrencyUnit {
     /// Derivation index mint will use for unit
+    #[deprecated(
+        since = "0.14.0",
+        note = "This function is outdated; use `hashed_derivation_index` instead."
+    )]
```
thesimplekid, line 594: I don't think we should deprecate this. Instead we keep the hard coded for the ones we already have and then hash anything else.

lescuer97: I don't think that would be a good idea. It would not make cdk spec compliant.  

I don't think breaking types here is a big deal thought. 

As an alternative, I can also just do a simple hack and return Some(u32). even thought there is no option. This would mantain type consistent

crates/cashu/src/nuts/nut00/mod.rs
```diff
@@ -588,6 +594,15 @@ impl CurrencyUnit {
...
+    ///  Big endian encoded integer of the first 4 bytes of the sha256 hash of the unit string.
+    pub fn hashed_derivation_index(&self) -> u32 {
+        use bitcoin::hashes::sha256;
+
+        let unit_str = self.to_string().to_lowercase();
+        let bytes = <sha256::Hash as BitcoinHash>::hash(unit_str.as_bytes());
+        // Take the first 4 bytes and convert to u32 (big endian) make sure the integer
+        u32::from_be_bytes([bytes[0], bytes[1], bytes[2], bytes[3]]) & !(1 << 31)
+    }
 }
```
vnprc, line 622: I am concerned that this is not a hardened solution. It would derive different derivation indexes for currency unit strings that are almost but not quite the same: 
- leading and trailing whitespace
- capitalization differences
- functionally equivalent unicode character combinations

The version in this PR also doesn't include a reserved range for the existing hard coded currency units. Over a long enough timeframe this will introduce derivation path conflicts with sats, millisats, USD, and EUR.

As written in this PR, the `hashed_derivation_index` function would open the door to bugs and exploits. Resolving those issues in a backward compatible way would be more complex and difficult than just addressing the issues now before any hashed currency unit derivation paths are introduced in the wild.

I just removed the draft setting on #1181. I will rebase it to the latest commit on main. I think we should get 1181 merged and rebase this commit on top of it. Y'all should also review the NUT spec PR: https://github.com/cashubtc/nuts/pull/292. Once I get some feedback on it I would be happy to open a PR to nutshell to get the spec finalized.

lescuer97:
@vnprc  @crodas  could you take a look at this pr again pls. I think it is ready now

thesimplekid:
> @crodas could you re run the test? I think there was an error

Its not an error. Cargo.lock file needs to be updated since you added a dep.

crodas:
Absolutely!

crodas (dismissed):

crates/cashu/src/nuts/nut00/mod.rs
thesimplekid: > I don't think that would be a good idea. It would not make cdk spec compliant.

There is no merged spec so this is not an issue. 

I still have concerns about removing support of the legacy method for known units. Both nutshell and cdk use the same paths now even though it is not spec'd. So it would be possible to keep that in the spec as a legacy path. The known units keep the current method anything else goes this way. 

Right now if only the signatory db is lost from before this spec where there are legacy keysets after this change these keysets would not be recoverable and the mint could not sign or verify ecash from them.

lescuer97:
@thesimplekid  rebased. 

I will do some final tests around compatibility and it should be ready

crodas (approved):

thesimplekid (approved):

PR #1270, opened 2025-11-14
feat: add nightly flag

lescuer97:
### Description

Some times when running `just format` or `just clippy` you will get alerts about not using nightly on your toolchain. Setting the flag directly in the command avoids confusion. 

-----

### Notes to the reviewers

-----

#### CHANGED
Justfile
#### FIXED

----

### Checklist


thesimplekid:
We don't want to do this because it would enforce having the nightly toolchain installed. We have the ci task that will check nightly fmt but pr ci will pass with stable fmt https://github.com/cashubtc/cdk/pull/1260

PR #1277, opened 2025-11-15
Update Wallet::fetch_mint_info

crodas:
### Description

Make `fetch_mint_info` fetch the mint_info through the mint metadata cache instead of calling the Mint Connector directly.

This will make sure to persist the data correctly in other instances and the database

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/mod.rs
+++ b/crates/cdk/src/wallet/mod.rs
@@ -284,72 +284,66 @@ impl Wallet {
     /// Query mint for current mint information
     #[instrument(skip(self))]
     pub async fn fetch_mint_info(&self) -> Result<Option<MintInfo>, Error> {
-        match self.client.get_mint_info().await {
-            Ok(mint_info) => {
-                // If mint provides time make sure it is accurate
-                if let Some(mint_unix_time) = mint_info.time {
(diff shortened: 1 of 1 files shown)


crates/cdk/src/wallet/mod.rs
```diff
@@ -284,72 +284,66 @@ impl Wallet {
     /// Query mint for current mint information
     #[instrument(skip(self))]
     pub async fn fetch_mint_info(&self) -> Result<Option<MintInfo>, Error> {
```
thesimplekid, line 286: We should change this name and comment. Seems like it always makes an http call which it used to but not anymore. 

At one point I tried to be consistent with naming load would never make an http call, get would try to get it locally and if not make an http call, fetch would always make an http call. I don't think we strictly follow this anymore with the metadata cache though.

crodas: But we still always go http, even now.

thesimplekid: Doesn't load from cache get it from there if we have it. Not a big thing anyway we can leave it.

PR #1303, opened 2025-11-19
Wallet get unissued mint quotes

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/bolt12.rs
+++ b/crates/cdk-integration-tests/tests/bolt12.rs
@@ -463,3 +463,145 @@ async fn test_attempt_to_mint_unpaid() {
         }
     }
 }
+
+/// Tests the check_all_mint_quotes functionality for Bolt12 quotes
+///
+/// This test verifies that:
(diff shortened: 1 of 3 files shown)


crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -568,6 +570,39 @@ ON CONFLICT(id) DO UPDATE SET
...
+                state,
+                expiry,
+                secret_key,
+                payment_method,
+                amount_issued,
+                amount_paid
+            FROM
+                mint_quote
+            WHERE
+                (amount_paid > amount_issued AND amount_paid > 0)
```
thesimplekid: @vnprc This is the sql I took from your PR but I don't think it is correct. The wallet db does not know the correct amount paid it needs to ask the mint for that so I think what the wallet should consider an unpaid quote is just where the amount issue is zero.

crodas (dismissed):

PR #1304, opened 2025-11-19
Fix race condition when concurrent payments are processed for the same payment_id

crodas:
### Description

This is a continuation of #1252

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/mod.rs
+++ b/crates/cdk/src/mint/mod.rs
@@ -766,7 +766,6 @@ impl Mint {
                             wait_payment_response.payment_id
                         );
                         // This is fine - another concurrent request already processed this payment
-                        // The in-memory check at line 712 can miss this due to stale data
                     }
                     Err(e) => return Err(e.into()),
                 }


crates/cdk/src/mint/mod.rs
```diff
@@ -749,14 +749,27 @@ impl Mint {
...
+                    Ok(total_paid) => {
+                        pubsub_manager.mint_quote_payment(mint_quote, total_paid);
+                    }
+                    Err(database::Error::Duplicate) => {
+                        tracing::info!(
+                            "Payment ID {} already processed (caught race condition)",
+                            wait_payment_response.payment_id
+                        );
+                        // This is fine - another concurrent request already processed this payment
+                        // The in-memory check at line 712 can miss this due to stale data
```
thesimplekid:
```suggestion
```

crates/cdk/src/mint/ln.rs
```diff
@@ -83,15 +83,30 @@ impl Mint {
...
+                            pubsub_manager.mint_quote_payment(quote, total_paid);
+                        }
+                    }
+                    Err(database::Error::Duplicate) => {
+                        tracing::debug!(
+                            "Payment ID {} already processed (caught race condition in check_mint_quote_paid)",
+                            payment.payment_id
+                        );
+                        // This is fine - another concurrent request already processed this payment
+                        // The in-memory check at line 49 can miss this due to stale data
```
thesimplekid:
```suggestion
```

PR #1329, opened 2025-11-22
Remove max_order from keyset database schema

### LINKED ISSUE #1074: Remove max order for mint keyset info
We currently no longer use max order as we do not assume power of two keysets. However, we still have max order in the db and on the keyset info they are just set to zero. I think this is confusing and can lead to using it accidentally. Can we remove t his field and column in db?

cc @crodas

crodas:
### Description

Fixes #1074

Remove the max_order column from the keyset table across the mint database layer, replacing it with the amounts field that was already being used. This simplifies the schema by eliminating redundant data.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/integration_tests_pure.rs
+++ b/crates/cdk-integration-tests/tests/integration_tests_pure.rs
@@ -565,7 +565,11 @@ async fn test_swap_overpay_underpay_fee() {
         .expect("Failed to create test mint");
 
     mint_bob
-        .rotate_keyset(CurrencyUnit::Sat, (0..32).map(|n| 2u64.pow(n)).collect(), 1)
+        .rotate_keyset(
+            CurrencyUnit::Sat,
+            cdk_integration_tests::standard_keyset_amounts(32),
+            1,
+        )
         .await
         .unwrap();
 
@@ -640,7 +644,11 @@ async fn test_mint_enforce_fee() {
         .expect("Failed to create test mint");
 
     mint_bob
-        .rotate_keyset(CurrencyUnit::Sat, (0..32).map(|n| 2u64.pow(n)).collect(), 1)
+        .rotate_keyset(
+            CurrencyUnit::Sat,
+            cdk_integration_tests::standard_keyset_amounts(32),
+            1,
+        )
         .await
         .unwrap();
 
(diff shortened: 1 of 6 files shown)


crates/cdk-mint-rpc/src/mint_rpc_cli/subcommands/rotate_next_keyset.rs
```diff
@@ -36,19 +36,28 @@ pub async fn rotate_next_keyset(
...
 ) -> Result<()> {
+    let amounts = if let Some(amounts_str) = &sub_command_args.amounts {
+        amounts_str
+            .split(',')
+            .map(|s| s.trim().parse::<u64>())
+            .collect::<Result<Vec<u64>, _>>()?
+    } else {
+        vec![]
+    };
+
```
thesimplekid, line 47: Shouldn't proto handle turning this into a vec?

crates/cdk-integration-tests/tests/integration_tests_pure.rs
```diff
@@ -565,7 +565,7 @@ async fn test_swap_overpay_underpay_fee() {
         .expect("Failed to create test mint");
 
     mint_bob
-        .rotate_keyset(CurrencyUnit::Sat, 32, 1)
+        .rotate_keyset(CurrencyUnit::Sat, (0..32).map(|n| 2u64.pow(n)).collect(), 1)
```
thesimplekid: We do this in multiple places both in tests and app code. We should make it a fn so we can comment what it is an reuse it.

crodas: Fixed in fd24e689a54c7f17363e39ce32a175199c8d6388

crates/cdk-mint-rpc/src/mint_rpc_cli/subcommands/rotate_next_keyset.rs
crodas: Proto takes a vector. Working with a string with comma-separated numbers would be weird.  Am I missing something?

thesimplekid: Oh this is the cli that we read a string with comma separation. 

think we can have clap so the parsing for use though

```rust
#[arg(short, long, value_delimiter = ',')]
amounts: Vec<u64>,
```

But we can keep it as it is I don't mind.

thesimplekid:
Seems like there is a doc ci issue @crodas

crodas:
@thesimplekid just pushed a fix.

PR #1337, opened 2025-11-25
Simplify increment_issued_quote implementation

### LINKED ISSUE #1066: Remove amount issued check from database that is responsibility of the mint
It was brought up in the dev call that while this is a light weight check, it is not the responsibility of the storage layer to do checks like this and it should be done in the mint. It currently is done in the mint and this is a redundant check that should be removed. 

https://github.com/cashubtc/cdk/blob/f2f54253958fb93f715cca12da7ce303bf9efb64/crates/cdk-sql-common/src/mint/mod.rs#L693-L716

cc @asmogo @crodas

crodas:
### Description

Fixes #1066

Removed explicit overflow and over-issuance checks from application code

Refactors the increment_issued_quote method to perform atomic SQL update instead of read-modify-write pattern. This simplifies the code by using SQL's atomic increment operation (amount_issued + :amount_issued) rather than fetching, calculating, and updating separately.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-common/src/database/mint/test/mint.rs
thesimplekid, line 1: When removing tests we should document why its safe to remove them. Since we are not longer expecting the DB to check this do we have tests in the mint that check these scenarios?

crodas:
Closed in favour of #1446

PR #1344, opened 2025-11-25
fix: fees

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk/src/wallet/proofs.rs
```diff
@@ -438,53 +438,67 @@ impl Wallet {
...
+                return Err(Error::InsufficientFunds);
+            }
+
+            let additional = Wallet::select_proofs(
+                remaining_amount,
+                remaining_proofs,
+                active_keyset_ids,
+                fees_and_keyset_amounts,
+                false,
+            )?;
```
davidcaseria, line 492: Do we need to do the full proof selection here? Or is there a smarter way to add what we need?

PR #1408, opened 2025-12-11
Self hosted

thesimplekid:
### Description

-----
- [x] ffi tests should be on self hosted
- [ ] regtest tests are not taking full advantage of caching this should be fixed but should not block merge if they seem to run okay.
### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/src/init_regtest.rs
+++ b/crates/cdk-integration-tests/src/init_regtest.rs
@@ -420,10 +420,48 @@ pub async fn start_regtest_end(
             // Wait for CLN to sync after previous block generation to ensure UTXOs are available
             cln_client.wait_chain_sync().await?;
 
-            cln_client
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -306,116 +218,71 @@ jobs:
...
-    timeout-minutes: 30
-    needs: pre-commit-checks
-    strategy:
-      fail-fast: true
-      matrix:
-        build-args: [
-            # Core library - all features EXCEPT swagger (which breaks MSRV)
-            '-p cdk --features "mint,wallet,auth,nostr,bip353,tor,prometheus"',
+  # msrv-build:
+  #   name: "MSRV build"
```
thesimplekid: these should be uncommented and run on self hosted

PR #1414, opened 2025-12-13
Batch Minting Implementation (NUT-XX)

vnprc:
## Summary

This PR implements batch minting functionality as specified in [NUT-XX](https://github.com/cashubtc/nuts/pull/323), allowing clients to mint multiple quotes in a single request. The implementation supports both bolt11 and bolt12 payment methods with comprehensive validation and spending condition support.

**Key Changes:**
- Added batch minting endpoints: `POST /v1/mint/{method}/batch` and `POST /v1/mint/{method}/check`
- Implemented payment method validation ensuring all quotes in a batch use the same payment method
- Added NUT-20 spending condition support for batch minting with signature validation
- Extended protocol types with new batch-specific structs and error codes
- Implemented bolt12 batch minting with signature validation and lifecycle validation

### Protocol Extensions (NUT-XX)

**New Types:**
- `BatchMintSettings` in NUT-06 for capability advertisement
- `BatchMintRequest` / `BatchQuoteStatusRequest` / `BatchQuoteStatusResponse` for batch operations
- Added `MintBolt11Batch` and `MintBolt12Batch` to NUT-19 path enumeration

**Error Codes:**
- `BatchEmpty` - empty quote array
- `BatchSizeExceeded` - exceeds 100 quote limit
- `DuplicateQuoteIds` - duplicate quotes in single request
- `PaymentMethodMismatch` - quotes use different payment methods
- `EndpointMethodMismatch` - quote method doesn't match URL path
- `SignatureInvalid` / `SignatureMissing` / `SignatureUnexpected` - NUT-20 validation errors
- `BatchUnitMismatch` - quotes use different currency units

### API Endpoints

**Batch Mint:**
```
POST /v1/mint/{method}/batch
```
- Mints multiple quotes atomically
- Validates all quotes use same payment method as URL path
- Supports bolt11 and bolt12 with spending conditions
- Returns unified `MintResponse` with all blind signatures

**Batch Quote Status:**
```
POST /v1/mint/{method}/check
```
- Checks status of multiple quotes in one call
- Omits unknown quotes from response (per spec)
- Returns heterogeneous response array supporting both bolt11 and bolt12

### Validation Rules

**Request Validation:**
1. Quotes array must not be empty
2. Maximum 100 quotes per batch
3. No duplicate quote IDs within a batch
4. All quotes must use same payment method
5. All quotes must use same currency unit
6. Payment method must match URL path (`{method}`)
7. Signature array (if present) must match quote count

**Quote Lifecycle Validation:**
- All quotes must be in `PAID` state
- Quotes must not be expired
- Quotes must not already be issued
- For bolt12: validates pubkey ownership through signature verification

**Spending Condition Support (NUT-20):**
- Validates signatures when quotes have pubkey locks
- Rejects signatures on unlocked quotes
- Requires signatures for all locked quotes in batch
- Verifies each signature against corresponding quote's pubkey

### Bolt12 Specifics

The bolt12 implementation includes:
- Signature validation for quotes with pubkey locks (NUT-20)
- Support for multiple quotes locked to different pubkeys in a single batch
- Proper route validation (quotes with pubkeys must come through `/mint/bolt12/batch`)

### Database Changes

**Wallet Database:**
- Added `add_mint_quote()` method for persisting quotes

**Mint Database:**
- Quote retrieval by quote ID (standard `get_mint_quote()`)
- Batch operations retrieve multiple quotes sequentially within a transaction

### Architecture

The implementation follows a unified pattern:
1. Request validation (size, duplicates, payment method)
2. Quote retrieval and lifecycle validation
3. Signature verification (if NUT-20 conditions present)
4. Atomic minting through existing single-mint infrastructure
5. Unified response construction

Core logic is shared between bolt11 and bolt12 through `process_batch_mint_request()`, with payment-method-specific validation extracted into helpers.

## Testing

All tests passing:

```bash
# Integration tests (pure)
OPENSSL_LIB_DIR=$(pkg-config --variable=libdir openssl) \
  RUSTFLAGS="-C link-arg=-Wl,-rpath,$OPENSSL_LIB_DIR" \
  CDK_TEST_DB_TYPE=memory \
  cargo test -p cdk-integration-tests --test integration_tests_pure -- --test-threads 1 batch_mint

# Wallet batch mint tests
cargo test -p cdk --tests wallet_batch_mint
```

**Test Coverage:**
- Basic batch minting with bolt11 quotes
- Batch minting with spending conditions (NUT-20)
- Multiple spending conditions in single batch
- Quote status checking
- Error cases: empty batch, oversized batch, duplicate quotes, payment method mismatches
- Bolt12 signature validation for locked quotes
- Double-spend prevention (quote reuse)

## Migration Guide

**For Mint Operators:**
- Batch minting is opt-in via NUT-XX capability advertisement
- Maximum batch size is currently hardcoded to 100 quotes (TODO: make configurable)
- Specify supported payment methods in `BatchMintSettings.methods`
- No database migrations required

**For Wallet Developers:**
- Use `POST /v1/mint/{method}/batch` instead of multiple single-mint calls
- Maximum 100 quotes per batch request (currently hardcoded)
- Handle new error codes for batch validation failures
- For bolt12: include signatures for all quotes with pubkeys

## Breaking Changes

None. This is a purely additive feature behind new endpoints.

## Related

- Implements NUT-XX batch minting specification
- Extends NUT-20 spending conditions to batch context
- Bolt12 integration follows NUT-25 patterns

### DIFF
--- a/crates/cdk-integration-tests/tests/batch_mint.rs
+++ b/crates/cdk-integration-tests/tests/batch_mint.rs
@@ -1,9 +1,14 @@
 //! Batch Mint Tests [NUT-XX]
 //!
-//! This file contains tests for the batch minting functionality per NUT-XX specification:
-//! https://github.com/cashubtc/nuts/issues/XX
+//! This file contains tests for the batch minting functionality per NUT-XX specification.
 //!
 //! ## Test Coverage
+//!
+//! **Protocol Parsing:**
+//! - test_batch_mint_parses_unlocked_quotes: Parse unlocked quote requests
+//! - test_batch_mint_parses_locked_quotes: Parse locked quote requests
+//! - test_batch_mint_parses_mixed_locked_unlocked: Parse mixed requests
+//! - test_batch_mint_round_trip_serialization: Serialization round-trip
 //!
 //! **Handler Validation Tests:**
 //! - test_batch_mint_handler_rejects_empty_quotes: Empty quote lists are rejected
@@ -14,28 +19,57 @@
 //! **NUT-20 Signature Validation:**
 //! - test_batch_mint_rejects_invalid_nut20_signatures: Invalid signatures are rejected
 //! - test_batch_mint_rejects_signature_without_pubkey: Signature on unlocked quote is rejected
+//! - test_batch_mint_rejects_reordered_outputs_signature_scope: Reordered outputs fail signature validation
+//! - test_batch_mint_rejects_signature_missing_blinded_message: Missing blinded messages fail signature validation
+//! - test_batch_mint_rejects_missing_signature_for_locked_quote: Locked quotes without signatures are rejected
 //!
 //! **Quote Validation:**
 //! - test_batch_mint_rejects_unpaid_quotes: Unpaid quotes are rejected
 //! - test_batch_mint_enforces_single_payment_method: All quotes must have same payment method
 //! - test_batch_mint_enforces_single_currency_unit: All quotes must have same unit
+//! - test_batch_mint_rejects_output_unit_mismatch: Outputs must match quote currency unit
 //! - test_batch_mint_validates_url_path_payment_method: Quotes must match endpoint payment method
+//! - test_batch_mint_rejects_unparsable_quote_id: Invalid quote IDs are rejected
+//! - test_batch_mint_one_valid_one_unknown_quote: Batch fails atomically with unknown quotes
 //!
-//! **Protocol Parsing:**
-//! - test_batch_mint_parses_unlocked_quotes: Parse unlocked quote requests
-//! - test_batch_mint_parses_locked_quotes: Parse locked quote requests
-//! - test_batch_mint_parses_mixed_locked_unlocked: Parse mixed requests
-//! - test_batch_mint_round_trip_serialization: Serialization round-trip
+//! **Amount Validation:**
+//! - test_batch_mint_bolt11_amount_mismatch: Bolt11 total amounts must match
+//! - test_batch_mint_bolt11_quote_amounts_length_mismatch: Quote amounts array length must match
+//! - test_batch_mint_bolt11_quote_amount_mismatch_individual: Individual quote amounts must match quote values
+//! - test_batch_mint_empty_outputs_array: Empty outputs are rejected
+//!
+//! **BOLT12 Support:**
+//! - test_batch_mint_bolt12_requires_quote_amounts: BOLT12 requires quote_amounts field
+//! - test_batch_mint_bolt12_rejects_over_requested_amount: Cannot mint more than available
+//! - test_batch_mint_bolt12_total_mismatch: Output amounts must match requested amounts
+//! - test_batch_mint_bolt12_incremental_minting: BOLT12 quotes support multiple batch operations
+//!
+//! **Batch Check Endpoint:**
+//! - test_batch_check_returns_only_known_quotes: Unknown quotes omitted from response
+//! - test_batch_check_empty_quotes_array: Empty quotes rejected
+//! - test_batch_check_all_unknown_quotes: Returns empty array when all quotes unknown
+//! - test_batch_check_mixed_paid_unpaid_states: Returns quotes with their respective states
+//! - test_batch_check_rejects_duplicate_quote_ids: Duplicate quote IDs rejected
+//! - test_batch_check_rejects_over_limit: Batch size limit enforced
+//!
+//! **Spending Conditions (NUT-11 P2PK and NUT-14 HTLC):**
+//! - test_batch_mint_with_p2pk_spending_conditions: P2PK conditions preserved through batch mint
+//! - test_batch_mint_with_htlc_spending_conditions: HTLC conditions preserved through batch mint
+//! - test_batch_mint_with_mixed_spending_conditions: Mixed P2PK and HTLC conditions
 //!
 //! **Happy Path Tests:**
-//! - test_batch_mint_successful_two_quotes_happy_path: Successful batch mint with two paid quotes
+//! - test_batch_mint_successful_two_quotes_happy_path: Successful batch mint combining two quotes
+//! - test_batch_mint_locked_quote_valid_signature: Successful locked quote with valid NUT-20 signature
+//! - test_batch_mint_unlocked_quotes_with_null_signatures: Unlocked quotes with explicit null signatures
 //!
-//! **Coverage Areas:**
-//! - Validation: empty batches, duplicates, size limits, state requirements
-//! - Signature handling: NUT-20 signature verification, pubkey validation
-//! - Quote constraints: payment method/unit consistency, endpoint validation
-//! - Protocol compliance: JSON parsing and serialization
-//! - Success scenarios: batch minting with multiple paid quotes
+//! **Coverage Summary:**
+//! - Validation: empty batches, duplicates, size limits, state requirements, amount balancing
+//! - Signature handling: NUT-20 signature verification, pubkey validation, output ordering
+//! - Quote constraints: payment method/unit consistency, endpoint validation, atomic failure
+//! - Protocol compliance: JSON parsing, serialization, check endpoint behavior
+//! - BOLT12: partial minting, incremental operations, amount validation
+//! - Spending conditions: P2PK, HTLC, and mixed condition support
+//! - Success scenarios: batch minting with multiple quotes, locked/unlocked quotes
 
 use std::collections::{HashMap, HashSet};
 use std::str::FromStr;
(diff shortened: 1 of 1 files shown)


crates/cdk-common/src/mint.rs
```diff
@@ -746,3 +746,205 @@ mod offer_serde {
...
+    pub fn state(&self) -> MintQuoteState {
+        self.state
+    }
+}
+
+impl<Q> From<MintQuoteBolt12Response<Q>> for MintQuoteBolt12BatchStatusResponse<Q> {
+    fn from(value: MintQuoteBolt12Response<Q>) -> Self {
+        Self::from_quote(value)
+    }
+}
```
thesimplekid: Requests and responses defined in the nuts should be in the cashu crate in a mod for the nut, nutxx for now.

crates/cashu/src/nuts/nut06.rs
```diff
@@ -270,6 +271,34 @@ impl MintInfo {
...
+        }
+    }
+}
+
+impl BatchMintSettings {
+    /// Returns true when no batch capabilities should be advertised
+    pub fn is_empty(&self) -> bool {
+        self.methods.is_empty()
+    }
+}
```
thesimplekid: Move this to its own mod nutxx.

crates/cdk-axum/src/router_handlers.rs
```diff
@@ -643,6 +693,199 @@ pub(crate) async fn post_restore(
...
+        "bolt12" => PaymentMethod::Bolt12,
+        _ => return Err(into_response(cdk::error::Error::UnsupportedPaymentMethod)),
+    };
+
+    #[cfg(feature = "auth")]
+    {
+        let route = match payment_method {
+            PaymentMethod::Bolt11 => RoutePath::MintBolt11,
+            PaymentMethod::Bolt12 => RoutePath::MintBolt12,
+            PaymentMethod::Custom(_) => unreachable!(),
```
thesimplekid: Would rather just return an error here then use unreachable.

crates/cdk-axum/src/router_handlers.rs
```diff
@@ -643,6 +693,199 @@ pub(crate) async fn post_restore(
...
+        return Err(into_response(cdk::error::Error::BatchSizeExceeded));
+    }
+
+    // Validation: no duplicate quotes
+    let mut unique_quotes = std::collections::HashSet::new();
+    for quote_id in &payload.quote {
+        if !unique_quotes.insert(quote_id.clone()) {
+            return Err(into_response(cdk::error::Error::DuplicateQuoteIdInBatch));
+        }
+    }
```
thesimplekid: We should not do this verification here in axum. We should let cdk handle it.

crates/cdk-axum/src/router_handlers.rs
```diff
@@ -643,6 +693,199 @@ pub(crate) async fn post_restore(
...
+
+    // Validation: signature array (if present) matches quotes length
+    if let Some(ref signatures) = payload.signature {
+        if signatures.len() != payload.quote.len() {
+            return Err(into_response(
+                cdk::error::Error::BatchSignatureCountMismatch,
+            ));
+        }
+    }
```
thesimplekid: The axum layer should be very thin just passing the request to cdk where it is validated.

crates/cdk-axum/src/router_handlers.rs
```diff
@@ -643,6 +693,199 @@ pub(crate) async fn post_restore(
...
+        "bolt11" => PaymentMethod::Bolt11,
+        "bolt12" => PaymentMethod::Bolt12,
+        _ => return Err(into_response(cdk::error::Error::UnsupportedPaymentMethod)),
+    };
+    #[cfg(feature = "auth")]
+    {
+        let route = match payment_method {
+            PaymentMethod::Bolt11 => RoutePath::MintBolt11,
+            PaymentMethod::Bolt12 => RoutePath::MintBolt12,
+            PaymentMethod::Custom(_) => unreachable!(),
```
thesimplekid: Would prefer a real error.

crates/cdk-axum/src/router_handlers.rs
```diff
@@ -643,6 +693,199 @@ pub(crate) async fn post_restore(
...
+        let quote_id = match QuoteId::from_str(quote_id_str) {
+            Ok(id) => id,
+            Err(_) => {
+                // Invalid quote ID format, skip
+                continue;
+            }
+        };
+
+        // Check quote status and get quote response
+        match state.mint.check_mint_quote(&quote_id).await {
```
thesimplekid: We shouldn't loop here, instead we should add a `check_mint_quotes` fn to cdk and just pass the request tot that. `check_mint_quote` should then just become a very thin wrapper around `check_mint_quotes` without its own logic.

crates/cdk-common/src/error.rs
```diff
@@ -313,6 +313,48 @@ pub enum Error {
...
+    BatchInvalidSignature,
+    /// Signature provided for unlocked quote (no pubkey)
+    #[error("Signature provided for unlocked quote")]
+    BatchUnexpectedSignature,
+    /// Batch size exceeds limit (max 100 quotes)
+    #[error("Batch size exceeds limit (max 100 quotes)")]
+    BatchSizeExceeded,
+    /// Signature missing for locked quote
+    #[error("Signature missing for locked quote")]
+    BatchSignatureMissing,
```
thesimplekid: I don't think we need specific batch errors for many of these and should reuse errors we already have.

crates/cdk-common/src/error.rs
```diff
@@ -437,18 +479,33 @@ pub struct ErrorResponse {
     /// Human readable description
     #[serde(default)]
     pub detail: String,
+    /// Optional quote identifier
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub quote: Option<String>,
```
thesimplekid, line 469: I don't think we should add this as it quickly becomes inaccurate as soon as two quotes in a batch have an issue is it is not useful. But more of a discussion for spec.

crates/cdk-common/src/mint.rs
```diff
@@ -746,3 +746,205 @@ mod offer_serde {
...
+    /// Create a batch entry from a Bolt11 quote
+    pub fn from_bolt11(response: MintQuoteBolt11Response<String>) -> Self {
+        Self::Bolt11(response)
+    }
+
+    /// Create a batch entry from a Bolt12 quote
+    pub fn from_bolt12(response: MintQuoteBolt12Response<String>) -> Self {
+        let batch_response = MintQuoteBolt12BatchStatusResponse::from(response);
+        Self::Bolt12(batch_response)
+    }
```
thesimplekid: Do we need these we have the `From` impl.

crates/cdk-common/src/mint.rs
```diff
@@ -746,3 +746,205 @@ mod offer_serde {
...
+
+    /// Quote state (UNPAID, PAID, ISSUED)
+    pub fn state(&self) -> MintQuoteState {
+        match self {
+            Self::Bolt11(response) => response.state,
+            Self::Bolt12(response) => response.state(),
+        }
+    }
```
thesimplekid: Is this used anywhere? I don't think it is and we should remove it. State is not a real think on bolt12 its for backwards compatibility so we should not surface it unless we have to.

crates/cdk-integration-tests/tests/batch_mint.rs
```diff
@@ -0,0 +1,1147 @@
...
+    );
+
+    let localstore = Arc::new(database);
+    let mut mint_builder = MintBuilder::new(localstore.clone());
+
+    mint_builder = mint_builder
+        .with_name("test mint".to_string())
+        .with_description("test mint".to_string());
+
+    mint_builder
```
thesimplekid: Can't we use some of our existing fake mints instead of creating a new one.

crates/cdk-redb/src/wallet/mod.rs
```diff
@@ -315,6 +316,24 @@ impl WalletDatabase for WalletRedbDatabase {
...
+            table
+                .insert(
+                    quote.id.as_str(),
+                    serde_json::to_string(&quote).map_err(Error::from)?.as_str(),
+                )
+                .map_err(Error::from)?;
+        }
+        write_txn.commit().map_err(Error::from)?;
+        Ok(())
+    }
```
thesimplekid: We already have this fn on line 760

crates/cdk-common/src/database/wallet.rs
```diff
@@ -139,6 +139,9 @@ pub trait Database: Debug {
     /// Get mint quote from storage
     async fn get_mint_quote(&self, quote_id: &str) -> Result<Option<WalletMintQuote>, Self::Err>;
 
+    /// Add mint quote to storage
+    async fn add_mint_quote(&self, quote: WalletMintQuote) -> Result<(), Self::Err>;
+
```
thesimplekid: This should not be added here we already have it on the tx. Line 59

crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -231,9 +231,9 @@ where
         query(
                 r#"
     INSERT INTO mint_quote
-    (id, mint_url, amount, unit, request, state, expiry, secret_key, payment_method, amount_issued, amount_paid)
+    (id, mint_url, amount, unit, request, state, expiry, secret_key, payment_method, amount_issued, amount_paid, spending_condition)
     VALUES
-    (:id, :mint_url, :amount, :unit, :request, :state, :expiry, :secret_key, :payment_method, :amount_issued, :amount_paid)
+    (:id, :mint_url, :amount, :unit, :request, :state, :expiry, :secret_key, :payment_method, :amount_issued, :amount_paid, :spending_condition)
```
thesimplekid: Why do we need to store the spending condition? The existing method it gets passed at runtime as an argument to the mint fn.

crates/cdk/src/mint/issue/mod.rs
```diff
@@ -547,156 +553,420 @@ impl Mint {
...
+
+                for available in &mintable_per_quote {
+                    let minted = std::cmp::min(*available, remaining);
+                    per_quote.push(minted);
+                    remaining = remaining.checked_sub(minted).ok_or(Error::AmountOverflow)?;
+                }
+
+                per_quote
+            }
+            _ => unreachable!(),
```
thesimplekid: Use a real error.

crates/cdk/src/mint/issue/mod.rs
```diff
@@ -547,156 +553,420 @@ impl Mint {
...
+
+        let output_unit = output_unit.ok_or(Error::UnsupportedUnit)?;
+        ensure_cdk!(output_unit == unit, Error::UnsupportedUnit);
 
         let operation = Operation::new_mint();
+        let associated_quote = if origin == MintRequestOrigin::Single {
+            quote_ids.first().cloned()
+        } else {
+            None
+        };
```
thesimplekid: We should still add the quote, we know how much each quote is for so we can split the message up per quote.

crates/cdk/src/mint/issue/mod.rs
```diff
@@ -547,156 +553,420 @@ impl Mint {
...
+                let mut remaining = outputs_amount;
+                let mut per_quote = Vec::with_capacity(mintable_per_quote.len());
+
+                for available in &mintable_per_quote {
+                    let minted = std::cmp::min(*available, remaining);
+                    per_quote.push(minted);
+                    remaining = remaining.checked_sub(minted).ok_or(Error::AmountOverflow)?;
+                }
+
+                per_quote
```
thesimplekid: I think this could be problematic if a large payment came in while the wallet is making the request it did not account for and that is the first quote in the list then all the outputs would be allocated minting for that quote and leaving none for the rest of the quotes. Worse the wallet has no way of knowing this even happened without further checks of mint quotes. 

Should the wallet state in the request how much it expects for each quote.

crates/cdk-axum/src/router_handlers.rs
vnprc: > check_mint_quote should then just become a very thin wrapper around check_mint_quotes without its own logic.

`check_mint_quote` returns errors, `check_mint_quotes` ignores them. So I left it as is, with the batch function calling the individual function and discarding errors. However, this function has been moved out of the axum module, as directed.

crates/cdk-common/src/error.rs
vnprc: Agree it's a spec discussion but if you have 100 quotes in a batch and 1 of them fails validation it could be very difficult to identify the failing quote. This is not a perfect solution but it helps in the worst case.

crates/cdk-common/src/mint.rs
vnprc: It's used in wallet/issue/batch.rs:337 in the `execute_batch_mint` function:

```
            for status in &batch_status.0 {
                match status.state() {
                    MintQuoteState::Paid => (),
                    MintQuoteState::Unpaid => return Err(Error::UnpaidQuote),
                    MintQuoteState::Issued => (),
                }
            }
```

If you want to remove it how should I validate the bolt12 quote is mintable?

crates/cdk-sql-common/src/wallet/mod.rs
vnprc: good catch, i was not suspicious enough of the llm output

crates/cdk/src/mint/issue/mod.rs
vnprc: yep. good idea

PR #1441, opened 2025-12-17
fix: race condition in connection pool causing "no such table" errors

thesimplekid:
### Description

-----
This fixes a race condition in the connection pool that caused intermittent "no such table: mint_quote_payments" errors, particularly on slower CI runners.
## Root Cause
The `get_timeout()` function in `pool.rs` had a TOCTOU (time-of-check-time-of-use) race condition. When acquiring a connection:

1. Thread A pops a connection from the queue
2. Thread A releases the mutex BEFORE incrementing `in_use` counter
3. Thread B acquires the mutex, sees empty queue, releases mutex
4. Thread B checks `in_use < max_size` (0 < 1) - passes because Thread A hasn't incremented yet
5. Thread B creates a NEW connection

For in-memory SQLite, each `Connection::open_in_memory()` creates a SEPARATE, INDEPENDENT database. Migrations only run on the first connection created during `SQLMintDatabase::new()`. When the race condition triggers, additional connections point to fresh databases without any tables - hence "no such table: mint_quote_payments".

## The Fix

Move `increment_connection_counter()` to BEFORE releasing the mutex in both code paths:

1. When popping an existing resource from the queue
2. When creating a new resource

This ensures other threads immediately see the updated count and wait instead of creating duplicate connections. Added error handling to decrement the counter if resource creation fails, preserving the behavior from PR #1367.
      
image

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sql-common/src/pool.rs
+++ b/crates/cdk-sql-common/src/pool.rs
@@ -237,9 +237,7 @@ where
                 .wait_timeout(resources, timeout)
                 .map_err(|_| Error::Poison)
                 .and_then(|(lock, timeout_result)| {
-                    #[allow(unreachable_code)]
                     if timeout_result.timed_out() {
-                        panic!("Debug for a backtrace");
                         tracing::warn!(
                             "Timeout waiting for the resource (pool size: {}). Waited {} ms",
                             self.max_size,


crates/cdk-sql-common/src/pool.rs
```diff
@@ -194,8 +194,12 @@ where
         loop {
             if let Some((stale, resource)) = resources.pop() {
```
crodas:
```suggestion
            while let Some((stale, resource)) = resources.pop() {
```

This would be the correct intent; otherwise, it would try the last element and immediately wait if that resource is stale. This loop would keep retrying if there are any other element in the resource queue

crates/cdk-sql-common/src/pool.rs
```diff
@@ -207,25 +210,36 @@ where
...
             }
 
             resources = self
                 .waiter
                 .wait_timeout(resources, timeout)
                 .map_err(|_| Error::Poison)
                 .and_then(|(lock, timeout_result)| {
+                    #[allow(unreachable_code)]
                     if timeout_result.timed_out() {
+                        panic!("Debug for a backtrace");
```
crodas:
```suggestion
```

Remove the debug panic

crates/cdk-sql-common/src/pool.rs
```diff
@@ -207,25 +210,36 @@ where
...
+                    }
+                }
             }
 
             resources = self
                 .waiter
                 .wait_timeout(resources, timeout)
                 .map_err(|_| Error::Poison)
                 .and_then(|(lock, timeout_result)| {
+                    #[allow(unreachable_code)]
```
crodas:
```suggestion
```

This is not longer needed

PR #1446, opened 2025-12-18
Mint: Add Acquired<T> type for compile-time row lock enforcement

### LINKED ISSUE #1404: Do we still check for valid state transitions in db or we expect that in application
Do we still check for valid state transitions in db or we expect that in application

_Originally posted by @thesimplekid in https://github.com/cashubtc/cdk/pull/1375#discussion_r2585379758_

crodas:
### Description

Fixes #1404 and #1066

Add a LockedRows tracking system to enforce that resources must be read (and locked) before being modified within a transaction. This prevents race conditions and ensures data consistency.

Key changes:
- Add locked_row module with RowId enum and LockedRows tracker
- Change add_mint_quote to return the quote and lock it
- Change increment_mint_quote_amount_paid to take quote by value instead of quote_id, returning the updated quote
- Change increment_mint_quote_amount_issued to take quote by value instead of quote_id, returning the updated quote
- Add get_proofs_states to ProofsTransaction trait
- Auto-lock quotes when fetched via get_mint_quote, get_melt_quote, and related lookup methods
- Add state transition checks before updating proof states in melt and swap sagas

This ensures the database layer can reject modifications to resources that weren't properly read first, forcing callers to always work with fresh data from the database.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/melt/shared.rs
+++ b/crates/cdk/src/mint/melt/shared.rs
@@ -278,7 +278,10 @@ pub async fn load_melt_quotes_exclusively(
         .get_melt_quote(quote_id)
         .await
         .map_err(|e| match e {
-            database::Error::Locked => database::Error::Duplicate,
+            database::Error::Locked => {
+                tracing::warn!("Quote {quote_id} is locked by another process");
+                database::Error::Duplicate
+            }
             e => e,
         })?
         .ok_or(Error::UnknownQuote)?;
@@ -288,7 +291,10 @@ pub async fn load_melt_quotes_exclusively(
         tx.get_melt_quotes_by_request_lookup_id(request_lookup_id)
             .await
             .map_err(|e| match e {
-                database::Error::Locked => database::Error::Duplicate,
+                database::Error::Locked => {
+                    tracing::warn!("Quotes with request_lookyup_id {request_lookup_id} is locked by another process");
+                    database::Error::Duplicate
+                }
                 e => e,
             })?
     } else {


crates/cdk-sql-common/src/mint/mod.rs
thesimplekid, line 1: I do not think this method of introducing a testing feature and silent no ops in production is correct. It add more complicity to the code base a introduces a divergence between testing code and  production function. There is a place for checks like this but in those cases we should use built in rust patterns like debug asserts not build our own. 

If the goal is to make sure we use our own traits and fns correctly we should do that through the type system, by introducing a locked quote type or similar. If the goal is to better test the mint for race conditions we should add more integration tests where we hold quotes and proofs in pending states where they should be locked and ensure the mint acts correctly.

crates/cdk-common/src/database/mint/test/mod.rs
```diff
@@ -57,56 +56,6 @@ where
     keyset_id
 }
 
-/// State transition test
-pub async fn state_transition<DB>(db: DB)
-where
-    DB: Database<crate::database::Error> + KeysDatabase<Err = crate::database::Error>,
-{
```
thesimplekid, line 64: Why was this test removed and there a similar test elsewhere, as this is something that should be tests?

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -1000,37 +1014,14 @@ where
...
-        };
-
-        // Calculate new amount_paid with overflow check
-        let new_amount_paid = current_amount_paid
-            .checked_add(amount_paid)
-            .ok_or_else(|| database::Error::AmountOverflow)?;
+        let current_amount_paid = quote.amount_paid();
+        let new_amount_paid = quote
+            .increment_amount_paid(amount_paid)
+            .map_err(|_| Error::AmountOverflow)?;
```
thesimplekid: I think the increment should not happen here but in the mint and we change the db fn from increment to set amount paid.

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -1070,55 +1061,18 @@ where
...
-            current_amount_paid
-                .checked_sub(new_amount_issued)
-                .ok_or(Error::Internal("Over-issued not allowed".to_owned()))?;
-
-            new_amount_issued
-        };
+    ) -> Result<(), Self::Err> {
+        let new_amount_issued = quote
+            .increment_amount_issued(amount_issued)
+            .map_err(|_| Error::AmountOverflow)?;
```
thesimplekid: Same comment as amount paid

thesimplekid: Another option would be just to remove these two fns for a more general update_mint_quote fn that can do both.

crates/cdk-common/src/database/mint/test/mod.rs
crodas: I think we test in the mint codebase, and we agreed to move this test off the database.

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -970,192 +984,82 @@ where
...
-        )?
-        .bind("payment_id", payment_id.clone())
-        .fetch_one(&self.inner)
-        .await?;
-
-        if exists.is_some() {
-            tracing::error!("Payment ID already exists: {}", payment_id);
-            return Err(database::Error::Duplicate);
-        }
-
```
thesimplekid, line 1030: Removing this check and relying on the in-memory duplicate check in add_payment() breaks multi-instance deployments. If two instances load the same quote, both will pass the in-memory check, but one will fail here with a generic DB constraint error instead of DuplicatePaymentId. The callers in ln.rs:92 and mod.rs:761 explicitly catch DuplicatePaymentId to handle this race gracefully that no longer works.

crates/cdk/src/mint/ln.rs
```diff
@@ -57,61 +56,55 @@ impl Mint {
...
-                    .await
-                {
-                    Ok(total_paid) => {
-                        quote.increment_amount_paid(amount_paid)?;
-                        quote.add_payment(amount_paid, payment.payment_id.clone(), unix_time())?;
+                let amount_paid = to_unit(payment.payment_amount, &payment.unit, &new_quote.unit)?;
+
+                match new_quote.add_payment(amount_paid, payment.payment_id.clone(), None) {
+                    Ok(()) => {
+                        tx.update_mint_quote(&mut new_quote).await?;
```
thesimplekid, line 87: This DuplicatePaymentId catch only works for single-instance races (caught by add_payment() in-memory check). In multi-instance deployments, both instances pass add_payment(), then one fails at update_mint_quote() with a DB constraint error which is not DuplicatePaymentId.

crates/cdk/src/mint/mod.rs
```diff
@@ -749,25 +749,23 @@ impl Mint {
...
-                    Ok(total_paid) => {
-                        pubsub_manager.mint_quote_payment(mint_quote, total_paid);
+                match mint_quote.add_payment(
+                    payment_amount_quote_unit,
+                    wait_payment_response.payment_id.clone(),
+                    None,
+                ) {
+                    Ok(()) => {
+                        tx.update_mint_quote(mint_quote).await?;
+                        pubsub_manager.mint_quote_payment(mint_quote, mint_quote.amount_paid());
```
thesimplekid, line 759: This has the same multi instance issue.

crates/cdk-common/src/mint.rs
```diff
@@ -377,6 +377,24 @@ impl Operation {
...
+/// This approach allows business rule validation to happen in the domain model while
+/// keeping the database layer focused purely on persistence.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Hash)]
+pub struct MintQuoteChange {
+    /// New payments added since the quote was loaded or last persisted.
+    pub payments: Option<Vec<IncomingPayment>>,
+    /// New issuance amounts recorded since the quote was loaded or last persisted.
+    pub issuances: Option<Vec<Amount>>,
+}
+
```
thesimplekid, line 378: I don't think the change-set pattern is appropriate here. The database must be the synchronization point for duplicate detection in multi-instance deployments. In-memory validation creates:
 - Inconsistent error types (domain vs DB errors)
 - False sense of correctness (passes in-memory, fails at DB)

crates/cdk-sql-common/src/mint/mod.rs
crodas: We are not, remember the quote is read as exclusive, signaling any other instance of this intent to update.

Also, the INSERT would fail on duplicate entry.

crates/cdk/src/mint/ln.rs
crodas: Keep in mind this is for Acquired<MintQuote> not MintQuote.

crates/cdk-sql-common/src/mint/mod.rs
thesimplekid: > Also, the INSERT would fail on duplicate entry.

With a different error. But yeah you're right using the for update we shouldn't hit this.

crates/cdk-common/src/mint.rs
thesimplekid: This comment also is incorrect and using for update makes it work.

crates/cdk-sql-common/src/mint/mod.rs
crodas: I would make it use the same error

crates/cdk/src/mint/melt/shared.rs
```diff
@@ -241,6 +244,74 @@ pub async fn process_melt_change(
...
+            e => e,
+        })?
+        .ok_or(Error::UnknownQuote)?;
+
+    // Lock any other quotes so they cannot be modified
+    let locked_quotes = if let Some(request_lookup_id) = quote.request_lookup_id.as_ref() {
+        tx.get_melt_quotes_by_request_lookup_id(request_lookup_id)
+            .await
+            .map_err(|e| match e {
+                database::Error::Locked => database::Error::Duplicate,
```
thesimplekid: We should log that its locked.

crates/cdk/src/mint/melt/shared.rs
```diff
@@ -241,6 +244,74 @@ pub async fn process_melt_change(
...
+/// # Returns
+///
+/// The loaded and locked melt quote, ready for state transitions.
+///
+/// # Errors
+///
+/// * [`Error::UnknownQuote`] if no quote exists with the given ID. * [`Error::Database(Duplicate)`]
+///   if another quote with the same lookup ID is already pending or paid, indicating a conflicting
+///   concurrent melt operation.
+#[inline(always)]
```
thesimplekid:
```suggestion
```

I think we're over using inline in general and here specifically its an async fn with multiple db calls I don't think there is a benefit to inline that.

crates/cdk-common/src/database/mint/test/mod.rs
thesimplekid: Can you link it? When removing tests especially an important one we should be careful that is intentional and there is the expected test converge and we know where it is.

crodas:
> Can you link it? When removing tests especially an important one we should be careful that is intentional and there is the expected test converge and we know where it is.

You are right, I will add unit tests for `check_state_transition` in `crates/cdk-common/src/state.rs`

crates/cdk/src/mint/melt/shared.rs
crodas: Fixed in 9f5b58eb95e26a381103b20f13ae8db25cbcc9c1

PR #1466, opened 2025-12-27
P2pk receive wallet

lescuer97:
### Description

Add public key generation with counter for wallet.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-cli/src/sub_commands/get_public_keys.rs
+++ b/crates/cdk-cli/src/sub_commands/get_public_keys.rs
@@ -59,7 +59,8 @@ pub async fn get_public_keys(
 
     let list_public_keys = wallet.get_public_keys().await?;
     if list_public_keys.is_empty() {
-        println!("\npublic not found!\n");
+        println!("\n public not found! \n");
+        return Ok(());
     }
     println!("\npublic keys found:\n");
     for public_key in list_public_keys {
--- a/crates/cdk-ffi/src/types/wallet.rs
+++ b/crates/cdk-ffi/src/types/wallet.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 
+use cdk_common::bitcoin;
 use serde::{Deserialize, Serialize};
 
 use super::amount::{Amount, SplitTarget};
@@ -100,13 +101,15 @@ pub struct P2PKSigningKey {
 impl TryFrom<P2PKSigningKey> for cdk_common::wallet::P2PKSigningKey {
     type Error = crate::error::FfiError;
 
-    fn try_from(key: P2PKSigningKey) -> Result<Self, Self::Error> {
+    fn try_from(key: P2PKSigningKey) -> Result<Self, FfiError> {
         Ok(Self {
             pubkey: key.pubkey.try_into()?,
             derivation_path: key
                 .derivation_path
                 .parse()
-                .expect("Invalid derivation path"),
+                .map_err(|e: bitcoin::bip32::Error| FfiError::Internal {
+                    error_message: e.to_string(),
+                })?,
             derivation_index: key.derivation_index,
             created_time: key.created_time,
         })
--- a/crates/cdk-redb/src/wallet/mod.rs
+++ b/crates/cdk-redb/src/wallet/mod.rs
@@ -1567,7 +1567,7 @@ impl WalletDatabase<database::Error> for WalletRedbDatabase {
             .map_err(Error::from)?
             .flatten()
             .filter_map(|(_k, v)| serde_json::from_str::<wallet::P2PKSigningKey>(v.value()).ok())
-            .max_by_key(|key| key.created_time);
+            .max_by_key(|key| key.derivation_index);
 
         Ok(latest_key)
     }


crates/cdk-common/src/database/wallet/mod.rs
```diff
@@ -191,4 +192,11 @@ where
         direction: Option<TransactionDirection>,
         unit: Option<CurrencyUnit>,
     ) -> Result<Vec<Transaction>, Err>;
+
+    // stores p2pk signing keys for the wallet
+    async fn add_p2pk_key(&self, pubkey: &PublicKey, derivation_path: String, derivation_index: u32) -> Result<(), Self::Err>;
+    /// Get a stored P2PK signing key by pubkey.
+    async fn get_p2pk_key(&self, pubkey: &PublicKey) -> Result<Option<wallet::P2PKSigningKey>, Self::Err>;
+    /// List all stored P2PK signing keys.
+    async fn list_p2pk_keys(&self) -> Result<Vec<wallet::P2PKSigningKey>, Self::Err>;
```
thesimplekid: I think we need to have a discussion and have some guidelines on when we should add new fns to the db traits and when we should take advantage of the KVStore. I think there is a balance between over using the store and making our db traits too complicated.

lescuer97: > I think we need to have a discussion and have some guidelines on when we should add new fns to the db traits and when we should take advantage of the KVStore. I think there is a balance between over using the store and making our db traits too complicated.

could you explain the difference between KVStore and the traits? what do you think it should be the parameters on choosing one or the other?

thesimplekid: The kvstore is general so we can just store arbitrary data in it without having to add fns on the trait. With the drawback of a different migration pattern and less structured data. So I'm not necessarily saying we should do it here but there is a discussion worth having about when to use the Kv and when to add new fns and db tables.

lescuer97: yeah, I understand. I would use the KVstore when you have data that is clearly independent. EX: Information about a specific npub that you sent money to comes up in my mind

lescuer97:
@thesimplekid @crodas  I think this is know in a decent state. I know I'm lacking tests. I want to get your opinion on the implementation before I do that.

crates/cdk-cli/src/sub_commands/get_public_keys.rs
```diff
@@ -0,0 +1,42 @@
...
+
+        match latest_public_key {
+            Some(key) => {
+                println!("\npublic key found! 🎉\n");
+
+                println!("public key: {}", key.pubkey.to_hex());
+                println!("derivation path: {}", key.derivation_path);
+            }
+            None => {
+                println!("\npublic key not found! 🤔\n");
```
thesimplekid: Lets remove the emojis

crates/cdk-cli/Cargo.toml
```diff
@@ -22,6 +22,7 @@ anyhow.workspace = true
 bip39.workspace = true
 bitcoin.workspace = true
 cdk = { workspace = true, default-features = false, features = ["wallet", "auth", "nostr", "bip353"]}
+cdk-common = {workspace = true}
```
thesimplekid: We shouldn't need this should be able to get what we need from cdk

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -376,6 +377,14 @@ impl TryFrom<Proofs> for TransactionId {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct P2PKSigningKey {
+    pub pubkey: PublicKey,
+    pub derivation_path: DerivationPath,
+    pub derivation_index: u32,
+    pub created_time: u64,
+}
```
thesimplekid: This should need doc comments not sure why clippy hasn't caught it

crates/cdk-ffi/src/database.rs
```diff
@@ -249,6 +264,10 @@ pub trait WalletDatabaseTransaction: Send + Sync {
         primary_namespace: String,
         secondary_namespace: String,
     ) -> Result<Vec<String>, FfiError>;
+
+    // async fn add_p2pk_key(&self, pubkey: PublicKey, derivation_path: String, derivation_index: u32) -> Result<(), FfiError>;
+    // async fn get_p2pk_key(&self, pubkey: PublicKey) -> Result<Option<P2PKSigningKey>, FfiError>;
+    // async fn list_p2pk_keys(&self) -> Result<Vec<P2PKSigningKey>, FfiError>;
```
thesimplekid: Out of date

crates/cdk-ffi/src/database.rs
```diff
@@ -1902,6 +2011,19 @@ impl WalletDatabaseTransaction for FfiWalletTransaction {
...
+    //     Ok(())
+    // }
+    // async fn get_p2pk_key(&self, pubkey: PublicKey) -> Result<Option<P2PKSigningKey>, FfiError> {
+    //     todo!("Implement get_p2pk_key");
+    //     Ok(None)
+    // }
+    // async fn list_p2pk_keys(&self) -> Result<Vec<P2PKSigningKey>, FfiError> {
+    //     todo!("Implement list_p2pk_keys");
+    //     Ok(vec![])
+    // }
```
thesimplekid: out of date

crates/cdk/src/wallet/mod.rs
```diff
@@ -757,6 +760,24 @@ impl Wallet {
...
     }
+
+    // try to get secret key from p2pk signing key in localstore
+    async fn get_signing_key(&self, pubkey: &PublicKey) -> Result<Option<SecretKey>, Error> {
+        let signing = self.localstore.get_p2pk_key(pubkey).await?;
+        if let Some(signing) = signing {
+            // make xpriv from seed
+            let xpriv = Xpriv::new_master(Network::Bitcoin, &self.seed)?;
+            // use seed to derive the secret key
+            // let key_path = DerivationPath::from_str(&signing.derivation_path)?;
```
thesimplekid: Outdated

crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -1984,6 +1988,40 @@ impl MultiMintWallet {
...
+            .await?;
+        Ok(pubkey)
+    }
+
+    pub async fn p2pk_public_key(
+        &self,
+        pubkey: &PublicKey,
+    ) -> Result<Option<P2PKSigningKey>, database::Error> {
+        return self.localstore.get_p2pk_key(pubkey).await;
+    }
```
thesimplekid: Think we should rename this get private key?

crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -1984,6 +1988,40 @@ impl MultiMintWallet {
 
         wallet.melt_human_readable_quote(address, amount_msat).await
     }
+    pub async fn generate_public_key(&self) -> Result<PublicKey, Error> {
```
thesimplekid: These fns should also have doc comments

thesimplekid: I wonder if we should have a fn to get the most recent pubkey especially with p2bk the user might want to just reuse the pubkey.

thesimplekid (changes requested):

crates/cdk/src/wallet/multi_mint_wallet.rs
lescuer97: this does not give a private key, so no

crates/cdk/src/wallet/multi_mint_wallet.rs
lescuer97: I don't think it's a bad idea.

lescuer97: I made a function for getting the latest public key

crates/cdk/src/wallet/multi_mint_wallet.rs
lescuer97: I don't expose right now a way to get the secret key. this is a private method.

I'll take suggestions on naming

thesimplekid: > this does not give a private key, so no

get_public_key i meant

crates/cdk-redb/src/wallet/mod.rs
```diff
@@ -601,6 +613,96 @@ impl WalletDatabase<database::Error> for WalletRedbDatabase {
...
+            .open_table(P2PK_SIGNING_KEYS_TABLE)
+            .map_err(Error::from)?;
+        table
+            .insert(
+                pubkey.to_bytes().as_slice(),
+                serde_json::to_string(&wallet::P2PKSigningKey {
+                    pubkey: *pubkey,
+                    derivation_path,
+                    derivation_index,
+                    created_time: 0,
```
thesimplekid: We need to set the time here. we can use `unix_time()`

crates/cdk/src/wallet/receive.rs
```diff
@@ -99,8 +99,20 @@ impl Wallet {
...
+                            None => {
+                                let secret_key_option = self.get_signing_key(&pubkey).await?;
+                                if let Some(secret_key) = secret_key_option {
+                                    // cache secret key so it only has to be locked up one for the duration of the receive operation
+                                    p2pk_signing_keys
+                                        .insert(pubkey.x_only_public_key(), secret_key.clone());
+
+                                    proof.sign_p2pk(secret_key.to_owned().clone())?;
+                                }
+                            }
```
thesimplekid: We don't error if we do not have the required keys to sign. If we cannot get enough sigs for a proof we should not send it to the mint.

crates/cdk/src/wallet/multi_mint_wallet.rs
lescuer97: changed the named functions plus added missing ones to ffi

crates/cdk/src/wallet/receive.rs
lescuer97: I added an `enough_signatures` function to the Proof struct that will check. Needed to add a new error to the Nut 10 file.

I'm getting a typo error from a `meeting agenda`. I assume this is a mistake

thesimplekid: > I'm getting a typo error from a meeting agenda. I assume this is a mistake

Rebase will solve this

crates/cdk-redb/src/wallet/mod.rs
```diff
@@ -601,6 +613,96 @@ impl WalletDatabase<database::Error> for WalletRedbDatabase {
...
     }
+
+    #[instrument(skip(self))]
+    async fn add_p2pk_key(
+        &self,
+        pubkey: &PublicKey,
+        derivation_path: DerivationPath,
+        derivation_index: u32,
+    ) -> Result<(), database::Error> {
+        // let txn = self.txn()?;
```
thesimplekid:
```suggestion
```

crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -1311,6 +1312,79 @@ where
...
+        &self,
+        pubkey: &PublicKey,
+        derivation_path: DerivationPath,
+        derivation_index: u32,
+    ) -> Result<(), Error> {
+        let conn = self.pool.get().map_err(|e| Error::Database(Box::new(e)))?;
+        let query_str = r#"
+        INSERT INTO p2pk_signing_key (pubkey, derivation_index, derivation_path)
+        VALUES (:pubkey, :derivation_index, :derivation_path)
+        "#
```
thesimplekid, line 1667: This is missing the created time. 

We should add some db tests for the new functions that would have caught this.

crates/cdk-cli/src/sub_commands/get_public_keys.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub async fn get_public_keys(
+    multi_mint_wallet: &MultiMintWallet,
+    sub_command_args: &GetPublicKeysSubCommand,
+) -> Result<()> {
+    let list_public_keys = multi_mint_wallet.get_public_keys().await?;
+    if sub_command_args.latest {
+        // keys are ordered by creation time, so the first one is the latest
+        let latest_public_key = list_public_keys.first().cloned();
+
```
thesimplekid: There is a `get_latest_pubke` fn on the MultiMintWallet we should use that if the latest arg is passed.

crates/cdk-cli/src/sub_commands/get_public_keys.rs
```diff
@@ -0,0 +1,42 @@
+use anyhow::{Ok, Result};
```
thesimplekid: We don't need the `Ok` import here.

crates/cdk-cli/src/sub_commands/get_public_keys.rs
```diff
@@ -0,0 +1,42 @@
...
+            None => {
+                println!("\npublic key not found!\n");
+            }
+        }
+
+        return Ok(());
+    }
+
+    println!("\npublic keys found:\n");
+    for public_key in list_public_keys {
```
thesimplekid, line 66: We should check if the list is empty here and print no keys found if so.

crates/cashu/src/nuts/nut00/mod.rs
```diff
@@ -379,6 +381,42 @@ impl Proof {
     pub fn y(&self) -> Result<PublicKey, Error> {
         Ok(hash_to_curve(self.secret.as_bytes())?)
     }
+
+    /// checks if the signature spend conditions are met
+    #[cfg(feature = "wallet")]
+    pub fn enough_signatures(&self) -> Result<(), Error> {
+        if let Ok(secret) = <crate::secret::Secret as TryInto<crate::nuts::nut10::Secret>>::try_into(
+            self.clone().secret,
```
thesimplekid: We already have a `verify_p2pk()` fn that we use for the mint. I think we can just reuse that and remove this fn?

crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -1984,6 +1988,49 @@ impl MultiMintWallet {
...
+        let mut last_derivation_index = 0;
+
+        for public_key in public_keys {
+            if public_key.derivation_index > last_derivation_index {
+                last_derivation_index = public_key.derivation_index;
+            }
+        }
+        last_derivation_index += 1;
+
+        let derivation_path = DerivationPath::from_str(&format!("m/0/{}", last_derivation_index))?;
```
thesimplekid: We should have a purpose identifier. We could reuse the cashu purpose of 129372' and then use 1 for the next where on secrets we use 0.

crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -1984,6 +1988,49 @@ impl MultiMintWallet {
 
         wallet.melt_human_readable_quote(address, amount_msat).await
     }
+
+    /// generates and stores public key in database
+    pub async fn generate_public_key(&self) -> Result<PublicKey, Error> {
```
thesimplekid: Why is this only here on the multi-mint wallet. Shouldn't it be on the `Wallet` and then the Multi-mint wallet can call that. Generally this is how we use the multi-mint wallet where it is just a dumb wrapper around the Wallet.

crates/cdk/src/wallet/multi_mint_wallet.rs
```diff
@@ -1984,6 +1988,49 @@ impl MultiMintWallet {
...
+
+    /// gets list of stored public keys in database
+    pub async fn get_public_keys(&self) -> Result<Vec<P2PKSigningKey>, database::Error> {
+        return self.localstore.list_p2pk_keys().await;
+    }
+
+    /// Gets the latest generated P2PK signing key (most recently created)
+    pub async fn get_latest_public_key(&self) -> Result<Option<P2PKSigningKey>, database::Error> {
+        self.localstore.latest_p2pk().await
+    }
```
thesimplekid: These fns should be on the `Wallet` as well. The `Wallet~ needs to be able to usable without the multi mint wallet.

crates/cdk/src/wallet/multi_mint_wallet.rs
thesimplekid: I thought about this a bit more and we should not reuse the cashu purpose. We should only use that where its spec'd in the nuts which this is not.

crates/cdk-sql-common/src/wallet/mod.rs
lescuer97: yeah, I had a default param in the migration, that is why I had not caught it. I will add some tests

crates/cdk/src/wallet/multi_mint_wallet.rs
lescuer97: I used the new 1 as we said with the cashu purpose.

thesimplekid: > I used the new 1 as we said with the cashu purpose.

I think was a bad idea from me. What if something else gets added to the spec using that purpose, we would then conflict.

crates/cdk/src/wallet/multi_mint_wallet.rs
lescuer97: I might have a miss understanding on how this works. I didn't want to put the functions inside the Wallet mostly because it would mean that the keys are dependent of each mint. 
Should I just allow for both?

thesimplekid: The keys won't really rely on the wallet since there is nothing in the derivation path linking them to a mint. But yeah we need the fns on the `Wallet` so can just make this on both the `Wallet` and `MultimintWallet`. The logic can be in a shared helper fn.

lescuer97:
@thesimplekid  rebased

crates/cdk/src/wallet/p2pk.rs
```diff
@@ -0,0 +1,63 @@
...
+        if public_key.derivation_index > last_derivation_index {
+            last_derivation_index = public_key.derivation_index;
+        }
+    }
+    last_derivation_index += 1;
+
+    let derivation_path = DerivationPath::from(vec![
+        ChildNumber::from_hardened_idx(0)?,
+        ChildNumber::from_hardened_idx(last_derivation_index)?,
+    ]);
```
thesimplekid: Shouldn't we still have a purpose here even though we said we shouldnt reuse the cashu one?

lescuer97: yeah we probably should

lescuer97: Added a purpose and solved conflicts

crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -1991,3 +2068,48 @@ fn sql_row_to_transaction(row: Vec<Column>) -> Result<Transaction, Error> {
...
+    }
+
+    async fn kv_list(
+        &self,
+        primary_namespace: &str,
+        secondary_namespace: &str,
+    ) -> Result<Vec<String>, Error> {
+        crate::keyvalue::kv_list(&self.pool, primary_namespace, secondary_namespace).await
+    }
+}
```
thesimplekid: Make a rebase issue why are these added here?

crates/cdk-ffi/src/types/wallet.rs
```diff
@@ -717,3 +756,5 @@ impl From<cdk::WalletKey> for WalletKey {
         }
     }
 }
+
+pub use cdk_common::wallet::{WalletSaga, WalletSagaState};
```
thesimplekid: Not sure why these are here?

crates/cdk/src/wallet/mod.rs
```diff
@@ -870,6 +874,48 @@ impl Wallet {
...
+            let xpriv = Xpriv::new_master(Network::Bitcoin, &self.seed)?;
+            return Ok(Some(SecretKey::from(
+                xpriv
+                    .derive_priv(&SECP256K1, &signing.derivation_path)?
+                    .private_key,
+            )));
+        }
+
+        Ok(None)
+    }
```
thesimplekid, line 943: Now that we have the wallet trait new wallet fns that are intended for the public api should go there.

crates/cdk/src/wallet/p2pk.rs
```diff
@@ -0,0 +1,140 @@
...
+
+use bitcoin::bip32::{ChildNumber, DerivationPath, Xpriv};
+use bitcoin::Network;
+use cdk_common::database::{self, WalletDatabase};
+use cdk_common::wallet::P2PKSigningKey;
+use cdk_common::{PublicKey, SECP256K1};
+
+use crate::error::Error;
+
+const CASHU_PURPOSE: u32 = 129373;
```
thesimplekid: This is not the cashu purpose

crates/cdk/src/wallet/p2pk.rs
```diff
@@ -0,0 +1,140 @@
...
+const P2PK_PURPOSE: u32 = 10;
```
thesimplekid: Maybe should link to coco/spec of why we use these.

crates/cdk/src/wallet/p2pk.rs
```diff
@@ -0,0 +1,140 @@
...
+) -> Result<Vec<P2PKSigningKey>, database::Error> {
+    localstore.list_p2pk_keys().await
+}
+
+/// Gets the latest generated P2PK signing key (most recently created)
+pub async fn get_latest_public_key(
+    localstore: &Arc<dyn WalletDatabase<database::Error> + Send + Sync>,
+) -> Result<Option<P2PKSigningKey>, database::Error> {
+    localstore.latest_p2pk().await
+}
```
thesimplekid: Why do we need these aren't they on the wallet? 

I think the ones currently in mod should be moved here and these should be removed.

crates/cdk/src/wallet/mod.rs
```diff
@@ -49,6 +52,7 @@ mod mint_connector;
 mod mint_metadata_cache;
 #[cfg(feature = "npubcash")]
 mod npubcash;
+pub mod p2pk;
```
thesimplekid: I think this should not be pub and should be used through the wallet.

crates/cdk-sql-common/src/wallet/mod.rs
lescuer97: 100% rebase issue. I'll fix it

crates/cdk/src/wallet/p2pk.rs
lescuer97: Im going to change the name.

crates/cdk-ffi/src/types/wallet.rs
lescuer97: don't know either. I remove them

crates/cdk-sql-common/src/wallet/mod.rs
lescuer97: removed

crates/cdk/src/wallet/mod.rs
lescuer97: I added the functions to the trait

crates/cdk/src/wallet/p2pk.rs
lescuer97: changed the name to make it clearer

crates/cdk/src/wallet/p2pk.rs
lescuer97: added some comments on why we use this. plus the PR link

crates/cdk/src/wallet/p2pk.rs
lescuer97: I remved them and put the in the wallet file

crates/cdk/src/wallet/mod.rs
lescuer97: done!

crates/cdk-ffi/src/types/wallet.rs
```diff
@@ -84,6 +84,46 @@ impl From<SendKind> for cdk::wallet::SendKind {
...
+impl TryFrom<P2PKSigningKey> for cdk_common::wallet::P2PKSigningKey {
+    type Error = crate::error::FfiError;
+
+    fn try_from(key: P2PKSigningKey) -> Result<Self, Self::Error> {
+        Ok(Self {
+            pubkey: key.pubkey.try_into()?,
+            derivation_path: key
+                .derivation_path
+                .parse()
+                .expect("Invalid derivation path"),
```
thesimplekid: We shouldn't panic here we're in a try from anyway return an error.

crates/cdk-cli/src/sub_commands/get_public_keys.rs
```diff
@@ -0,0 +1,70 @@
...
+                println!("\npublic key not found!\n");
+            }
+        }
+
+        return Ok(());
+    }
+
+    let list_public_keys = wallet.get_public_keys().await?;
+    if list_public_keys.is_empty() {
+        println!("\npublic not found!\n");
```
thesimplekid:
```suggestion
        println!("\npublic not found!\n");
        return Ok(());
```

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -950,6 +951,38 @@ pub trait Wallet: Send + Sync {
...
+    ) -> Result<Option<P2PKSigningKey>, Self::Error>;
+
+    /// gets list of stored public keys in database
+    async fn get_public_keys(&self) -> Result<Vec<P2PKSigningKey>, Self::Error>;
+
+    /// Gets the latest generated P2PK signing key (most recently created)
+    async fn get_latest_public_key(&self) -> Result<Option<P2PKSigningKey>, Self::Error>;
+
+    /// try to get secret key from p2pk signing key in localstore
+    async fn get_signing_key(&self, pubkey: &PublicKey) -> Result<Option<SecretKey>, Self::Error>;
```
thesimplekid, line 972: On second thought. And thinking about how we want to reduce our public api surface, maybe only generate and get latest should be in the public api and the other fns should be on the wallet as pub(crate) this way they can be removed from the trait and ffi?

crates/cdk-redb/src/wallet/mod.rs
```diff
@@ -1467,6 +1479,98 @@ impl WalletDatabase<database::Error> for WalletRedbDatabase {
...
+        let table = read_txn
+            .open_table(P2PK_SIGNING_KEYS_TABLE)
+            .map_err(Error::from)?;
+
+        let latest_key = table
+            .iter()
+            .map_err(Error::from)?
+            .flatten()
+            .filter_map(|(_k, v)| serde_json::from_str::<wallet::P2PKSigningKey>(v.value()).ok())
+            .max_by_key(|key| key.created_time);
```
thesimplekid: Here it is time

crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -1649,6 +1650,82 @@ where
...
+                Some(row)
+            })
+            .collect::<Vec<wallet::P2PKSigningKey>>())
+    }
+
+    #[instrument(skip(self))]
+    async fn latest_p2pk(&self) -> Result<Option<wallet::P2PKSigningKey>, Error> {
+        let conn = self.pool.get().map_err(|e| Error::Database(Box::new(e)))?;
+        let query_str = r#"
+        SELECT pubkey, derivation_index, derivation_path, created_time FROM p2pk_signing_key ORDER BY derivation_index DESC LIMIT 1
```
thesimplekid, line 1720: There is a difference by how we select the latest in redb vs sqlite. In redb we use time (https://github.com/cashubtc/cdk/pull/1466/changes#r3000818469) but in sql we use derivation index there should be the same, but we should still select on the same thing.

crates/cdk-common/src/wallet/mod.rs
lescuer97: I think the 2 that we could remove from the public are `get_signing_key` and `get_public_keys`. the others do need to there

crates/cdk-redb/src/wallet/mod.rs
lescuer97: fixed

crates/cdk-sql-common/src/wallet/mod.rs
lescuer97: fixed

crates/cdk-cli/src/sub_commands/get_public_keys.rs
lescuer97: fixed

crates/cdk-ffi/src/types/wallet.rs
lescuer97: fixed

Issue #1463, opened 2025-12-26
Add currency unit to the `Amount` type


thesimplekid:
This will be a fairly simple in complexity but large in the number of lines its touches.

PR #1470, opened 2025-12-28
feat: Generic unit on amount

thesimplekid:
### Description

-----

The Cashu protocol supports multiple currency units (Sat, Msat, Usd, Eur, etc.), but the previous Amount type was a simple u64 wrapper with no awareness of units. This created a subtle but dangerous class of bugs: nothing prevented code from accidentally adding 1000 satoshis to 500 millisatoshis, producing a nonsensical result of 1500 in an undefined unit. These bugs are especially insidious because the code compiles and runs without error—the incorrect math only manifests as wrong balances or failed payments in production.

This change makes unit mismatches impossible by encoding the currency unit into the type system. Amount<CurrencyUnit> now carries its unit at the type level, and arithmetic operations verify unit compatibility before proceeding. The compiler catches many mistakes statically, and runtime checks catch the rest with a clear UnitMismatch error rather than silent corruption.

The refactor maintains backwards compatibility by using Amount<()> (untyped) at serialization boundaries where the wire protocol expects a plain integer, while internal mint logic now uses Amount<CurrencyUnit> to get the safety guarantees. The with_unit() method provides a clear conversion point at the boundary between protocol parsing and application logic, making it explicit where unit context is being added.

This is particularly important for the mint's quote handling, where amounts flow between payment backends (which may use different units internally) and the core mint logic. By making MintQuote.amount_paid, MeltQuote.fee_reserve, and payment response types carry their units, we ensure that fee calculations, balance checks, and unit conversions are always performed correctly.

closes: https://github.com/cashubtc/cdk/issues/1463

- [ ] test with lnbits
- [x] wait on https://github.com/cashubtc/cdk/pull/1251

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-signatory/src/proto/server.rs
+++ b/crates/cdk-signatory/src/proto/server.rs
@@ -61,8 +61,7 @@ where
         for msg in blinded_messages {
             converted_messages.push(msg.try_into()?);
         }
-        let result = match signatory.blind_sign(converted_messages).await
-        {
+        let result = match signatory.blind_sign(converted_messages).await {
             Ok(blind_signatures) => proto::BlindSignResponse {
                 sigs: Some(proto::BlindSignatures {
                     blind_signatures: blind_signatures
@@ -93,8 +92,7 @@ where
         for p in proofs {
             converted_proofs.push(p.try_into()?);
         }
-        let result = match signatory.verify_proofs(converted_proofs).await
-        {
+        let result = match signatory.verify_proofs(converted_proofs).await {
             Ok(()) => proto::BooleanResponse {
                 success: true,
                 ..Default::default()


davidcaseria (approved):

crates/cashu/src/amount.rs
```diff
@@ -199,23 +264,31 @@ impl Amount {
     }
 
     /// Checked addition for Amount. Returns None if overflow occurs.
-    pub fn checked_add(self, other: Amount) -> Option<Amount> {
-        self.0.checked_add(other.0).map(Amount)
+    pub fn checked_add(self, other: Amount<()>) -> Option<Amount<()>> {
```
crodas, line 302: @thesimplekid  Pardon my ignorance, but wouldn't it be better if the other amount had the same generic, and we refuse to do the math if the currency units are different? Or the implementor can implement a way to unify the units if possible (from Sat to MSat, for instance).

This is a great chance in my opinion, to reduce confusions when doing math with different units

thesimplekid: If the other Amount had a unit there would be no need for a generic or a second unit type. This allows us to ensure our units match in the mint where it matters without losing copy and having to introduce clone in even more places throughout the wallet making this a much harder refactor.

thesimplekid: > Or the implementor can implement a way to unify the units if possible (from Sat to MSat, for instance).

There is already a convert to fn for this.

crodas (approved):

crodas (dismissed):

PR #1487, opened 2026-01-02
feat: Mint reporting rpc

### LINKED ISSUE #1093: Mint Feature Request: Add tos_url to GetInfo method | Add UpdateTosUrl Request
The tos_url is not currently returned in the GetInfo response object.
Additionally having a method to update the value would improve the gRPC

orangeshyguy21:
### Description

This PR splits the mint gRPC API into management and reporting services.

Mint reporting endpoints are introduced to expose mint records (balances, keysets, quotes, proofs, signatures, operations). 

Mint management endpoints were added (update_tos_url, create_backup).

-----

### Notes to the reviewers

There is one outstanding shortcoming with this PR
- get_keysets returns null values for valid_to, valid_from, and index
Changes are required to core traits to expose this information

Some features were not included in with this PR that are may be candidates for future improvement
- Backup auth database
- Restore database
- Auth reporting endpoints
- Update melt quote state endpoint 
- Authorization updates to allow gated access to certain rpc endpoints
- Server resource endpoints (Prometheus data) 

There are 'breaking' changes in this PR as the current proto file was renamed and reorganized. Consumers will be required to change their implementation to support this release.

-----

#### CHANGED
- Split mint RPC service into separate Management and Reporting gRPC services.
- Updated cdk-mint-rpc CLI to use the new reporting/management clients and reorganized commands accordingly.
- Balance information is no longer in `get-info` and is now on a dedicated `get-balances` method

#### ADDED
- New reporting RPC endpoints and CLI commands
  - balances
  - keysets
  - list/lookup mint quotes
  - list/lookup melt quotes
  - list proofs
  - list blind signatures
  - list operations.
- Added `tos_url` to `get_info` and added `update_tos_url` rpc method
- Database backup support (SQLite binary backup and SQL dump) exposed via a streaming management RPC.

----

### Checklist


orangeshyguy21:
> Can we do this issue as part of this? #1093

Yes, this issue gets addressed in the PR.

crates/cdk-sql-common/src/mint/backup.rs
thesimplekid, line 1: Introduction of a backup scheme should be in its own PR not part of the rpc pr

crates/cdk-sql-common/src/mint/filters.rs
thesimplekid, line 1: Adding these new query types and filters should be its own pr.

thesimplekid (changes requested):
We should split this up into multiple focused prs for better review.

crates/cdk-sql-common/src/mint/backup.rs
orangeshyguy21: Agree here. This is a good one to omit to narrow scope.

crates/cdk-sql-common/src/mint/filters.rs
orangeshyguy21: Roger. For the MVP we could just include paging - no where clause filtering

PR #1488, opened 2026-01-03
Introduce ProofsWithState for atomic proof state management

crodas:
### Description

This PR depends on #1490 

Replace individual proof state operations with a unified ProofsWithState type that enforces the invariant that all proofs in a set share the same state. This shifts responsibility for state consistency to the database layer and simplifies state transition logic in the saga implementations.

Changes to ProofsTransaction trait:
- add_proofs now returns Acquired<ProofsWithState>
- update_proofs_states(ys, state) -> update_proofs(&mut Acquired<ProofsWithState>)
- get_proofs_states(ys) -> get_proofs(ys) returning Acquired<ProofsWithState>

Benefits:
- State transitions validated in memory before persisting
- Eliminates scattered state checking in saga code
- Database layer guarantees proof state consistency
- Cleaner API with state encapsulated in the type

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/melt/melt_saga/mod.rs
+++ b/crates/cdk/src/mint/melt/melt_saga/mod.rs
@@ -248,17 +248,6 @@ impl MeltSaga<Initial> {
 
         let mut proofs = tx.get_proofs(&input_ys).await?;
 
-        let original_state = proofs.state;
-
-        if matches!(original_state, State::Pending | State::Spent) {
-            tx.rollback().await?;
-            return Err(if original_state == State::Pending {
-                Error::TokenPending
-            } else {
-                Error::TokenAlreadySpent
-            });
-        }
-
         if let Err(err) = Mint::update_proofs_state(&mut tx, &mut proofs, State::Pending).await {
             tx.rollback().await?;
             return Err(err);
--- a/crates/cdk/src/mint/proofs.rs
+++ b/crates/cdk/src/mint/proofs.rs
@@ -1,6 +1,6 @@
 use cdk_common::database::{Acquired, DynMintTransaction};
 use cdk_common::mint::ProofsWithState;
-use cdk_common::state::check_state_transition;
+use cdk_common::state::{self, check_state_transition};
 use cdk_common::{Error, State};
 
 use crate::Mint;
@@ -22,7 +22,11 @@ impl Mint {
         proofs: &mut Acquired<ProofsWithState>,
         new_state: State,
     ) -> Result<(), Error> {
-        check_state_transition(proofs.state, new_state).map_err(|_| Error::UnexpectedProofState)?;
+        check_state_transition(proofs.state, new_state).map_err(|err| match err {
+            state::Error::AlreadySpent => Error::TokenAlreadySpent,
+            state::Error::Pending => Error::TokenPending,
+            _ => Error::UnexpectedProofState,
+        })?;
 
         tx.update_proofs_state(proofs, new_state)
             .await
@@ -170,7 +174,7 @@ mod tests {
 
         let result = Mint::update_proofs_state(&mut tx, &mut acquired, State::Pending).await;
 
-        assert!(matches!(result, Err(Error::UnexpectedProofState)));
+        assert!(matches!(result, Err(Error::TokenPending)));
     }
 
     /// Test that update_proofs_state rejects invalid transition from Spent
@@ -215,7 +219,7 @@ mod tests {
 
         let result = Mint::update_proofs_state(&mut tx, &mut acquired, State::Pending).await;
 
-        assert!(matches!(result, Err(Error::UnexpectedProofState)));
+        assert!(matches!(result, Err(Error::TokenAlreadySpent)));
     }
 
     /// Test that ProofsWithState.state is updated after successful update
--- a/crates/cdk/src/mint/swap/swap_saga/mod.rs
+++ b/crates/cdk/src/mint/swap/swap_saga/mod.rs
@@ -4,7 +4,6 @@ use std::sync::Arc;
 use cdk_common::database::DynMintDatabase;
 use cdk_common::mint::{Operation, Saga, SwapSagaState};
 use cdk_common::nuts::BlindedMessage;
-use cdk_common::state::check_state_transition;
 use cdk_common::{database, Amount, Error, Proofs, ProofsMethods, PublicKey, QuoteId, State};
 use tokio::sync::Mutex;
 use tracing::instrument;
@@ -192,17 +191,6 @@ impl<'a> SwapSaga<'a, Initial> {
                 });
             }
         };
-
-        let original_state = new_proofs.state;
-
-        if check_state_transition(new_proofs.state, State::Pending).is_err() {
-            tx.rollback().await?;
-            return Err(if original_state == State::Pending {
-                Error::TokenPending
-            } else {
-                Error::TokenAlreadySpent
-            });
-        }
 
         let ys = match input_proofs.ys() {
             Ok(ys) => ys,


crates/cdk/src/mint/swap/swap_saga/mod.rs
```diff
@@ -209,24 +221,12 @@ impl<'a> SwapSaga<'a, Initial> {
         };
 
         // Verify proofs weren't already pending or spent
-        if ys.len() != original_proof_states.len() {
+        if ys.len() != new_proofs.len() {
             tracing::error!("Mismatched proof states");
             tx.rollback().await?;
             return Err(Error::Internal);
         }
```
thesimplekid: Can this check be removed now? Since we have the lock on the proofs from the add and then we check the state transition in the `set_new_state`?

crates/cdk/src/mint/swap/swap_saga/mod.rs
```diff
@@ -423,21 +423,10 @@ impl SwapSaga<'_, Signed> {
...
-            .await?
-            .into_iter()
-            .collect::<Option<Vec<_>>>()
-            .ok_or(Error::UnexpectedProofState)?
-        {
-            check_state_transition(current_state, State::Spent)
-                .map_err(|_| Error::UnexpectedProofState)?;
-        }
+        let mut proofs = tx.get_proofs(&self.state_data.ys).await?;
+        proofs.set_new_state(State::Spent)?;
```
thesimplekid: I think we need to call compensate on the error here since we would expect the error here and not in the `tx.update_proofs` like we used to.

crates/cdk/src/mint/swap/swap_saga/mod.rs
thesimplekid: If my understanding is correct and my comment on line 427, we can remove these since we expect the error there.

crates/cdk/src/mint/swap/swap_saga/mod.rs
```diff
@@ -177,25 +176,38 @@ impl<'a> SwapSaga<'a, Initial> {
...
+
+        let original_state = new_proofs.get_state();
+
+        new_proofs.set_new_state(State::Pending).map_err(|_| {
+            if original_state == State::Pending {
+                Error::TokenPending
+            } else {
+                Error::TokenAlreadySpent
+            }
+        })?;
```
thesimplekid: We should explicitly roll back the tx before returning the error. As is done on 214, other wise it leads to race conditions in tests.

crates/cdk/src/mint/swap/swap_saga/mod.rs
thesimplekid: These errors are no longer expected here but above on line 196.

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -329,13 +342,62 @@ where
...
+            FOR UPDATE
+            "#,
+        )?
+        .bind_vec("ys", ys.iter().map(|y| y.to_bytes().to_vec()).collect())
+        .fetch_all(&self.inner)
+        .await?;
+
+        if rows.is_empty() {
+            return Err(database::Error::ProofNotFound);
+        }
```
thesimplekid: We should enforce that the number of got proofs is == the number requested. I think this is safer then forcing the application to check this.

crates/cdk/src/mint/swap/swap_saga/mod.rs
```diff
@@ -423,21 +423,10 @@ impl SwapSaga<'_, Signed> {
...
-            .get_proofs_states(&self.state_data.ys)
-            .await?
-            .into_iter()
-            .collect::<Option<Vec<_>>>()
-            .ok_or(Error::UnexpectedProofState)?
-        {
-            check_state_transition(current_state, State::Spent)
-                .map_err(|_| Error::UnexpectedProofState)?;
-        }
+        let mut proofs = tx.get_proofs(&self.state_data.ys).await?;
```
thesimplekid: Should have compensate here as well or input proofs are stuck in pending.

crates/cdk/src/mint/swap/swap_saga/mod.rs
thesimplekid: Not introduced in this PR but noticed it. We should have the same pattern here on error. 
```rust
tx.add_completed_operation(
&self.state_data.operation,
&self.state_data.fee_breakdown.per_keyset,
)
.await?;
```

crates/cdk-sql-common/src/mint/mod.rs
crodas: Yes, I'm doing this right now.

crodas: Fixed in e5f358df02af5a4d1f9938a8069e8cfd50c7b23a

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: Fixed in 9dca0391846d48c3562133710f52887ddddd8792

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: Fixed in 0d761ef

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: Fixed in 0d761ef

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: Fixed in 0d761ef

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: Fixed in 0d761ef

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: Fixed in 0d761ef

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: Fixed in bbff7717

crates/cdk-sql-common/src/mint/proofs.rs
```diff
@@ -332,13 +346,62 @@ where
...
-    async fn get_proofs_states(
+    async fn get_proofs(
         &mut self,
         ys: &[PublicKey],
-    ) -> Result<Vec<Option<State>>, Self::Err> {
-        let mut current_states = get_current_states(&self.inner, ys, true).await?;
+    ) -> Result<Acquired<ProofsWithState>, Self::Err> {
+        if ys.is_empty() {
+            return Ok(ProofsWithState::new(vec![], State::Unspent).into());
+        }
```
thesimplekid, line 356: This seems a bit weird setting this to unspent, should we just error on an empty query?

crates/cdk-common/src/mint.rs
```diff
@@ -31,6 +33,87 @@ pub enum OperationKind {
...
+    /// # Errors
+    ///
+    /// Returns [`Error::UnexpectedProofState`] if the transition from the current
+    /// state to the new state is not permitted.
+    pub fn set_new_state(&mut self, new_state: State) -> Result<State, Error> {
+        check_state_transition(self.state, new_state).map_err(|_| Error::UnexpectedProofState)?;
+        let old_state = self.state;
+        self.state = new_state;
+        Ok(old_state)
+    }
```
thesimplekid: This introduces a third state update pattern to the codebase. Currently we have:

  1. MeltQuote (DB-first): State is updated in-memory after writing to [DB](https://github.com/cashubtc/cdk/blob/dd196a88048532796e3f35821e7cdfc31c5a329a/crates/cdk-sql-common/src/mint/quotes.rs#L927), ensuring the two are always in sync.
  2. MintQuote (change-tracking): Uses take_changes() to track and persist events. This is justified because MintQuote state is computed from payments/issuances rather than explicitly set.
  3. ProofsWithState (memory-first): Updates in-memory first via set_new_state(), then requires a separate update_proofs() call.

  Since proofs follow a state machine (Unspent → Pending → Spent) just like MeltQuote, it makes sense to use the same pattern:
```rust
  // Current
  proofs.set_new_state(State::Pending)?;
  tx.update_proofs(&mut proofs).await?;

  // Suggested (matches MeltQuote pattern)
  tx.update_proofs_state(&mut proofs, State::Pending).await?;
```

  This keeps persistence patterns consistent and eliminates the possibility of in-memory/DB state divergence if update_proofs() fails after set_new_state() succeeds.

crodas: @thesimplekid What about the changes in d8bd4fa17ac06d03f7496e5532178bb6227a7b8a ?

thesimplekid: There is still a difference here with how we handle melt quotes vs proofs. Melt quotes we check the transition within sql common(https://github.com/cashubtc/cdk/blob/44fee4c4d1903b3d0911014fbff3b78293c9d3ea/crates/cdk-sql-common/src/mint/quotes.rs#L898). Where the proofs we call check transition in the mint before calling update.

crodas: Yeah, this is duplicated and will be removed in another PR. I think the consensus was to move all checks off the storage layer. This was missed, I'll fix it later

crates/cdk/src/mint/swap/swap_saga/mod.rs
```diff
@@ -423,28 +416,22 @@ impl SwapSaga<'_, Signed> {
...
+                return Err(err.into());
             }
+        };
+        if check_state_transition(proofs.state, State::Spent).is_err() {
+            tx.rollback().await?;
+            self.compensate_all().await?;
+            return Err(Error::TokenAlreadySpent);
+        }
+
+        match tx.update_proofs_state(&mut proofs, State::Spent).await {
```
thesimplekid: I think we can do better then calling check state and then update. If we have a `set_state` fn on the `Acquired<ProofsWithState>` that takes a db tx. It can check its a valid transition and update the memory state and db state in one fn.  Same pattern could be applied to the MeltQuote, and a similar one even to MintQuote.

crates/cdk-common/src/mint.rs
thesimplekid: > this is duplicated

As far as I can tell we only call it there and not in the mint so its not a duplicate.

> I think the consensus was to move all checks off the storage layer.

Yes but we need to make sure they are not missed and the storage and memory are updated together (see review comment).

crates/cdk/src/mint/swap/swap_saga/mod.rs
```diff
@@ -177,15 +178,29 @@ impl<'a> SwapSaga<'a, Initial> {
...
+        if check_state_transition(new_proofs.state, State::Pending).is_err() {
             tx.rollback().await?;
-            return Err(match err {
-                database::Error::Duplicate => Error::TokenPending,
-                database::Error::AttemptUpdateSpentProof => Error::TokenAlreadySpent,
-                _ => Error::Database(err),
+            return Err(if original_state == State::Pending {
+                Error::TokenPending
+            } else {
+                Error::TokenAlreadySpent
```
thesimplekid: Can't we remove this check as its done in the `Mint::update_proofs_state`? We should make sure that method returns the correct error.

crates/cdk/src/mint/proofs.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// # Errors
+    ///
+    /// - [`Error::UnexpectedProofState`] if the state transition is invalid
+    /// - [`Error::TokenAlreadySpent`] if the database rejects the update (proofs already spent)
+    pub async fn update_proofs_state(
+        tx: &mut DynMintTransaction,
+        proofs: &mut Acquired<ProofsWithState>,
+        new_state: State,
+    ) -> Result<(), Error> {
+        check_state_transition(proofs.state, new_state).map_err(|_| Error::UnexpectedProofState)?;
```
thesimplekid: We should return the correct error based on the current state. 

Ex from what we did before
```rust
 if matches!(original_state, State::Pending | State::Spent) {
            tx.rollback().await?;
            return Err(if original_state == State::Pending {
                Error::TokenPending
            } else {
                Error::TokenAlreadySpent
            });
        }
```

crates/cdk/src/mint/proofs.rs
```diff
@@ -0,0 +1,263 @@
...
+                cdk_common::database::Error::AttemptUpdateSpentProof
+                | cdk_common::database::Error::AttemptRemoveSpentProof => Error::TokenAlreadySpent,
+                err => err.into(),
+            })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use cdk_common::mint::Operation;
```
thesimplekid, line 43: Nice adding tests.

crates/cdk/src/mint/melt/melt_saga/mod.rs
```diff
@@ -245,48 +245,22 @@ impl MeltSaga<Initial> {
...
+            tx.rollback().await?;
+            return Err(if original_state == State::Pending {
+                Error::TokenPending
+            } else {
+                Error::TokenAlreadySpent
+            });
+        }
 
-        if has_forbidden_state {
+        if let Err(err) = Mint::update_proofs_state(&mut tx, &mut proofs, State::Pending).await {
```
thesimplekid: Can't we remove this check as its done in the Mint::update_proofs_state? We should make sure that method returns the correct error.

crodas: Fixed in 36b0dc6e2d9644b0ab3ff8a5b48b50036e86797d

crates/cdk/src/mint/proofs.rs
crodas: Fixed in 36b0dc6e2d9644b0ab3ff8a5b48b50036e86797d

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: Fixed in 36b0dc6e2d9644b0ab3ff8a5b48b50036e86797d

PR #1511, opened 2026-01-09
fix: publish notifications after tx commit

asmogo:
### Description

We want to send subscribers updates after a transaction is committed. 

Similar to what we do [here](https://github.com/cashubtc/cdk/blob/main/crates/cdk/src/mint/issue/mod.rs#L727)

There’s still some refactoring to be done in `handle_mint_quote_payment` and `pay_mint_quote`.  

-----

### Notes to the reviewers

-----

#### FIXED
* State change updates for notification subscribers.

----

### Checklist


crodas (approved):
LGTM

PR #1522, opened 2026-01-12
Simplify subscription quote fetching by removing try_join_all

crodas:
### Description

Replace batch query execution with inline sequential processing for melt and mint quote lookups. This reduces code complexity by handling each quote type immediately within the match statement instead of collecting futures and joining them at the end.

This is a continuation of #1514

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/subscription.rs
+++ b/crates/cdk/src/mint/subscription.rs
@@ -76,12 +76,12 @@ impl MintPubSubSpec {
                     if let Some(mint_quote) =
                         self.get_mint_quote(uuid).await.map_err(|e| e.to_string())?
                     {
-                        let mint_quote = match mint_quote.payment_method.as_str() {
-                            "bolt11" => {
+                        let mint_quote = match idx {
+                            NotificationId::MintQuoteBolt11(_) => {
                                 let response: MintQuoteBolt11Response<QuoteId> = mint_quote.into();
                                 response.into()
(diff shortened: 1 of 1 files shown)


crates/cdk/src/mint/subscription.rs
```diff
@@ -57,75 +57,46 @@ impl MintPubSubSpec {
...
-                    mint_queries.push(self.get_mint_quote(uuid))
-                }
-                NotificationId::MeltQuoteBolt12(uuid) => {
-                    melt_queries.push(self.db.get_melt_quote(uuid))
+                NotificationId::MintQuoteBolt11(uuid) | NotificationId::MintQuoteBolt12(uuid) => {
+                    if let Some(mint_quote) =
+                        self.get_mint_quote(uuid).await.map_err(|e| e.to_string())?
+                    {
+                        let mint_quote = match mint_quote.payment_method.as_str() {
+                            "bolt11" => {
```
thesimplekid: Why do we match on the string can't we use the enum?

thesimplekid (changes requested):

crates/cdk/src/mint/subscription.rs
crodas: You were right, in the past the enum was lost, now we have it there. I'll update it.

PR #1529, opened 2026-01-14
feat(ffi): init_logging

asmogo:
### Description

improves logging. tested with go and swift bindings.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
Thanks looks good to me. I can test with android. I think we might need to add some stuff for that but can do that as a follow up.

PR #1530, opened 2026-01-14
remove flat_error

asmogo:
### Description

removes usage of `#[uniffi(flat_error)]`

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1346,6 +1346,8 @@ dependencies = [
  "serde_json",
  "thiserror 2.0.17",
  "tokio",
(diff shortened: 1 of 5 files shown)


thesimplekid:
> As a side note, https://github.com/cashubtc/cdk/pull/1529 also works with swift, but the log prefixes are somehow broken.

I've seen this before when opening in logs in a text editor so I don't think its something specefic to ffi or swift. 

We should address it as part of cleaning up our logging as I mentioned on one of the recent dev calls.

Issue #1532, opened 2026-01-14
Review ffi logging with android


thesimplekid:
We improved it here for swift and go but I think for android we need to do more then just log to stdout. https://github.com/cashubtc/cdk/pull/1529

https://github.com/rust-mobile/android_logger-rs

PR #1533, opened 2026-01-15
Android logging

thesimplekid:
### Description

-----

Having the `FfiError` with a message feild conflicts with kotlin code so rename it to avoid conflicts.

Removing the ANSI codes in the logging should resolve formatting in places that cannot handle it ie ios.

Add android log crate so logs are piped to logcat. 

closes: https://github.com/cashubtc/cdk/issues/1532

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,7 +84,7 @@ dependencies = [
  "amplify_derive",
  "amplify_num",
  "ascii",
- "getrandom 0.2.16",
+ "getrandom 0.2.17",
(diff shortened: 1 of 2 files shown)


asmogo:
swit logs are looking great. 

ack [b4bc921](https://github.com/cashubtc/cdk/pull/1533/commits/b4bc9213ff1130822c56dc8f9f2e11f17d99d48c)

PR #1540, opened 2026-01-19
feat: adding supabase db

asmogo:
### Description

Introducing supabase db to the cdk wallet. 
* added supabase crate
* moved oicd client to common crate
* auto refresh jwt token in the supabase crate
* exported constructor, wrapper and helper functions (set_jwt_token, set_refresh_token and set_token_expiration)

Todo: 

- [x] move migrations to common package + auto migrate 
- [x] add `call_rpc` function to invoke custom db rpc
- [x] cleanup 
- [x] remove `SetTokenExpiration` (set exp internally based on access token)
- [x] replace extra fields from other applications (dead_code) using `serde(flatten)` ?. 
- [x] test with supabase provided authentication 
- [x] make use of the ffi database macro `crate::impl_ffi_wallet_database!`

This example reuses the keycloak token for auth wallet and supabase.

My supabase is configured to use the jwt token from keycloak for user authentication and RBAC  (golang): 

```go 
// ProvideWallet creates the CDK FFI wallet instance with CAT token authentication
func ProvideWallet(params WalletParams) (WalletResult, error) {
	cfg := params.Config
	log := logger.WithComponent("wallet_provider")
	cdk_ffi.InitLogging("debug")

	ctx := context.Background()

	catToken, err := params.KeycloakClient.Login(ctx, cfg.Wallet.Username, cfg.Wallet.Password)
	if err != nil {
		return WalletResult{}, fmt.Errorf("failed to login to keycloak for wallet: %w", err)
	}
	log.Info("Obtained CAT token for wallet authentication")
	apiKey := "<my-supabase-public-anon-key>"

	wdb, err := cdk_ffi.WalletSupabaseDatabaseWithOidc("https://supabase.domain.com",
		apiKey,
		"https://auth.domain.com/realms/my-application-realm/.well-known/openid-configuration",
		&cfg.Keycloak.Verification.ClientID)

	if err != nil {
		return WalletResult{}, fmt.Errorf("failed to create supabase db concetion: %w", err)
	}
	wdb.SetJwtToken(&catToken.AccessToken)
	wdb.SetRefreshToken(&catToken.RefreshToken)
	
	exp := uint64(time.Now().Add(time.Duration(catToken.ExpiresIn)).Unix())
	wdb.SetTokenExpiration(&exp)

	wallet, err := cdk_ffi.NewWallet(
		cfg.Wallet.Mint,
		cdk_ffi.CurrencyUnitUsd{},
		cfg.Wallet.Mnemonic,
		wdb,
		cdk_ffi.WalletConfig{
			TargetProofCount: nil,
		},
	)

	if err != nil {
		return WalletResult{}, fmt.Errorf("failed to create wallet: %w", err)
	}
    // Set CAT token for wallet authentication (NUT-21 and NUT-22 auth wallet)
	if err := wallet.SetCat(catToken.AccessToken); err != nil {
		return WalletResult{}, fmt.Errorf("failed to set CAT token: %w", err)
	}

	// Set refresh token
	if err := wallet.SetRefreshToken(catToken.RefreshToken); err != nil {
		return WalletResult{}, fmt.Errorf("failed to set refresh token: %w", err)
	}

	return WalletResult{
		Wallet:          wallet,
	}, nil
}
```

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-supabase/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+tracing.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+lightning-invoice.workspace = true
+uuid.workspace = true
+reqwest = { workspace = true }
+base64 = "0.22"
+url.workspace = true
+once_cell.workspace = true
+hex = "0.4"
```
thesimplekid: Can we use the hex provided by cdk_common::util? 
```suggestion
```

crates/cdk-supabase/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+urlencoding = "2.1"
+cdk-sql-common = { workspace = true }
+aes-gcm = "0.10"
+rand = "0.8"
+pbkdf2 = "0.12"
+hmac = "0.12"
+sha2 = "0.10"
```
thesimplekid:
```suggestion
```

Do we only use sha256? Think we can get that from bitcoin.

crates/cdk-supabase/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+uuid.workspace = true
+reqwest = { workspace = true }
+base64 = "0.22"
+url.workspace = true
+once_cell.workspace = true
+hex = "0.4"
+urlencoding = "2.1"
+cdk-sql-common = { workspace = true }
+aes-gcm = "0.10"
+rand = "0.8"
```
thesimplekid:
```suggestion
rand.workspace = true
```

crates/cdk-supabase/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+bitcoin.workspace = true
+thiserror.workspace = true
+tokio = { workspace = true, features = ["rt-multi-thread"] }
+tracing.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+lightning-invoice.workspace = true
+uuid.workspace = true
+reqwest = { workspace = true }
+base64 = "0.22"
```
thesimplekid:
```suggestion
```

We can get this from bitcoin.

crates/cashu/Cargo.toml
```diff
@@ -11,10 +11,11 @@ license.workspace = true
 readme = "README.md"
 
 [features]
-default = ["mint", "wallet"]
+default = ["mint", "wallet", "auth"]
 swagger = ["dep:utoipa"]
 mint = []
 wallet = []
+auth = ["dep:regex"]
```
thesimplekid: I think we should just skip the feature and always include it. Both cdk mint and cdk wallet use the auth feature so in effect this feature ends up being always needed.

crates/cdk-common/Cargo.toml
```diff
@@ -17,6 +17,7 @@ test = []
 bench = []
 wallet = ["cashu/wallet", "dep:uuid"]
 mint = ["cashu/mint", "dep:uuid"]
+auth = ["cashu/auth", "dep:reqwest", "dep:jsonwebtoken", "http"]
```
thesimplekid: Same comment here both the mint and wallet enable this feature so think we should just include it.

crates/cdk-supabase/migrations/001_initial_schema.sql
```diff
@@ -0,0 +1,409 @@
...
+RETURNS void
+LANGUAGE plpgsql
+SECURITY DEFINER
+AS $$
+BEGIN
+    EXECUTE query;
+END;
+$$;
+
+GRANT EXECUTE ON FUNCTION exec_sql(TEXT) TO authenticated;
```
thesimplekid: Should this be granted to all authenticated users? Isn't this an issue in the multi tenant model?

crates/cdk-supabase/migrations/002_add_saga_and_operations.sql
thesimplekid, line 1: I think we need a new migration since this was merged https://github.com/cashubtc/cdk/pull/1761

crates/cdk-supabase/src/wallet.rs
```diff
@@ -0,0 +1,2742 @@
...
+                }
+                None => return Err(DatabaseError::ProofNotUnspent),
+            }
+
+            // Update proof state to Reserved with operation_id
+            let update = serde_json::json!({
+                "state": State::Reserved.to_string(),
+                "used_by_operation": op_id_str,
+            });
+            let patch_path = format!("rest/v1/proof?y=eq.{}", url_encode(&y_hex));
```
thesimplekid: I think this is not atomic as the get of the proof to check the state and the patch to update the state are separation operations without a transactions. In sql we account for this by on the update passing in the state Unspent as a filter and then checking the rows affected is what we expect. I think we need to do similar here?

crates/cdk-supabase/src/wallet.rs
```diff
@@ -0,0 +1,2742 @@
...
+                }
+            }
+            None => return Err(DatabaseError::UnknownQuote),
+        }
+
+        let update = serde_json::json!({
+            "used_by_operation": op_id_str,
+        });
+        let patch_path = format!("rest/v1/melt_quote?id=eq.{}", url_encode(quote_id));
+        let (status, response_text) = self.patch_request(&patch_path, &update).await?;
```
thesimplekid: This has the same issue as the reserve proofs where two operations can observe it as free and then update successfully so we should filter on the patch that the operation_id is null.

crates/cdk-supabase/src/wallet.rs
```diff
@@ -0,0 +1,2742 @@
...
+            )));
+        }
+        Ok(())
+    }
+
+    async fn add_mint_quote(&self, quote: MintQuote) -> Result<(), DatabaseError> {
+        let item: MintQuoteTable = quote.try_into()?;
+        let (status, response_text) = self
+            .post_request("rest/v1/mint_quote?on_conflict=id,wallet_id", &item)
+            .await?;
```
thesimplekid: This is an unconditional write and does not use the opportunistic locking that we use for SQL by checking the expected version. 

https://github.com/cashubtc/cdk/blob/1505652c49e0bd49567e9ecb540398762a093c79/crates/cdk-sql-common/src/wallet/mod.rs#L1103

crates/cdk-supabase/src/wallet.rs
```diff
@@ -0,0 +1,2742 @@
...
+            )));
+        }
+        Ok(())
+    }
+
+    async fn add_melt_quote(&self, quote: wallet::MeltQuote) -> Result<(), DatabaseError> {
+        let item: MeltQuoteTable = quote.try_into()?;
+        let (status, response_text) = self
+            .post_request("rest/v1/melt_quote?on_conflict=id,wallet_id", &item)
+            .await?;
```
thesimplekid: Same concern as the add_mint_quote

crates/cdk-supabase/src/wallet.rs
```diff
@@ -0,0 +1,2742 @@
...
+                        }
+                    }
+                }
+
+                // Decrypt C
+                if let Ok(encrypted_c) = hex::decode(&p.c) {
+                    if let Ok(decrypted_c) = self.decrypt(&encrypted_c).await {
+                        p.c = hex::encode(decrypted_c);
+                    }
+                }
```
thesimplekid: Ref

crates/cdk-supabase/src/wallet.rs
```diff
@@ -0,0 +1,2742 @@
...
+        let (status, text) = self.get_request(&path).await?;
+
+        if !status.is_success() {
+            return Err(DatabaseError::Internal(format!(
+                "get_reserved_proofs failed: HTTP {}",
+                status
+            )));
+        }
+
+        if let Some(items) = Self::parse_response::<ProofTable>(&text)? {
```
thesimplekid, line 1942: Do we need to decrypt here in other get proof fns we do for example https://github.com/cashubtc/cdk/pull/1540/changes#r2994674273.

crates/cdk-supabase/migrations/001_initial_schema.sql
asmogo: You are right. This will bypass all RLS rules.

crates/cdk-supabase/migrations/005_add_p2pk_signing_key.sql
thesimplekid, line 1: We should bump the schema version to 5 here technically right?

asmogo: Thanks. Bumped the schema version based on the migration readme. Now I will try to set up a fresh wallet db with the supabase cli to verify that this is actually working.

asmogo: I had to rename the migration files and separate the functions from the schema to make this compatible with supabase CLI.

crates/cdk/src/wallet/payment_request.rs
```diff
@@ -133,7 +133,7 @@ impl Wallet {
                             .await
                             .map_err(|e| Error::Custom(format!("Publish Nostr event: {e}")))?;
 
-                        println!(
```
asmogo, line 136: I am not sure why changes were made to this file.

thesimplekid: its okay looks like just replacing print for tracing which is more correct.

Issue #1552, opened 2026-01-21
Add mint version that generates keyset to the keysets table


thesimplekid:
Mostly just as metadata that could be useful if we change derivation method.

PR #1556, opened 2026-01-22
feat: track cdk version in keysets

a1denvalu3:
- Adds `cdk_version` column to `keyset` table (Postgres/SQLite).
- Populates field with `env!("CARGO_PKG_VERSION")` on keyset generation.
- Updates internal structs and proto definitions.
- Allows internal tracking of CDK version per keyset.
- 
closes: https://github.com/cashubtc/cdk/issues/1552

### DIFF
--- a/crates/cdk-common/src/common.rs
+++ b/crates/cdk-common/src/common.rs
@@ -268,10 +268,10 @@ impl std::str::FromStr for IssuerVersion {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let (implementation, version_str) = match s.split_once('/') {
-            Some((impl_str, ver_str)) => (impl_str.to_string(), ver_str),
-            None => ("cdk".to_string(), s),
-        };
+        let (implementation, version_str) = s
+            .split_once('/')
+            .ok_or(Error::Custom(format!("Invalid version string: {}", s)))?;
+        let implementation = implementation.to_string();
 
         let parts: Vec<&str> = version_str.splitn(3, '.').collect();
         if parts.len() != 3 {
@@ -481,14 +481,6 @@ mod tests {
 
     #[test]
     fn test_version_parsing() {
-        // Test legacy format (defaults to cdk)
-        let v = IssuerVersion::from_str("0.1.0").unwrap();
-        assert_eq!(v.implementation, "cdk");
-        assert_eq!(v.major, 0);
-        assert_eq!(v.minor, 1);
-        assert_eq!(v.patch, 0);
-        assert_eq!(v.to_string(), "cdk/0.1.0");
-
         // Test explicit cdk format
         let v = IssuerVersion::from_str("cdk/1.2.3").unwrap();
         assert_eq!(v.implementation, "cdk");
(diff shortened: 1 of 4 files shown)


crates/cdk-common/src/common.rs
```diff
@@ -368,3 +435,112 @@ pub struct FeeReserve {
     /// Percentage expected fee
     pub percent_fee_reserve: f32,
 }
+
+/// CDK Version
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct CdkVersion {
+    /// Implementation name (e.g., "cdk", "nutshell")
+    pub implementation: String,
```
thesimplekid: Can we move these above the test mod? The test mod should be last in the file.

crates/cdk/src/cdk_version_test.rs
thesimplekid, line 1: Can we include these tests in the test mod of crates/cdk-common/src/common.rs where the CdkVersion is defined, we shouldn't need a separate mod here.

crates/cdk-signatory/src/proto/signatory.proto
```diff
@@ -64,6 +64,7 @@ message KeySet {
   Keys keys = 5;
   optional uint64 final_expiry = 6;
   uint64 version = 7;
+  optional string cdk_version = 8;
```
thesimplekid: I think this should just be version? In theory this should be general for all signatory though the spec is not merged. @lescuer97

a1denvalu3: just "version" would be confusing.

thesimplekid: Yeah true, software version? I'm not sure just something not cdk specefic I think.

lescuer97: how version is used right now is to specify the order of the keysets. 

`mint_version` is probably fine.

lescuer97: `issuer` would be a good name here

a1denvalu3: `IssuerVersion` maybe?

crates/cdk-signatory/src/common.rs
```diff
@@ -155,6 +157,7 @@ pub fn create_new_keyset<C: secp256k1::Signing>(
         derivation_path_index,
         amounts: amounts.to_owned(),
         input_fee_ppk,
+        issuer_version: IssuerVersion::from_str(env!("CARGO_PKG_VERSION")).ok(),
```
thesimplekid: This will be just the version number and not include the implementation

crates/cdk-common/src/common.rs
```diff
@@ -204,6 +204,124 @@ impl Default for QuoteTTL {
...
+    }
+}
+
+impl std::str::FromStr for IssuerVersion {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (implementation, version_str) = match s.split_once('/') {
+            Some((impl_str, ver_str)) => (impl_str.to_string(), ver_str),
+            None => ("cdk".to_string(), s),
```
thesimplekid: I'm not sure it makes sense to default to cdk. For example a nutshell migration would only have the numbers and with this it would default to cdk.

a1denvalu3: > For example a nutshell migration would only have the numbers and with this it would default to cdk.

I have changed the migration script to include the implementation prefix

thesimplekid (approved):
Thank you 

ACK 54fbd5b490f5127344b2beb85a65828c5c82cc95
@lescuer97 Note for the signatory spec the version is of the format String/0.0.0

PR #1569, opened 2026-01-25
feat(wallet): fetch_mint_quote

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut06.rs
+++ b/crates/cashu/src/nuts/nut06.rs
@@ -9,9 +9,10 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use super::nut01::PublicKey;
 use super::nut17::SupportedMethods;
 use super::nut19::CachedEndpoint;
-use super::{nut04, nut05, nut15, nut19, MppMethodSettings};
-#[cfg(feature = "auth")]
-use super::{AuthRequired, BlindAuthSettings, ClearAuthSettings, ProtectedEndpoint};
+use super::{
+    nut04, nut05, nut15, nut19, AuthRequired, BlindAuthSettings, ClearAuthSettings,
+    MppMethodSettings, ProtectedEndpoint,
+};
 use crate::util::serde_helpers::deserialize_empty_string_as_none;
 use crate::CurrencyUnit;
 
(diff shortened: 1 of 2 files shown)


crates/cdk/src/wallet/issue/bolt11.rs
```diff
@@ -157,6 +157,42 @@ impl Wallet {
...
+                // Update the state of the existing quote
+                existing_quote.state = response.state;
+                existing_quote
+            }
+            None => {
+                // Create a new quote from the mint response
+                MintQuote::new(
+                    quote_id.to_string(),
+                    self.mint_url.clone(),
+                    PaymentMethod::Known(KnownMethod::Bolt11),
```
thesimplekid: I think we should have the fn accept a payment method? This way we can determine  what method it is and what endpoint to call.

asmogo: fixed https://github.com/cashubtc/cdk/pull/1569/commits/1de68e4de6b47c46236a8bb5c3ebe361a020e60c

crates/cdk/src/wallet/issue/mod.rs
```diff
@@ -69,4 +70,122 @@ impl Wallet {
...
+
+                match existing_quote {
+                    Some(mut existing) => {
+                        // Update the existing quote with new state
+                        existing.state = response.state;
+                        existing
+                    }
+                    None => {
+                        // Create a new quote from the response
+                        MintQuote::new(
```
thesimplekid, line 460: Maybe this is an issue we need to review for our custom method. But I think for custom we should use the amount_paid and amount_issued api. It can be used for more method kinds (ie onchain) as they are repayable and the state should only be used for the bolt12 as a legacy thing.

asmogo: > the state should only be used for the bolt12 as a legacy thing.

you're talking about bolt11, right?

thesimplekid: Yes sorry bolt11.

thesimplekid (approved):

PR #1570, opened 2026-01-25
feat(ffi): npubx.cash

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -534,9 +534,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.15.3"
+version = "1.15.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


crates/cdk-ffi/src/npubcash.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/// Parse a Nostr secret key from either hex or nsec format
+fn parse_nostr_secret_key(key: &str) -> Result<nostr_sdk::Keys, FfiError> {
+    // Try parsing as nsec (bech32) first
+    if key.starts_with("nsec") {
+        nostr_sdk::Keys::parse(key)
+            .map_err(|e| FfiError::internal(format!("Invalid nsec key: {}", e)))
+    } else {
+        // Try parsing as hex
```
thesimplekid, line 250: NIT: These comments aren't needed as they are evident by the code. In general I think comments should should tell the "why" something is done if it is not evident, not the what as this should be evident from the code. Of course this is not a strict rule, but I've noticed my llm and over-commenting a bit and we should be careful to ensure the comment are actually useful.

asmogo: I regularly remove generated comments as they are often very verbose. 
Thanks. I'll try to take extra care next time.

Issue #1543, opened 2026-01-19
Review if we can remove reqwest for bitreq


thesimplekid:
We should consider moving to btitreq if we can. It is a newer http lib with the goal of having minimal dependencies and keeping an msrv both of which reqwest are not true for reqwest.  It was created by a maintainer of rust-bitcoin and other rust bitcoin projects have expressed intrest in moving to it. 

https://docs.rs/bitreq/latest/bitreq/

thesimplekid:
We support bitreq by default now

PR #1573, opened 2026-01-26
Abstract HTTP client behind cdk_common::HttpClient

crodas:
This code is just a chore that would make #1543 and #1567 much easier, since all the   HTTP code lives in a single place.

Introduce a unified HttpClient wrapper in cdk-common that abstracts reqwest usage across the codebase. This provides a cleaner API and centralizes HTTP client configuration.

Changes:
- Add http module to cdk-common with HttpClient, HttpClientBuilder, and supporting types for requests/responses/errors
- Update cdk examples to use cdk_common::HttpClient instead of reqwest::Client directly
- Update cdk-cli, cdk-fake-wallet, cdk-integration-tests, and cdk-npubcash to use the new HttpClient
- Simplify wallet transport and OIDC client to use HttpClient
- Remove direct reqwest dependency from cdk dev-dependencies

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -348,6 +348,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert-json-diff"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e4f2b81832e72834d7518d8487a0396a28cc408186a2e8854c0f98011faf12"
+dependencies = [
+ "serde",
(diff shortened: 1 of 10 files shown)


crates/cdk-http-client/README.md
thesimplekid, line 1: Can we just add a note that this crate is only meant to be consumed internally by cdk so we don't make guarantees about breaking the API.

PR #1580, opened 2026-01-27
fix: just quick-check

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cashu/src/nuts/nut26/encoding.rs
```diff
@@ -883,1297 +882,6 @@ mod tests {
         );
     }
 
-    #[test]
```
thesimplekid, line 886: Don't think these should have been removed?

PR #1581, opened 2026-01-28
feat: add wallet trait

crodas:
### Description

Introduce a modular trait system for wallet implementations that enables flexible composition of wallet capabilities. This allows different wallet implementations to share a common interface while supporting selective trait implementation.

The trait hierarchy consists of:
- WalletTypes: Base trait defining associated types (Amount, Proofs, etc.)
- WalletBalance: Balance query operations
- WalletMintInfo: Mint information and keyset operations
- WalletMint: Minting operations (payment -> tokens)
- WalletMelt: Melting operations (tokens -> payment)
- WalletReceive: Token receiving
- WalletProofs: Proof state management
- Wallet: Super-trait combining all capabilities with blanket impl

Implements all traits for both cdk::Wallet and cdk-ffi::Wallet, enabling generic code that works with any wallet implementation.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/wallet/mod.rs
+++ b/crates/cdk-common/src/wallet/mod.rs
@@ -917,11 +917,20 @@ pub trait Wallet: Send + Sync {
         network: bitcoin::Network,
     ) -> Result<Self::MeltQuote, Self::Error>;
 
-    /// Check a mint quote status (alias for `check_mint_quote_status`)
-    async fn check_mint_quote(
+    /// Get a melt quote for a human-readable address (alias for `melt_human_readable_quote`)
+    #[cfg(all(feature = "bip353", not(target_arch = "wasm32")))]
+    async fn melt_human_readable(
         &self,
-        quote_id: &str,
-    ) -> Result<Self::MintQuote, Self::Error> {
+        address: &str,
+        amount_msat: Self::Amount,
+        network: bitcoin::Network,
+    ) -> Result<Self::MeltQuote, Self::Error> {
+        self.melt_human_readable_quote(address, amount_msat, network)
+            .await
+    }
+
+    /// Check a mint quote status (alias for `check_mint_quote_status`)
+    async fn check_mint_quote(&self, quote_id: &str) -> Result<Self::MintQuote, Self::Error> {
         self.check_mint_quote_status(quote_id).await
     }
 
--- a/crates/cdk/src/wallet/mod.rs
+++ b/crates/cdk/src/wallet/mod.rs
@@ -74,7 +74,7 @@ pub use bip321::{
 };
 pub use builder::WalletBuilder;
 pub use cdk_common::wallet as types;
-pub use cdk_common::wallet::{ReceiveOptions, SendMemo, SendOptions, Wallet as WalletTrait};
+pub use cdk_common::wallet::{ReceiveOptions, SendMemo, SendOptions};
 pub use keysets::KeysetFilter;
 pub use melt::{MeltConfirmOptions, MeltOutcome, PendingMelt, PreparedMelt};
 pub use mint_connector::transport::Transport as HttpTransport;
--- a/crates/cdk/src/wallet/wallet_trait.rs
+++ b/crates/cdk/src/wallet/wallet_trait.rs
@@ -12,11 +12,11 @@ use cdk_common::nuts::{
     AuthProof, CurrencyUnit, Id, KeySetInfo, MeltOptions, MintInfo, PaymentMethod, Proofs,
     SpendingConditions,
 };
+use cdk_common::subscription::WalletParams;
 use cdk_common::wallet::{
     MeltQuote, MintQuote, ReceiveOptions, Restored, SendOptions, Transaction, TransactionDirection,
     TransactionId, Wallet as WalletTrait,
 };
-use cdk_common::subscription::WalletParams;
 use cdk_common::Amount;
 use tracing::instrument;
 use uuid::Uuid;
@@ -319,10 +319,7 @@ impl WalletTrait for super::Wallet {
     }
 
     #[instrument(skip(self, params))]
-    async fn subscribe(
-        &self,
-        params: WalletParams,
-    ) -> Result<ActiveSubscription, Self::Error> {
+    async fn subscribe(&self, params: WalletParams) -> Result<ActiveSubscription, Self::Error> {
         self.subscribe(params).await
     }
 
--- a/crates/cdk-ffi/src/wallet.rs
+++ b/crates/cdk-ffi/src/wallet.rs
@@ -258,10 +258,7 @@ impl Wallet {
     /// **Note:** The mint quote must be known to the wallet (stored locally) for this
     /// function to work. If the quote is not stored locally, use `fetch_mint_quote`
     /// instead.
-    pub async fn check_mint_quote_status(
-        &self,
-        quote_id: String,
-    ) -> Result<MintQuote, FfiError> {
+    pub async fn check_mint_quote_status(&self, quote_id: String) -> Result<MintQuote, FfiError> {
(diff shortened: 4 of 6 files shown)


crates/cdk/src/wallet/traits.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    async fn total_reserved_balance(&self) -> Result<Self::Amount, Self::Error> {
+        self.total_reserved_balance().await
+    }
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+impl WalletMintInfo for Wallet {
```
thesimplekid: Do we see a use case where we would want multiple traits to implement only part of the api? I don't think I see one and one trait would be simpler.

crodas: Is it any better at 1a2fd1fcd960696fd5fb3beb0ac60921b9d4bf32?

thesimplekid: i don't see how this change addresses the comment of 1 wallet trait vs the multiple it still has multiple. To be clear I not strongly against having multiple traits if there is a reason for it, just one trait would be simpler but if others see a use case for the multiple I'm okay with that.

thesimplekid:
@crodas Now that we have released 0.15 I think we should turn our attention back to this and get it merged. Looks like it needs a rebase is there anything else that is outstanding with it? 

cc @ye0man

crates/cdk/src/wallet/wallet_trait.rs
```diff
@@ -0,0 +1,599 @@
...
+                *ttl
+            })
+            .await?
+            .mint_info
+            .clone();
+
+        Ok(mint_info)
+    }
+
+    // === Keysets (moved from keysets.rs) ===
```
thesimplekid: We don't need comments like this
```suggestion
```

crates/cdk/src/wallet/wallet_trait.rs
```diff
@@ -0,0 +1,599 @@
...
+            })
+            .await?
+            .active_keysets
+            .iter()
+            .min_by_key(|k| k.input_fee_ppk)
+            .map(|ks| (**ks).clone())
+            .ok_or(Error::NoActiveKeyset)
+    }
+
+    // === Minting (moved from issue/mod.rs) ===
```
thesimplekid:
```suggestion
```

crates/cdk/src/wallet/wallet_trait.rs
```diff
@@ -0,0 +1,599 @@
...
+
+    fn mint_url(&self) -> MintUrl {
+        self.mint_url.clone()
+    }
+
+    fn unit(&self) -> CurrencyUnit {
+        self.unit.clone()
+    }
+
+    // === Balance (moved from balance.rs) ===
```
thesimplekid:
```suggestion
```

crates/cdk/src/wallet/wallet_trait.rs
```diff
@@ -0,0 +1,599 @@
...
+    async fn total_balance(&self) -> Result<Amount, Self::Error> {
+        let balance = self
+            .localstore
+            .get_balance(
+                Some(self.mint_url.clone()),
+                Some(self.unit.clone()),
+                Some(vec![State::Unspent]),
+            )
+            .await?;
+        Ok(Amount::from(balance))
```
thesimplekid: This trait impl follows two different patterns: functions like `total_balance` have their logic inline here, while functions like `mint()` delegate to inherent methods on `Wallet`. We should unify on the delegation pattern. Move the inline logic to inherent methods on `Wallet`, and have the trait impl always delegate. This way downstream consumers can call methods on the concrete `Wallet` without needing to use `WalletTrait`.

crates/cdk/src/wallet/balance.rs
```diff
@@ -1,34 +1,2 @@
...
-    }
-
-    /// Total reserved balance
-    #[instrument(skip(self))]
-    pub async fn total_reserved_balance(&self) -> Result<Amount, Error> {
-        Ok(self.get_reserved_proofs().await?.total_amount()?)
-    }
-}
+// Balance methods have been moved to the WalletTrait implementation
+// in wallet_trait.rs. Import `use cdk::wallet::WalletTrait;` to use them.
```
thesimplekid: Empty files like this with a comment shouldn't be left in, but per my other comment we should have fns here.

crates/cdk/examples/auth_wallet.rs
thesimplekid, line 1: Using delegation these examples should not need to import `WalletTrait`

crates/cdk/src/wallet/mod.rs
```diff
@@ -74,6 +74,7 @@ pub use bip321::{
 };
 pub use builder::WalletBuilder;
 pub use cdk_common::wallet as types;
+pub use cdk_common::wallet::{ReceiveOptions, SendMemo, SendOptions, Wallet as WalletTrait};
```
thesimplekid:
```suggestion
pub use cdk_common::wallet::{ReceiveOptions, SendMemo, SendOptions};
```

I don't think we should re-export and promote the trait to be used publicity. The trait should be used internally by use to keep the ffi inline but public use should not be encouraged.

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -285,6 +290,110 @@ impl MintQuote {
...
+    pub proofs_to_swap: Proofs,
+    /// Swap fee
+    pub swap_fee: Amount,
+    /// Input fee
+    pub input_fee: Amount,
+    /// Input fee without swap
+    pub input_fee_without_swap: Amount,
+    /// Metadata
+    pub metadata: HashMap<String, String>,
+}
```
thesimplekid: Having these owned types is required for the trait and ffi but it has the downside of adding extra clones since we cannot borrow. This is one reason I suggest we dont encourage the public use of the trait and the Wallet should be used directly.

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -562,6 +671,242 @@ impl FromStr for OperationKind {
...
+
+    /// Prepare a melt operation
+    async fn prepare_melt(
+        &self,
+        quote_id: &str,
+        metadata: HashMap<String, String>,
+    ) -> Result<Self::PreparedMelt, Self::Error>;
+
+    /// Prepare a melt operation with specific proofs
+    async fn prepare_melt_proofs(
```
thesimplekid, line 821:
```suggestion
    async fn prepare_melt_data(
        &self,
        quote_id: &str,
        metadata: HashMap<String, String>,
    ) -> Result<Self::PreparedMelt, Self::Error>;

    /// Prepare a melt operation with specific proofs
    async fn prepare_melt_proofs_data(
```

We should name this fns so its clear they use the ones with the owned type and not the borrowed like the rust api can.

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -562,6 +671,242 @@ impl FromStr for OperationKind {
...
+    async fn receive_proofs(
+        &self,
+        proofs: Proofs,
+        options: ReceiveOptions,
+        memo: Option<String>,
+        token: Option<String>,
+    ) -> Result<Self::Amount, Self::Error>;
+
+    /// Prepare a send transaction
+    async fn prepare_send(
```
thesimplekid, line 775:
```suggestion
    async fn prepare_send_data(
```

crodas:
@thesimplekid  Does a473c0bd889417cf5199090659000de67b28769d make it better than your suggestion? I think so, lmk what you think so we can merge this PR.

PR #1582, opened 2026-01-28
feat(wallet): remove multi mint wallet

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/wallet_repository.rs
+++ b/crates/cdk/src/wallet/wallet_repository.rs
@@ -427,7 +427,7 @@ impl WalletRepository {
         let mut by_unit: BTreeMap<CurrencyUnit, cdk_common::Amount> = BTreeMap::new();
         for (key, amount) in balances {
             let entry = by_unit.entry(key.unit).or_insert(cdk_common::Amount::ZERO);
-            *entry = *entry + amount;
+            *entry += amount;
         }
         Ok(by_unit)
     }


crates/cdk/src/wallet/wallet_repository.rs
```diff
@@ -0,0 +1,961 @@
...
+        Ok(balances)
+    }
+    /// Get total balance across all wallets
+    #[instrument(skip(self))]
+    pub async fn total_balance(&self) -> Result<cdk_common::Amount, Error> {
+        let balances = self.get_balances().await?;
+        Ok(balances
+            .values()
+            .fold(cdk_common::Amount::ZERO, |acc, &x| acc + x))
+    }
```
thesimplekid: This could sum different units since it just takes the value of the `get_balances` BTreeMap without checking the unit. It could return a BTreeMap of <Unit, Amount>?

crates/cdk/src/wallet/wallet_repository.rs
```diff
@@ -0,0 +1,961 @@
...
+    pub amount_sent: cdk_common::Amount,
+    /// Amount received at target
+    pub amount_received: cdk_common::Amount,
+    /// Fees paid
+    pub fees_paid: cdk_common::Amount,
+    /// Source balance after transfer
+    pub source_balance_after: cdk_common::Amount,
+    /// Target balance after transfer
+    pub target_balance_after: cdk_common::Amount,
+}
```
thesimplekid: Are these actually used, I do not see where they are?

crates/cdk/src/wallet/wallet_repository.rs
```diff
@@ -0,0 +1,961 @@
...
+            wallets: Arc::new(RwLock::new(BTreeMap::new())),
+            proxy_config: None,
+            shared_tor_transport: Some(TorAsync::new()),
+        };
+
+        // Automatically load wallets from database
+        wallet.load_wallets().await?;
+
+        Ok(wallet)
+    }
```
thesimplekid: We should remove the multiple news and instead use a builder pattern like we do in other places.

crates/cdk/src/wallet/wallet_repository.rs
```diff
@@ -0,0 +1,961 @@
...
+        if !wallets.contains_key(&key) {
+            return Err(Error::UnknownWallet(key));
+        }
+
+        // Check if this is the last wallet for the mint
+        let is_last_wallet = wallets.keys().filter(|k| k.mint_url == mint_url).count() == 1;
+
+        if is_last_wallet {
+            self.localstore.remove_mint(mint_url).await?;
+        }
```
thesimplekid: Would it be safer to not remove the wallet from the db? I think having this case of it sometimes removes from the db isn't safe and that should be explicit.

crates/cdk/src/wallet/wallet_repository.rs
```diff
@@ -0,0 +1,961 @@
...
+    }
+
+    /// Add a mint to the repository
+    ///
+    /// Fetches the mint info to discover all supported currency units and creates
+    /// a wallet for each unit. Returns all created wallets.
+    #[instrument(skip(self))]
+    pub async fn add_mint(&self, mint_url: MintUrl) -> Result<Vec<Wallet>, Error> {
+        self.add_mint_with_config(mint_url, None).await
+    }
```
thesimplekid: There is some inconsistency with the naming of the API in a few places we use `add_mint` but then to get the wallet we say `get_wallet`. I think we should align this and it should be `add_wallet`.

crates/cdk/src/wallet/wallet_repository.rs
```diff
@@ -0,0 +1,961 @@
...
+        &self,
+        mint_url: &MintUrl,
+        unit: CurrencyUnit,
+    ) -> Result<Wallet, Error> {
+        if let Ok(wallet) = self.get_wallet(mint_url, &unit).await {
+            return Ok(wallet);
+        }
+
+        self.create_wallet(mint_url.clone(), unit, None).await
+    }
```
thesimplekid: Should we remove this and just force the caller to be explicit about if they are getting a wallet or creating one. In cdk-cli we already have an additional wrapper of a fn by the same name that does this, I think it makes more sense to be done like that.

thesimplekid (changes requested):

PR #1599, opened 2026-02-03
refactor: remove `auth` feature flag from codebase

### LINKED ISSUE #1199: Remove auth feature and just always include auth fns
I think it is best we just remove the auth feature and include that functionality by default. It does not add too much extra code or deps and reduces our tests matrix by reducing features. Currenly in our mintd bins we just include auth anyway and disable it at run time.

crodas:
### Description

Fixes #1199 

The `auth` feature was gating authentication code (NUT-21/NUT-22) behind a compile-time flag across the entire workspace. This removes the feature entirely, making auth code always compiled. This simplifies the build matrix and eliminates ~400 lines of dual code paths (`#[cfg(feature = "auth")]` / `#[cfg(not(feature = "auth"))]`).

Changes across 43 files in 13 crates:
- Remove `auth` feature definitions from all Cargo.toml files
- Make `strum`, `strum_macros`, `regex` (cashu) and `jsonwebtoken` (cdk) unconditional dependencies
- Remove all `#[cfg(feature = "auth")]` gates from source files
- Merge dual code paths, keeping auth versions and deleting non-auth fallbacks
- Simplify combined cfgs (e.g. `#[cfg(all(feature = "auth", feature = "postgres"))]` → `#[cfg(feature = "postgres")]`)
- Remove auth-specific CI checks from flake.nix

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/mint_metadata_cache.rs
+++ b/crates/cdk/src/wallet/mint_metadata_cache.rs
@@ -34,7 +34,7 @@ use cdk_common::database::{self, WalletDatabase};
 use cdk_common::mint_url::MintUrl;
 use cdk_common::nuts::{KeySetInfo, Keys};
 use cdk_common::parking_lot::RwLock;
-use cdk_common::{KeySet, MintInfo};
+use cdk_common::{CurrencyUnit, KeySet, MintInfo};
 use tokio::sync::Mutex;
 use web_time::Instant;
 
@@ -589,14 +589,18 @@ impl MintMetadataCache {
             if let std::collections::hash_map::Entry::Vacant(e) =
                 new_metadata.keys.entry(keyset_info.id)
             {
-                let keyset = if let Some(client) = client.as_ref() {
-                    client.get_mint_keyset(keyset_info.id).await?
-                } else if let Some(auth_client) = auth_client.as_ref() {
+                let keyset = if keyset_info.unit == CurrencyUnit::Auth {
                     auth_client
+                        .as_ref()
+                        .ok_or(Error::Internal)?
                         .get_mint_blind_auth_keyset(keyset_info.id)
                         .await?
                 } else {
-                    return Err(Error::Internal);
+                    client
+                        .as_ref()
+                        .ok_or(Error::Internal)?
+                        .get_mint_keyset(keyset_info.id)
+                        .await?
                 };
 
                 // Verify the keyset ID matches the keys


crates/cdk/src/wallet/mint_metadata_cache.rs
```diff
@@ -601,17 +591,11 @@ impl MintMetadataCache {
             {
                 let keyset = if let Some(client) = client.as_ref() {
                     client.get_mint_keyset(keyset_info.id).await?
+                } else if let Some(auth_client) = auth_client.as_ref() {
+                    auth_client
+                        .get_mint_blind_auth_keyset(keyset_info.id)
+                        .await?
                 } else {
```
thesimplekid: Should this be an else if like this? Shouldn't we check what unit the keyset is and if its auth use auth client if not use the other client?

crates/cdk-mintd/src/env_vars/mod.rs
```diff
@@ -76,26 +74,22 @@ impl Settings {
...
+        // Parse auth database configuration from environment variables
+        self.auth_database = Some(crate::config::AuthDatabase {
+            postgres: Some(
+                self.auth_database
+                    .clone()
+                    .unwrap_or_default()
+                    .postgres
+                    .unwrap_or_default()
+                    .from_env(),
+            ),
```
thesimplekid, line 88: Should it always set to Some?

crodas: I think the reason why it is needed is to go give a meaningful error when it is `None` from the config, https://github.com/crodas/cdk/blob/feauture/remove-wallet-feature-auth/crates/cdk-mintd/src/lib.rs#L735-L737

But I can make it not an Option

crates/cdk/src/wallet/mint_metadata_cache.rs
crodas: Is it better in bd6b77226cde7fabb5ff91e26bf8e2e7dc4bab6b ?

thesimplekid: Yes thanks.

crates/cdk-mintd/src/env_vars/mod.rs
thesimplekid: Lets leave it for now since that's how it was and focus on just removing the feature and we can follow up.

PR #1608, opened 2026-02-07
Add bitreq

lescuer97:
### Description
use bitreq for the non wasm targets and use reqwest for wasm. in the http client crate.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


lescuer97:
okey, just to clear thing up.  I will  then keep this code but add back the features for `bitreq` and `reqwest`.  

and reimplement the trait for reqwest. 

does this sound correct?

thesimplekid:
Rebasing on https://github.com/cashubtc/cdk/pull/1894 will pin the correct deps for msrv. 

If you could also squash this branch.

crodas (dismissed):

crodas:
> Verified findings approved for disclosure:
> 
> * **Tor BIP353 DNS lookups bypass the Tor transport** (high) - Tor users resolving BIP353 payment addresses leak the queried DNS name over the local network instead of routing it through Tor, compromising the privacy/anonymity expected from `TorHttpClient`.
>   Unanchored locations included in summary:
>   
>   * crates/cdk/src/wallet/mint_connector/http_client.rs:283

This should be addressed in another PR

crates/cdk-http-client/Cargo.toml
```diff
@@ -19,17 +39,25 @@ tracing.workspace = true
 url.workspace = true
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-reqwest = { workspace = true }
-regex = { workspace = true }
+bitreq = { version = "0.3.4", features = ["async", "async-https-rustls", "json-using-serde", "proxy"], optional = true }
```
thesimplekid: Should add default features false

crates/cdk-cli/src/sub_commands/cat_device_login.rs
```diff
@@ -6,6 +6,7 @@ use cdk::mint_url::MintUrl;
 use cdk::nuts::MintInfo;
 use cdk::wallet::WalletRepository;
 use cdk::OidcClient;
+use cdk_http_client::RequestBuilderExt;
```
thesimplekid: We should not have to import it from cdk as downstream users should only need cdk and not our other crates so I think we need to add a rexport in cdk?

crates/cdk-http-client/src/request_builder_ext.rs
thesimplekid, line 1: I think this extension trait add more complexity then is need and we can do without it. Would resolve https://github.com/cashubtc/cdk/pull/1608/changes#r3451434098

https://github.com/lescuer97/cdk/pull/19

crates/cdk-http-client/src/transport.rs
```diff
@@ -0,0 +1,359 @@
...
+
+#[cfg(all(feature = "bitreq", not(target_arch = "wasm32")))]
+/// Bitreq-backed transport implementation.
+pub type BitreqTransport = Async;
+
+#[cfg(all(feature = "reqwest", not(target_arch = "wasm32")))]
+/// Reqwest-backed transport implementation.
+pub type ReqwestTransport = Async;
+
+#[cfg(feature = "tor")]
```
thesimplekid: This both violates the [code style](https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md#sub-modules) and does not compile `cargo check -p cdk-http-client --features tor --all-targets`

This needs to fixed since hyper was updated and added to ci.

thesimplekid (changes requested):
> > Verified findings approved for disclosure:
> > 
> > * **Tor BIP353 DNS lookups bypass the Tor transport** (high) - Tor users resolving BIP353 payment addresses leak the queried DNS name over the local network instead of routing it through Tor, compromising the privacy/anonymity expected from `TorHttpClient`.
> >   Unanchored locations included in summary:
> >   
> >   * crates/cdk/src/wallet/mint_connector/http_client.rs:283
> 
> This should be addressed in another PR

This introduces a regression in this PR then before this PR the bip353 delegated to the transport so it would use tor. So why allow the introduction of  a regression to be fixed later? 

There are three bot reviews that do not seem to be addressed? Do they require code changes or over eager review? 

- [x] https://github.com/cashubtc/cdk/pull/1608#discussion_r3439293423
- [x] https://github.com/cashubtc/cdk/pull/1608#discussion_r3439741459
- [x] https://github.com/cashubtc/cdk/pull/1608#discussion_r3435613396

crates/cdk-http-client/src/request_builder_ext.rs
crodas: Merged your PR in 58aab06d511ddbb24f09b810268ec8134c4521c4

crodas:
> > > Verified findings approved for disclosure:
> > > 
> > > * **Tor BIP353 DNS lookups bypass the Tor transport** (high) - Tor users resolving BIP353 payment addresses leak the queried DNS name over the local network instead of routing it through Tor, compromising the privacy/anonymity expected from `TorHttpClient`.
> > >   Unanchored locations included in summary:
> > >   
> > >   * crates/cdk/src/wallet/mint_connector/http_client.rs:283
> > 
> > 
> > This should be addressed in another PR
> 
> This introduces a regression in this PR then before this PR the bip353 delegated to the transport so it would use tor. So why allow the introduction of a regression to be fixed later?
> 
> There are three bot reviews that do not seem to be addressed? Do they require code changes or over eager review?
> 
> * [x]  [#1608 (comment)](https://github.com/cashubtc/cdk/pull/1608#discussion_r3439293423)
> * [x]  [#1608 (comment)](https://github.com/cashubtc/cdk/pull/1608#discussion_r3439741459)
> * [x]  [#1608 (comment)](https://github.com/cashubtc/cdk/pull/1608#discussion_r3435613396)

Yes, I think the DNS was moved towards the HTTP client and removed from the correct layer. I'll fix it now, following the code from main

thesimplekid:
I pushed a follow-up to refine the HTTP backend feature model from the earlier review discussion. Rather than removing reqwest, the implementation now allows bitreq and reqwest to coexist as additive features, with reqwest taking precedence when both are enabled because it supports the fuller set of behavior, including SOCKS proxies and invalid-certificate handling. bitreq remains the lightweight default backend for native builds, but its more limited proxy support is now validated explicitly so unsupported configurations fail clearly instead of behaving inconsistently. I also tightened the related backend behavior by preserving no-redirect handling, improving WASM request error handling, and simplifying the request builder API by moving the fluent methods onto inherent implementations instead of requiring an extension trait import.

I think we can clean up these features a bit better in a follow up PR by splitting the wallet and mint into separate crates. 

@crodas Can you review my changes.

crodas (dismissed):
LGTM

thesimplekid:
@codex review

thesimplekid (approved):

PR #1615, opened 2026-02-10
Add WASM-compatible WebSocket client abstraction

crodas:
### Description

Move WebSocket dependency (tokio-tungstenite) from cdk to cdk-http-client and introduce a platform-agnostic ws module with native (tokio-tungstenite) and WASM (ws_stream_wasm) backends. This enables WebSocket subscriptions on WASM targets, which were previously gated behind cfg(not(wasm32)).

Flatten wallet subscription module by merging ws.rs into subscription.rs now that the stream client is unconditionally compiled.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-http-client/src/ws/wasm.rs
+++ b/crates/cdk-http-client/src/ws/wasm.rs
@@ -1,13 +1,21 @@
-//! WASM WebSocket implementation using ws_stream_wasm
+//! WASM WebSocket implementation using web-sys + wasm-bindgen
 
-use futures::{SinkExt, StreamExt};
-use ws_stream_wasm::{WsMessage, WsMeta};
+use std::cell::RefCell;
+use std::rc::Rc;
+
+use futures::StreamExt;
+use futures_channel::{mpsc, oneshot};
+use wasm_bindgen::prelude::*;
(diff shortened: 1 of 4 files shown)


crates/cdk-http-client/src/ws/wasm.rs
```diff
@@ -0,0 +1,160 @@
...
+    url: &str,
+    headers: &[(&str, &str)],
+) -> Result<(WsSender, WsReceiver), WsError> {
+    if !headers.is_empty() {
+        tracing::warn!(
+            "WebSocket headers are not supported on WASM (browser limitation). \
+             {} header(s) will be ignored.",
+            headers.len()
+        );
+    }
```
thesimplekid, line 92: Won't just discarding the header lead to difficult to debug errors specifically around auth? I don't think logging and just dropping is a good solution.

crodas: @thesimplekid Can this be handled as part of https://github.com/cashubtc/cdk/issues/1682 and merge this as is?

thesimplekid: Sure, but need to make sure it is fixed before 0.16.

PR #1656, opened 2026-02-19
update lnd

lescuer97:
### Description

The send payment function being used has been depracated. I'm updating to use the recommended one.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-lnd/src/lib.rs
+++ b/crates/cdk-lnd/src/lib.rs
@@ -603,7 +603,8 @@ impl MintPayment for Lnd {
                                 payment_lookup_id: payment_identifier,
                                 payment_proof: payment_preimage,
                                 status: response_status,
-                                total_spent: Amount::new(total_msat as u64, CurrencyUnit::Msat).convert_to(unit)?,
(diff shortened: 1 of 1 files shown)


crates/cdk-lnd/src/lib.rs
```diff
@@ -545,53 +545,69 @@ impl MintPayment for Lnd {
...
+                            };
+
+                            let payment_identifier =
+                                PaymentIdentifier::PaymentHash(*bolt11.payment_hash().as_ref());
+
+                            return Ok(MakePaymentResponse {
+                                payment_lookup_id: payment_identifier,
+                                payment_proof: payment_preimage,
+                                status: response_status,
+                                total_spent: Amount::new(total_msat as u64, CurrencyUnit::Msat),
```
thesimplekid: We want to use convert_to here to convert back to the unit of the mint.  The unit is passed into the make make_payment fn.

lescuer97: done!

PR #1671, opened 2026-02-24
feat: add Blink Lightning backend (cdk-blink)

Forte11Cuba:
### Description

Add Blink as a Lightning backend for cdk-mintd, supporting both BTC (sats) and USD (cents) wallets via Blink's GraphQL API.

New crate: cdk-blink — implements the MintPayment trait with:
- Dual wallet support (BTC + USD) auto-discovered from account
- Invoice creation/payment using the correct wallet per unit
- Unit-aware fee probing (lnInvoiceFeeProbe for BTC, lnUsdInvoiceFeeProbe for USD)
- Currency conversion via currencyConversionEstimation for USD quotes
- Correct USD settlement handling (Blink's settlementAmount already includes fees for USD wallet)
- Lazy payment detection (no WebSocket stream, status checked on demand)

Modified: cdk-mintd — Blink configuration, env vars, and setup integration.

-----

### Notes to the reviewers
  - Blink does not support paid_invoices_stream, so wait_payment_event() returns a pending future (same pattern as the lazy detection path in CDK core via check_mint_quote_payments).
  - For USD wallet, settlementAmount already includes the fee (unlike BTC where settlementAmount + settlementFee are separate). This is handled in both  get_transaction_total_spent and parse_outgoing_transactions.
  - Fee reserve min_fee_reserve is enforced post-conversion in the target unit (cents for USD, sats for BTC) to avoid underestimation from msat-based  calculation.

-----

#### ADDED

 - New cdk-blink crate: Blink Lightning backend supporting BTC and USD wallets
  - Blink configuration support in cdk-mintd (config file + env vars)

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/Cargo.toml
+++ b/crates/cdk-integration-tests/Cargo.toml
@@ -30,7 +30,7 @@ cdk-redb = { workspace = true }
 cdk-fake-wallet = { workspace = true }
 cdk-common = { workspace = true, features = ["mint", "wallet", "http"] }
 cdk-http-client = { workspace = true }
-cdk-mintd = { workspace = true, features = ["cln", "lnd", "fakewallet", "grpc-processor", "lnbits", "management-rpc", "sqlite", "postgres", "ldk-node", "prometheus"] }
+cdk-mintd = { workspace = true, features = ["cln", "lnd", "fakewallet", "grpc-processor", "lnbits", "blink", "management-rpc", "sqlite", "postgres", "ldk-node", "prometheus"] }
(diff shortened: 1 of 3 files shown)


thesimplekid:
We don't plan to add more ln backends to the main repo right now. 

The way we should support backends like blink is through the grpc interface. The good news is most of your work here is not lost as the MintPayment implementation just needs to be moved into a grpc server. 

There is a template repo here https://github.com/thesimplekid/cdk-template-payment-processor. I am happy to work with you on this as having good support for blink with both usd and sats would be great.

PR #1681, opened 2026-02-26
feat: add inactive and expired keysets to fake wallet test mint ( #1596 )

### LINKED ISSUE #1596: Fake wallet: add inactive and expired keysets
Downstream testing would be easier if "fake wallet" supported more "real world" keyset scenarios, such as:

- inactive - Reason: Expired final_expiry
- inactive - Reason: mint rotation (no final_expiry)
- v1 and v2 keysets
- A non-sat keyset (eg: usd)

Forte11Cuba:
### Description

Add inactive and expired keysets to the fake wallet test mint so downstream wallet tests can exercise keyset filtering logic (closes #1596 ).          
                                                            
The fake mint now performs two keyset rotations at startup, producing:
- **Sat V1 (active)** — current working keyset
- **Sat V0 (inactive, `final_expiry` in the past)** — expired keyset scenario
- **Sat V1 (inactive, no `final_expiry`)** — rotation-deactivated scenario
- **USD V1 (active)** — non-sat keyset (unchanged)
- **Auth V1 (active)** — auth keyset (unchanged)

To enable this, `run_mintd_with_shutdown` was decomposed into `build_mint_from_settings` + `start_mint_services`, allowing callers to interact with
the mint between build and service start.

A bug where `final_expiry` was hardcoded to `None` in `DbSignatory::rotate_keyset` was also fixed as part of this work.

-----

### Notes to the reviewers

- The `Mint::rotate_keyset()` signature gained a `final_expiry: Option<u64>` parameter. All existing callers pass `None` so behavior is unchanged.
- `MintdInstance` uses a manual `Debug` impl because `DynMintAuthDatabase` doesn't implement `Debug`.
- The new `build_mint_from_settings` / `start_mint_services` public API is only used by `start_fake_mint.rs` for now, but could be useful for other test binaries or custom mint setups.

-----

#### CHANGED

- `Mint::rotate_keyset()` now takes an additional `final_expiry: Option<u64>` parameter

#### ADDED

- `build_mint_from_settings()` and `start_mint_services()` public API in `cdk-mintd` for composable mint setup
- `MintdInstance` struct exposing the built mint before services start
- Inactive and expired test keysets in the fake wallet mint binary
- `test_rotate_keyset_with_expiry` integration test

#### FIXED

- `final_expiry` passthrough in `DbSignatory` — was hardcoded to `None`, now correctly uses the value from `RotateKeyArguments`

----

### Checklist

### DIFF
--- a/crates/cdk-mintd/src/lib.rs
+++ b/crates/cdk-mintd/src/lib.rs
@@ -4,6 +4,7 @@
 // std
 use std::collections::HashMap;
 use std::env::{self};
+use std::fmt;
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -1351,12 +1352,15 @@ pub struct MintdInstance {
     auth_localstore: Option<cdk_common::database::DynMintAuthDatabase>,
 }
 
-impl std::fmt::Debug for MintdInstance {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+impl fmt::Debug for MintdInstance {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("MintdInstance")
             .field("mint", &"<Mint>")
             .field("config_mint_info", &self.config_mint_info)
-            .field("auth_localstore", &self.auth_localstore.as_ref().map(|_| "<auth_db>"))
+            .field(
+                "auth_localstore",
+                &self.auth_localstore.as_ref().map(|_| "<auth_db>"),
+            )
             .finish()
     }
 }
--- a/crates/cdk-integration-tests/src/bin/start_fake_mint.rs
+++ b/crates/cdk-integration-tests/src/bin/start_fake_mint.rs
@@ -135,20 +135,14 @@ async fn start_fake_mint(
     // Run the mint in a separate task
     let handle = tokio::spawn(async move {
         // Build the mint first
-        let instance = match cdk_mintd::build_mint_from_settings(
-            &temp_dir,
-            &settings,
-            None,
-            None,
-        )
-        .await
-        {
-            Ok(instance) => instance,
-            Err(e) => {
-                eprintln!("Failed to build mint: {e}");
-                return;
-            }
-        };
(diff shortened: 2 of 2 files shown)


crates/cdk-mintd/src/lib.rs
```diff
@@ -1339,15 +1340,39 @@ pub async fn run_mintd(
...
+/// [`Mint`] (e.g. rotate keysets) before starting the HTTP server via
+/// [`start_mint_services`].
+pub struct MintdInstance {
+    /// The fully-built mint.
+    pub mint: Arc<Mint>,
+    config_mint_info: cdk::nuts::MintInfo,
+    auth_localstore: Option<cdk_common::database::DynMintAuthDatabase>,
+}
+
+impl fmt::Debug for MintdInstance {
```
thesimplekid, line 1355: Why is it required to split this? Can't the rotate just be called on the `Mint`?

crates/cdk-mintd/src/lib.rs
```diff
@@ -1339,15 +1340,39 @@ pub async fn run_mintd(
...
+///
+/// Returned by [`build_mint_from_settings`] so that callers can interact with the
+/// [`Mint`] (e.g. rotate keysets) before starting the HTTP server via
+/// [`start_mint_services`].
+pub struct MintdInstance {
+    /// The fully-built mint.
+    pub mint: Arc<Mint>,
+    config_mint_info: cdk::nuts::MintInfo,
+    auth_localstore: Option<cdk_common::database::DynMintAuthDatabase>,
+}
```
thesimplekid, line 1353: I think it maybe better to modify the mint_builder with a fn that adds the keysets in the configuration we want rather then using the rotate_keyset.

Forte11Cuba:
Thanks for the review!     

I've separated the final_expiry fix into its own PR: #1686 

Regarding the MintBuilder approach that makes sense. I'll rework this PR to add keyset configuration directly in the MintBuilder instead of splitting run_mintd_with_shutdown. Will also look into exposing it via mintd settings/env for Docker.

Forte11Cuba:
Closing in favor of #1687, which implements this via `MintBuilder` as suggested in the review.

PR #1686, opened 2026-02-27
fix: pass final_expiry through in DbSignatory::rotate_keyset

Forte11Cuba:
### Description

The `final_expiry` field from `RotateKeyArguments` was hardcoded to `None` in  `DbSignatory::rotate_keyset`, causing any expiry set during keyset rotation to be silently discarded. Now correctly forwards `args.final_expiry` to `create_new_keyset`.

-----

### Notes to the reviewers

Single-line fix. The `create_new_keyset` function already accepts  `final_expiry: Option<u64>` and `RotateKeyArguments` already carries the field it just wasn't being passed through.

-----

[CHANGELOG](https://github.com/cashubtc/cdk/blob/main/CHANGELOG.md) Updates

#### FIXED

- `final_expiry` passthrough in `DbSignatory::rotate_keyset` was hardcoded to `None`, now correctly uses the value from `RotateKeyArguments`

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/mod.rs
+++ b/crates/cdk/src/mint/mod.rs
@@ -1214,7 +1214,8 @@ mod tests {
     #[tokio::test]
     async fn mint_mod_rotate_keyset_with_expiry() {
         let mut supported_units = HashMap::new();
-        supported_units.insert(CurrencyUnit::default(), (0, 32));
+        let amounts: Vec<u64> = (0..32).map(|i| 2u64.pow(i)).collect();
+        supported_units.insert(CurrencyUnit::default(), (0, amounts));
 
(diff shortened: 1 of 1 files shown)


thesimplekid:
Looks like we also need to set it here. Ideally we could add a test were we call rotate on the mint with an expiry and check it is set correctly. We do not need to use the grpc for this test, just rotate with an expiry directly on the mint. 

https://github.com/cashubtc/cdk/blob/e3f8dd0bab37127faf9426833b851ff43e7cdd98/crates/cdk/src/mint/keysets/mod.rs#L92

thesimplekid (approved):
Thanks for adding the tests. 

LGTM

Issue #1695, opened 2026-03-02
Regtest CLN nodes cannot create invoices


TheMhv:
When i try to create invoices from some regtest CLN node with command
```bash
just ln-cln1 invoice 1000 label "Test payment"
```

I got this error

```
{
   "code": -32602,
   "message": "expiry: should be an unsigned 64 bit integer: invalid token (see logs for details)"
}
error: Recipe `ln-cln1` failed with exit code 1
```

====

Obs: I was using nix regtest environment with `nix develop .#regtest`

TheMhv:
Apparently, this occurs because the quotes was not passed correctly from shell script.

PR #1696, opened 2026-03-02
fix(justfile): set positional arguments for bash script commands

TheMhv:
### Description

Fix for the issue: https://github.com/cashubtc/cdk/issues/1695

Set positional arguments for prevent quotes bugs when try to run bash commands with just

-----

### Notes to the reviewers

Just set positional arguments config and change args parse from bash calls inside justfile

-----

#### FIXED
- https://github.com/cashubtc/cdk/issues/1695

----

### Checklist

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,3 +1,5 @@
+set positional-arguments
+
 alias b := build
 alias c := check
 alias t := test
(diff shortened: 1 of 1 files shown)


SatsAndSports:
I don't know enough about `just` to review this, but I do like `bash` improvements including `"$@" to expand the positional parameters correctly

Thanks!

PR #1698, opened 2026-03-03
Batch minting

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut29.rs
+++ b/crates/cashu/src/nuts/nut29.rs
@@ -1,0 +1,290 @@
+//! NUT-29: Batch Mint Tokens
+//!
+//! <https://github.com/cashubtc/nuts/blob/main/29.md>
+
+use bitcoin::secp256k1::schnorr::Signature;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+
+use super::nut01::Error as Nut01Error;
+use super::{PublicKey, SecretKey};
+use crate::{Amount, BlindedMessage};
+
+/// Error types for batch operations
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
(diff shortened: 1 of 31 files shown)


crates/cashu/src/nuts/nut29.rs
```diff
@@ -0,0 +1,290 @@
...
+/// NUT-29 Settings for mint info
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Default)]
+#[cfg_attr(feature = "swagger", derive(utoipa::ToSchema))]
+pub struct Settings {
+    /// Maximum number of quotes allowed in a single batch request
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_batch_size: Option<u64>,
+    /// Supported payment methods for batch minting
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub methods: Option<Vec<String>>,
```
crodas, line 33: Can we implement a Default ourselves with some sane defaults? Because there are max inputs and outputs, what if people send a very large payload? Attempting to decode it would fail.

Or perhaps there is a general max payload size, so we don't even attempt to decode a large payload. I'm just thinking in terms of general defense against adversarial requests. If there is a maximum global, as suggested here, check that the max_batch_size does not go above those general settings.

What do you think @thesimplekid ? Other than this, I would approve this PR as it LGTM.

thesimplekid: We set the default in mintd. I don't think we need to set it in cashu.

https://github.com/thesimplekid/cdk/blob/cc80866cc83252e0d41992a8429a28749d8a313c/crates/cdk-mintd/src/lib.rs#L74

Issue #1670, opened 2026-02-23
Provide dart bindings with uniffi


thesimplekid:
Currently we have flutter bindings using flutter rust bridge (https://github.com/cashubtc/cdk-flutter). However, we would ideally drop support in this in favor of using uniffi like we do for our other binding. I believe the most up to date version is being worked on by the PDK team [here](https://github.com/chavic/uniffi-dart). It looks like the second team has had some success with it (?) [here](https://gitlab.com/ark-bitcoin/bark-ffi-bindings).

PR #1706, opened 2026-03-06
Swift and dart bindings tests in ci

crodas:
### Description

Prototype to have a monorepo for all the bindings

-----

Closes: https://github.com/cashubtc/cdk/pull/1539, https://github.com/cashubtc/cdk/issues/1670

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/bindings/swift/generate-bindings.sh
+++ b/bindings/swift/generate-bindings.sh
@@ -22,14 +22,6 @@ mkdir -p "$XCFRAMEWORK_DIR"
 
 cd "$RUST_DIR"
 
-# Install targets if needed
-echo "📦 Ensuring Rust targets are installed..."
-rustup target add aarch64-apple-ios
-rustup target add x86_64-apple-ios
-rustup target add aarch64-apple-ios-sim
-rustup target add aarch64-apple-darwin
-rustup target add x86_64-apple-darwin
-
 # Set cross-compilation toolchain for Apple targets when building on Linux.
 # cc-rs (used by build scripts like secp256k1-sys) picks CC_<target> to find the C compiler.
 # Without these it falls back to the host "cc" (Linux GCC) which doesn't understand


bindings/swift/generate-bindings.sh
```diff
@@ -0,0 +1,222 @@
...
+
+cd "$RUST_DIR"
+
+# Install targets if needed
+echo "📦 Ensuring Rust targets are installed..."
+rustup target add aarch64-apple-ios
+rustup target add x86_64-apple-ios
+rustup target add aarch64-apple-ios-sim
+rustup target add aarch64-apple-darwin
+rustup target add x86_64-apple-darwin
```
thesimplekid: We need to define this in a nix and not use rustup

bindings/dart/flutter_example/lib/main.dart
thesimplekid, line 1: I think this flutter example should live else where. We should not commit a full flutter example to the cdk repo. Instead we should have a more basic just dart example/test.

bindings/swift/Sources/Cdk/CashuDevKit.swift
thesimplekid, line 1: I don't think we should commit generated files and they should be added to .gitignore

crates/cdk-ffi/Cargo.toml
```diff
@@ -28,7 +28,7 @@ serde = { workspace = true, features = ["derive", "rc"] }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["sync", "rt", "rt-multi-thread"] }
-uniffi = { version = "0.29", features = ["cli", "tokio"] }
+uniffi = { version = "0.30", features = ["cli", "tokio"] }
```
thesimplekid: Can this be a workspace dep, we use it in multiple places?

bindings/dart/generate-bindings.sh
thesimplekid, line 1: For this to work in ci we'll need to add a new shell to the flake that has dart available. 

https://github.com/cashubtc/cdk-flutter/blob/master/flake.nix

bindings/swift/generate-bindings.sh
```diff
@@ -0,0 +1,222 @@
...
+    export AR_aarch64_apple_ios=$OSXCROSS_BIN/$DARWIN_TRIPLE-ar
+    export CFLAGS_aarch64_apple_ios="-isysroot $IOS_SDK -target arm64-apple-ios14.0"
+
+    export CC_aarch64_apple_ios_sim=/usr/bin/clang
+    export AR_aarch64_apple_ios_sim=$OSXCROSS_BIN/$DARWIN_TRIPLE-ar
+    export CFLAGS_aarch64_apple_ios_sim="-isysroot $IOS_SIM_SDK -target arm64-apple-ios14.0-simulator"
+
+    export CC_x86_64_apple_ios=/usr/bin/clang
+    export AR_x86_64_apple_ios=$OSXCROSS_BIN/$DARWIN_X86_TRIPLE-ar
+    export CFLAGS_x86_64_apple_ios="-isysroot $IOS_SIM_SDK -target x86_64-apple-ios14.0-simulator"
```
thesimplekid: Should define this in a flake and not use rustup

bindings/README.md
```diff
@@ -0,0 +1,16 @@
...
+The purpose of this project is to expose the CDK Wallet and its associated
+traits so they can be consumed in target languages as abstract classes or
+interfaces. This enables developers to extend or implement language-specific
+functionality while relying on the shared Rust core.
+
+A separate crate is provided for each target language. These crates serve as
+integration layers where Rust code can be adapted to the conventions and native
+capabilities of each platform or language runtime.
+
+[1]: https://gitlab.com/ark-bitcoin/bark-ffi-bindings
```
thesimplekid: Should make this more what it is and how to use it and when we need to modify it. We can credit bark at the end of the readme.

crates/cdk-ffi/src/database.rs
```diff
@@ -1908,6 +1908,32 @@ pub enum WalletDbBackend {
     },
 }
 
+/// Unified wallet storage: either a built-in Rust backend or a custom
+/// foreign-language implementation of the `WalletDatabase` callback interface.
+#[derive(uniffi::Enum)]
```
thesimplekid, line 2077: We use the enum because of a uniffi limitation of being able to use a trait directly, I'm guessing? Can we just add a comment explaining that? 

Should we add some constructors on the enum, new_sqlite(path), new_psgl(url) for example? Think that would be easier for downstream to use.

bindings/dart/rust/uniffi-bindgen.rs
```diff
@@ -0,0 +1,130 @@
...
+        true,              // library_mode = true (auto-locate UDL from dependencies)
+    )
+    .expect("Failed to generate dart bindings");
+
+    // Post-generation patching to fix uniffi-dart codegen issues
+    let generated_path = out_dir.join("cdk_ffi.dart");
+    patch_generated(&generated_path);
+}
+
+/// Patches the generated cdk_ffi.dart to fix uniffi-dart codegen issues.
```
thesimplekid: Can we just add a little more detail on why we need to patch and if it might change in the future as I understand this is required but it could be fragile, so to save us some time debugging in the future. Also if we could track what we need in order to avoid patching to get the upstream that would be great.

bindings/dart/rust/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[[bin]]
+name = "uniffi-bindgen"
+path = "uniffi-bindgen.rs"
+
+
+[dependencies]
+cdk-ffi.workspace = true
+uniffi = { version = "=0.30.0", features = ["cli"] }
+uniffi-dart = { git = "https://github.com/Uniffi-Dart/uniffi-dart" }
```
thesimplekid: We should pick a tag, looks like payjoin is using v0.1.0+v0.30.0, do we need main. If so whatever commit is working for us.

crates/cdk-ffi/src/database.rs
crodas: > We use the enum because of a uniffi limitation of being able to use a trait directly, I'm guessing? Can we just add a comment explaining that?


Yeah, I'll expand on that. But that is the reason

thesimplekid:
I notice the flake.lock is not changed after adding the dart overlay. This could be what is causing the rate limiting as this needs to be updated. `nix flake lock` should be run and the new lock file committed.

bindings/dart/lib/src/generated/cdk_ffi.dart
thesimplekid, line 1: I think generated files should not be commited.

Package.swift
thesimplekid, line 1: Does this need to be at the root level?

crodas: [I think so](https://gitlab.com/ark-bitcoin/bark-ffi-bindings/-/blob/master/Package.swift)

bindings/dart/lib/src/generated/cdk_ffi.dart
crodas: I think it should, at least people do that in [other repos.](https://gitlab.com/ark-bitcoin/bark-ffi-bindings/-/blob/master/dart/lib/src/generated/bark.dart?ref_type=heads)

bindings/swift/Tests/CdkTests.swift
```diff
@@ -0,0 +1,57 @@
...
+final class CdkTests: XCTestCase {
+    private var wallet: Wallet!
+    private var dbPath: String!
+
+    override func setUp() async throws {
+        try await super.setUp()
+        let tempDir = FileManager.default.temporaryDirectory
+        dbPath = tempDir.appendingPathComponent(UUID().uuidString + ".sqlite").path
+        wallet = try Wallet(
+            mintUrl: "https://testnut.cashu.space",
```
thesimplekid:
```suggestion
            mintUrl: "https://testnut.cashudevkit.org",
```

bindings/dart/lib/src/generated/cdk_ffi.dart
thesimplekid: We removed them

Package.swift
thesimplekid: Per discussion we're not going to keep this here and will publish from dedicated repo.

crodas: True, I forgot it. I'm on it now.

bindings/swift/generate-bindings.sh
crodas: [I took a different path](https://github.com/cashubtc/cdk/pull/1706/changes#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R159-R181) to be compatible with nix and non-nix systems

bindings/dart/generate-bindings.sh
crodas: [I took a different path](https://github.com/cashubtc/cdk/pull/1706/changes#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R159-R181) to be compatible with nix and non-nix systems

bindings/swift/generate-bindings.sh
thesimplekid: I don't think we should maintain two ways to do it and should be nix only.

bindings/dart/rust/rust-toolchain.toml
```diff
@@ -0,0 +1,29 @@
+[toolchain]
+channel = "1.90.0"
```
thesimplekid, line 2:
```suggestion
channel = "1.94.0"
```

bindings/dart/rust/rust-toolchain.toml
thesimplekid, line 1: Should these be here its a bit awkward to have two in the workspace?

bindings/swift/rust/rust-toolchain.toml
thesimplekid, line 1: Same comment as dart.

bindings/README.md
crodas: Is it better at a3cc95f24f8bdbcf722f7a8363ba3c731253444f?

bindings/swift/resources/Info-iOS.plist
thesimplekid, line 1: Im not sure we need the full XC framework and IOS stuff since we're not releasing from here and only want to run a smoke test of the building and running tests of swift bindings on macos runner.

bindings/swift/generate-bindings.sh
```diff
@@ -0,0 +1,275 @@
...
+            dependencies: ["CashuDevKitFFI"],
+            path: "bindings/swift/Sources/Cdk"
+        ),
+        .testTarget(
+            name: "CdkTests",
+            dependencies: ["Cdk"],
+            path: "bindings/swift/Tests"
+        ),
+    ]
+)
```
thesimplekid, line 256: I think if we don't build the XC framework we don't need this.

bindings/swift/generate-bindings.sh
```diff
@@ -0,0 +1,275 @@
...
+    export CFLAGS_aarch64_apple_ios="-isysroot $IOS_SDK -target arm64-apple-ios14.0"
+
+    export CC_aarch64_apple_ios_sim=/usr/bin/clang
+    export AR_aarch64_apple_ios_sim=$OSXCROSS_BIN/$DARWIN_TRIPLE-ar
+    export CFLAGS_aarch64_apple_ios_sim="-isysroot $IOS_SIM_SDK -target arm64-apple-ios14.0-simulator"
+
+    export CC_x86_64_apple_ios=/usr/bin/clang
+    export AR_x86_64_apple_ios=$OSXCROSS_BIN/$DARWIN_X86_TRIPLE-ar
+    export CFLAGS_x86_64_apple_ios="-isysroot $IOS_SIM_SDK -target x86_64-apple-ios14.0-simulator"
+fi
```
thesimplekid, line 55: Maybe should just remove this and only support the macos runner?

bindings/swift/Tests/CdkTests.swift
```diff
@@ -0,0 +1,59 @@
+import XCTest
+@testable import Cdk
+
+final class CdkTests: XCTestCase {
```
thesimplekid, line 4: XCTest seems to be the legacy testing method so should consider moving to swift testing.

.github/workflows/ci.yml
```diff
@@ -477,3 +477,50 @@ jobs:
...
+      - uses: cachix/cachix-action@v16
+        with:
+          name: cashudevkit
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+          useDaemon: false
+        continue-on-error: true
+      - name: Build native library and generate bindings
+        run: nix develop -L .#bindings --command just binding-dart
+      - name: Run Dart tests
+        run: nix develop -L .#bindings --command just test-dart
```
thesimplekid, line 498: These should be nix build defined in the flake to take advantage of caching.

.github/workflows/ci.yml
```diff
@@ -477,3 +477,50 @@ jobs:
...
+        uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: cachix/cachix-action@v16
+        with:
+          name: cashudevkit
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+          useDaemon: false
+        continue-on-error: true
+      - name: Build XCFramework and generate bindings
+        run: nix develop -L .#bindings --command just binding-swift
```
thesimplekid, line 516: These should be nix build defined in the flake to take advantage of caching.

PR #1707, opened 2026-03-06
cdk-cli: added non-interactive mode

### LINKED ISSUE #1701: Add non-interactive modes to cdk-cli
This would make the cli more bot friendly then it currently is.

0xEgao:
### Description

Fixes #1701

### Notes to the reviewers

-----

#### ADDED
Added Non-interactive mode for the cdk-cli

#### FIXED

----

### Checklist

* [] I ran `just final-check` before committing

### DIFF
--- a/crates/cdk-cli/README.md
+++ b/crates/cdk-cli/README.md
@@ -202,8 +202,14 @@ cdk-cli receive --nostr-key <nostr_key> --relay wss://relay.example.com
 ### Lightning Payments
 
 ```bash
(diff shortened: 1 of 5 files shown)


thesimplekid:
This is a good start but I think we need to check that each functions has the needed args that it does not require interactive prompts. For example I think on the mpp branch of melt we do not have the needed args to complete this non-interactively. 

We also need to work on how the output of the command is handle and how agent friendly that is or could be made, could also be a separate pr. 

We could even add some tests for this, but I won't require that as part of this pr.

PR #1715, opened 2026-03-07
refactor(cdk): unify melt and melt_async

### LINKED ISSUE #1692: Combine melt_async and melt
These two functions should really be the same we can follow the same model as the wallet where a future gets returned and in the non async case we await it and in the async case we do not. https://github.com/cashubtc/cdk/blob/6d3ea882afb9451c6f84f13e5f2a0cc05d6ad57c/crates/cdk/src/wallet/melt/mod.rs#L426-L498 

https://github.com/cashubtc/cdk/blob/6d3ea882afb9451c6f84f13e5f2a0cc05d6ad57c/crates/cdk/src/mint/melt/mod.rs#L545-L710

0xEgao:
### Description

Fixes #1692

I refactored `cdk/src/mint/melt/mod.rs` and now `melt` and `melt_async` share one returned future path, with `melt` awaiting it, while `melt_async` spawning it.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/fake_wallet.rs
+++ b/crates/cdk-integration-tests/tests/fake_wallet.rs
@@ -1138,14 +1138,15 @@ async fn test_fake_mint_multiple_unit_swap() {
         .await
         .unwrap();
 
-    let mut proof_streams =
-        wallet_usd.proof_stream(mint_quote.clone(), SplitTarget::default(), None);
-
-    let usd_proofs = proof_streams
-        .next()
+    let usd_proofs = wallet_usd
+        .wait_and_mint_quote(
+            mint_quote,
+            SplitTarget::default(),
+            None,
+            Duration::from_secs(30),
+        )
         .await
-        .expect("payment")
-        .expect("no error");
+        .unwrap();
 
     let active_keyset_id = wallet.fetch_active_keyset().await.unwrap().id;
     let fee_and_amounts = (0, ((0..32).map(|x| 2u64.pow(x)).collect::<Vec<_>>())).into();


crates/cdk/src/mint/melt/mod.rs
```diff
@@ -604,107 +709,18 @@ impl Mint {
...
-            request: Some(quote.request.to_string()),
-            unit: Some(quote.unit),
-        })
+    /// Process melt and return an outcome with an optional awaitable pending future.
+    #[instrument(skip_all)]
+    pub async fn melt_outcome(
+        &self,
+        melt_request: &MeltRequest<QuoteId>,
+    ) -> Result<MeltOutcome, Error> {
+        self.melt_with_outcome_impl(melt_request).await
```
thesimplekid: We can remove melt_async and melt_outcome and just have melt.

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -33,7 +35,150 @@ mod tests;
...
+    type Output = Result<MeltQuoteBolt11Response<QuoteId>, Error>;
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send>>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        Box::pin(self.wait())
+    }
+}
+
 impl Mint {
+    async fn melt_with_outcome_impl(
```
thesimplekid: Can this just be melt?

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -33,7 +35,150 @@ mod tests;
...
+                    tracing::error!(
+                        "Background melt completion failed for quote {}: {}",
+                        quote_id_for_log,
+                        e
+                    );
+                }
+            }
+
+            result
+        });
```
thesimplekid: Why do we have these 2 async blocks? I think we should just have one spawn?

crates/cdk/src/mint/melt/mod.rs
0xEgao: Yeah done, have renamed this as melt.

crates/cdk/src/mint/melt/mod.rs
0xEgao: Yeah that's what the pr is about i increased the noise earlier but now have removed this function, now only one `melt` fn exists.

crates/cdk/src/mint/melt/mod.rs
0xEgao: Yeah have moved back to 1 async block(just one spawn), 2 are overkill here.

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -33,6 +35,48 @@ mod tests;
...
 use melt_saga::MeltSaga;
 
+/// Outcome of mint melt processing.
+#[derive(Debug)]
+pub enum MeltOutcome {
+    /// Melt completed immediately.
+    Paid(MeltQuoteBolt11Response<QuoteId>),
+    /// Melt was accepted and is completing in background.
+    Pending(PendingMelt),
+}
```
thesimplekid: We can actually remove this enum, currently the Paid branch of this is unused since we also return the pending with the completion. In the wallet we need the enum since we don't know if we will get a paid or pending response but in the case of the mint we control this.

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -33,6 +35,48 @@ mod tests;
...
+    response: MeltQuoteBolt11Response<QuoteId>,
+    completion: tokio::task::JoinHandle<Result<MeltQuoteBolt11Response<QuoteId>, Error>>,
+}
+
+impl PendingMelt {
+    /// Return the immediate pending response (NUT-05 style).
+    pub fn into_pending_response(self) -> MeltQuoteBolt11Response<QuoteId> {
+        self.response
+    }
+
```
thesimplekid, line 55: we should add a pending_respons(&self) -> &MeltQuoteBolt11Response fn that does not consome the `PendingMelt` this way someone could inspect it but still await if they want.

thesimplekid (changes requested):

crates/cdk/src/mint/melt/mod.rs
0xEgao: Done, have removed this enum

crates/cdk/src/mint/melt/mod.rs
0xEgao: Done

crates/cdk-axum/src/custom_handlers.rs
```diff
@@ -426,16 +426,14 @@ pub async fn post_melt_custom(
...
+    let pending = state.mint.melt(&payload).await.map_err(into_response)?;
+
     let res = if respond_async {
         // Asynchronous processing - return immediately after setup
-        state
-            .mint
-            .melt_async(&payload)
-            .await
-            .map_err(into_response)?
+        pending.pending_response().clone()
```
thesimplekid: nit: we could avoid the clone using the `into_pending_response`

0xEgao: Done added that

PR #1721, opened 2026-03-09
Fix: Remove test-all command from Justfile

TheMhv:
### Description

The command `just test-all` doesn't seem to work. This PR try to fix this calling the other tests with right arguments.

edit: Remove this command because it's heavy to be useful.

-----

### Notes to the reviewers

- ~~Added `test-units` and `test-pure` to `test-all` command~~ 
- ~~Change the default database to **sqlite** (memory and postgresql does not work on my machine)~~
- Remove `test-all` command from justfile
- Lint the `justfile`

-----

#### CHANGED
- ~~`justfile` test-all command~~

#### REMOVED
- `justfile` **test-all** command

#### FIXED

----

### Checklist

### DIFF
--- a/justfile
+++ b/justfile
@@ -15,10 +15,10 @@ new-migration target name:
         echo "Error: target must be either 'mint' or 'wallet'"
         exit 1
     fi
-    
+
     timestamp=$(date +%Y%m%d%H%M%S)
     migration_path="./crates/cdk-sql-common/src/{{target}}/migrations/${timestamp}_{{name}}.sql"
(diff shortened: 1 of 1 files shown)


TheMhv:
> Honestly,  I think the just test all is probably to heavy to be useful and should maybe just be removed.

I agree. I was using before every PR but take a lot of time. Now i just use `final-check` and `test-pure` (that take a long time too).

I can change this PR propouse to remove then. What do you think?

PR #1750, opened 2026-03-20
feat: add p2pk_signing_keys to SendOptions

vnprc:
Summary 

Wallets that hold P2PK-locked proofs (received from another party, or minted with spending conditions) currently have no way to spend them through the standard prepare_send / confirm flow — they must manually call swap() and handle signing themselves. This PR closes that gap.

Changes:

- Add `p2pk_signing_keys: Vec<SecretKey>` to `SendOptions` (defaults to empty, fully backward-compatible)
- Extract P2PK/HTLC signing logic from `receive/saga/mod.rs` into a new `sign_proofs()` utility in `wallet/util.rs`, and call it from both the receive and send sagas
- Call `sign_proofs(&mut proofs_to_swap, &opts.p2pk_signing_keys)` in `PreparedSend::confirm()` before the swap, so the mint accepts the spend
- Mirror the field in cdk-ffi's `SendOptions` struct and both `From` conversions (matching the existing `ReceiveOptions` pattern)

Usage

```
// Wallet holds P2PK-locked proofs locked to `secret`
let prepared = wallet
    .prepare_send(
        Amount::from(10),
        SendOptions {
            p2pk_signing_keys: vec![secret],
            ..Default::default()
        },
    )
    .await?;

let token = prepared.confirm(None).await?;
```

Test plan

- 4 unit tests in `wallet/util.rs`: correct key signs, wrong key is a no-op, plain proof is a no-op, empty key vec is a no-op
- Integration test `test_p2pk_send_options_signing_keys`: funds a wallet, swaps clean proofs for P2PK-locked proofs via the raw mint API, then exercises the full prepare_send / confirm flow with signing keys — verifies the recipient receives the correct amount

### DIFF
--- a/crates/cdk-integration-tests/tests/integration_tests_pure.rs
+++ b/crates/cdk-integration-tests/tests/integration_tests_pure.rs
@@ -2228,3 +2228,124 @@ async fn test_p2pk_send_options_signing_keys() {
         "Bob should receive exactly the send amount"
     );
 }
+
+/// Regression test for the exact-denomination short-circuit bug in `p2pk_signing_keys`.
+///
+/// When a wallet holds P2PK-locked proofs whose total exactly equals the requested
+/// send amount, `prepare_send` takes a short-circuit path: all proofs go directly
+/// to `proofs_to_send` (no swap needed), so `proofs_to_swap` is empty.
+/// `confirm` only signs `proofs_to_swap`, so signing is skipped entirely and the
+/// P2PK-locked proofs flow out in the token unchanged.
+///
+/// Fix: when `p2pk_signing_keys` is non-empty, force any proofs in `proofs_to_send`
+/// into `proofs_to_swap` so they are always signed and unlocked via a real swap.
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_p2pk_signing_keys_exact_denomination_short_circuit() {
+    setup_tracing();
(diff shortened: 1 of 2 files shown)


crates/cdk-integration-tests/tests/integration_tests_pure.rs
```diff
@@ -2112,3 +2113,239 @@ async fn test_p2bk_multi_key_receive() {
...
+/// Regression test for the exact-denomination short-circuit bug in `p2pk_signing_keys`.
+///
+/// When a wallet holds P2PK-locked proofs whose total exactly equals the requested
+/// send amount, `prepare_send` takes a short-circuit path: all proofs go directly
+/// to `proofs_to_send` (no swap needed), so `proofs_to_swap` is empty.
+/// `confirm` only signs `proofs_to_swap`, so signing is skipped entirely and the
+/// P2PK-locked proofs flow out in the token unchanged.
+///
+/// Fix: when `p2pk_signing_keys` is non-empty, force any proofs in `proofs_to_send`
+/// into `proofs_to_swap` so they are always signed and unlocked via a real swap.
```
thesimplekid, line 2241: I think we should address this. Sending proofs with a pubkey lock as well as their signature can have a negative privacy implication. So we should not just fall back to including it. This should be an explicit option to include tokens that have locks.

PR #1753, opened 2026-03-20
fix(wallet): restore scans all keysets including inactive

### LINKED ISSUE #1752: cdk-restore inactive keysets issue
## Description

`fn restore()` in `wallet/mod.rs` only scans active keysets. If a mint has rotated keysets, proofs signed with the old (now inactive) keyset are never found during restore.

`restore()` should iterate all keysets for the wallet's unit, not just active ones. The `active` filter makes sense for normal operations (you want to mint with the current keyset), but during restore you need to scan everything.

Could be a separate method like `get_all_mint_keysets()` that skips the `keyset.active` check, or just an optional parameter on the existing one.

Forte11Cuba:
### Description

restore() in wallet/mod.rs only scanned active keysets via load_mint_keysets(). If a mint has rotated keysets, proofs signed with the old (now inactive) keyset were never found during restore.

Added get_all_mint_keysets() which returns all keysets for the wallet's unit regardless of active status, and updated restore() to use it.

Closes #1752     

-----

### Notes to the reviewers
            
The fix is minimal a new method get_all_mint_keysets() in keysets.rs that drops the keyset.active filter, and a one-line change in restore() to call it. The active filter remains correct for all other operations (mint, swap, melt) where you want the current keyset.  - cdk: restore() now scans inactive keysets so proofs from rotated keysets are recovered                                                              

-----

#### FIXED

- cdk: restore() now scans inactive keysets so proofs from rotated keysets are recovered

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/src/init_pure_tests.rs
+++ b/crates/cdk-integration-tests/src/init_pure_tests.rs
@@ -407,10 +407,7 @@ pub async fn create_test_wallet_for_mint(mint: Mint) -> Result<Wallet> {
 /// Create a test wallet connected directly to a mint with a specific seed
 ///
 /// Useful for restore tests where two wallets must share the same seed.
-pub async fn create_test_wallet_for_mint_with_seed(
-    mint: Mint,
-    seed: [u8; 64],
-) -> Result<Wallet> {
+pub async fn create_test_wallet_for_mint_with_seed(mint: Mint, seed: [u8; 64]) -> Result<Wallet> {
     let connector = DirectMintConnection::new(mint.clone());
 
     let mint_info = mint.mint_info().await?;


crates/cdk/src/wallet/keysets.rs
```diff
@@ -64,6 +64,38 @@ impl Wallet {
...
+            })
+            .collect::<Vec<_>>();
+
+        if !keysets.is_empty() {
+            Ok(keysets)
+        } else {
+            Err(Error::UnknownKeySet)
+        }
+    }
+
```
lescuer97: maybe it would be better to change the contract for `get_mint_keysets` to have `only_active` boolean param. this way we don't repeat that much code

Forte11Cuba: You're right. I'll fix it.

crates/cdk/src/wallet/keysets.rs
```diff
@@ -29,16 +29,18 @@ impl Wallet {
...
     ///
     /// Checks the metadata cache for keysets. If cache is not populated,
-    /// fetches from mint and updates cache. Returns error if no active keysets found.
+    /// fetches from mint and updates cache. When `only_active` is true,
+    /// returns only active keysets; when false, returns all keysets
+    /// (e.g. for restore, which needs to scan rotated keysets).
     #[instrument(skip(self))]
     #[inline(always)]
-    pub async fn get_mint_keysets(&self) -> Result<Vec<KeySetInfo>, Error> {
+    pub async fn get_mint_keysets(&self, only_active: bool) -> Result<Vec<KeySetInfo>, Error> {
```
lescuer97: wait for @thesimplekid  to say something but he might prefer a struct for the flag. It makes it clearer for when calling the function

crates/cdk/src/wallet/keysets.rs
```diff
@@ -49,7 +60,12 @@ impl Wallet {
             .keysets
             .values()
             .filter_map(|keyset| {
-                if keyset.unit == self.unit && keyset.active {
+                let dominated = keyset.unit == self.unit;
+                let dominated = match filter {
+                    KeysetFilter::Active => dominated && keyset.active,
+                    KeysetFilter::All => dominated,
+                };
+                if dominated {
```
thesimplekid: I think dominated is unclear and we can make this cleaner using early returns. 

```rust
if keyset.unit != self.unit {
    return None;
}

if matches!(filter, KeysetFlter::Active) && !keyset.active {
    return None;
}

Some((*keyset.clone()).clone())

````

PR #1758, opened 2026-03-22
feat(cln): add expose_private_channels config option

### LINKED ISSUE #1757: cdk-cln: Add `exopose_private_channels` configuration option for invoice creation
When running cdk-mintd with the CLN backend, mint quote invoices are created via CLN's `invoice` RPC without passing the `exposeprivatechannels` parameter.

The value is currently hardcoded in `crates/cdk-cln/src/lib.rs`:

```rust

  let invoice_response = cln_client
      .call_typed(&InvoiceRequest {
          amount_msat,
          description: description.unwrap_or_default(),
          label: label.clone(),
          expiry: unix_expiry.map(|t| t - time_now),
          fallbacks: None,
          preimage: None,
          cltv: None,
          deschashonly: None,
          exposeprivatechannels: None, // always None, not configurable
      })
```
 CLN's default behavior is to omit private channel route hints when there is sufficient inbound capacity on public
  channels. However, this default behavior does not always work correctly in practice. As a result, even when the
  mint node has significant inbound liquidity on private channels, invoices do not include route hints for those
  channels, preventing payers from routing payments through them.

### Proposed Solution

Add an expose_private_channels option to the mintd config that passes exposeprivatechannels: true to CLN's invoice
   RPC.
```
  [cln]
  rpc_path = "/path/to/lightning-rpc"
  expose_private_channels = true  # default: false
```
 ### implementation Notes

  The InvoiceRequest in cln-rpc (0.4.0 currently used, as well as latest 0.5.0 and unreleased 0.6.0 on master)
  defines exposeprivatechannels as Option<Vec<ShortChannelId>>, so there is no way to pass a boolean true through
  the typed API. This is a known limitation of CLN's code generator (msggen) which does not support oneOf JSON
  schemas. The same issue has been reported on the gRPC side (ElementsProject/lightning#7601).

  Therefore, when expose_private_channels = true, the implementation should use call_raw to construct JSON directly
  and pass exposeprivatechannels: true to CLN.

  ### Files to Change

  - cdk-mintd/src/config.rs — add field to Cln config struct
  - cdk-mintd/src/env_vars/cln.rs — CDK_MINTD_CLN_EXPOSE_PRIVATE_CHANNELS env var
  - cdk-mintd/src/setup.rs — pass config value to backend
  - cdk-cln/src/lib.rs — add field to Cln struct, call_raw branch for invoice creation
  - cdk-mintd/example.config.toml — document the option

4xvgal:
## Summary

- Add `expose_private_channels` config option to cdk-cln backend (default: `false`)
- When enabled, passes `exposeprivatechannels: true` to CLN's `invoice` RPC so private channels become route hint candidates in bolt11 invoices
- Uses `call_raw` to bypass cln-rpc's typed API limitation (`Option<Vec<ShortChannelId>>` cannot represent boolean `true`)
- Includes integration test that verifies private channels are included as route hint candidates

## Context

CLN's default behavior for route hint selection is non-deterministic — it depends on internal heuristics around inbound capacity, channel count, and other factors. In practice, this means mint nodes with private channel inbound liquidity may intermittently fail to receive payments, because CLN omits the private channel route hints that payers need for routing. This option lets mint operators explicitly include private channels as route hint candidates.

See also: [ElementsProject/lightning#7601](https://github.com/ElementsProject/lightning/issues/7601) for the upstream cln-rpc typing limitation.

Closes #1757

## CHANGED

- `Cln::new()` now accepts `expose_private_channels: bool` parameter

## ADDED

- `expose_private_channels` field in `[cln]` config section
- `CDK_MINTD_CLN_EXPOSE_PRIVATE_CHANNELS` environment variable
- `ClnClient::open_private_channel()` for integration tests
- `create_cln_backend_with_options()` helper for integration tests
- Integration test `test_expose_private_channels`

## Test plan

- [x] `cargo test -p cdk-cln` passes
- [x] `cargo clippy -p cdk-cln -- -D warnings` passes
- [ ] `cargo fmt` passes
- [x] Regtest integration test (`test_expose_private_channels`): opens private channel between CLN-1 and CLN-2, generates 10 invoices with `expose_private_channels=true`, verifies at least one invoice includes private channel route hint (6/10 in local test)
- [ ] CI integration tests pass (existing CLN tests should not break since default is `false`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/cashu/src/nuts/auth/nut22.rs
+++ b/crates/cashu/src/nuts/auth/nut22.rs
@@ -2,10 +2,9 @@
 
 use std::fmt;
 
-use bitcoin::base64::alphabet;
 use bitcoin::base64::engine::general_purpose::{self, GeneralPurposeConfig};
 use bitcoin::base64::engine::GeneralPurpose;
-use bitcoin::base64::Engine;
+use bitcoin::base64::{alphabet, Engine};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
--- a/crates/cdk-payment-processor/src/bin/payment_processor.rs
+++ b/crates/cdk-payment-processor/src/bin/payment_processor.rs
@@ -111,7 +111,10 @@ async fn main() -> anyhow::Result<()> {
                     };
 
                     let kv_store = Arc::new(MintSqliteDatabase::new(":memory:").await?);
-                    Arc::new(cdk_cln::Cln::new(cln_settings.rpc_path, fee_reserve, false, kv_store).await?)
+                    Arc::new(
+                        cdk_cln::Cln::new(cln_settings.rpc_path, fee_reserve, false, kv_store)
+                            .await?,
+                    )
                 }
                 #[cfg(feature = "fake")]
                 "FAKEWALLET" => {


crates/cdk-cln/src/lib.rs
```diff
@@ -582,20 +585,31 @@ impl MintPayment for Cln {
...
+                let invoice_response: InvoiceResponse = if self.expose_private_channels {
+                    let mut params = serde_json::to_value(&request).map_err(Error::from)?;
+                    params["exposeprivatechannels"] = serde_json::Value::Bool(true);
+                    cln_client
+                        .call_raw("invoice", &params)
+                        .await
+                        .map_err(Error::from)?
+                } else {
+                    cln_client.call_typed(&request).await.map_err(Error::from)?
+                };
```
thesimplekid, line 612: Would another option be on the start up of the mint list the cln channels and then pass in the channel ides on the exposeprivatechannels field of the request? I'm not necessarily requesting we do that instead this is probably simpler just trying to understand the api.

4xvgal: That would be cleaner from a cdk<-> cln API perspective. however, fetching short cheannel ids only at start up wouldn't detect channels opened after launch.

The root cause is that cln-rpc doesn't expose the needed fields in its tyepd API yet. Once the upstream issue is resolved, we won't need the startup-based approach or the raw JSON workaround in this PR.
  
For now, I think it makes sense to go with json-raw approach and switch over when the typed API catches up.

4xvgal:
> > Would it be better to loop until the first private channel hint is found, with an upper bound of 100 attempts?
> 
> Yeah I think this makes sense if you saw it fail 4/10 times locally would like to reduce flakiness where we can if we know there is a potential source of it, since we already have some issues with our ci just because its so heavy.

Updated. The test now iterates up to 100 times, decoding each invoice and checking for a private channel route hint. It breaks on the first match.

4xvgal:
Update: 

it's coming from CLN's `msggen' is not suppot oneOf type.
that bug fixing cause breaking at this moment. two maintainers claim this bug and i opened up PR which is stay in current approach and fix the bug for future fields.

I guess exposing private channel by jsonRAW will remain until refactor cln msggen
I'll keep update when CLN got updated

PR #1760, opened 2026-03-23
feat: add home page to mintd

thesimplekid:
### Description

-----

It is common for users to click on the mint url
and see nothing and think the mint is not working
not realizing they need to use that url only within a cashu wallet as it is api only. This adds a minimal home page showing basic mint info and a note to use a cashu wallet.

image

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-axum/src/router_handlers.rs
+++ b/crates/cdk-axum/src/router_handlers.rs
@@ -337,7 +337,7 @@ const CSS: &str = r#"
 body {
   background: var(--bg);
   color: var(--text-primary);
-  font-family: 'Outfit', -apple-system, sans-serif;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', sans-serif;
   min-height: 100vh;
   -webkit-font-smoothing: antialiased;
 }
@@ -474,7 +474,7 @@ body {
 
 .version-chip {
   font-size: 11px;
-  font-family: 'JetBrains Mono', monospace;
+  font-family: ui-monospace, 'SFMono-Regular', 'SF Mono', 'Cascadia Code', 'Segoe UI Mono', monospace;
   font-weight: 500;
   color: var(--text-muted);
   background: var(--surface);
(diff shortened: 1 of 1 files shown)


crates/cdk-axum/src/router_handlers.rs
```diff
@@ -309,6 +309,847 @@ pub(crate) async fn post_restore(
...
+    let markup = html! {
+        (maud::DOCTYPE)
+        html lang="en" {
+            head {
+                title { (name) }
+                meta charset="utf-8";
+                meta name="viewport" content="width=device-width, initial-scale=1";
+                link rel="preconnect" href="https://fonts.googleapis.com";
+                link href="https://fonts.googleapis.com/css2?family=Outfit:wght@400;500;600;700;800&family=JetBrains+Mono:wght@400;500&display=swap" rel="stylesheet";
+                style { (maud::PreEscaped(CSS)) }
```
thesimplekid, line 931: I think we would like to avoid the google fonts.

crates/cdk-mintd/Cargo.toml
```diff
@@ -11,7 +11,7 @@ rust-version.workspace = true
 readme = "README.md"
 
 [features]
-default = ["management-rpc", "cln", "lnd", "lnbits", "fakewallet", "grpc-processor", "sqlite"]
+default = ["management-rpc", "cln", "lnd", "lnbits", "fakewallet", "grpc-processor", "sqlite", "cdk-axum/info-page"]
```
thesimplekid: We should add a info-page feature to mintd

crates/cdk-axum/src/router_handlers.rs
Egge21M: Yep, third party fonts are a privacy issue. Either bundle the font or rely on the users system font

thesimplekid: I removed the google fonts in https://github.com/cashubtc/cdk/pull/1760/commits/4c5e4a3dbd80e3afecd2c1cfc20c389bc56d0e36 I don't notice any changes my llm tells me there is a css fallback already.

Issue #1310, opened 2025-11-19
Consider removeing clone from Mint and Wallet structs


thesimplekid:

TheMhv:
@thesimplekid We can estimate a date for review this thing?

PR #1762, opened 2026-03-23
chore: Removes `clone` from Wallet struct

TheMhv:
### Description

It's a part from https://github.com/cashubtc/cdk/issues/1310

Removes `clone` from `Wallet` struct

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-cli/src/sub_commands/update_mint_url.rs
+++ b/crates/cdk-cli/src/sub_commands/update_mint_url.rs
@@ -1,8 +1,11 @@
+use std::sync::Arc;
+
 use anyhow::Result;
 use cdk::mint_url::MintUrl;
 use cdk::nuts::CurrencyUnit;
 use cdk::wallet::WalletRepository;
 use clap::Args;
+use tokio::sync::Mutex;
 
 #[derive(Args)]
 pub struct UpdateMintUrlSubCommand {
@@ -22,12 +25,17 @@ pub async fn update_mint_url(
         new_mint_url,
     } = sub_command_args;
 
-    let mut wallet = wallet_repository
-        .get_wallet(&sub_command_args.old_mint_url, unit)
-        .await?
-        .clone();
+    let wallet = Arc::new(Mutex::new(
+        Arc::try_unwrap(
+            wallet_repository
+                .get_wallet(&sub_command_args.old_mint_url, unit)
+                .await?,
+        )
+        .expect("Unable to unwrap wallet"),
+    ));
 
-    wallet.update_mint_url(new_mint_url.clone()).await?;
+    let mut wallet_lock = wallet.lock().await;
+    wallet_lock.update_mint_url(new_mint_url.clone()).await?;
 
     println!("Mint Url changed from {old_mint_url} to {new_mint_url}");
 
--- a/crates/cdk-cli/src/sub_commands/update_mint_url.rs
+++ b/crates/cdk-cli/src/sub_commands/update_mint_url.rs
@@ -1,11 +1,8 @@
-use std::sync::Arc;
-
 use anyhow::Result;
 use cdk::mint_url::MintUrl;
 use cdk::nuts::CurrencyUnit;
 use cdk::wallet::WalletRepository;
 use clap::Args;
-use tokio::sync::Mutex;
 
 #[derive(Args)]
 pub struct UpdateMintUrlSubCommand {
@@ -25,17 +22,9 @@ pub async fn update_mint_url(
         new_mint_url,
     } = sub_command_args;
 
-    let wallet = Arc::new(Mutex::new(
-        Arc::try_unwrap(
-            wallet_repository
-                .get_wallet(&sub_command_args.old_mint_url, unit)
-                .await?,
-        )
-        .expect("Unable to unwrap wallet"),
-    ));
-
-    let mut wallet_lock = wallet.lock().await;
-    wallet_lock.update_mint_url(new_mint_url.clone()).await?;
+    wallet_repository
+        .update_mint_url(old_mint_url, new_mint_url, unit)
+        .await?;
 
     println!("Mint Url changed from {old_mint_url} to {new_mint_url}");
 
--- a/crates/cdk-integration-tests/src/init_pure_tests.rs
+++ b/crates/cdk-integration-tests/src/init_pure_tests.rs
@@ -573,7 +573,7 @@ pub async fn create_mint_with_limits(limits: Option<(usize, usize)>) -> Result<M
     Ok(mint)
 }
 
-pub async fn create_test_wallet_for_mint(mint: Mint) -> Result<Wallet> {
+pub async fn create_test_wallet_for_mint(mint: Mint) -> Result<Arc<Wallet>> {
     let seed = Mnemonic::generate(12)?.to_seed_normalized("");
     create_test_wallet_for_mint_with_seed(mint, seed).await
 }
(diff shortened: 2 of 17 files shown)


crates/cdk/src/wallet/mod.rs
```diff
@@ -243,7 +243,7 @@ impl Wallet {
         localstore: Arc<dyn WalletDatabase<database::Error> + Send + Sync>,
         seed: [u8; 64],
         target_proof_count: Option<usize>,
-    ) -> Result<Self, Error> {
+    ) -> Result<Arc<Self>, Error> {
```
thesimplekid: New should not create an Arc it should be Self

thesimplekid (changes requested):

crates/cdk-ffi/src/wallet.rs
```diff
@@ -49,20 +49,20 @@ impl Wallet {
...
+                })?)
+                .unit(unit.into())
+                .localstore(localstore)
+                .seed(seed)
+                .target_proof_count(config.target_proof_count.unwrap_or(3) as usize)
+                .build()
+                .map_err(FfiError::from)?,
+        );
+
+        Ok(Self { inner: wallet })
```
thesimplekid: nit: Doesn't really matter but think this change was unnecessary and the effect is the same.

crates/cdk-integration-tests/src/init_pure_tests.rs
```diff
@@ -399,7 +399,7 @@ pub async fn create_mint_with_limits(limits: Option<(usize, usize)>) -> Result<M
     Ok(mint)
 }
 
-pub async fn create_test_wallet_for_mint(mint: Mint) -> Result<Wallet> {
+pub async fn create_test_wallet_for_mint(mint: Mint) -> Result<Arc<Wallet>> {
```
thesimplekid, line 576: This shouldn't be changed to an Arc. If it needs to be shared it can be Arc'd later.

crates/cdk-integration-tests/src/init_pure_tests.rs
```diff
@@ -466,7 +468,7 @@ fn create_temp_dir(prefix: &str) -> Result<PathBuf> {
 }
 
 pub async fn fund_wallet(
-    wallet: Wallet,
+    wallet: Arc<Wallet>,
```
thesimplekid: I think this could be just &Wallet since we don't spawn the wallet or need ownership. But its test code and cloning the Arc is cheap so it doesn't matter.

crates/cdk/examples/npubcash.rs
```diff
@@ -97,7 +97,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Subscribe to quote updates and wait for the single payment
     let mut stream =
-        wallet.npubcash_proof_stream(SplitTarget::default(), None, Duration::from_secs(5));
+        wallet
+            .clone()
```
thesimplekid: We should prefer Arc::clone just to be explict about what is being cloned. There is a clippy rule for this I would like to turn on but have to fix it in a bunch of places so should try and not introduce more. 

```suggestion
       Arc::clone( wallet)
```

crates/cdk-cli/src/sub_commands/update_mint_url.rs
```diff
@@ -22,12 +25,17 @@ pub async fn update_mint_url(
...
+            wallet_repository
+                .get_wallet(&sub_command_args.old_mint_url, unit)
+                .await?,
+        )
+        .expect("Unable to unwrap wallet"),
+    ));
 
-    wallet.update_mint_url(new_mint_url.clone()).await?;
+    let mut wallet_lock = wallet.lock().await;
+    wallet_lock.update_mint_url(new_mint_url.clone()).await?;
```
thesimplekid: This would be okay in the current cli context, but I think its fragile. For example if we added a daemon mode to the cli then we would no longer guarantee there  is only one ref to the Wallet Arc and the try_unwrap would fail and with the expect we would panic. 

I think this exposes an issue with the WalletRepo api that we should address. We may need to add interior mutability to the wallet using something like an ArcSwap.  

@crodas What do you think?

thesimplekid (changes requested):

crates/cdk-integration-tests/tests/wallet_saga.rs
```diff
@@ -103,15 +105,15 @@ async fn test_reserved_proofs_excluded_from_selection() -> Result<()> {
...
     // Fund wallet
     let initial_amount = Amount::from(2000);
-    fund_wallet(wallet.clone(), initial_amount.into(), None).await?;
+    fund_wallet(&wallet, initial_amount.into(), None).await?;
 
     // Prepare two sends concurrently
-    let wallet1 = wallet.clone();
-    let wallet2 = wallet.clone();
+    let wallet1 = Arc::clone(&wallet);
+    let wallet2 = Arc::clone(&wallet);
```
thesimplekid, line 116: Thanks for going back and fixing this pattern of using the Arc::clone even in the places I didn't explicitly call out.

thesimplekid (commented):
This looks pretty good I think we just need to resolve https://github.com/cashubtc/cdk/pull/1762#discussion_r2978081076

crates/cdk-cli/src/sub_commands/update_mint_url.rs
crodas: I agree. Let's add a interior mutability.

crodas: In a following up PR

crates/cdk/src/wallet/streams/npubcash.rs
```diff
@@ -114,7 +115,7 @@ pub struct WalletNpubCashProofStream {
 impl WalletNpubCashProofStream {
     /// Create a new NpubCash proof stream for a single wallet
     pub fn new(
-        wallet: Wallet,
+        wallet: Arc<Wallet>,
```
crodas, line 118: This could be improved by using lifetime in other [streams](https://github.com/cashubtc/cdk/blob/main/crates/cdk/src/wallet/streams/payment.rs#L26). Not a blocker; this can be revised later. 

Other than this little issue, overall, this looks good to me.

TheMhv: We discussed about this on last [Weekly Meeting](https://github.com/cashubtc/cdk/pull/1806), we can do this improvement in another PR.

crodas: Yes

thesimplekid: I know we said that but why should we introduce a API change we don't want to only have to change it later. I think its better we just do it correctly.

crates/cdk-cli/src/sub_commands/update_mint_url.rs
```diff
@@ -22,12 +25,17 @@ pub async fn update_mint_url(
...
-        .get_wallet(&sub_command_args.old_mint_url, unit)
-        .await?
-        .clone();
+    let wallet = Arc::new(Mutex::new(
+        Arc::try_unwrap(
+            wallet_repository
+                .get_wallet(&sub_command_args.old_mint_url, unit)
+                .await?,
+        )
+        .expect("Unable to unwrap wallet"),
```
lescuer97: This message is going to get bubbled to the user of the CLI. Probably something more descriptive would help in case of debugging in the future

PR #1766, opened 2026-03-24
refactor: inject version header at client level instead of per request

### LINKED ISSUE #1765: cdk-mint-cli: Client should inject header instead of each request
We recently required a version header in the grpc of the mint management, I forgot to include this in most of the request. To avoid this we could add the injection of the header in the client instead of per request. Highlighted in this PR https://github.com/cashubtc/cdk/pull/1764. This should probably be the case for the signatory and MintPayment grpcs as well.

Forte11Cuba:
### Description

Replace per-request `with_version_header()` calls with a client-level `VersionInterceptor` that automatically injects the version header on every outgoing gRPC request. Applied to all three gRPC clients: mint-rpc CLI, payment-processor, and signatory.
                                                                                                                                                      
This adds a `VersionInterceptor` struct in `cdk-common` that implements `tonic::service::Interceptor` (same pattern as `MacaroonInterceptor` in `cdk-lnd`), and uses it via `Client::with_interceptor()` at client construction time.
                                                                                                                                                      
Closes #1765 

-----

### Notes to the reviewers

- The existing `create_version_inject_interceptor` closure in `cdk-common/src/grpc.rs` was replaced by the new `VersionInterceptor` struct because closures return opaque types that can't be named in struct fields or type aliases.
- The signatory client had a `Box::leak` on every request to create a `&'static str` for the version. This is now done once at client construction.   
- A new `InterceptedCdkMintClient` type alias was added to `cdk-mint-rpc/src/lib.rs` to keep the 12 subcommand signatures clean.

-----

#### CHANGED

- gRPC version header injection moved from per-request to client-level interceptor for mint-rpc CLI, payment-processor, and signatory clients         

#### ADDED

- `VersionInterceptor` struct in `cdk-common::grpc` implementing `tonic::service::Interceptor`                                                        
- `InterceptedCdkMintClient` type alias in `cdk-mint-rpc`

#### REMOVED

- `with_version_header()` helper functions from mint-rpc CLI, payment-processor, and signatory clients                                                
- `create_version_inject_interceptor()` closure (replaced by `VersionInterceptor` struct)

#### FIXED

- Per-request `Box::leak` memory leak in signatory gRPC client

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/grpc.rs
+++ b/crates/cdk-common/src/grpc.rs
@@ -29,10 +29,7 @@ impl VersionInterceptor {
     pub fn new(header: &'static str, version: impl AsRef<str>) -> Self {
         Self {
             header,
-            value: version
-                .as_ref()
-                .parse()
-                .expect("Invalid protocol version"),
+            value: version.as_ref().parse().expect("Invalid protocol version"),
         }
     }
 }


crates/cdk-common/src/grpc.rs
```diff
@@ -1,24 +1,44 @@
...
-    move |mut request: Request<()>| {
+    value: AsciiMetadataValue,
+}
+
+impl VersionInterceptor {
+    /// Create a new `VersionInterceptor`.
+    ///
+    /// # Panics
+    /// Panics if `version` is not a valid gRPC metadata ASCII value.
+    pub fn new(header: &'static str, version: &'static str) -> Self {
```
thesimplekid:
```suggestion
    pub fn new(header: &'static str, version: impl AsRef<str>) -> Self {
```

I think the version here does not need to be a static string. Changing this would allow us to remove the use of box leak.

crates/cdk-signatory/src/proto/client.rs
```diff
@@ -79,8 +71,12 @@ impl SignatoryRpcClient {
                 .await?
         };
 
+        let version_str = (proto::Constants::SchemaVersion as u8).to_string();
+        let version: &'static str = Box::leak(version_str.into_boxed_str());
```
thesimplekid: Changing the version of the interceptor to not be a static string we can remove the box leak here.

PR #1769, opened 2026-03-25
feat(wallet): default-on token-bucket rate limiting

### LINKED ISSUE #1642: Investigate adding rate limiting to the wallet
Currently the cdk wallet does not do any rate limiting. We know from our ts counter parts with heavier use mints can rate limit aggressively. Coco has some rate limiting integrated we should look at how they do it and see if it makes sense to integrate something similar.

https://github.com/cashubtc/coco/blob/master/packages/core/infra/RequestRateLimiter.ts

@crodas Can I assign this to you?

Micah-Shallom:
Closes #1642

This PR adds preemptive rate limiting to the wallet at the transport level, using a token bucket (same approach as Coco's `RequestRateLimiter`). Default-on with 20 capacity / 20 per minute; NUT-22 blind-auth shares the same bucket so mint-side rate limits count everything against one budget.

Usage through `WalletBuilder`:

```rust
WalletBuilder::new()
    .with_rate_limiting_config(config) // custom (defaults: 20 capacity, 20/min)
    .without_rate_limiting()           // opt out (default-on)
```

### Not covered here

- `WalletRepository` creates wallets through its own code paths that don't go through `WalletBuilder` yet, can be a follow-up
- No reactive 429 backoff, the limiter is proactive only

### DIFF
--- a/crates/cdk/src/wallet/builder.rs
+++ b/crates/cdk/src/wallet/builder.rs
@@ -11,6 +11,11 @@ use crate::error::Error;
 use crate::mint_url::MintUrl;
 use crate::nuts::CurrencyUnit;
 use crate::wallet::auth::{AuthMintConnector, AuthWallet};
+use crate::wallet::mint_connector::transport;
+use crate::wallet::mint_connector::transport::rate_limit::TokenBucket;
+use crate::wallet::mint_connector::{
+    AuthHttpClient, RateLimitConfig, RateLimitedAuthHttpClient, RateLimitedHttpClient,
+};
 use crate::wallet::mint_metadata_cache::MintMetadataCache;
 use crate::wallet::{HttpClient, MintConnector, SubscriptionManager, Wallet};
 
@@ -25,6 +30,8 @@ pub struct WalletBuilder {
     seed: Option<[u8; 64]>,
     use_http_subscription: bool,
     client: Option<Arc<dyn MintConnector + Send + Sync>>,
+    rate_limiter: Option<TokenBucket>,
+    auth_cat: Option<String>,
     metadata_cache_ttl: Option<Duration>,
     metadata_cache: Option<Arc<MintMetadataCache>>,
     metadata_caches: HashMap<MintUrl, Arc<MintMetadataCache>>,
@@ -51,6 +58,8 @@ impl Default for WalletBuilder {
             auth_connector: None,
             seed: None,
             client: None,
+            rate_limiter: Some(TokenBucket::new(RateLimitConfig::default())),
+            auth_cat: None,
             metadata_cache_ttl: Some(Duration::from_secs(3600)),
             use_http_subscription: false,
             metadata_cache: None,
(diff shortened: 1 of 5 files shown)


crodas:
On the same page, if throttled, I think there should be a config to retry X times.

Micah-Shallom:
thanks for your feedback @crodas ....
@thesimplekid do you think it will be nice to add a second line of defence sort off in a situation where the mint throws a 429 which we expect not to happen....so if mint  returns HTTP 429, the wallet can wait and retry instead of failing

Micah-Shallom:
hii @thesimplekid  pushed an updated version of this......ready for your review whenever you've got a window.
cc @crodas

crates/cdk/src/wallet/auth/auth_wallet.rs
```diff
@@ -73,6 +73,26 @@ impl AuthWallet {
...
+    ) -> Self {
+        Self {
+            mint_url,
+            localstore,
+            metadata_cache,
+            protected_endpoints: Arc::new(RwLock::new(protected_endpoints)),
+            refresh_token: Arc::new(RwLock::new(None)),
+            auth_client,
+            oidc_client: Arc::new(RwLock::new(oidc_client)),
+        }
```
TheMhv, line 93: I think that you can do that on the `new` function, adding a new `Option<Arc<dyn AuthMingConnector....>>` parameter that's `None` by default, with that, you can use a pattern matching to do the same thing here.

crates/cdk/src/wallet/mint_connector/http_client.rs
```diff
@@ -646,6 +646,17 @@ where
...
+    pub fn with_transport(mint_url: MintUrl, transport: T, cat: Option<AuthToken>) -> Self {
+        Self {
+            transport: transport.into(),
+            mint_url,
+            cat: Arc::new(RwLock::new(
+                cat.unwrap_or(AuthToken::ClearAuth("".to_string())),
+            )),
+        }
+    }
 }
```
TheMhv: You can do the same thing, that i suggested before, here

crates/cdk/src/wallet/builder.rs
```diff
@@ -162,39 +183,21 @@ impl WalletBuilder {
...
-        self.auth_wallet = Some(AuthWallet::new(
-            mint_url,
-            Some(AuthToken::ClearAuth(cat)),
-            localstore,
-            metadata_cache,
-            HashMap::new(),
-            None,
-        ));
+        self.auth_cat = Some(cat);
+        self.auth_wallet = None;
```
TheMhv, line 217: Why this became as None?

TheMhv (changes requested):
You need to fix some lint errors that you can find running `just final-check` or `cargo fmt --all -- --check`.

ye0man:
@Micah-Shallom this has become higher priority now to address rate limiting issues with our native wallet, now that it's in beta. I've added @crodas as an assignee to pick this up as it's a bit stale

crodas:
Closed in favor of #2251

PR #1797, opened 2026-03-30
Add Kotlin JVM bindings

crodas:
### Description

Add Kotlin/JVM bindings for CDK using UniFFI, following the same pattern as the existing Dart and Swift bindings. Includes a Gradle project with JNA-based native library loading, a generate-bindings.sh build script, and JUnit 5 integration tests (initial balance, mint flow) mirroring the Dart test suite.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


bindings/kotlin/.gitignore
thesimplekid, line 1: Should we just keep one .gitignore?

bindings/kotlin/generate-bindings.sh
thesimplekid, line 1: Is this used? I don't see where it is in ci but might just be missing it. I do see it in the justfile. Do we need both?

.github/workflows/ci.yml
```diff
@@ -493,10 +493,40 @@ jobs:
...
+      - name: Build native library and generate bindings
+        run: |
+          nix build -L .#kotlin-bindings
+          mkdir -p bindings/kotlin/cdk-jvm/src/main/kotlin
+          cp -r result/kotlin/org bindings/kotlin/cdk-jvm/src/main/kotlin/
+          mkdir -p bindings/kotlin/cdk-jvm/src/main/resources
+          cp result/resources/libcdk_ffi.* bindings/kotlin/cdk-jvm/src/main/resources/
+          chmod -R u+w bindings/kotlin/cdk-jvm/src/main/kotlin bindings/kotlin/cdk-jvm/src/main/resources
+      - name: Run Kotlin tests
+        run: nix develop -L .#bindings --command just test-kotlin
```
thesimplekid, line 517: Think this could all be defined in the flake and then just one nix build command here. But not something I'm gonna block over this is okay too.

bindings/dart/rust/rust-toolchain.toml
thesimplekid, line 1: Can we just remove these? Doesn't the flake set this?

bindings/kotlin/generate-bindings.sh
crodas: That was left over of the pre nix era. I'll remove them now

justfile
```diff
@@ -840,20 +840,41 @@ ffi-release-kotlin VERSION:
...
+  chmod -R u+w bindings/dart/lib/src/generated/
 
 # Run Dart binding tests
 test-dart:
   #!/usr/bin/env bash
   set -euo pipefail
-  cd "{{justfile_directory()}}/bindings/dart"
+  cd "{{justfile_directory()}}"
+  cp rust-toolchain.toml bindings/dart/rust/rust-toolchain.toml
+  trap 'rm -f bindings/dart/rust/rust-toolchain.toml' EXIT
```
thesimplekid, line 858: Why do we need this. The toolchains should be defined by nix so there isn't a need for the toolchain.toml

.github/workflows/ci.yml
```diff
@@ -497,6 +497,25 @@ jobs:
...
+      - uses: cachix/cachix-action@v16
+        with:
+          name: cashudevkit
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+          useDaemon: false
+        continue-on-error: true
+      - name: Build native library and generate bindings
+        run: nix develop -L .#bindings --command just binding-kotlin
+      - name: Run Kotlin tests
+        run: nix develop -L .#bindings --command just test-kotlin
```
thesimplekid, line 517: These should be nix build not develop. Weren't they already in a previous commit?

bindings/kotlin/cdk-jvm/build.gradle.kts
thesimplekid, line 1: Why is this different then the one we're already using for our bindings.

bindings/kotlin/gradle/wrapper/gradle-wrapper.jar
thesimplekid, line 1: Should we gitignore this?

thesimplekid: Looks like its also committed in cdk-kotlin so maybe we need it.

justfile
crodas: The test wants the rust-toolchain.toml at the same level as the crate that it is compiling

.github/workflows/ci.yml
crodas: binding-kotlin is calling `nix build` internally. I want one command to build. Is it wrong?

justfile
thesimplekid: Oh is it that dart package native_toolchain_rust we're using that expects it. Okay I see can't really get around that then.

.github/workflows/ci.yml
thesimplekid: Actually I guess the test itself needs a network connection so that can't be in a nix build so we have to use develop. Its okay we can keep it like this.

PR #1799, opened 2026-03-30
chore: Removes `clone` from Mint struct

TheMhv:
### Description

It's a part from https://github.com/cashubtc/cdk/issues/1310

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/src/init_pure_tests.rs
+++ b/crates/cdk-integration-tests/src/init_pure_tests.rs
@@ -30,12 +30,12 @@ use tracing_subscriber::EnvFilter;
 use uuid::Uuid;
 
 pub struct DirectMintConnection {
-    pub mint: Mint,
+    pub mint: Arc<Mint>,
     auth_wallet: Arc<RwLock<Option<AuthWallet>>>,
 }
 
 impl DirectMintConnection {
-    pub fn new(mint: Mint) -> Self {
+    pub fn new(mint: Arc<Mint>) -> Self {
         Self {
             mint,
             auth_wallet: Arc::new(RwLock::new(None)),
@@ -398,11 +398,11 @@ pub fn setup_tracing() {
         .try_init();
 }
 
-pub async fn create_and_start_test_mint() -> Result<Mint> {
+pub async fn create_and_start_test_mint() -> Result<Arc<Mint>> {
     create_mint_with_limits(None).await
 }
 
-pub async fn create_mint_with_fee(fee_ppk: u64) -> Result<Mint> {
+pub async fn create_mint_with_fee(fee_ppk: u64) -> Result<Arc<Mint>> {
     // Read environment variable to determine database type
     let db_type = env::var("CDK_TEST_DB_TYPE").expect("Database type set");
 
@@ -475,9 +475,11 @@ pub async fn create_mint_with_fee(fee_ppk: u64) -> Result<Mint> {
 
     let quote_ttl = QuoteTTL::new(10000, 10000);
 
-    let mint = mint_builder
-        .build_with_seed(localstore.clone(), &mnemonic.to_seed_normalized(""))
-        .await?;
+    let mint = Arc::new(
+        mint_builder
+            .build_with_seed(localstore.clone(), &mnemonic.to_seed_normalized(""))
+            .await?,
+    );
 
     mint.set_quote_ttl(quote_ttl).await?;
 
@@ -486,7 +488,7 @@ pub async fn create_mint_with_fee(fee_ppk: u64) -> Result<Mint> {
     Ok(mint)
 }
 
-pub async fn create_mint_with_limits(limits: Option<(usize, usize)>) -> Result<Mint> {
+pub async fn create_mint_with_limits(limits: Option<(usize, usize)>) -> Result<Arc<Mint>> {
     // Read environment variable to determine database type
     let db_type = env::var("CDK_TEST_DB_TYPE").expect("Database type set");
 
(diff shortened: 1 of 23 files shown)


crates/cdk/src/test_helpers/mint.rs
```diff
@@ -73,7 +73,7 @@ pub(crate) fn should_fail_for(operation: &str) -> bool {
 ///     // Use the mint for testing
 /// }
 /// ```
-pub async fn create_test_mint() -> Result<Mint, Error> {
+pub async fn create_test_mint() -> Result<Arc<Mint>, Error> {
```
thesimplekid, line 76: the create should return just `Mint` and the caller can `Arc` it if it needs to.

crates/cdk/src/test_helpers/nut10.rs
```diff
@@ -15,7 +17,7 @@ use crate::Error;
 
 /// Test mint wrapper with convenient access to common keyset info
 pub struct TestMintHelper {
-    pub mint: Mint,
```
thesimplekid, line 18: Does this need to be an `Arc<Mint>`? We don' have `Clone` on `TestMintHelper` so I think maybe not.

thesimplekid (changes requested):
Since we don't need the Mint to be cheap to clone anymore we should remove some of these inner Arc's if we can 

https://github.com/cashubtc/cdk/blob/a46f99802f5bb4b85ff6a535d4987e1b9514b2a2/crates/cdk/src/mint/mod.rs#L57-L78

I'm not sure all of the arc's added in the tests are needed I think some of them can still take a mint maybe needing a lifetime. But theyre also tests so if Arc's are best thats fine.

crates/cdk/src/test_helpers/nut10.rs
TheMhv: Yes, because we need `Arc::clone` on

```rust
    pub fn mint(&self) -> Arc<Mint> {
        Arc::clone(&self.mint)
    }
```

That's is used on most of tests

crates/cdk-integration-tests/src/init_pure_tests.rs
```diff
@@ -396,19 +401,22 @@ pub async fn create_mint_with_limits(limits: Option<(usize, usize)>) -> Result<M
...
 /// Create a test wallet connected directly to a mint with a specific seed
 ///
 /// Useful for restore tests where two wallets must share the same seed.
-pub async fn create_test_wallet_for_mint_with_seed(mint: Mint, seed: [u8; 64]) -> Result<Wallet> {
-    let connector = DirectMintConnection::new(mint.clone());
+pub async fn create_test_wallet_for_mint_with_seed(
+    mint: Arc<Mint>,
+    seed: [u8; 64],
+) -> Result<Wallet> {
+    let connector = DirectMintConnection::new(Arc::clone(&mint));
```
crodas: You can just keep it as `mint.clone()`

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -588,7 +588,10 @@ impl Mint {
     ///
     /// Uses MeltSaga typestate pattern for atomic transaction handling with automatic rollback on failure.
     #[instrument(skip_all)]
-    pub async fn melt(&self, melt_request: &MeltRequest<QuoteId>) -> Result<PendingMelt, Error> {
+    pub async fn melt(
+        self: std::sync::Arc<Self>,
```
crodas: You can just keep ast &Arc<Self> to avoid cloning to use it one time

crates/cdk/src/mint/melt/tests/p2pk_sigall_spending_conditions_tests.rs
```diff
@@ -118,7 +120,7 @@ async fn test_p2pk_sig_all_requires_transaction_signature() {
     println!("✓ Melting WITHOUT signature failed verification as expected");
 
     // Also verify the actual melt fails
-    let melt_result = mint.melt(&melt_request_no_sig).await;
+    let melt_result = Arc::clone(&mint).melt(&melt_request_no_sig).await;
```
crodas: When https://github.com/cashubtc/cdk/pull/1799/changes#r3373558924 happens you should just use it without cloning it. Although arc clones are cheap, they are not free.

crates/cdk/src/mint/melt/tests/p2pk_sigall_spending_conditions_tests.rs
```diff
@@ -161,7 +163,12 @@ async fn test_p2pk_sig_all_requires_transaction_signature() {
     println!("✓ P2PK SIG_ALL spending conditions verified successfully");
 
     // Perform the actual melt - this also verifies spending conditions internally
-    let melt_response = mint.melt(&melt_request).await.unwrap().await.unwrap();
+    let melt_response = Arc::clone(&mint)
+        .melt(&melt_request)
```
crodas: Same as https://github.com/cashubtc/cdk/pull/1799/changes#r3373562180

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -616,7 +619,7 @@ impl Mint {
             .ok_or(Error::UnknownQuote)?;
 
         let init_saga = MeltSaga::new(
-            std::sync::Arc::new(self.clone()),
+            std::sync::Arc::clone(&self),
```
crodas: You can just `self.clone()`

crodas (changes requested):
A few suggestions:

* `&Arc<Self>` instead of `Arc<Self>`. It rarely makes sense to take ownership. Even when it does, we can always clone internally. &Arc<Self> would make it easy to use without cloning outside
* `Arc::clone(&mint) vs `mint.clone()`, I prefer the later. Can we decide on one as a code style @thesimplekid ?

thesimplekid:
Would still be good to remove clone but these are well stale.

PR #1805, opened 2026-04-01
Move fee calculation to cdk-common and add fee methods to Wallet trait

crodas:
### Description

Move `calculate_fee` and `ProofsFeeBreakdown` from `cdk/src/fees.rs` to `cdk-common/src/wallet/fees.rs` so downstream consumers of cdk-common can compute fees without depending on the full cdk crate.

Add `get_proofs_fee`, `get_proofs_fee_by_count`, and `get_proofs_with` to the Wallet trait, and make `get_proofs_by_states` a default method that delegates to `get_proofs_with`. Expose corresponding FFI bindings.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/wallet/mod.rs
+++ b/crates/cdk-common/src/wallet/mod.rs
@@ -8,6 +8,7 @@ use async_trait::async_trait;
 use bitcoin::bip32::DerivationPath;
 use bitcoin::hashes::{sha256, Hash, HashEngine};
 use cashu::amount::{FeeAndAmounts, KeysetFeeAndAmounts, SplitTarget};
+use cashu::nut00::ProofsMethods;
 use cashu::nuts::nut07::ProofState;
 use cashu::nuts::nut18::PaymentRequest;
 use cashu::nuts::{AuthProof, Keys};
@@ -22,6 +23,7 @@ use crate::nuts::{
 };
 use crate::{Amount, Error};
 
+pub use crate::fees;
 pub mod saga;
 
 pub use saga::{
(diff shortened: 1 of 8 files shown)


crates/cdk-common/src/wallet/mod.rs
```diff
@@ -798,6 +800,28 @@ pub trait Wallet: Send + Sync {
         keyset_id: Id,
     ) -> Result<Self::Amount, Self::Error>;
 
+    /// Fee required to redeem proof set by count per keyset
+    async fn get_proofs_fee_by_count(
+        &self,
+        proofs_per_keyset: HashMap<Id, u64>,
+    ) -> Result<fees::ProofsFeeBreakdown, Self::Error>;
```
thesimplekid, line 807: Do we think this fn should be in the public api, Is it even used privately? Imagine most will use the `get_proofs_fee`

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -992,7 +1016,9 @@ pub trait Wallet: Send + Sync {
     /// Returns all proofs whose state matches any of the given states.
     /// The `Spent` state is typically excluded since spent proofs are removed
     /// from the database.
-    async fn get_proofs_by_states(&self, states: Vec<State>) -> Result<Proofs, Self::Error>;
+    async fn get_proofs_by_states(&self, states: Vec<State>) -> Result<Proofs, Self::Error> {
+        self.get_proofs_with(Some(states), None).await
```
thesimplekid, line 1020: We're inconsistent about if the traits have an implementation here or just a definition. In cdk-coommon I think they should only have a definition.

thesimplekid:
@crodas Think this maybe this just fell of your radar? Do we still need it think so?

crodas:
We certainly needed but we agreed last week to work on the wallet documentation and planning first. I'm working on understanding what's really needed and what is an internal function used only for Rust.

Once that document is done and approved, I'll adjust this PR accordingly

PR #1808, opened 2026-04-02
fix: replace blocking Condvar pool with async Semaphore

thesimplekid:
The SQLite connection pool used std::sync::Condvar to wait for free connections, which blocks the OS thread. Under the default pool size of 1, concurrent requests can block all Tokio worker threads, deadlocking the runtime the task holding the connection can never be scheduled to release it. Replace with tokio::sync::Semaphore so waiters yield to the async runtime instead of blocking.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ ldk-node = "0.7.0"
 serde = { version = "1", features = ["derive", "rc"] }
 serde_json = "1"
 thiserror = { version = "2" }
-tokio = { version = "1", default-features = false, features = ["rt", "macros", "test-util", "sync"] }
+tokio = { version = "1", default-features = false, features = ["rt", "macros", "test-util", "sync", "time"] }
 tokio-util = { version = "0.7.11", default-features = false }
 tower = "0.5.2"
 tower-http = { version = "0.6.1", features = ["compression-full", "decompression-full", "cors", "trace", "fs"] }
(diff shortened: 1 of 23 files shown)


crodas (dismissed):

lescuer97 (approved):
I would be lying if I said that I completely understand what is going on. 

but it does more or less makes sense

crodas:
I'm having issues with the connection pool. What's left for this to be ready @thesimplekid ? We can revisit this fix. Mine is unrelated, but I'm suspecting the pool is sharing the same resource twice.

thesimplekid:
I know my ai flagged: 

## Summary
I reviewed PR #1808 in `cashubtc/cdk`. The PR refactors database connection pooling logic inside `cdk-sql-common/src/pool.rs` by switching from threading primitives (`Condvar`) to asynchronous Rust semantics using `tokio::sync::Semaphore`. In doing so, it introduces a vulnerability: the `get_timeout` function, which is supposed to wait asynchronously up to a configured `timeout` duration for an available connection, completely ignores the `timeout` parameter during semaphore acquisition (`self.semaphore.clone().acquire_owned().await`). Under load or exhaustion, this causes the task to block indefinitely rather than returning the expected `Error::Timeout`. This breaks the system's ability to gracefully degrade or fail fast when limits are hit, leading to resource exhaustion, starvation of other dependent tasks, and potential deadlocks in production environments, making this a high-severity DoS vulnerability introduced by this PR.

I think there were some other things last time I asked my agent to review it as well.

crodas (dismissed):

crodas:
Can you please also check my latest changes @thesimplekid ?

crodas (approved):

PR #1834, opened 2026-04-03
NUT-XX: Mint Quote Lookup by Public Key

### LINKED ISSUE #1746: New Nut: Implement get quotes by pubkeys
https://github.com/cashubtc/nuts/pull/341#pullrequestreview-3977365369

TheMhv:
### Description

The CDK implementation of [NUT-XX: Mint Quote Lookup by Public Key](https://github.com/cashubtc/nuts/pull/341)

Closes: https://github.com/cashubtc/cdk/issues/1746

-----

### Notes to the reviewers

- Create a new route for mint quote `/mint/quote/{method}/pubkey`
- Create a new `get_mint_quotes_by_pubkey` database function and add your test
- Create a migration to add index into pubkey from mint_quote table in Database
- Create the struct for Request defined in [NUT](https://github.com/cashubtc/nuts/pull/341)

-----

#### ADDED
- A new NUT file `nuts/nutxx.rs`
- A new route `v1/mint/quote/{method}/pubkey`
- Migrations for add index to pubkey field in mint_quote table

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nutxx.rs
+++ b/crates/cashu/src/nuts/nutxx.rs
@@ -1,0 +1,14 @@
+//! NUT-XX: Mint Quote Lookup by Public Key
+//!
+//! <https://github.com/cashubtc/nuts/blob/get-quotes-by-pubkeys/xx.md>
+
+use serde::{Deserialize, Serialize};
+
+/// Mint quote by pubkey request [NUT-XX]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MintQuoteByPubkeyRequest {
+    /// Pubkeys
+    pub pubkeys: Vec<String>,
+    /// Signatures
+    pub pubkey_signatures: Vec<String>,
+}
--- a/crates/cashu/src/nuts/nutxx.rs
+++ b/crates/cashu/src/nuts/nutxx.rs
@@ -2,6 +2,7 @@
 //!
 //! <https://github.com/cashubtc/nuts/blob/get-quotes-by-pubkeys/xx.md>
 
+use bitcoin::secp256k1::schnorr::Signature;
 use serde::{Deserialize, Serialize};
 
 use super::PublicKey;
@@ -18,10 +19,20 @@ pub const MAX_LOOKUP_PUBKEYS: usize = 50;
 /// Mint quote by pubkey request [NUT-XX]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct MintQuoteByPubkeyRequest {
-    /// Pubkeys
-    pub pubkeys: Vec<String>,
-    /// Signatures
-    pub pubkey_signatures: Vec<String>,
+    /// NUT-20 public keys to look up quotes for
+    pub pubkeys: Vec<PublicKey>,
+    /// Schnorr signatures, in the same order as `pubkeys`
+    pub pubkey_signatures: Vec<Signature>,
+}
+
+/// Mint quote by pubkey response [NUT-XX]
+///
+/// Generic over the quote representation so the mint can answer with its own response type
+/// without this crate depending on the unified `MintQuoteResponse` enum in `cdk-common`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MintQuoteByPubkeyResponse<T> {
+    /// Quotes locked to the requested pubkeys, in [NUT-04] response format
+    pub quotes: Vec<T>,
 }
 
 /// Build the message a wallet signs to prove control of `pubkey`.
@@ -75,12 +86,10 @@ mod tests {
         );
     }
 
-    /// The signed message is the pre-image, not its digest: `sign`/`verify` hash internally.
+    /// A signature produced over the pre-image verifies, and one bound to a different mint
+    /// does not — this is what stops a signature being replayed at another mint.
     #[test]
-    fn test_signature_covers_a_single_hash() {
-        use bitcoin::hashes::sha256::Hash as Sha256Hash;
-        use bitcoin::hashes::Hash;
-
+    fn test_sign_and_verify_is_mint_bound() {
         let (mint_pubkey, _) = fixed_keys();
         let secret_key = SecretKey::generate();
         let pubkey = secret_key.public_key();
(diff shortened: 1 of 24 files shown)


crates/cashu/src/nuts/nutxx.rs
```diff
@@ -0,0 +1,25 @@
...
+use crate::{quote_id::QuoteId, BlindSignature, PublicKey};
+
+/// Mint quote by pubkey request [NUT-XX]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(feature = "swagger", derive(utoipa::ToSchema))]
+pub struct MintQuoteByPubkeyRequest {
+    /// Pubkeys
+    pub pubkeys: Vec<PublicKey>,
+    /// Signatures
+    pub pubkeys_signatures: Vec<BlindSignature>,
```
thesimplekid: These aren't going to be blind signatures. In for nut20 we just use a string https://github.com/cashubtc/cdk/blob/42232b38e82852f38b6979421e4bb729915b282c/crates/cashu/src/nuts/nut04.rs#L47-L48 though I think we know theyre all 64 bytes so could use a slice.

crates/cashu/src/nuts/nutxx.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+/// Mint quote by pubkey response [NUT-XX]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(feature = "swagger", derive(utoipa::ToSchema))]
+pub struct MintQuoteByPubkeyResponse {
+    /// Quotes
+    pub quotes: Vec<QuoteId>,
+}
```
thesimplekid: Per the nut pr the response is a Vec<MintQuoteResponse> not just a list of QuoteIds https://github.com/cashubtc/cdk/blob/42232b38e82852f38b6979421e4bb729915b282c/crates/cdk-common/src/mint_quote.rs#L78-L88

crates/cdk-sql-common/src/mint/quotes.rs
```diff
@@ -1194,6 +1194,52 @@ where
         Ok(mint_quotes)
     }
 
+    async fn get_mint_quotes_by_pubkey(
+        &self,
+        pubkeys: &[PublicKey],
+    ) -> Result<Vec<MintQuote>, Self::Err> {
```
thesimplekid, line 1387: We should add a guard to check if the list is empty like we added in this pr. https://github.com/cashubtc/cdk/pull/1822

crates/cdk-sql-common/src/mint/quotes.rs
thesimplekid, line 1: We should add db unit tests for the new fn. https://github.com/cashubtc/cdk/blob/42232b38e82852f38b6979421e4bb729915b282c/crates/cdk-common/src/database/mint/test/mod.rs#L188

crates/cdk-sql-common/src/mint/quotes.rs
```diff
@@ -1194,6 +1194,52 @@ where
...
+                pubkey,
+                created_time,
+                amount_paid,
+                amount_issued,
+                payment_method,
+                request_lookup_id_kind,
+                extra_json
+            FROM
+                mint_quote
+            WHERE pubkey IN (:pubkeys)
```
thesimplekid, line 1417: Do we have an index on pubkey? If not we should add it.

thesimplekid (changes requested):

crates/cdk-common/src/mint.rs
thesimplekid, line 1: Why do we need to add Serialize to these?

crates/cdk-axum/src/custom_handlers.rs
```diff
@@ -545,6 +546,53 @@ pub async fn cache_post_batch_mint(
...
+
+    let response = state
+        .mint
+        .get_mint_quote_by_pubkey(request.pubkeys)
+        .await
+        .map_err(into_response)?;
+
+    Ok(Json(response).into_response())
+}
+
```
thesimplekid: We keep the logic out of cdk-axum the fn added to the `Mint`  should do the verification and db get.

crates/cashu/src/nuts/nutxx.rs
```diff
@@ -0,0 +1,18 @@
...
+use crate::PublicKey;
+
+/// Mint quote by pubkey request [NUT-XX]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "swagger", derive(utoipa::ToSchema))]
+pub struct MintQuoteByPubkeyRequest {
+    /// Pubkeys
+    pub pubkeys: Vec<PublicKey>,
+    /// Signatures
+    pub pubkeys_signatures: Vec<Signature>,
```
thesimplekid: I think we should keep bitcoin types out of the public api. We can just use string or `[u8; 64]`

thesimplekid (changes requested):

crates/cdk-common/src/mint.rs
TheMhv: ~It's a typo fix for old response type that i forget to remove. I will remove then.~

It's because `into_response` needs to serialize the `MintQuote` into a Json object on return
```rust
    let response = state
        .mint
        .get_mint_quote_by_pubkey(pubkeys, signatures)
        .await
        .map_err(into_response)?; // <-- Will return a MintQuote

    Ok(Json(response).into_response()) // <-- This needs to serialize MintQuote into JSON
```

thesimplekid: These are internal types we should not be returning them.

c03rad0r:
Thanks for driving this forward. I have an ESP32-S3 device with an on-device ecash wallet (using the nucula library) that I'd like to connect directly to a CDK mint supporting NUT-XX ("Get quotes by pubkeys", cashubtc/nuts#341). My current workaround routes tokens through a stratum proxy translator, but direct mint queries from the ESP32 would simplify the architecture significantly.

I can offer testing against actual ESP32 hardware with my wallet integration. The secp256k1 Schnorr signing is already available on-device. I'm very much looking forward to integrating this — let me know if there's a branch I should target for integration testing.

TheMhv:
- Implemented https://github.com/cashubtc/nuts/pull/391 on commit 35cb21df6b9ce51600ab3c3a8f805e40ff13dea1

robwoodgate:
One behavioral question not covered by the CDK bot:

get_mint_quote_by_pubkey ignores the {method} path segment when filtering, so a /v1/mint/quote/bolt11/pubkey call returns matching quotes of every method. 

I actually think a method agnostic lookup is probably better, as method is now in quote struct. So have raised the question on the NUT PR.

TheMhv:
> I actually think a method agnostic lookup is probably better, as method is now in quote struct. So have raised the question on the NUT PR.

Me too, so i pushed a separated commit only to remove `{method}` param from request endpoint, we can drop this commit if your suggestion isn't approved.

vnprc:
blast from the past. what changed since github.com/cashubtc/cdk/pull/937?

vnprc:
answered my own question: the nut adds authentication proving the requestor controls the locking privkey. nice.

concept ack, obviously

vnprc:
I opened a PR to fix up this PR: https://github.com/TheMhv/cdk/pull/8

vnprc:
This PR only implements the mint side. We also need a wallet impl to call the new API. I'm happy to build it unless you want to do it @TheMhv. Just let me know.

After testing I think this PR needs a filter on quote ~`state`~ `amount_paid - amount_issued` and `updated_at` fields. This lets the caller limit the query to recent, unpaid quotes.

~I also recommend adding a timestamp to the signed auth message and a mint-side expiration to invalidate stale requests. The current `msg = "Cashu_MintQuoteLookup_v1" ‖ mint_pubkey_hex ‖ wallet_pubkey_hex` spec allows a stolen signed request to be replayed forever.~ nevermind, already proposed and removed in the NUT spec

vnprc:
Also worth noting: this new spec would enable me to delete ~2000 lines of code across 6 repos from the hashpool impl. It also eliminates an entire class of errors: failure to deliver the quote ID from the pool back to the miner currently results in stranded ehash that the miner can't recover.

TheMhv:
> This PR only implements the mint side. We also need a wallet impl to call the new API. I'm happy to build it unless you want to do it...

We already have a PR for this thing, that i need to update, feel free to contribute:
- https://github.com/cashubtc/cdk/pull/1932

ye0man (changes requested):
based on feedback from the wider team and to remain consistent with the spec, we should undo `948578d` and add back method param

TheMhv:
@cdk-bot review

TheMhv:
@cdk-bot review

TheMhv:
@cdk-bot review

crates/cdk-sql-common/src/mint/quotes.rs
```diff
@@ -1381,6 +1381,59 @@ where
         Ok(mint_quotes)
     }
 
+    async fn get_mint_quotes_by_pubkey(
```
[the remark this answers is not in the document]
TheMhv: @thesimplekid @crodas I don't know how we can solve this things without adding a rate limit or quote result limit.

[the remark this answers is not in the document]
thesimplekid: Yeah brought this up on the spec a bit ago. https://github.com/cashubtc/nuts/pull/341#issuecomment-5303998652

thesimplekid:
Can we clean up the commit history here.

PR #1835, opened 2026-04-03
feat: add p2pk_signing_keys to SendOptions

vnprc:
Add `p2pk_signing_keys: Vec<SecretKey>` and `allow_locked_proofs: bool` to `SendOptions` so wallets holding P2PK-locked proofs can spend them through the standard `prepare_send` / `confirm` flow without a separate manual swap.

Behaviour
---------
- When `p2pk_signing_keys` is non-empty and `allow_locked_proofs` is false (the default), all selected proofs are routed through a swap so the resulting token contains fresh, unconditioned proofs.
- Setting `allow_locked_proofs: true` opts in to the less-private passthrough path: proofs are signed and retain their spending conditions in the token. This is useful for multi-hop or offline flows where the next holder intentionally receives signed-but-still-locked proofs.

SIG_ALL incompatibility
-----------------------
`SigFlag::SigAll` is incompatible with passthrough signing: the signature would need to commit to swap outputs that do not exist at signing time. `confirm()` now calls `enforce_sig_flag()` on the proofs-to-send set and returns `nut11::Error::SigAllNotSupportedHere` early rather than silently producing an unspendable token.

Changes
-------
- `cdk-common`: add `p2pk_signing_keys` and `allow_locked_proofs` to `SendOptions`, with doc comments explaining the SigAll incompatibility.
- `cdk/wallet/util.rs`: extract `sign_proofs()` with four unit tests covering correct key, wrong key, plain proof, and empty-keys cases.
- `cdk/wallet/send/saga/mod.rs`: shadow `force_swap` to `true` in `internal_prepare` when signing keys are provided and passthrough is not opted in; call `sign_proofs()` in `confirm()` for both `proofs_to_swap` (default path) and `proofs_to_send` (passthrough path); add SigAll check with an explanatory block comment before the passthrough signing block.
- `cdk-ffi`: add `p2pk_signing_keys` and `allow_locked_proofs` to FFI `SendOptions` and update both `From` conversions.
- Integration tests: add four pure tests covering the normal signing flow, the exact-denomination short-circuit regression, the passthrough opt-in, and the SigAll rejection.

### Description

Add `p2pk_signing_keys: Vec<SecretKey>` and `allow_locked_proofs: bool` to `SendOptions` so wallets holding P2PK-locked proofs can spend them through the standard `prepare_send` / `confirm` flow without a separate manual swap.

**Default behaviour** (`allow_locked_proofs: false`): all selected proofs are routed through a swap, producing a clean token with no spending conditions. The caller provides signing keys only to authorize the swap inputs.

**Passthrough** (`allow_locked_proofs: true`): proofs are signed in-place and retain their spending conditions. Useful for multi-hop or offline flows where the next holder intentionally receives signed-but-still-locked proofs.

**SigAll rejection**: `SigFlag::SigAll` is incompatible with passthrough because the signature would need to commit to swap outputs that don't exist at signing time. `confirm()` detects this via `enforce_sig_flag()` and returns `nut11::Error::SigAllNotSupportedHere` early rather than silently producing an unspendable token.

Also fixes a force-swap short-circuit bug where proofs summing exactly to the send amount bypassed the swap entirely, causing locked proofs to escape into the token unsigned.

## Test plan

- [ ] `cargo test -p cdk wallet::util` — 5 unit tests pass
- [ ] `test_p2pk_send_options_signing_keys` — normal signing flow
- [ ] `test_p2pk_signing_keys_exact_denomination_short_circuit` — force-swap regression
- [ ] `test_p2pk_allow_locked_proofs_passthrough` — passthrough opt-in
- [ ] `test_p2pk_allow_locked_proofs_rejects_sig_all` — SigAll rejection

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/wallet/mod.rs
+++ b/crates/cdk-common/src/wallet/mod.rs
@@ -335,6 +335,10 @@ pub struct SendOptions {
     pub metadata: HashMap<String, String>,
     /// Use P2BK (NUT-28)
     pub use_p2bk: bool,
+    /// Signing keys for P2PK-locked input proofs; auto-detected from the wallet keyring if omitted
+    pub p2pk_signing_keys: Vec<SecretKey>,
+    /// If `true`, sign and forward P2PK-locked proofs directly without swapping them for fresh ones
+    pub allow_locked_proofs: bool,
 }
 
 /// Send memo
--- a/crates/cdk-common/src/wallet/mod.rs
+++ b/crates/cdk-common/src/wallet/mod.rs
@@ -337,8 +337,18 @@ pub struct SendOptions {
     pub use_p2bk: bool,
     /// Signing keys for P2PK-locked input proofs; auto-detected from the wallet keyring if omitted
     pub p2pk_signing_keys: Vec<SecretKey>,
-    /// If `true`, sign and forward P2PK-locked proofs directly without swapping them for fresh ones
-    pub allow_locked_proofs: bool,
+    /// How P2PK-locked input proofs should be handled during send
+    pub p2pk_locked_proof_send_mode: P2PKLockedProofSendMode,
+}
+
+/// Send behavior for selected P2PK-locked input proofs
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+pub enum P2PKLockedProofSendMode {
+    /// Swap locked proofs into fresh proofs before creating the token
+    #[default]
+    Swap,
+    /// Sign locked proofs and include them directly in the token
+    SignAndSend,
 }
 
 /// Send memo
--- a/crates/cdk-ffi/src/types/wallet.rs
+++ b/crates/cdk-ffi/src/types/wallet.rs
@@ -160,6 +160,10 @@ pub struct SendOptions {
     pub max_proofs: Option<u32>,
     /// Metadata
     pub metadata: HashMap<String, String>,
+    /// Signing keys for P2PK-locked input proofs
+    pub p2pk_signing_keys: Vec<SecretKey>,
+    /// Allow P2PK-locked proofs to be included directly in the token without a swap
+    pub allow_locked_proofs: bool,
 }
 
 impl Default for SendOptions {
@@ -173,13 +177,23 @@ impl Default for SendOptions {
             max_proofs: None,
             metadata: HashMap::new(),
             use_p2bk: false,
+            p2pk_signing_keys: Vec::new(),
+            allow_locked_proofs: false,
         }
     }
 }
 
-impl From<SendOptions> for cdk::wallet::SendOptions {
-    fn from(opts: SendOptions) -> Self {
-        cdk::wallet::SendOptions {
+impl TryFrom<SendOptions> for cdk::wallet::SendOptions {
+    type Error = FfiError;
+
+    fn try_from(opts: SendOptions) -> Result<Self, Self::Error> {
+        let p2pk_signing_keys = opts
+            .p2pk_signing_keys
+            .into_iter()
+            .map(TryInto::try_into)
+            .collect::<Result<Vec<_>, _>>()?;
+
+        Ok(cdk::wallet::SendOptions {
             memo: opts.memo.map(Into::into),
             conditions: opts.conditions.and_then(|c| c.try_into().ok()),
             amount_split_target: opts.amount_split_target.into(),
@@ -188,7 +202,9 @@ impl From<SendOptions> for cdk::wallet::SendOptions {
             max_proofs: opts.max_proofs.map(|p| p as usize),
             metadata: opts.metadata,
             use_p2bk: opts.use_p2bk,
-        }
+            p2pk_signing_keys,
+            allow_locked_proofs: opts.allow_locked_proofs,
+        })
     }
 }
 
@@ -203,6 +219,8 @@ impl From<cdk::wallet::SendOptions> for SendOptions {
             max_proofs: opts.max_proofs.map(|p| p as u32),
             metadata: opts.metadata,
             use_p2bk: opts.use_p2bk,
+            p2pk_signing_keys: opts.p2pk_signing_keys.into_iter().map(Into::into).collect(),
+            allow_locked_proofs: opts.allow_locked_proofs,
         }
     }
 }
--- a/crates/cdk-ffi/src/types/wallet.rs
+++ b/crates/cdk-ffi/src/types/wallet.rs
@@ -142,6 +142,40 @@ impl From<cdk::wallet::SendKind> for SendKind {
     }
 }
 
+/// FFI-compatible P2PK locked proof send mode
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, uniffi::Enum, Default,
+)]
+pub enum P2PKLockedProofSendMode {
+    /// Swap locked proofs into fresh proofs before creating the token
+    #[default]
+    Swap,
+    /// Sign locked proofs and include them directly in the token
+    SignAndSend,
+}
+
+impl From<P2PKLockedProofSendMode> for cdk::wallet::P2PKLockedProofSendMode {
+    fn from(mode: P2PKLockedProofSendMode) -> Self {
+        match mode {
+            P2PKLockedProofSendMode::Swap => cdk::wallet::P2PKLockedProofSendMode::Swap,
+            P2PKLockedProofSendMode::SignAndSend => {
+                cdk::wallet::P2PKLockedProofSendMode::SignAndSend
+            }
+        }
+    }
+}
+
+impl From<cdk::wallet::P2PKLockedProofSendMode> for P2PKLockedProofSendMode {
+    fn from(mode: cdk::wallet::P2PKLockedProofSendMode) -> Self {
+        match mode {
+            cdk::wallet::P2PKLockedProofSendMode::Swap => P2PKLockedProofSendMode::Swap,
+            cdk::wallet::P2PKLockedProofSendMode::SignAndSend => {
+                P2PKLockedProofSendMode::SignAndSend
+            }
+        }
+    }
+}
+
 /// FFI-compatible Send options
 #[derive(Debug, Clone, Serialize, Deserialize, uniffi::Record)]
 pub struct SendOptions {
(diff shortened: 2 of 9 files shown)


crates/cdk-common/src/wallet/mod.rs
```diff
@@ -323,6 +323,42 @@ pub struct SendOptions {
...
+    ///
+    /// # SIG_ALL incompatibility
+    ///
+    /// Proofs whose spending condition carries `SigFlag::SigAll` cannot be passed through.
+    /// A `SIG_ALL` signature must commit to the swap outputs, which are unknown at signing time.
+    /// The recipient cannot construct valid outputs for such a proof, so any redemption attempt
+    /// would be rejected by the mint. `confirm` will return
+    /// [`cashu::nuts::nut11::Error::SigAllNotSupportedHere`] if any passthrough proof uses
+    /// `SIG_ALL`.
+    pub allow_locked_proofs: bool,
```
lescuer97: this flag doesn't make sense. If we wasnt to support p2pk locked proofs in send (which we should) we just want to accept them and sign them with either the local keyring for p2pk or the outsite `p2pk_signing_keys`.

crates/cdk/src/wallet/send/saga/mod.rs
```diff
@@ -316,6 +317,11 @@ impl<'a> SendSaga<'a, Initial> {
         let is_exact_or_offline =
             exact_proofs || opts.send_kind.is_offline() || opts.send_kind.has_tolerance();
 
+        // When p2pk_signing_keys are provided and passthrough is not opted in, all proofs must go
+        // through a swap so the token contains fresh, unconditioned proofs.
+        let force_swap =
+            force_swap || (!opts.p2pk_signing_keys.is_empty() && !opts.allow_locked_proofs);
```
lescuer97: let's not force swap. Mints are starting to have fees. this is an extra cost for the user. 

this force swaps even if there are no locked proofs.

crates/cdk/src/wallet/util.rs
```diff
@@ -1,5 +1,98 @@
...
+use crate::{Error, SECP256K1};
+
+/// Sign P2PK-locked proofs using the provided signing keys.
+///
+/// For each proof with a recognised NUT-10 secret:
+/// - P2PK: signs the data key (slot 0) and any condition keys (slots 1+)
+/// - HTLC: signs condition keys (slots 1+) only; preimage injection is the caller's responsibility
+///
+/// Proofs without a NUT-10 secret, or with no matching signing key, are left unchanged.
+pub(crate) fn sign_proofs(
```
lescuer97, line 82: ~~we already have a sign_proof function called `sign_p2pk`~~

If we want to use this function you should also use the existing keyring to detect needed keys for signing.

lescuer97 (commented):
some comments around the code and proper flow

crates/cdk-common/src/wallet/mod.rs
lescuer97: If we want to keep the flag because of some privacy concerns. the swap should only include the p2pk locked proofs.

crates/cdk/src/wallet/send/saga/mod.rs
lescuer97: where does this force_swap actually get's triggered?

crates/cdk-common/src/wallet/mod.rs
thesimplekid: I think this flag does make sense for the privacy reason and we should default to not including proofs with witnesses as it is more private and I'm not convinced all wallets would handle receiving a proof with a witness correctly even though they should.

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -323,6 +323,42 @@ pub struct SendOptions {
...
+    /// corresponding secret keys here so they are signed before the swap in `confirm`.
+    ///
+    /// When non-empty and [`SendOptions::allow_locked_proofs`] is `false` (the default), all
+    /// selected proofs are routed through a swap so the resulting token contains fresh,
+    /// unconditioned proofs.
+    pub p2pk_signing_keys: Vec<SecretKey>,
+    /// Allow P2PK-locked proofs to be included directly in the outgoing token without a swap.
+    ///
+    /// By default (`false`), providing [`SendOptions::p2pk_signing_keys`] forces a swap so the
+    /// token always contains fresh, unlocked proofs. Setting this to `true` opts in to the
```
thesimplekid: While I appreciate good doc comments, I think these are too verbose and should be tightened up.

crates/cdk/src/wallet/send/saga/mod.rs
thesimplekid: see https://github.com/cashubtc/cdk/pull/1835/changes#r3037329809. We should default to swapping locked proofs

CHANGELOG.md
thesimplekid, line 1: Can we remove edits to the changelog I will do that at release

thesimplekid (changes requested):

crates/cdk-common/src/wallet/mod.rs
vnprc: I originally coded it up without the flag but after tsk's suggestion I plan to force swaps on all ehash withdrawals for privacy reasons. It is a useful flag for other use cases like offline send, so we should def keep it. Good idea bro.

vnprc: > If we want to keep the flag because of some privacy concerns. the swap should only include the p2pk locked proofs.

Interesting point. Upon reflection, I don't think restricting the mixing of locked and unlocked proofs when the flag is true is valuable. You already give up your privacy when you include a single locked proof in the token. Disallowing additional unlocked proofs in the token doesn't improve the privacy situation but it does significantly reduce the flexibility (and usefulness) of this function in the cdk library. I don't believe this is a good change to make.

crates/cdk/src/wallet/send/saga/mod.rs
vnprc: good point, i will update the PR to only swap locked proofs and then include all unlocked proofs in the token

crates/cdk/src/wallet/util.rs
vnprc: smart!

vnprc:
thanks y'all. i rebased against main and all comments addressed. ready for another look

vnprc:
@thesimplekid any ETA on this? should i rebase?

crates/cdk/src/wallet/send/saga/mod.rs
```diff
@@ -88,6 +92,96 @@ use crate::{Amount, Error, Wallet};
...
+
+        // Extract the data key and check explicit keys first.
+        let Ok(secret) = <crate::secret::Secret as TryInto<crate::nuts::nut10::Secret>>::try_into(
+            proof.secret.clone(),
+        ) else {
+            out.push(proof);
+            continue;
+        };
+
+        let Ok(data_key) = crate::nuts::PublicKey::from_str(secret.secret_data().data()) else {
```
thesimplekid: This signability filter only checks the P2PK data key, but P2PK conditions can also require additional `pubkeys` via `n_sigs`. A proof where the wallet has the data key but not enough additional required keys will pass this filter, then fail later at swap/confirm with a mint rejection instead of being excluded as unsignable. Can this reuse the same requirements logic as verification, or sign and verify before treating the proof as selectable?

crates/cdk/src/wallet/send/saga/mod.rs
```diff
@@ -444,13 +594,47 @@ impl<'a> SendSaga<'a, Prepared> {
...
+                if sig_flag == SigFlag::SigAll {
+                    return Err(crate::nuts::nut11::Error::SigAllNotSupportedHere.into());
+                }
+                let keys = merge_keyring_keys(
+                    self.wallet,
+                    &final_proofs_to_send,
+                    &options.p2pk_signing_keys,
+                )
+                .await?;
+                if !keys.is_empty() {
```
thesimplekid, line 792: In passthrough mode this signs only when a matching key is found, but it does not fail if no key is found or if the proof is only partially signed. That can create a token containing locked proofs that the recipient cannot redeem. Before creating the token, this should verify the passthrough proofs satisfy their spending conditions and return an error if they do not.

crates/cdk-ffi/src/types/wallet.rs
```diff
@@ -188,6 +194,12 @@ impl From<SendOptions> for cdk::wallet::SendOptions {
             max_proofs: opts.max_proofs.map(|p| p as usize),
             metadata: opts.metadata,
             use_p2bk: opts.use_p2bk,
+            p2pk_signing_keys: opts
+                .p2pk_signing_keys
+                .into_iter()
+                .filter_map(|k| k.try_into().ok())
```
thesimplekid: This silently drops invalid FFI secret keys. `ReceiveOptions` uses a fallible conversion and returns an error for invalid `p2pk_signing_keys`; send should do the same so callers do not get a later `InsufficientFunds` or mint rejection caused by a key that was ignored during conversion.

thesimplekid (changes requested):
Noticed a few things and yes could you squash and rebase and we can get it merged thanks.

vnprc:
ready for re-review. my agent also found an ffi bug that i will be opening a different pr to fix

vnprc:
related bug fix: #1970

vnprc:
oh shit my agent force pushed this branch. woops. i will try to reconcile the two force pushes.

vnprc:
ah ok nm, it looks like codex did the right thing. please let me know if the force push was destructive. i tried to put guardrails in place but it seems like controlling agents is an impossible task. ¯\\\_(ツ)_/¯

thesimplekid:
We can avoid needed the direct secp dep by just changing the error to not use transparent, I did this in this commit https://github.com/cashubtc/cdk/pull/1835/commits/bf2757f365be3606736e4521b0b3b2d339b519bd. I also think it is more explixt and better if we use a enum vs the bool, I added that here https://github.com/cashubtc/cdk/pull/1835/commits/c2d29dcde7db97fa3c8ac6ca3a059c8bd30af411. If you're happy with those changes I can merge and do any further rebasing once CI completes.

vnprc:
LGTM

I have confirmed that hashpool mints ehash with this PR branch as it is. Sorry about the dependency churn, I need to work on my agent harness and put more human in the loop before pushing to a PR branch. Thanks for all the fee and enum improvements! Great stuff!

thesimplekid (approved):

PR #1876, opened 2026-04-11
fix(cdk-cli): parse --unit consistently and scope balance by currency

b-l-u-e:
This PR improves how the global `--unit` flag is parsed and how `balance` uses it, so the CLI no longer misleads users about which currency unit they are acting on. Plural aliases and a typo check align input with Cashu’s standard units; `--unit` is global so it works both before and after the subcommand.

### issues i faced 

**Ambiguous or misleading unit handling**

A typo or odd string could still parse as a custom unit, so what i typed did not match what i saw on the screen. For example, a mistyped unit could still show **sat** balances as if nothing was wrong:

```text
cdk --unit satt balance
 http://127.0.0.1:8085 100 sat

Total balance across all wallets: 179 sat
```

**`balance` and `--unit`**

`--unit` did not line up with what i expected for `balance`. Asking for non-matching unit could still show **sat** totals. And putting `--unit` after the subcommand failed entirely:

```text
cdk --unit dollarz balance
 http://127.0.0.1:8085 100 sat

Total balance across all wallets: 179 sat
```

```text
cdk balance --unit usd
error: unexpected argument '--unit' found

Usage: cdk-cli balance

For more information, try '--help'.
```

Whereas putting `--unit` before the command still did not filter to that unit when you only held sat:

```text
cdk --unit usd balance
0: http://127.0.0.1:8085 100 sat
1: https://fake.thesimplekid.dev 79 sat

Total balance across all wallets: 179 sat
```

**Empty `--unit`**

Passing an empty value looked like “no unit,” but the tool still behaved like a normal **sat** run, which is not the same as omitting `--unit` on purpose:

```text
cdk --unit "" balance
0: http://127.0.0.1:8085 100 sat

Total balance across all wallets: 179 sat
```

### What this PR changes

- Stricter parsing: plural aliases, explicit error for `satt`, clear error for empty `--unit` (omit the flag for default `sat`).
- `balance` respects the resolved unit (default `sat` when `--unit` is omitted).
- `global = true` on `--unit` so `cdk balance --unit usd` and `cdk --unit usd balance` are equivalent.

### DIFF
--- a/crates/cdk-cli/src/sub_commands/balance.rs
+++ b/crates/cdk-cli/src/sub_commands/balance.rs
@@ -7,9 +7,13 @@ use cdk::wallet::WalletRepository;
 use cdk::Amount;
 use cdk_common::wallet::WalletKey;
 
-pub async fn balance(wallet_repository: &WalletRepository) -> Result<()> {
-    // Show individual mint balances
-    let mint_balances = mint_balances(wallet_repository).await?;
+/// Print balances for each mint for the given [`CurrencyUnit`] (e.g. sat from `cdk balance`, or
+/// `--unit usd` for USD-only wallets).
+pub async fn balance(
+    wallet_repository: &WalletRepository,
+    filter_unit: &CurrencyUnit,
+) -> Result<()> {
+    let mint_balances = mint_balances(wallet_repository, filter_unit).await?;
 
     if !mint_balances.is_empty() {
         // Aggregate totals per currency unit
@@ -29,6 +33,8 @@ pub async fn balance(wallet_repository: &WalletRepository) -> Result<()> {
                 println!("  {} {}", total, unit);
             }
         }
+    } else {
+        println!("No balance for unit {filter_unit}.");
     }
 
     Ok(())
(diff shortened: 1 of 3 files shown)


crates/cdk-cli/src/sub_commands/balance.rs
```diff
@@ -7,9 +7,13 @@ use cdk::wallet::WalletRepository;
...
 use cdk_common::wallet::WalletKey;
 
-pub async fn balance(wallet_repository: &WalletRepository) -> Result<()> {
-    // Show individual mint balances
-    let mint_balances = mint_balances(wallet_repository).await?;
+/// Print balances for each mint for the given [`CurrencyUnit`] (e.g. sat from `cdk balance`, or
+/// `--unit usd` for USD-only wallets).
+pub async fn balance(
+    wallet_repository: &WalletRepository,
+    filter_unit: &CurrencyUnit,
```
lescuer97, line 14: I think you want this to be and Option and filter if it's not None

lescuer97 (commented):
logic change

thesimplekid:
Close for https://github.com/cashubtc/cdk/pull/1930

PR #1878, opened 2026-04-11
Move empty IN clause guard into bind_vec

crodas:
### Description

Instead of scattering empty-vec checks at each call site, bind_vec now returns Result and errors with EmptyInClause when given an empty vector.  This prevents invalid SQL from being generated at the source.

Fixes #1858 and #1813

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sql-common/src/mint/auth/mod.rs
+++ b/crates/cdk-sql-common/src/mint/auth/mod.rs
@@ -224,10 +224,6 @@ where
         &mut self,
         protected_endpoints: Vec<ProtectedEndpoint>,
     ) -> Result<(), database::Error> {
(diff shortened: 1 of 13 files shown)


crates/cdk-sql-common/src/mint/auth/mod.rs
```diff
@@ -334,7 +330,7 @@ where
         }
```
thesimplekid: Can we remove these guards now that we have this. Suspect there are mote then this one.

thesimplekid (changes requested):

thesimplekid (approved):

PR #1879, opened 2026-04-12
feat: add Go bindings build and tests to CI

Forte11Cuba:
### Description

Add Go bindings build and tests to CDK CI, following the same pattern used by Dart, Kotlin, and Swift bindings.

- Create `cdk-ffi-go` wrapper crate in `bindings/go/rust/`
- Add `generate-bindings.sh` script using `uniffi-bindgen-go` (v0.6.0+v0.30.0)
- Add `binding-go` and `test-go` justfile recipes
- Add `go-binding-tests` job to CI workflow
- Add Go (`pkgs.go`) to the Nix `#bindings` dev shell
- Exclude `cdk-ffi-go` from MSRV builds

Closes #1860

-----

### Notes to the reviewers

-----

#### ADDED

- Added Go FFI bindings (`cdk-ffi-go`) and CI job to build and verify them

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -516,6 +516,25 @@ jobs:
       - name: Run Kotlin tests
         run: nix develop -L .#bindings --command just test-kotlin
 
+  go-binding-tests:
+    name: "Go binding tests"
+    runs-on: self-hosted
+    timeout-minutes: 30
+    needs: pre-commit-checks
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+      - uses: cachix/cachix-action@v16
+        with:
+          name: cashudevkit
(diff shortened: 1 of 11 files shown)


asmogo (commented):
Thanks for this contribution.

We need to install `uniffi-bindgen-go` in the CI environment. This can be done like [this](https://github.com/cashubtc/cdk-go/blob/main/scripts/install-uniffi-bindgen-go.sh). 

If this does not work, we would have to install the package directly on the runners.

asmogo:
With this new approach, are we building the bindings twice? Once for testing inside CDK and once for publishing the actual bindings in the dedicated downstream repositories? I believe we could improve on that process.

PR #1884, opened 2026-04-13
fix(cdk-lnd): honor mint quote TTL on LND invoices

orangeshyguy21:
### Description

LND invoices were always created with the default 24h expiry, ignoring `mint_ttl`. Set `lnrpc::Invoice.expiry` from the quote TTL and read back the actual expiry from the bolt11, matching CLN/LNBits behavior.

-----

### Notes to the reviewers

The `lnrpc::Invoice.expiry` field expects a duration in seconds (not a unix timestamp). Since the mint passes an absolute unix timestamp (`unix_expiry`), we convert it to a relative duration via `saturating_sub(unix_time())`, same pattern CLN and LNBits already use.

-----

#### FIXED

- Fixed LND invoices ignoring `mint_ttl` and always using 24h default expiry

----

### Checklist

* [X] I followed the [code style guidelines](https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md)
* [X] I ran `just quick-check` before committing

### DIFF
--- a/crates/cdk-lnd/src/lib.rs
+++ b/crates/cdk-lnd/src/lib.rs
@@ -23,7 +23,7 @@ use cdk_common::payment::{
     MintPayment, OutgoingPaymentOptions, PaymentIdentifier, PaymentQuoteResponse, SettingsResponse,
     WaitPaymentResponse,
 };
-use cdk_common::util::hex;
+use cdk_common::util::{hex, unix_time};
 use cdk_common::Bolt11Invoice;
 use error::Error;
 use futures::{Stream, StreamExt};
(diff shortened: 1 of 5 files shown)


crates/cdk-lnd/src/lib.rs
```diff
@@ -634,6 +634,9 @@ impl MintPayment for Lnd {
                 let invoice_request = lnrpc::Invoice {
                     value_msat: u64::from(amount_msat) as i64,
                     memo: description,
+                    expiry: unix_expiry
+                        .map(|t| t.saturating_sub(unix_time()) as i64)
+                        .unwrap_or(0),
```
lescuer97: my understanding of this is that it the function fails for what ever reason it will just expire intermediately. I don't thing that is the behavior we expect

thesimplekid: Yeah was thinking the same thing. But needed to check the lnd api to confirm.

orangeshyguy21: LND interprets 0 as unset, and falls back to it's default expiry of 86400s

However, the intent to use default in the unhappy path is implicit here and could be more explicit by omitting the parameter instead.

Will update the PR, thanks for the review

PR #1875, opened 2026-04-10
refactor(cdk): use polymorphic enums for mint and melt quotes

thesimplekid:
### Description

-----
- [x] depends on https://github.com/cashubtc/cdk/pull/1873
- [x] depends on https://github.com/cashubtc/cdk/pull/1897
- [x] depends on https://github.com/cashubtc/cdk/pull/1900
- [x] depends on https://github.com/cashubtc/cdk/pull/1910
### Notes to the reviewers

-----

#### FIXED

----

### Checklist

* [ ] I followed the [code style guidelines](https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md)
* [ ] I ran `just quick-check` before committing
* [ ] If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`)

PR #1885, opened 2026-04-13
Async melt streams

thesimplekid:
### Description

- [x] depends on https://github.com/cashubtc/cdk/pull/1875
-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/melt/mod.rs
+++ b/crates/cdk/src/mint/melt/mod.rs
@@ -1,16 +1,21 @@
 use std::future::Future;
 use std::pin::Pin;
 use std::str::FromStr;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
 
+use cdk_common::database::DynMintDatabase;
 use cdk_common::melt::MeltQuoteRequest;
 use cdk_common::mint::MeltPaymentRequest;
 use cdk_common::nut00::KnownMethod;
 use cdk_common::nut05::MeltMethodOptions;
+use cdk_common::nuts::nut17::{Kind, NotificationPayload};
 use cdk_common::payment::{
     Bolt11OutgoingPaymentOptions, Bolt12OutgoingPaymentOptions, CustomOutgoingPaymentOptions,
     OutgoingPaymentOptions,
 };
 use cdk_common::quote_id::QuoteId;
+use cdk_common::subscription::Params;
 use cdk_common::{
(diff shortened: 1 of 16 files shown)


crates/cdk/src/mint/melt/shared.rs
```diff
@@ -119,19 +119,31 @@ pub async fn rollback_melt_quote(
...
-            .await?;
-
-        if previous_state != MeltQuoteState::Pending {
-            tracing::warn!(
-                "Unexpected quote state during rollback: expected Pending, got {}",
-                previous_state
-            );
+        match quote.state {
+            MeltQuoteState::Pending => {
+                tx.update_melt_quote_state(&mut quote, MeltQuoteState::Unpaid, None)
```
lescuer97, line 125: shouldn't this be pending?

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -77,6 +106,74 @@ impl std::future::IntoFuture for PendingMelt {
 }
 
 impl Mint {
+    async fn load_terminal_melt_response(
```
lescuer97, line 109: what does load_terminal mean?

crates/cdk/src/mint/mod.rs
```diff
@@ -750,6 +802,129 @@ impl Mint {
...
+            return Ok(());
+        };
+
+        match quote.state {
+            MeltQuoteState::Pending => {}
+            MeltQuoteState::Paid => {
+                tracing::info!(
+                    "Ignoring failed payment event for already paid melt quote {}",
+                    quote_id
+                );
```
lescuer97, line 886: should this be a warning? I feel this a case that should not happen

lescuer97 (commented):
just an initial set of questions

lescuer97 (commented):
just an initial set of questions

crates/cdk/src/mint/melt/shared.rs
thesimplekid: This is a rollback path so we actually expect the quote to be in pending and the payment failed so we use this fn to reset it to unpaid. I've added some comments one this.

crates/cdk/src/mint/melt/mod.rs
thesimplekid: Renamed and added comments

crates/cdk/src/mint/mod.rs
thesimplekid: It can happen is running multiple mint instances.

PR #1886, opened 2026-04-14
Add dart publish workflow, prebuilt binary support, and README

crodas:
### Description

Add a GitHub Actions workflow (publish.yml) for the cdk-dart repo that automates the full Dart release pipeline: updates Cargo.toml versions, builds native FFI libraries for 8 targets (Linux x86_64/arm64, macOS arm64, Windows x86_64, Android arm64/armv7/x86_64, iOS arm64), generates Dart bindings via uniffi-bindgen, and creates a GitHub release with tag.

Extend the Dart build hook (build.dart) to detect and use prebuilt binaries from a `prebuilt/<target-triple>/` directory, skipping the cargo build when available. Falls back to building from source otherwise.

Enable npubcash and bip353 features on cdk-ffi for the Dart bindings.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


bindings/dart/.github/workflows/publish.yml
```diff
@@ -0,0 +1,226 @@
...
+  update-cargo:
+    name: "Update CDK version in Cargo.toml"
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: master
```
thesimplekid: cdk uses main can we keep it the same.

.github/workflows/dart-publish.yml
```diff
@@ -0,0 +1,226 @@
...
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: native-${{ matrix.target }}
+          path: dist/
+
+  generate-bindings:
+    name: "Generate Dart bindings"
+    needs: [update-cargo]
+    runs-on: ubuntu-22.04
```
thesimplekid: We switch between ubuntu-22.04 and ubuntu-latest, why?

thesimplekid: We have had issues on the GH runners with running out of space during builds so there is a preference for nix defined on the self hosted. However, for the releases if this works we can keep it since we have the other os builds anyway.

.github/workflows/dart-publish.yml
thesimplekid, line 1: This changes the logic from our existing bindings flow where we pull cdk into those repos to we push the built bindings into the repos. I think this is a good change as keeps more logic here inline with our tests and makes those very thin just for release. In follow up PRs we should change the other (swift/kotlin) ones so we all match as well as create a high level doc explaining this flow.

crodas: No, this is for building, testing, and keeping it in sync with the external release-only repo as we agreed.

The release-only repositories are going to be like https://github.com/crodas/cdk-dart, which pulls cdk in from crate, and prebuilds, so people without Rust can install.

crodas:
Here is an example in action from my personal repos:

* [Monorepo task](https://github.com/crodas/cdk/actions/runs/24490921032)
* [the dart built triggered by the above](https://github.com/crodas/cdk-dart)

@thesimplekid @Forte11Cuba ^

.github/workflows/dart-publish.yml
```diff
@@ -0,0 +1,299 @@
...
+      release_tag:
+        description: 'Tag to create for release (e.g. v0.17.0)'
+        required: true
+      cdk_version:
+        description: 'CDK dependency version (e.g. 0.17.0). Defaults to release_tag without the "v" prefix.'
+        required: false
+
+env:
+  CARGO_TERM_COLOR: always
+  CDK_DART_REPO: crodas/cdk-dart
```
thesimplekid, line 15: Lets change this to cashubtc before merge.

.github/workflows/dart-publish.yml
```diff
@@ -0,0 +1,299 @@
...
+            https://x-access-token:${{ secrets.DART_DEPLOY_KEY_TEST }}@github.com/${{ env.CDK_DART_REPO }}.git cdk-dart
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+
+      - name: Install cross (Android)
+        if: contains(matrix.target, 'android')
+        run: cargo install cross --git https://github.com/cross-rs/cross
```
thesimplekid, line 177: Should probably pin a commit hash or version?

bindings/dart/README.md
thesimplekid, line 1: Can we add some docs to cover what we disused on the call of how this works. It is maintained and updated here and pushed to cdk-dart.

thesimplekid (commented):
Can you also document what secrets we need to set where?

bindings/dart/README.md
crodas: Absolutely

crodas:
Chopped a few minutes off the building process 
image

bindings/dart/README.md
crodas: Is the README complete enough?

bindings/dart/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+| Name | Purpose |
+|---|---|
+| `DART_DEPLOY_KEY` | Personal access token (PAT) with `repo` scope on the `cdk-dart` target repo. Used to clone, push, and create releases. |
+
+#### How to create the PAT
+
+1. Go to **GitHub → Settings → Developer settings → Personal access tokens → Fine-grained tokens**.
+2. Create a token scoped to the `cdk-dart` repository with **Contents** (read/write) and **Metadata** (read) permissions.
+3. Add it as a repository secret named `DART_DEPLOY_KEY` in the monorepo.
```
thesimplekid, line 65: Done

bindings/dart/README.md
```diff
@@ -0,0 +1,77 @@
...
+## Pre-built binaries
+
+Pre-built native libraries for all supported platforms are available as [GitHub release assets](https://github.com/cashubtc/cdk-dart/releases).
+
+Supported targets:
+
+| Platform | Architecture |
+|----------|-------------|
+| Linux | x86_64, aarch64 |
+| macOS | x86_64, aarch64 |
```
thesimplekid:
```suggestion
| macOS | aarch64 |
```

bindings/dart/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Building
+
+The Rust native library is built automatically when you run `dart pub get` or `dart run`. No manual compilation step is needed.
+
+If you're in a Nix environment, OpenSSL paths are detected automatically from `NIX_CFLAGS_COMPILE` and `NIX_LDFLAGS`.
+
+## Pre-built binaries
+
+Pre-built native libraries for all supported platforms are available as [GitHub release assets](https://github.com/cashubtc/cdk-dart/releases).
```
thesimplekid, line 36: I don't think we upload to releases we just upload in the repo which is fine should just correct docs.

.github/workflows/dart-publish.yml
```diff
@@ -0,0 +1,311 @@
...
+        working-directory: cdk-dart
+        run: |
+          TAG="${{ inputs.release_tag }}"
+          RELEASE_BRANCH="release/v${TAG#v}"
+          DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+          git checkout "${DEFAULT_BRANCH}"
+          git merge "${RELEASE_BRANCH}" --squash
+          git commit -m "release: ${TAG}"
+          git tag -f "${TAG}"
+          git push origin "${DEFAULT_BRANCH}" --tags -f
```
thesimplekid: I think we should avoid force pushing to the default branch and overwriting tags.

thesimplekid (changes requested):

bindings/dart/README.md
crodas: We do, but I can stop doing it

https://github.com/crodas/cdk-dart/releases

.github/workflows/dart-publish.yml
crodas: I had it this way because I was testing. I will remove the forcing today

bindings/dart/README.md
thesimplekid: The assets aren't there though. We do tag releases as we should. The assets are only here https://github.com/crodas/cdk-dart/tree/master/prebuilt

crodas: I see the assets within the zip / tgz under releases

image

thesimplekid: Oh well I guess that counts. GH just zips the whole repo on release. I thought it meant up uploads them as assets, like we do for our binarys. Doesn't matter anyway. 

image

crodas: I see, I would explore creating a zip package with each prebuild for each platform, otherwise the download zip is big. That's a good idea. I'll make it happen now/

thesimplekid: I don't think anyone is going to use these zips so doesn't really matter.

thesimplekid (approved):

PR #1890, opened 2026-04-15
Feat/prepare melt token

### LINKED ISSUE #1888: Add `melt_token` fn to wallet
Currently cdk has the `prepare_melt` fn that will select proofs for a melt quote from the internal wallet and we have the `prepare_melt_proofs` command that accepts proofs if a wallet wants to do its own coin selection. However, it would be helpful to add a `prepare_melt_token` command that accepts a token directly. This would allow the wallet to skip decoding the token which because of keysetv2 requires keyset state and offload that to the wallet during the prepare_melt_token step. This would be specifically helpful to the flow in Numo without much added overhead to cdk. 

- [ ] add to wallet trait 
- [ ] expose in ffi

Forte11Cuba:
### Description

Adds `prepare_melt_token` to the `Wallet` trait, core implementation, and FFI bindings.
 
Currently CDK has `prepare_melt` (auto-selects proofs) and `prepare_melt_proofs` (accepts raw proofs), but there's no way to pass an encoded token  directly. 

This new method internalizes that complexity, following the same pattern as `receive()`.                                                                                                            
                                                          
Closes #1888 

-----

### Notes to the reviewers

-----

#### ADDED

- Added `prepare_melt_token` to `Wallet` trait and FFI bindings, allowing callers to pass an encoded token (cashuA/cashuB) directly for melt operations 

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/melt/mod.rs
+++ b/crates/cdk/src/wallet/melt/mod.rs
@@ -654,6 +654,7 @@ impl Wallet {
         ensure_cdk!(self.mint_url == token.mint_url()?, Error::IncorrectMint);
 
         let keysets_info = self.load_mint_keysets().await?;
+        println!("{:?}", keysets_info);
         let proofs = token.proofs(&keysets_info)?;
 
         self.prepare_melt_proofs(quote_id, proofs, metadata).await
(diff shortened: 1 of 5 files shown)


crates/cdk/src/wallet/melt/mod.rs
```diff
@@ -581,6 +582,29 @@ impl Wallet {
...
+    /// and delegates to [`prepare_melt_proofs`](Wallet::prepare_melt_proofs).
+    #[instrument(skip(self, encoded_token, metadata))]
+    pub async fn prepare_melt_token(
+        &self,
+        quote_id: &str,
+        encoded_token: &str,
+        metadata: HashMap<String, String>,
+    ) -> Result<PreparedMelt<'_>, Error> {
+        let token = Token::from_str(encoded_token)?;
+
```
thesimplekid, line 597: Can we add tests for this.

thesimplekid: added in [ba5509c](https://github.com/cashubtc/cdk/pull/1890/commits/ba5509c8a7a9f03a683add021fc59bce693a8660)

Issue #1895, opened 2026-04-16
Enforce final expiry of keysetv2 keyset


thesimplekid:
Currently the expiry of keysetv2 is not enforced by the mint and and ecash for a keyset that has expired is still spendable. However, we should take a more strict approach here and ecash should become unspendable after the expiry. This gives the expiry a stronger meaning. 

Having a strictly enforced expiry also allows a mint operator the stop running a mint in an honest way by setting all keysets to inactive and running the mint until all keysets are expired. It also allows more unique use cases like ecash with a short expiry to encourage spending for marketing or maybe at some in person event when the ecash is only good for the weekend. 

CDK currently also enforces that there is 1 active keyset. However, for the spinning down a mint use case outlined above we should remove this.

thesimplekid:
@gudnuf were you planning to work on this or just bringing up the issues you noticed?

gudnuf:
> @gudnuf were you planning to work on this or just bringing up the issues you noticed? 

I'm planing to work on it. I was going to make two PRs. One that allows starting a mint without active keysets and the other to enforce expiry

thesimplekid:
> > [@gudnuf](https://github.com/gudnuf) were you planning to work on this or just bringing up the issues you noticed?
> 
> I'm planing to work on it. I was going to make two PRs. One that allows starting a mint without active keysets and the other to enforce expiry

Sounds perfect thanks.

thesimplekid:
> One that allows starting a mint without active keysets

Since we realized in the spec we are required to have an active keyset, I think the way to achieve the goal of shutting down a mint is to disable minting and keep the keyset active until it expires. This has the same effect of running a mint with no active keysets.

So I think we should revise the goal here to just start enforcing the expiry. We also need to think about how mint start up should work, since we should not create a new active keyset with a longer expiry if it was the intention of the operator to shutdown.

gudnuf:
> So I think we should revise the goal here to just start enforcing the expiry. 

So, for example, if there's one keyset that is expired, then the keys response would still say `active: true`, but the ecash would be unusable. Right?

There's also a race condition to consider where a mint quote can be issued before expiry, then if the keyset expires before the quote is paid the invoice would be paid but no ecash could be issued.  Maybe this is just a wallet problem  because it still sucks if you mint ecash and then it expires soon after.

> We also need to think about how mint start up should work, since we should not create a new active keyset with a longer expiry if it was the intention of the operator to shutdown.

[...]

PR #1899, opened 2026-04-16
Feat/relax active keyset requirement

gudnuf:
### Description

This is the first phase of #1895. In this PR I make it so that a mint operator can deactivate a keyset by ID and allow the mint to start with no active keysets. The mint will still auto-rotate when no keysets for a unit exist or when the config changes compared to an active keyset. Rotation will not occur only if the keyset was explicitly deactivated.

-----

### Notes to the reviewers

I did this separate from enforcing expiry so that a mint operator can simply deactivate a keyset without an expiry and then leave the mint up until they feel like rugging. I could see an argument for only allowing no active keysets if they were deactivated due to expiry, but imo a mint should be able to run with no active keysets regardless of expiry.

If this is approved, then I will open a followup that adds expiry enforcement by disabling expired keysets and then extending input/output validation to disallow any operations for expired keysets 

-----

#### CHANGED

Keyset rotation no longer happens automatically if keysets are deliberately deactivated

#### ADDED

Ability to deactivate keysets

#### REMOVED

Startup validation that checks for active keysets

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/database/mint/mod.rs
+++ b/crates/cdk-common/src/database/mint/mod.rs
@@ -129,6 +129,9 @@ pub trait KeysDatabaseTransaction<'a, Error>: DbTransactionFinalizer<Err = Error
     /// Add Active Keyset
     async fn set_active_keyset(&mut self, unit: CurrencyUnit, id: Id) -> Result<(), Error>;
 
+    /// Deactivate a specific keyset by ID without activating a replacement
+    async fn deactivate_keyset(&mut self, id: Id) -> Result<(), Error>;
+
     /// Add [`MintKeySetInfo`]
     async fn add_keyset_info(&mut self, keyset: MintKeySetInfo) -> Result<(), Error>;
 }
--- a/crates/cdk-signatory/src/common.rs
+++ b/crates/cdk-signatory/src/common.rs
@@ -30,6 +30,16 @@ pub async fn init_keysets(
     for (unit, keysets) in keysets_by_unit {
         // We only care about units that are supported
         if let Some((input_fee_ppk, amounts)) = supported_units.get(&unit) {
+            // If no keyset is currently active for this unit, the operator
+            // deliberately deactivated it. Respect that and skip reactivation.
+            if !keysets.iter().any(|k| k.active) {
+                tracing::info!(
+                    "No active keyset for unit {}, respecting deliberate deactivation",
+                    unit
+                );
+                continue;
+            }
+
             let mut keysets = keysets;
             keysets.sort_by_key(|b| std::cmp::Reverse(b.derivation_path_index));
 
--- a/crates/cdk/src/mint/builder.rs
+++ b/crates/cdk/src/mint/builder.rs
@@ -570,9 +570,23 @@ impl MintBuilder {
                     }
                 }
             } else {
-                // No active keyset for this unit
-                tracing::info!("Rotating keyset for unit {} (no active keyset found)", unit);
-                rotate = true;
+                // No active keyset for this unit. Check if any keyset (active or not) has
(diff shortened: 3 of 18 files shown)


crates/cdk-signatory/src/common.rs
```diff
@@ -30,6 +30,16 @@ pub async fn init_keysets(
     for (unit, keysets) in keysets_by_unit {
         // We only care about units that are supported
         if let Some((input_fee_ppk, amounts)) = supported_units.get(&unit) {
+            // If no keyset is currently active for this unit, the operator
+            // deliberately deactivated it. Respect that and skip reactivation.
+            if !keysets.iter().any(|k| k.active) {
```
gudnuf, line 35: this I'm not sure about. Claude claims that `init_keysets` could be removed because it is now redundant with the mint builder.

crates/cdk/src/mint/builder.rs
```diff
@@ -570,9 +570,23 @@ impl MintBuilder {
...
+                        unit
+                    );
+                } else {
+                    tracing::info!(
+                        "No keyset history for unit {}, bootstrapping initial keyset",
+                        unit
+                    );
+                    rotate = true;
+                }
             }
```
thesimplekid, line 590: I think this does not account for keysetv2 and those will reactivate?

crates/cdk-signatory/src/common.rs
thesimplekid: I think we have some preexisting split logic that should be cleaned up but for this I think as it is, is okay.

crates/cdk-common/src/database/mint/mod.rs
```diff
@@ -129,6 +129,9 @@ pub trait KeysDatabaseTransaction<'a, Error>: DbTransactionFinalizer<Err = Error
     /// Add Active Keyset
     async fn set_active_keyset(&mut self, unit: CurrencyUnit, id: Id) -> Result<(), Error>;
 
+    /// Deactivate a specific keyset by ID without activating a replacement
+    async fn deactivate_keyset(&mut self, id: Id) -> Result<(), Error>;
+
```
thesimplekid, line 134: This spec will need to be updated to support this. But its not merged yet so thing its okay CDK goes first.

gudnuf: Ahh I missed that

NUT-02 says "Mints can have multiple keysets at the same time but MUST have at least one active keyset".

So this would be a breaking change

crates/cdk/src/mint/builder.rs
gudnuf: why do you say that? This looks correct to me:
1. load all keysets
2. for each supported unit: if there are inactive keysets then don't reactivate them

One confusing thing I noticed about this function is that that `let active_keysets  = signatory.keysets().await?;` is all keysets not just active

gudnuf: also btw, I think v2 is covered by the integration tests already if its true the builder uses v2 keysets by default

crates/cdk-common/src/database/mint/mod.rs
thesimplekid: Sorry looks like I forgot to link the Mint remote signer spec that I meant https://github.com/cashubtc/nuts/pull/250. 

But yes thats a good catch on NUT-02 as well. I guess the current way to spin down a mint is to disable minting but keep the keyset active.

lescuer97: I'll update the spec if needed

gudnuf:
closing because nut02 requires one active keyset. I will open a different one that just enforces keyset expiry

PR #1915, opened 2026-04-21
fix: remove state from MintQuoteCustomResponse and update default extra

asmogo:
… quote

- Add #[serde(default)] to MintQuoteCustomResponse::state so mints that omit the field (returning amount_paid/amount_issued instead) deserialize correctly, defaulting to QuoteState::Unpaid instead of failing
- Replace extra.unwrap_or_default() (empty string) with extra.as_deref().unwrap_or("{}") in wallet/issue and cdk-payment-processor so serde_json::from_str never receives an empty string, which caused an EOF parse error when extra was None

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut04.rs
+++ b/crates/cashu/src/nuts/nut04.rs
@@ -377,6 +377,7 @@ pub struct MintQuoteCustomResponse<Q> {
     /// Currency unit
     pub unit: Option<CurrencyUnit>,
     /// Quote State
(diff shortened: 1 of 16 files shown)


thesimplekid:
@asmogo I just squashed and rebased this. Changes LGTM will just let ci finish.

thesimplekid (approved):

PR #1917, opened 2026-04-22
fix(wallet): auto-release reserved proofs when melt confirm fails

### LINKED ISSUE #1891: Feature Request: Auto-cancel reserved proofs when melt confirm() fails
When `PreparedMelt::confirm()` fails, reserved proofs remain stuck in `PendingSpent` with no automatic compensation. The wallet's only recovery `reclaim_pending_proofs()` is unsafe because it also revokes unclaimed sent tokens (`confirm_send` stores sent proofs as `PendingSpent`).

The compensation infrastructure already exists:
- `handle_failure()` (mod.rs:1041-1059) — unreserves proofs, but only for specific quote states
- `cancel_prepared_melt()` (mod.rs:712-747) — full cancellation mechanism                                                                             
- Crash recovery (resume.rs:227-255) — uses saga compensation correctly                                                                               

The gap: none of these run on the normal error path of `confirm()`.                                                                                   

### Why melt is more vulnerable than send                                                                                                             

Send naturally forces a split pattern (`prepare_send` → `confirm_send` / `cancel_send`) because developers need the `PreparedSend` struct. Melt invites chaining:
                                                                                                                                                      
```rust         
wallet.prepare_melt(&id, HashMap::new()).await?.confirm().await?;
// If confirm fails, PreparedMelt is consumed — no cancel possible
                                                                                                                                                      
Melt also involves external Lightning payments that fail for reasons outside wallet control.
                                                                                                                                                      
Proposed fix                                                                                                                                          
Wire existing compensation into confirm's error path:

match self.saga.wallet.confirm_prepared_melt_with_options(...).await {
    Ok(finalized) => Ok(finalized),                                                                                                                   
    Err(e) => {
        let _ = self.saga.wallet.cancel_prepared_melt(                                                                                                
            self.operation_id, self.proofs, self.proofs_to_swap
        ).await;                                                                                                                                      
        Err(e)
    }                                                                                                                                                 
}

Forte11Cuba:
### Description
                                                                                                                                                      
When `PreparedMelt::confirm()` failed, reserved proofs were left stuck in the `Pending` state with no automatic compensation. The only prior recovery path `reclaim_pending_proofs()` is unsafe because it also revokes unclaimed                                                                            
sent tokens. Consumers were forced to call `check_all_pending_proofs()` manually or reach into raw DB state.                                                                                                                  

This PR wires saga compensation into the error path of `PreparedMelt::confirm`, `confirm_prefer_async_with_options`, and `PendingMelt::wait`, so the chained `wallet.prepare_melt(...).await?.confirm().await?` pattern self-heals on failure. Compensation delegates to `resume_melt_saga`, which consults the mint before acting this prevents the double-spend edge case where a pre-melt swap has already consumed proofs (the naive "unreserve everything" approach would                                                                          
have reverted `Spent` - `Unspent`).

Additionally, `cancel_prepared_melt` is hardened: it now filters by the proof's current state and only reverts `Reserved` / `Pending` - `Unspent`, mirroring the `RevertProofReservation` compensation pattern. Proofs a successful swap step already moved to `Spent` are preserved.

Closes #1891

-----

### Notes to the reviewers

-----

#### CHANGED

- `Wallet::cancel_prepared_melt` now filters by current proof state (only reverts `Reserved`/`Pending` - `Unspent`). Proofs already `Spent` (e.g. by a prior swap) are preserved, preventing a double-spend vector.
- `PreparedMelt`'s `#[must_use]` message updated to reflect auto-release on confirm failure. 

#### ADDED
        
- Private helper `Wallet::compensate_failed_melt_confirm` that delegates to `resume_melt_saga` for safe error-path compensation, including mint status check before reverting proofs.                                                                                                               
- Auto-compensation on failure for `PreparedMelt::confirm_with_options`, `PreparedMelt::confirm_prefer_async_with_options`, and `PendingMelt::wait` — reserved proofs no longer stay stuck in `Pending` when these methods return `Err`.

#### FIXED

- Reserved proofs no longer leak into `Pending` state when `PreparedMelt::confirm()` fails via the chained `prepare_melt().await?.confirm().await?`      
----

### Checklist

### DIFF
--- a/crates/cdk-integration-tests/tests/fake_wallet.rs
+++ b/crates/cdk-integration-tests/tests/fake_wallet.rs
@@ -1819,13 +1819,11 @@ async fn test_fake_mint_duplicate_proofs_melt() {
     }
 }
 
-/// Regression test for cashubtc/cdk#1891.
-///
 /// Exercises the chained `prepare_melt(...).await?.confirm().await?` pattern
-/// that forced consumers to reach for `check_all_pending_proofs()` when the
-/// melt failed. After the fix, confirm() must auto-release reserved proofs on
-/// any failure path, so balance should be fully recovered without any
-/// explicit recovery call.
+/// that used to force consumers to reach for `check_all_pending_proofs()` when
+/// the melt failed. confirm() must auto-release reserved proofs on any failure
+/// path, so balance should be fully recovered without any explicit recovery
+/// call.
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 async fn test_chained_confirm_auto_releases_proofs_on_failure() {
     let wallet = Wallet::new(
--- a/crates/cdk-integration-tests/tests/fake_wallet.rs
+++ b/crates/cdk-integration-tests/tests/fake_wallet.rs
@@ -1819,6 +1819,79 @@ async fn test_fake_mint_duplicate_proofs_melt() {
     }
 }
 
+/// Regression test for cashubtc/cdk#1891.
+///
+/// Exercises the chained `prepare_melt(...).await?.confirm().await?` pattern
+/// that forced consumers to reach for `check_all_pending_proofs()` when the
+/// melt failed. After the fix, confirm() must auto-release reserved proofs on
+/// any failure path, so balance should be fully recovered without any
+/// explicit recovery call.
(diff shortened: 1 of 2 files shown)


thesimplekid:
close for https://github.com/cashubtc/cdk/pull/1929

PR #1918, opened 2026-04-22
feat: implement prepare_melt_token in wallet and ffi

GEET3001:
Implemented the prepare_melt_token method in the Wallet and cdk-ffi layers. This feature allows the SDK to process raw Cashu token strings directly during the melt flow.

By offloading token decoding to the wallet core, the SDK can use its existing keyset state to correctly interpret KeysetsV2 proof amounts. This simplifies the implementation for mobile and external clients (specifically requested for the Numo flow) as they no longer need to perform manual token parsing or coin selection for this use case.

Addresses #1888.

Notes to the reviewers
The implementation uses Token::from_str to support both V3 and V4 tokens.
It performs a check to ensure the token's mint matches the wallet's mint before proceeding.
load_mint_keysets() is called to ensure the wallet has the metadata required to resolve proof amounts from the token.
The FFI layer was updated to ensure this new method is available for mobile development (Swift/Kotlin).
Suggested CHANGELOG Updates
ADDED
Wallet::prepare_melt_token to process raw token strings for melting.
FFI exposure for prepare_melt_token in cdk-ffi.


thesimplekid:
This is already started in https://github.com/cashubtc/cdk/pull/1890

PR #1919, opened 2026-04-22
Feat/prepare melt token v2

### LINKED ISSUE #1888: Add `melt_token` fn to wallet

GEET3001:
### Description

Implements prepare_melt_token on the Wallet and exposes it in the FFI layer, as requested in #1888.

Currently, cdk provides prepare_melt (internal coin selection) and prepare_melt_proofs (caller-supplied proofs). This PR adds prepare_melt_token which accepts a raw encoded Cashu token string directly. The wallet decodes it internally, which is necessary because KeysetsV2 proof amounts require keyset state that only the wallet core has access to. This is specifically helpful for the Numo flow.

Closes #1888.

-----

### Notes to the reviewers

Token::from_str handles both V3 (cashuA...) and V4 (cashuB...) token formats automatically.
A unit check (Error::UnsupportedUnit) is performed before the mint URL check — this prevents a misleading IncorrectMint error when the real issue is a currency mismatch.
load_mint_keysets() is called to ensure KeysetsV2 proof amounts are resolved correctly before being passed to the melt saga.
FFI bindings (cdk-ffi) have been updated so mobile clients (Swift/Kotlin) can access this method.

-----

#### ADDED
- `Wallet::prepare_melt_token` — accepts a raw encoded token string and prepares a melt 
  without requiring the caller to decode the token or perform coin selection.
- FFI exposure of `prepare_melt_token` in `cdk-ffi`.
- Unit tests for `prepare_melt_token` token parsing and validation logic.
#### FIXED

----

### Checklist

*[ ]  I followed the [code style guidelines](https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md)
*[ ]I ran just quick-check before committing
*[ ]  If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (crates/cdk-ffi)


thesimplekid:
This is already started in https://github.com/cashubtc/cdk/pull/1890 and you already opened a PR on this that I closed saying this.

Issue #1914, opened 2026-04-20
cdk-restore rate limit issue


Forte11Cuba:
## Summary

`Wallet::restore()` in `crates/cdk/src/wallet/mod.rs` fails opaquely against any nutshell mint with rate-limiting enabled (the upstream default since commit `44f0abd`). Three compounding problems:

1. **HTTP status code is dropped** 429s are laundered into `ErrorCode::Unknown(999)`, so callers can't detect rate-limit.
2. **Restore is not tunable** batch size (100), max-gap (3) and pacing are hardcoded.
3. **No low-level primitive** `restore()` is all-or-nothing; callers can't chunk/resume like `cashu-ts` does.

Observed against `mint.cubabitcoin.org` (stock nutshell). A restore of a 3-keyset × 2-unit mint fires ~36 requests against `/v1/restore` + `/v1/checkstate` in a tight loop, reliably tripping nutshell's default `mint_global_rate_limit_per_minute=60`.

## Problem 1 HTTP status is never checked

`http_post` / `http_get` in `crates/cdk/src/wallet/mint_connector/transport.rs:204-220` skip straight from `.send()` to `.text()` to `serde_json::from_str`. On a 429 with body `{"detail":"Rate limit exceeded."}` the body has no `code` field, so `ErrorResponse::from_value` falls back to a fabricated `ErrorCode::Unknown(999)` at `crates/cdk-common/src/error.rs:724-731`. The final error is `Unknown error response: code: 999, detail: {"detail":"Rate limit exceeded."}` — the 429 is gone.

The Tor transport (`tor_transport.rs:210`) does check status and returns `Error::HttpError(Some(status), text)`, so the two transports behave differently. `Error::is_definitive_failure` already handles 429 (`error.rs:500-503`), but it's unreachable from the reqwest path.

**Fix:** check `response.status()` before parsing; return `Error::HttpError(Some(status), body)` on non-2xx.

## Problem 2 `restore()` is not tunable

`crates/cdk/src/wallet/mod.rs:540-546` hardcodes `batch_size=100` and `max_gap=3`, with no delay between calls. Each batch also fires `/v1/restore` followed immediately by `/v1/checkstate` (`mod.rs:627`). Reference clients are more conservative: `cashu-ts` uses `BATCH_SIZE=200, MAX_GAP=2`; nutshell's own CLI defaults to `batch=25, empty_batch=2`.

[...]

PR #1920, opened 2026-04-22
fix(wallet): surface HTTP status code in Async transport

Forte11Cuba:
### Description

The Async transport in the wallet's mint connector was reading the response body without first checking the HTTP status code. Non-2xx responses whose body does not match the Cashu ErrorResponse schema for example nutshell's {"detail":"Rate limit exceeded."} on HTTP 429 fell into the fallback path in ErrorResponse::from_value (crates/cdk-common/src/error.rs:724-731) and were laundered into ErrorCode::Unknown(999). The original status was lost.                                                                                                                                                 

This meant that:
- Error::is_definitive_failure (which already handles 429 correctly at crates/cdk-common/src/error.rs:500-503) was unreachable from this path, so the error was classified as ambiguous and eligible for retry — the opposite of what you want on a rate-limited mint.
- retriable_http_request in http_client.rs:162-175 could not distinguish rate-limit errors from other failures. 

Only the Tor transport (tor_transport.rs:208-211) handled status correctly.

This PR aligns the Async transport with the Tor transport: check response.status() before consuming the body, and return Error::HttpError(Some(status), body) on non-2xx. The existing 2xx parsing path is preserved.

-----

### Notes to the reviewers

-----

#### CHANGED

- Async transport in cdk::wallet::mint_connector::transport now returns Error::HttpError(Some(status), body) on non-2xx HTTP responses, matching the  Tor transport's behavior.    

#### FIXED

- HTTP 429 responses from rate-limited mints (e.g. nutshell with mint_global_rate_limit_per_minute) no longer surface as ErrorCode::Unknown(999) and are now correctly classified as definitive failures. Partial fix for #1914.   
----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/mint_connector/transport.rs
+++ b/crates/cdk/src/wallet/mint_connector/transport.rs
@@ -170,14 +170,26 @@ impl Transport for Async {
         let response = request
             .send()
             .await
-            .map_err(|e| Error::HttpError(None, e.to_string()))?
+            .map_err(|e| Error::HttpError(None, e.to_string()))?;
+
+        let status = response.status();
+        let body = response
             .text()
             .await
             .map_err(|e| Error::HttpError(None, e.to_string()))?;
 
-        serde_json::from_str::<R>(&response).map_err(|err| {
+        if !(200..300).contains(&status) {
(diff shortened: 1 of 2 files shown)


thesimplekid:
@Forte11Cuba This would lose the cashu spec'd errors since they are 400. I added a commit to address this.

PR #1921, opened 2026-04-22
feat(wallet): preemptive client-side rate limiter

Forte11Cuba:
### Description

Adds a preemptive token-bucket rate limiter to the wallet's HTTP client, so the wallet stays under the mint's advertised (or assumed) rate limit and avoids triggering server-side 429s in the first place.  

Modeled after Coco's RequestRateLimiter:

- Token bucket with continuous refill, capped at capacity.  
- Defaults match Coco's WalletService: capacity=20, refill=20/min. Safely under nutshell's default global 60/min.
- FIFO ordering via tokio::sync::Mutex lock ordering.
- Waits (awaits) when tokens are exhausted  never rejects.
- No reactive handling of 429s  surfacing those correctly is handled by the companion transport fix.
- Asserts on zero capacity or zero refill in RateLimiter::new — both are configuration bugs (zero capacity would deadlock, zero refill would yield a non-finite wait).                                                                                                                                     

wallet.restore() against a nutshell mint with mint_global_rate_limit_per_minute=60 (the upstream default since 44f0abd) reliably trips the limit on  mints with a few keysets × units, because the scan fires dozens of /v1/restore + /v1/checkstate requests in a tight loop. 

- cdk::wallet::RateLimiter (re-exported from cdk::wallet::mint_connector).
- RateLimiter::new(capacity, refill_per_minute), RateLimiter::default(), pub async fn acquire(&self).
- HttpClient::with_rate_limiter(...); HttpClient::new / with_transport / with_proxy install RateLimiter::default() automatically.                     
- AuthHttpClient::with_rate_limiter(...) and AuthWallet::with_rate_limiter(...) so NUT-22 blind-auth traffic can share the same limiter.
- WalletBuilder::rate_limiter(Option<Arc<RateLimiter>>) for app-level sharing or opt-out.   

-----

### Notes to the reviewers

-----

#### CHANGED

- HttpClient, AuthHttpClient, and wallets built via WalletBuilder now install a default client-side RateLimiter (20 req/min, burst 20). Pass None via  HttpClient::with_rate_limiter / WalletBuilder::rate_limiter to opt out.
- WalletBuilder::set_auth_cat now defers AuthWallet construction to build(). Public signature unchanged. 

#### ADDED

- cdk::wallet::RateLimiter — preemptive token-bucket limiter for wallet HTTP requests.
- HttpClient::with_rate_limiter, AuthHttpClient::with_rate_limiter, AuthWallet::with_rate_limiter.
- WalletBuilder::rate_limiter to share a limiter across wallets targeting the same mint or opt out. 

#### FIXED

- Preemptive pacing avoids triggering mint-side 429s during bursty workloads like Wallet::restore(). Partial fix for #1914.

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/auth/auth_wallet.rs
+++ b/crates/cdk/src/wallet/auth/auth_wallet.rs
@@ -17,7 +17,7 @@ use crate::nuts::{
     nut12, AuthRequired, AuthToken, BlindAuthToken, CurrencyUnit, KeySetInfo, PreMintSecrets,
     Proofs, ProtectedEndpoint, State,
 };
-use crate::wallet::mint_connector::AuthHttpClient;
+use crate::wallet::mint_connector::{AuthHttpClient, RateLimiter};
 use crate::wallet::mint_metadata_cache::MintMetadataCache;
 use crate::{Amount, Error, OidcClient};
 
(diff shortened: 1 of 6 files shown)


thesimplekid:
Did you review the work started on this here https://github.com/cashubtc/cdk/pull/1769? That has gone  a bit stall so probably okay to take over but don't want to duplicate review work.

PR #1923, opened 2026-04-22
feat(wallet): add RestoreOptions to tune Wallet::restore batching

Forte11Cuba:
### Description

Makes Wallet::restore()'s batch size and gap limit configurable via a new RestoreOptions struct, as requested by @thesimplekid in the thread on #1914. Defaults match the NUT-13 spec recommendation (100 / 3), so existing callers see no behavior change.

Partial fix for #1914

-----

### Notes to the reviewers

-----

#### CHANGED

- nostr_backup::RestoreOptions (feature nostr) renamed to BackupRestoreOptions to free the RestoreOptions name for the main wallet API.
- Wallet::restore() is now a thin wrapper over the new Wallet::restore_with_opts(RestoreOptions::default()).

#### ADDED

- cdk::wallet::RestoreOptions (re-exported from cdk_common::wallet) with tunable batch_size and max_gap.
- Wallet::restore_with_opts(opts) for callers that want to override the NUT-13 defaults.

#### FIXED

- Callers can now shrink batch_size to be gentler on rate-limited mints while the wallet is being restored.

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/wallet/mod.rs
+++ b/crates/cdk-common/src/wallet/mod.rs
@@ -328,6 +328,30 @@ pub struct Restored {
     pub pending: Amount,
 }
 
+/// Options for [`crate::wallet::Wallet::restore_with_opts`].
+///
+/// Defaults match the NUT-13 spec recommendation
+/// (<https://github.com/cashubtc/nuts/blob/main/13.md#generate-blindedmessages>):
+/// a batch of 100 blinded messages and three consecutive empty batches to
+/// signal end-of-history. Callers that need more conservative pacing or
+/// different gap tolerance can override either field.
+#[derive(Debug, Clone)]
+pub struct NUT13Options {
+    /// Number of blinded messages to request per batch.
+    pub batch_size: u32,
+    /// Number of consecutive empty batches that terminate the scan.
+    pub max_gap: u32,
+}
+
+impl Default for NUT13Options {
+    fn default() -> Self {
+        Self {
+            batch_size: 100,
+            max_gap: 3,
+        }
+    }
+}
+
 /// Send options
 #[derive(Clone, Default)]
 pub struct SendOptions {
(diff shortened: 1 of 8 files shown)


crates/cdk-common/src/wallet/mod.rs
```diff
@@ -305,6 +305,36 @@ pub struct Restored {
...
+/// (<https://github.com/cashubtc/nuts/blob/main/13.md#generate-blindedmessages>):
+/// a batch of 100 blinded messages and three consecutive empty batches to
+/// signal end-of-history. Callers that need more conservative pacing or
+/// different gap tolerance can override either field.
+///
+/// Degenerate configurations (`batch_size == 0` and/or `max_gap == 0`) do
+/// not panic but produce a scan that terminates without recovering proofs:
+/// `max_gap == 0` skips the scan entirely, and `batch_size == 0` sends
+/// empty requests until `max_gap` is reached. Both are configuration
+/// errors; callers should use `Default` or positive values.
```
thesimplekid:
```suggestion
```

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -305,6 +305,36 @@ pub struct Restored {
...
+#[derive(Debug, Clone)]
+pub struct RestoreOptions {
```
thesimplekid: Maybe just call this NUT13Options so we don't have to rename the nostr at least in this PR?

crates/cdk/src/wallet/mod.rs
```diff
@@ -513,9 +513,29 @@ impl Wallet {
...
+    /// Restore proofs from the mint using the given [`RestoreOptions`].
+    ///
+    /// Scans each keyset in batches of `opts.batch_size` blinded messages
+    /// and stops after `opts.max_gap` consecutive empty batches. Lowering
+    /// `batch_size` trades scan latency for a gentler request pattern; the
+    /// mint's rate limit and the wallet's configured
+    /// [`RateLimiter`](crate::wallet::RateLimiter) are orthogonal pacing
+    /// controls that still apply.
+    #[instrument(skip(self))]
+    pub async fn restore_with_opts(&self, opts: RestoreOptions) -> Result<Restored, Error> {
```
thesimplekid: We need to add new fns to the Wallet trait and then expose them in ffi as well.

thesimplekid (changes requested):
Think this is a useful addition just a few comments.

thesimplekid:
@codex review

thesimplekid (approved):

PR #1931, opened 2026-04-23
feat(wallet): add offline ecash receiving with DLEQ verification

### LINKED ISSUE #1927: Wallet handle receive when offline
Currently, receiving in the CDK wallet requires the wallet to be online to perform a swap. However, Cashu enables offline receiving pending DLEQ verification and locking conditions. We should add this capability to CDK.

One way to implement this is by adding a new function to the wallet that accepts a token and specific conditions. These conditions would include the minimum locktime, a list of mints, and whether the token must be locked. The CDK wallet would then verify the token's DLEQ proof and ensure these conditions are met. Upon verification, it would add the proofs to the database in a new state, `PendingReceive`, and add them to the wallet's transaction list. Wallets should display these transactions with a warning indicating that they must be swapped for final settlement.

Since the proofs are not truly received until they are swapped, we would also need a function that checks for all pending receives and executes the swaps.

GEET3001:
## Description
Closes #1927

Currently, receiving in the CDK wallet requires the wallet to be online to perform a swap. However, Cashu enables offline receiving pending DLEQ verification and locking conditions. This PR adds that capability end-to-end across the core library, mobile FFI bindings, and CLI.

`receive_offline()` accepts a token and an `OfflineReceiveOptions` struct containing:
- `min_locktime` — the minimum locktime required on the token
- `require_locked` — whether the token must be P2PK locked to protect against double-spend by the sender while offline

The wallet verifies the token's DLEQ proof and checks these conditions entirely offline. Upon successful verification, the proofs are added to the database in the new `PendingReceive` state. Wallets should display these with a warning that they must be swapped for final settlement.

`finalize_pending_receives()` checks for all proofs in `PendingReceive` state and executes the swaps with the mint. This should be called whenever the wallet comes back online.

### Core Protocol
- Adds `PendingReceive` state to the `State` enum in `cashu/nut07`
- Adds `OfflineReceiveOptions` struct (`min_locktime`, `require_locked`)
- Adds `receive_offline()` method to the `Wallet` trait and core implementation:
  - Verifies DLEQ proofs on all tokens without going online
  - Validates locking conditions (locktime, P2PK)
  - Stores verified proofs in `PendingReceive` state in the database
- Adds `finalize_pending_receives()` to swap pending proofs with the mint
- Adds `total_pending_receive_balance()` to query the total amount sitting in the pending offline queue
- Fixes offline receive finalization losing the sender's memo and splitting one token into per-proof transactions. Proofs are now grouped by a UUID stamped at `receive_offline` time; `finalize_pending_receives` processes each group as a single receive operation and recovers the memo from the `OfflinePendingReceive` saga entry

### Database
- Adds SQLite and Postgres migrations to add `PENDING_RECEIVE` to the proof state `CHECK` constraint

### FFI Layer (cdk-ffi)
- Exposes `receive_offline()`, `finalize_pending_receives()`, and `total_pending_receive_balance()` over the UniFFI boundary so Swift/Kotlin consumers can use the full offline receive lifecycle
- Exposes `ProofState::PendingReceive` as a UniFFI enum variant
- Adds JSON encode/decode helpers for `OfflineReceiveOptions` for mobile callers
- Handles `PendingReceive` in all exhaustive match sites (test utils, FFI)

### CLI (cdk-cli)
- Adds `--offline` flag to the receive subcommand — routes the token to `receive_offline` instead of the standard online swap
- Adds new `finalize-receives` subcommand — sweeps all `PendingReceive` proofs and swaps them with the mint when the user is back online

### Bug Fix
- Fixes a non-exhaustive match compiler error in `check_spendable.rs` that was introduced when `PendingReceive` was added to the `State` enum

## Known limitations

- `receive_offline` depends on the keyset for the incoming token having been synced during a prior online session. This relies on #1957's `load_from_db` fallback in `MintMetadataCache::load()`.

  `is_populated` is a coarse boolean — it returns `true` if *any* keyset is cached, not specifically the one the incoming token uses. If the wallet has never seen that keyset (e.g. the mint rotated keys between the sender minting and the receiver syncing), `load_keyset_keys` returns `None`. Scoped fix in this PR: improved error message at the `load_keyset_keys` call site directing the user to go online once to sync. Follow-up: `is_populated` should be keyset-id-aware rather than a global boolean.

- Atomic offline receive: `receive_offline` does two DB writes with no transaction between them — a crash between them loses the memo but not funds. Needs a DB transaction API to fix properly.

- No dismissal API for stuck pending receives — UI wallets have no way to let users clear a `PendingReceive` proof they know is gone.

## Notes to reviewers
- `PendingReceive` is kept intentionally separate from `Pending` (used for in-flight send/melt) to avoid ambiguity in state machine transitions and to make it easy to display pending offline receives distinctly in wallet UIs.
- Database migrations update the `CHECK (state IN (...))` constraint on the proof table to include `PENDING_RECEIVE`, preventing constraint violations at runtime.
- The FFI layer is fully updated: `ProofState::PendingReceive` is exposed as a UniFFI enum variant, and all three new methods are callable from mobile/desktop consumers.
- The CLI `--offline` flag requires users to explicitly opt-in to offline receives, as bearer tokens accepted offline carry a double-spend risk unless `require_locked` is set.

### Added
- `State::PendingReceive` — new proof state for offline-received ecash awaiting final swap
- `Wallet::receive_offline(token, options)` — verify a token's DLEQ proof offline and store proofs in `PendingReceive` state
- `Wallet::finalize_pending_receives()` — sweep all `PendingReceive` proofs and swap them with the mint
- `Wallet::total_pending_receive_balance()` — query the total pending offline balance
- `OfflineReceiveOptions` struct with `min_locktime` and `require_locked` fields
- SQLite and Postgres DB migrations to add `PENDING_RECEIVE` to the proof state constraint
- FFI bindings for all of the above via `cdk-ffi`
- `cdk-cli` `receive --offline` flag
- `cdk-cli` `finalize-receives` subcommand

## Checklist
- [x] followed the code style guidelines
- [x] ran `just quick-check` before committing
- [x] wallet API changes are reflected in FFI bindings

### DIFF
--- a/crates/cdk-cli/src/sub_commands/receive.rs
+++ b/crates/cdk-cli/src/sub_commands/receive.rs
@@ -37,6 +37,9 @@ pub struct ReceiveSubCommand {
     /// Allow receiving from untrusted mints (mints not already in the wallet)
     #[arg(long, default_value = "false")]
     allow_untrusted: bool,
+    /// Only accept the token offline and place it into PendingReceive state
+    #[arg(long, default_value = "false")]
+    offline: bool,
 }
 
 pub async fn receive(
@@ -72,6 +75,7 @@ pub async fn receive(
                 &signing_keys,
                 &sub_command_args.preimage,
                 sub_command_args.allow_untrusted,
+                sub_command_args.offline,
                 unit,
             )
             .await?
@@ -114,6 +118,7 @@ pub async fn receive(
                     &signing_keys,
                     &sub_command_args.preimage,
                     sub_command_args.allow_untrusted,
+                    sub_command_args.offline,
                     unit,
                 )
                 .await
@@ -142,6 +147,7 @@ async fn receive_token(
     signing_keys: &[SecretKey],
     preimage: &[String],
     allow_untrusted: bool,
+    offline: bool,
     unit: &CurrencyUnit,
 ) -> Result<Amount> {
     let token: Token = Token::from_str(token_str)?;
@@ -162,15 +168,22 @@ async fn receive_token(
     // Get or create wallet for the token's mint
     let wallet = get_or_create_wallet(wallet_repository, &mint_url, unit).await?;
 
-    // Create receive options
-    let receive_options = ReceiveOptions {
-        p2pk_signing_keys: signing_keys.to_vec(),
-        preimages: preimage.to_vec(),
-        ..Default::default()
-    };
+    if offline {
+        let options = cdk_common::wallet::OfflineReceiveOptions {
+            minimum_locktime: None,
+            require_locked: false,
+        };
+        Ok(wallet.receive_offline(token_str, options).await?)
+    } else {
+        // Create receive options
+        let receive_options = ReceiveOptions {
+            p2pk_signing_keys: signing_keys.to_vec(),
+            preimages: preimage.to_vec(),
+            ..Default::default()
+        };
 
-    let received = wallet.receive(token_str, receive_options).await?;
-    Ok(received)
+        Ok(wallet.receive(token_str, receive_options).await?)
+    }
 }
 
 /// Receive tokens sent to nostr pubkey via dm
--- a/crates/cdk-common/src/wallet/mod.rs
+++ b/crates/cdk-common/src/wallet/mod.rs
@@ -490,6 +490,22 @@ impl fmt::Debug for ReceiveOptions {
     }
 }
 
+/// Offline Receive options
+///
+/// DLEQ proof verification is always performed for offline receives
+/// since it is the only way to verify token authenticity without
+/// contacting the mint.
+///
+/// The wallet is already bound to a specific mint URL, so mint
+/// filtering is handled automatically.
+#[derive(Debug, Clone, Default)]
+pub struct OfflineReceiveOptions {
+    /// Optional minimum locktime required for the token
+    pub minimum_locktime: Option<u64>,
+    /// Require the token to be P2PK locked
+    pub require_locked: bool,
+}
+
 /// Send Kind
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Default, Serialize, Deserialize)]
 pub enum SendKind {
@@ -859,6 +875,12 @@ pub trait Wallet: Send + Sync {
     /// Total pending balance of the wallet
     async fn total_pending_balance(&self) -> Result<Self::Amount, Self::Error>;
 
+    /// Total pending-receive balance of the wallet
+    ///
+    /// Proofs in this state have been DLEQ-verified offline but not yet
+    /// swapped with the mint. Call `finalize_pending_receives` to settle them.
+    async fn total_pending_receive_balance(&self) -> Result<Self::Amount, Self::Error>;
+
     /// Total reserved balance of the wallet
     async fn total_reserved_balance(&self) -> Result<Self::Amount, Self::Error>;
 
@@ -970,6 +992,16 @@ pub trait Wallet: Send + Sync {
         memo: Option<String>,
         token: Option<String>,
     ) -> Result<Self::Amount, Self::Error>;
+
+    /// Receive an encoded token offline without contacting the mint
+    async fn receive_offline(
+        &self,
+        encoded_token: &str,
+        options: OfflineReceiveOptions,
+    ) -> Result<Self::Amount, Self::Error>;
+
+    /// Finalize pending offline receives by attempting to swap them
+    async fn finalize_pending_receives(&self) -> Result<Self::Amount, Self::Error>;
 
     /// Prepare a send transaction
     async fn prepare_send(
--- a/crates/cdk-ffi/src/wallet.rs
+++ b/crates/cdk-ffi/src/wallet.rs
@@ -128,6 +128,12 @@ impl Wallet {
         Ok(balance.into())
     }
 
+    /// Get total pending receive balance
+    pub async fn total_pending_receive_balance(&self) -> Result<Amount, FfiError> {
+        let balance = self.inner.total_pending_receive_balance().await?;
+        Ok(balance.into())
+    }
+
     /// Get total reserved balance
     pub async fn total_reserved_balance(&self) -> Result<Amount, FfiError> {
         let balance = self.inner.total_reserved_balance().await?;
@@ -205,6 +211,32 @@ impl Wallet {
             .receive_proofs(cdk_proofs, options.try_into()?, memo, token)
             .await?;
         Ok(amount.into())
+    }
+
+    /// Receive an encoded token offline without contacting the mint
+    ///
+    /// Verifies the token's DLEQ proofs locally (no network call) and stores
+    /// the proofs as `PendingReceive` in the local database. Call
+    /// `finalize_pending_receives` once back online to settle them.
+    pub async fn receive_offline(
+        &self,
+        encoded_token: String,
+        options: OfflineReceiveOptions,
+    ) -> Result<Amount, FfiError> {
+        Ok(self
+            .inner
+            .receive_offline(&encoded_token, options.into())
+            .await?
+            .into())
+    }
+
+    /// Finalize pending offline receives by swapping with the mint
+    ///
+    /// Finds all proofs stored as `PendingReceive` (from prior calls to
+    /// `receive_offline`) and swaps them with the mint, making them fully
+    /// spendable. Requires an active network connection.
+    pub async fn finalize_pending_receives(&self) -> Result<Amount, FfiError> {
+        Ok(self.inner.finalize_pending_receives().await?.into())
     }
 
     /// Get all pending send operations
@@ -548,6 +580,11 @@ impl Wallet {
                 ProofState::Pending => self.inner.get_pending_proofs().await?,
                 ProofState::Reserved => self.inner.get_reserved_proofs().await?,
                 ProofState::PendingSpent => self.inner.get_pending_spent_proofs().await?,
+                ProofState::PendingReceive => {
+                    self.inner
+                        .get_proofs_by_states(vec![cdk::nuts::State::PendingReceive])
+                        .await?
+                }
                 ProofState::Spent => {
                     // CDK doesn't have a method to get spent proofs directly
                     // They are removed from the database when spent
--- a/crates/cdk-sql-common/src/wallet/migrations/sqlite/20260423000000_allow_pending_receive.sql
+++ b/crates/cdk-sql-common/src/wallet/migrations/sqlite/20260423000000_allow_pending_receive.sql
@@ -1,0 +1,39 @@
+-- Create a new table with the updated CHECK constraint
+CREATE TABLE IF NOT EXISTS proof_new (
+    y BLOB PRIMARY KEY,
+    mint_url TEXT NOT NULL,
+    state TEXT CHECK ( state IN ('SPENT', 'UNSPENT', 'PENDING', 'RESERVED', 'PENDING_SPENT', 'PENDING_RECEIVE' ) ) NOT NULL,
+    spending_condition TEXT,
+    unit TEXT NOT NULL,
+    amount INTEGER NOT NULL,
+    keyset_id TEXT NOT NULL,
+    secret TEXT NOT NULL,
+    c BLOB NOT NULL,
+    witness TEXT,
+    dleq_e BLOB,
+    dleq_s BLOB,
+    dleq_r BLOB,
+    used_by_operation TEXT,
+    created_by_operation TEXT,
+    p2pk_e BLOB
+);
+
+CREATE INDEX IF NOT EXISTS secret_index ON proof_new(secret);
+CREATE INDEX IF NOT EXISTS state_index ON proof_new(state);
+CREATE INDEX IF NOT EXISTS spending_condition_index ON proof_new(spending_condition);
+CREATE INDEX IF NOT EXISTS unit_index ON proof_new(unit);
+CREATE INDEX IF NOT EXISTS amount_index ON proof_new(amount);
+CREATE INDEX IF NOT EXISTS mint_url_index ON proof_new(mint_url);
+CREATE INDEX IF NOT EXISTS proof_used_by_operation_index ON proof_new(used_by_operation);
+CREATE INDEX IF NOT EXISTS proof_created_by_operation_index ON proof_new(created_by_operation);
+
+-- Copy data from old proof table to new proof table
+INSERT INTO proof_new (y, mint_url, state, spending_condition, unit, amount, keyset_id, secret, c, witness, dleq_e, dleq_s, dleq_r, used_by_operation, created_by_operation, p2pk_e)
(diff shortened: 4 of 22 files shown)


crates/cdk-common/src/wallet/mod.rs
```diff
@@ -1040,115 +1063,3 @@ pub struct P2PKSigningKey {
     /// Created time
     pub created_time: u64,
 }
-
-#[cfg(test)]
-mod tests {
```
TheMhv, line 1206: Why you removed those tests?

crates/cdk/src/wallet/receive/mod.rs
```diff
@@ -124,4 +124,165 @@ impl Wallet {
...
+        if !opts.trusted_mints.is_empty() {
+            ensure_cdk!(opts.trusted_mints.contains(&mint_url), Error::IncorrectMint);
+        }
+
+        if let Token::TokenV3(token) = &token {
+            ensure_cdk!(!token.is_multi_mint(), Error::MultiMintTokenNotSupported);
+        }
+
+        let keysets_info = self.load_mint_keysets().await?;
+        use cdk_common::ProofsMethods;
```
TheMhv: Where it is used? I think that can be removed

crates/cdk/src/wallet/receive/mod.rs
```diff
@@ -124,4 +124,165 @@ impl Wallet {
...
+                            );
+                        } else {
+                            return Err(Error::LocktimeNotProvided);
+                        }
+                    } else {
+                        return Err(Error::LocktimeNotProvided);
+                    }
+                } else {
+                    return Err(Error::LocktimeNotProvided);
+                }
```
TheMhv: I think that those else's statement can be unified

crates/cdk/src/wallet/receive/mod.rs
```diff
@@ -124,4 +124,165 @@ impl Wallet {
...
+            )
+            .await?;
+
+        if proofs_info.is_empty() {
+            return Ok(Amount::ZERO);
+        }
+
+        let proofs: Proofs = proofs_info.into_iter().map(|p| p.proof).collect();
+
+        self.receive_proofs(proofs, ReceiveOptions::default(), None, None)
```
TheMhv: You was using a default and null values as a argument. This is correct?

crates/cdk/src/wallet/receive/mod.rs
```diff
@@ -124,4 +124,165 @@ impl Wallet {
...
+            if opts.require_dleq {
+                ensure_cdk!(proof.dleq.is_some(), Error::DleqProofNotProvided);
+            }
+
+            if proof.dleq.is_some() {
+                let keys = self.load_keyset_keys(proof.keyset_id).await?;
+                let key = keys.amount_key(proof.amount).ok_or(Error::AmountKey)?;
+                proof.verify_dleq(key)?;
+            }
+
```
TheMhv: Since the DLEQ proof is aways required for offline reception, this verification needs to aways run on this function, those if's statements can be removed.

TheMhv (changes requested):
You need to fix formatting and some errors that you can find running `just final-check` command.

Can you implement some integration test for this `receive_offline` function too.

And please, return the wallet tests

crates/cdk-common/src/wallet/mod.rs
GEET3001: My apologies, that was an accidental deletion , i made a different branch for that . I have fully restored all the wallet unit tests to match main exactly.

crates/cdk/src/wallet/receive/mod.rs
GEET3001: I've removed the unused import and replaced it with a fully qualified syntax call where needed.

crates/cdk/src/wallet/receive/mod.rs
GEET3001: I have refactored and flattened this entire block using idiomatic `?` operator chains to make it much cleaner.

crates/cdk/src/wallet/receive/mod.rs
GEET3001: I've removed the `require_dleq` toggle entirely from the API. DLEQ verification is now unconditionally mandatory for all offline receives.

crates/cdk/src/wallet/receive/mod.rs
GEET3001: We pass `ReceiveOptions::default()` because we don't need any special options here, and `None` for the memo and token string because we are just sweeping the raw proofs we already have saved, rather than processing a brand new token string

crates/cdk-common/src/wallet/mod.rs
```diff
@@ -395,6 +395,21 @@ pub struct ReceiveOptions {
...
 }
 
+/// Offline Receive options
+///
+/// DLEQ proof verification is always performed for offline receives
+/// since it is the only way to verify token authenticity without
+/// contacting the mint.
+#[derive(Debug, Clone, Default)]
+pub struct OfflineReceiveOptions {
+    /// List of trusted mint URLs (if empty, all mints are accepted)
```
TheMhv: I think that is not good to accept any mint by default

crates/cdk/src/wallet/receive/mod.rs
```diff
@@ -124,4 +124,147 @@ impl Wallet {
...
+        let unit = token.unit().unwrap_or_default();
+        ensure_cdk!(unit == self.unit, Error::UnsupportedUnit);
+
+        let mint_url = token.mint_url()?;
+        ensure_cdk!(self.mint_url == mint_url, Error::IncorrectMint);
+
+        if !opts.trusted_mints.is_empty() {
+            ensure_cdk!(opts.trusted_mints.contains(&mint_url), Error::IncorrectMint);
+        }
```
TheMhv: If a wallet is created for the specific mint, isn't a trusted mint?

crates/cdk-common/src/wallet/mod.rs
GEET3001: I've removed the `trusted_mints` field entirely from `OfflineReceiveOptions`. Since a `Wallet` is always created for a specific mint and we already enforce `self.mint_url == token.mint_url()`, the field was redundant and the "accept any mint" default was unnecessary.

crates/cdk/src/wallet/receive/mod.rs
GEET3001: The wallet is already bound to `self.mint_url` and we check `self.mint_url == token.mint_url()` right above, so the `trusted_mints` check was completely redundant. I've removed `trusted_mints` from `OfflineReceiveOptions` and dropped this block.

GEET3001:
Integration tests added (as requested by @TheMhv)

Added two integration tests in crates/cdk-integration-tests/tests/integration_tests_pure.rs:

test_receive_offline_pending_then_finalize — full lifecycle test: token verified offline via DLEQ → stored as PendingReceive → finalized online via finalize_pending_receives() → spendable balance updated
test_receive_offline_rejects_token_without_dleq — security test: tokens without DLEQ proofs are rejected with DleqProofNotProvided
Both pass locally (CDK_TEST_DB_TYPE=memory).

Note on dependencies: This PR's offline path relies on keysets being available in the local cache. The correct behavior when the wallet is truly offline (no cached keysets) depends on the metadata cache improvements being worked on in the related issue. Once those land, receive_offline will work seamlessly without any prior online session

GEET3001:
just a heads up, it looks like the recent CI checks are failing due to a runner infrastructure issue. The logs for the Quick Check and Binding tests are all throwing a No space left on device (os error 28) error, which means the runner's disk is currently full.

Let me know when the runner has been cleared and I can re-trigger the checks, or feel free to re-trigger them on your end

crates/cdk/src/wallet/receive/mod.rs
```diff
@@ -123,4 +123,187 @@ impl Wallet {
...
+
+        use crate::wallet::ProofInfo;
+
+        let mut total = Amount::ZERO;
+
+        for proof_info in proofs_info {
+            let proof_y = proof_info.y;
+            let proof_clone = proof_info.proof.clone();
+
```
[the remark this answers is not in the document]
GEET3001: fix(wallet): preserve token memo and grouping through offline receive finalization

receive_offline now stamps all proofs with a shared operation UUID
(created_by_operation) and writes an OfflinePendingReceive saga holding
the original encoded token string.

finalize_pending_receives groups PendingReceive proofs by that UUID,
looks up the saga to recover the memo, and calls receive_proofs once
per group — producing one transaction record with the correct memo
instead of one per proof with no memo.

crash recovery (recover_incomplete_sagas) skips OfflinePendingReceive
sagas so it never compensates proofs the user hasn't collected yet.

Proofs stored before this fix (no created_by_operation) fall through
to a backward-compatible one-at-a-time path.

crates/cdk-sql-common/src/wallet/migrations/sqlite/20260423000000_allow_pending_receive.sql
TheMhv, line 1: You can simply add those fields into `proof` table instead all of this

crates/cdk/src/wallet/receive/mod.rs
```diff
@@ -128,4 +131,372 @@ impl Wallet {
         let token_str = Token::try_from(binary_token)?.to_string();
         self.receive(token_str.as_str(), opts).await
     }
+
+    /// Receive an encoded token offline without contacting the mint
+    #[instrument(skip_all)]
+    pub async fn receive_offline(
```
TheMhv, line 137: You need to create a new function to receive offline payments? Can't integrate with `receive` function?

crates/cdk-cli/src/sub_commands/receive.rs
```diff
@@ -162,15 +168,24 @@ async fn receive_token(
...
+    } else {
+        // Create receive options
+        let receive_options = ReceiveOptions {
+            p2pk_signing_keys: signing_keys.to_vec(),
+            preimages: preimage.to_vec(),
+            ..Default::default()
+        };
+
+        let received = wallet.receive(token_str, receive_options).await?;
+        Ok(received)
```
crodas:
```suggestion
        Ok(wallet.receive(token_str, receive_options).await?)
```

crates/cdk-cli/src/sub_commands/receive.rs
```diff
@@ -162,15 +168,24 @@ async fn receive_token(
...
-
-    let received = wallet.receive(token_str, receive_options).await?;
-    Ok(received)
+    if offline {
+        let options = cdk_common::wallet::OfflineReceiveOptions {
+            minimum_locktime: None,
+            require_locked: false,
+        };
+        let received = wallet.receive_offline(token_str, options).await?;
+        Ok(received)
```
crodas:
```suggestion
        Ok(wallet.receive_offline(token_str, options).await?)
```

crates/cdk-ffi/src/wallet.rs
```diff
@@ -207,6 +213,33 @@ impl Wallet {
...
+    pub async fn receive_offline(
+        &self,
+        encoded_token: String,
+        options: OfflineReceiveOptions,
+    ) -> Result<Amount, FfiError> {
+        let amount = self
+            .inner
+            .receive_offline(&encoded_token, options.into())
+            .await?;
+        Ok(amount.into())
```
crodas:
```suggestion
        Ok(self
            .inner
            .receive_offline(&encoded_token, options.into())
            .await?
            .into())
```

crates/cdk-ffi/src/wallet.rs
```diff
@@ -207,6 +213,33 @@ impl Wallet {
...
+    }
+
+    /// Finalize pending offline receives by swapping with the mint
+    ///
+    /// Finds all proofs stored as `PendingReceive` (from prior calls to
+    /// `receive_offline`) and swaps them with the mint, making them fully
+    /// spendable. Requires an active network connection.
+    pub async fn finalize_pending_receives(&self) -> Result<Amount, FfiError> {
+        let amount = self.inner.finalize_pending_receives().await?;
+        Ok(amount.into())
```
crodas:
```suggestion
        Ok(self.inner.finalize_pending_receives().await?)
```

crates/cdk/src/wallet/receive/mod.rs
```diff
@@ -128,4 +131,372 @@ impl Wallet {
...
+
+            total_amount += proof.amount;
+        }
+
+        use cdk_common::wallet::{
+            OperationData, ReceiveOperationData, ReceiveSagaState, WalletSaga, WalletSagaState,
+        };
+
+        use crate::nuts::State;
+        use crate::wallet::ProofInfo;
```
crodas, line 221: This use inside a function seems weird. Can we ban them in Clippy @thesimplekid ?

crodas (changes requested):
Overall it looks good. I left some minor comments

crates/cdk-sql-common/src/wallet/migrations/sqlite/20260423000000_allow_pending_receive.sql
GEET3001: This migration doesn't add any columns — it only widens the state CHECK constraint to allow PENDING_RECEIVE. SQLite doesn't support altering a CHECK constraint in place (no ALTER TABLE ... DROP/ADD CONSTRAINT), so the table has to be rebuilt. This follows the exact pattern already used in 20250314082116_allow_pending_spent.sql, which added PENDING_SPENT the same way. The Postgres version of this migration does use ALTER TABLE ... DROP/ADD CONSTRAINT since Postgres supports it. Happy to change the approach if you'd prefer dropping the CHECK constraint entirely, but I kept it consistent with the existing migrations

crates/cdk/src/wallet/receive/mod.rs
GEET3001: They do different things: receive() swaps with the mint immediately and returns a finalized, spendable amount, while receive_offline() never contacts the mint — it DLEQ-verifies, stores the proofs as PendingReceive, and defers the swap to finalize_pending_receives, so the returned amount isn't spendable yet. Their option sets are also disjoint (OfflineReceiveOptions has minimum_locktime/require_locked; ReceiveOptions has split target, p2pk keys, preimages, metadata), and the network/DLEQ policy differs. I kept them separate so the return semantics stay unambiguous.

 If you'd prefer a single entry point, I can add a ReceiveMode (or offline flag) to ReceiveOptions and dispatch internally — the shared decode/validate stays in one place and the two behaviors stay distinct. Let me know which you'd rather have.

crates/cdk-ffi/src/wallet.rs
GEET3001: Applied, but kept the .into() — finalize_pending_receives returns cdk::Amount and this method returns the FFI Amount, so it needs the From<CdkAmount> conversion (same as the receive_offline change above). Ended up as Ok(self.inner.finalize_pending_receives().await?.into()).

crates/cdk/src/wallet/receive/mod.rs
thesimplekid: Yeah fine with me if clippy supports it.

PR #1936, opened 2026-04-24
feat(cdk-axum): support redis-cluster for caching

### LINKED ISSUE #1578: Redis cluster
we should support redis-cluster for caching (and valkey) based on the configuration. 

https://docs.rs/redis/latest/redis/cluster/

GEET3001:
closes #1578

### Description
This PR adds support for Redis Cluster and Valkey Cluster to the cdk-axum HTTP caching backend.

Currently, the cache only supports a single-node Redis connection, which fails in clustered environments because it cannot handle "MOVED" redirections. This change introduces a cluster-aware client that automatically routes requests to the correct nodes and handles cluster topology changes.

### Key Changes
1. Introduced `RedisClient` enum to support both Standard and Cluster connection types.
2. Added `cluster` and `cluster-async` features to the redis dependency.
3. Updated `HttpCacheRedis` to use the appropriate connection methods for both single-node and cluster setups.
4. Added support for cluster configuration via environment variables:
   - `CDK_MINTD_CACHE_REDIS_USE_CLUSTER`: Enable cluster mode.
   - `CDK_MINTD_CACHE_REDIS_CLUSTER_NODES`: Comma-separated list of node URLs.
5. **Added `docker-compose.redis-cluster.yaml` to spin up a local 6-node Redis cluster (`grokzen/redis-cluster`) and a fully configured `mintd` service for easy local testing.**

### Notes to the reviewers
1. The `Config` struct in `redis.rs` was updated to make `connection_string` optional, as cluster users will use `cluster_nodes` instead. Backward compatibility is maintained for existing single-node setups.
2. The `HttpCache::from_config(config)` logic in `mod.rs` now intelligently selects the client type based on the provided configuration.
3. Added specific error logging for cluster-specific failures to aid in debugging distributed environments.

### Suggested CHANGELOG Updates
**ADDED**
- Redis Cluster and Valkey support for HTTP caching in `cdk-axum`.
- New environment variables for cluster configuration: `CDK_MINTD_CACHE_REDIS_USE_CLUSTER` and `CDK_MINTD_CACHE_REDIS_CLUSTER_NODES`.
- `docker-compose.redis-cluster.yaml` for testing cluster setups.

**FIXED**
- Cache failures in clustered Redis/Valkey environments due to lack of redirection handling.

### Checklist
- [x] I followed the code style guidelines
- [x] I ran `cargo check -p cdk-axum --features redis` and verified the build passes
- [x] If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`) — N/A (Mint-side cache change only)

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,7 +1,8 @@
 [files]
 extend-exclude = [
     "target/",
-    "bindings/dart/flutter_example"
+    "bindings/dart/flutter_example",
+    "Cargo.lock.msrv"
 ]
 
 [default]
--- a/crates/cdk-axum/Cargo.toml
+++ b/crates/cdk-axum/Cargo.toml
@@ -35,8 +35,10 @@ serde.workspace = true
 uuid.workspace = true
 sha2 = "0.10.8"
 maud = { workspace = true, optional = true }
-redis = { version = "0.31.0", features = [
+redis = { version = "0.31.0", default-features = false, features = [
     "tokio-rustls-comp",
+    "cluster-async",
+    "connection-manager",
 ], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/crates/cdk-axum/src/cache/backend/redis.rs
+++ b/crates/cdk-axum/src/cache/backend/redis.rs
@@ -5,13 +5,36 @@ use serde::{Deserialize, Serialize};
 
 use crate::cache::{HttpCacheKey, HttpCacheStorage};
 
+/// Redis client enum to handle both single-node and cluster Redis clients.
+///
+/// Both variants store a connection manager that automatically handles
+/// reconnection. The manager is cheaply cloneable (internally `Arc`-backed),
+/// so each cache operation simply clones it rather than opening a new TCP
+/// connection.
+#[derive(Clone)]
+pub enum RedisClient {
+    /// Single-node Redis client via [`redis::aio::ConnectionManager`].
+    Single(redis::aio::ConnectionManager),
+    /// Redis Cluster client via [`redis::cluster_async::ClusterConnection`].
+    Cluster(redis::cluster_async::ClusterConnection),
+}
+
+impl std::fmt::Debug for RedisClient {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Single(_) => write!(f, "Single"),
+            Self::Cluster(_) => write!(f, "Cluster"),
+        }
+    }
+}
+
 /// Redis cache storage for the HTTP cache.
 ///
 /// This cache storage backend uses Redis to store the cache.
 pub struct HttpCacheRedis {
     cache_ttl: Duration,
     prefix: Option<Vec<u8>>,
-    client: redis::Client,
+    client: RedisClient,
 }
 
 impl std::fmt::Debug for HttpCacheRedis {
@@ -19,6 +42,7 @@ impl std::fmt::Debug for HttpCacheRedis {
         f.debug_struct("HttpCacheRedis")
             .field("cache_ttl", &self.cache_ttl)
             .field("prefix", &self.prefix)
+            .field("client", &self.client)
             .finish_non_exhaustive()
     }
 }
@@ -26,16 +50,24 @@ impl std::fmt::Debug for HttpCacheRedis {
 /// Configuration for the Redis cache storage.
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct Config {
-    /// Commong key prefix
+    /// Common key prefix
     pub key_prefix: Option<String>,
 
-    /// Connection string to the Redis server.
+    /// Connection string to the Redis server (for single node).
+    #[serde(default)]
     pub connection_string: String,
+
+    /// Use Redis Cluster.
+    #[serde(default)]
+    pub use_cluster: bool,
+
+    /// Redis Cluster nodes (for cluster).
+    pub cluster_nodes: Option<Vec<String>>,
 }
 
 impl HttpCacheRedis {
     /// Create a new Redis cache.
-    pub fn new(client: redis::Client) -> Self {
+    pub fn new(client: RedisClient) -> Self {
         Self {
             client,
             prefix: None,
@@ -60,46 +92,56 @@ impl HttpCacheStorage for HttpCacheRedis {
     }
 
     async fn get(&self, key: &HttpCacheKey) -> Option<Vec<u8>> {
-        let mut conn = self
-            .client
-            .get_multiplexed_tokio_connection()
-            .await
-            .map_err(|err| {
-                tracing::error!("Failed to get redis connection: {:?}", err);
-                err
-            })
-            .ok()?;
-
         let mut db_key = self.prefix.clone().unwrap_or_default();
         db_key.extend(&**key);
 
-        conn.get(db_key)
-            .await
-            .map_err(|err| {
-                tracing::error!("Failed to get value from redis: {:?}", err);
-                err
-            })
-            .ok()?
+        match &self.client {
+            RedisClient::Single(conn) => conn
+                .clone()
+                .get(db_key)
+                .await
+                .map_err(|err| {
+                    tracing::error!("Failed to get value from redis: {}", err);
+                    err
+                })
+                .ok()?,
+            RedisClient::Cluster(conn) => conn
+                .clone()
+                .get(db_key)
+                .await
+                .map_err(|err| {
+                    tracing::error!("Failed to get value from redis cluster: {}", err);
+                    err
+                })
+                .ok()?,
+        }
     }
 
     async fn set(&self, key: HttpCacheKey, value: Vec<u8>) {
         let mut db_key = self.prefix.clone().unwrap_or_default();
         db_key.extend(&*key);
 
-        let mut conn = match self.client.get_multiplexed_tokio_connection().await {
-            Ok(conn) => conn,
-            Err(err) => {
-                tracing::error!("Failed to get redis connection: {:?}", err);
-                return;
+        match &self.client {
+            RedisClient::Single(conn) => {
+                let _: Result<(), _> = conn
+                    .clone()
+                    .set_ex(db_key, value, self.cache_ttl.as_secs())
+                    .await
+                    .map_err(|err| {
+                        tracing::error!("Failed to set value in redis: {}", err);
+                        err
+                    });
             }
-        };
-
-        let _: Result<(), _> = conn
-            .set_ex(db_key, value, self.cache_ttl.as_secs())
-            .await
-            .map_err(|err| {
-                tracing::error!("Failed to set value in redis: {:?}", err);
-                err
-            });
+            RedisClient::Cluster(conn) => {
+                let _: Result<(), _> = conn
+                    .clone()
+                    .set_ex(db_key, value, self.cache_ttl.as_secs())
+                    .await
+                    .map_err(|err| {
+                        tracing::error!("Failed to set value in redis cluster: {}", err);
+                        err
+                    });
+            }
+        }
     }
 }
--- a/crates/cdk-axum/src/cache/config.rs
+++ b/crates/cdk-axum/src/cache/config.rs
@@ -6,6 +6,10 @@ pub const ENV_CDK_MINTD_CACHE_BACKEND: &str = "CDK_MINTD_CACHE_BACKEND";
 pub const ENV_CDK_MINTD_CACHE_REDIS_URL: &str = "CDK_MINTD_CACHE_REDIS_URL";
 #[cfg(feature = "redis")]
 pub const ENV_CDK_MINTD_CACHE_REDIS_KEY_PREFIX: &str = "CDK_MINTD_CACHE_REDIS_KEY_PREFIX";
+#[cfg(feature = "redis")]
+pub const ENV_CDK_MINTD_CACHE_REDIS_USE_CLUSTER: &str = "CDK_MINTD_CACHE_REDIS_USE_CLUSTER";
+#[cfg(feature = "redis")]
+pub const ENV_CDK_MINTD_CACHE_REDIS_CLUSTER_NODES: &str = "CDK_MINTD_CACHE_REDIS_CLUSTER_NODES";
 
 pub const ENV_CDK_MINTD_CACHE_TTI: &str = "CDK_MINTD_CACHE_TTI";
 pub const ENV_CDK_MINTD_CACHE_TTL: &str = "CDK_MINTD_CACHE_TTL";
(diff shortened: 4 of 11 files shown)


thesimplekid:
https://github.com/cashubtc/cdk/blob/main/DEVELOPMENT.md#keeping-your-branch-up-to-date

crates/cdk-axum/src/cache/config.rs
```diff
@@ -70,14 +83,23 @@ impl Config {
                 // If Redis backend is selected, parse Redis configuration
                 #[cfg(feature = "redis")]
                 if matches!(self.backend, Backend::Redis(_)) {
-                    let connection_string = env::var(ENV_CDK_MINTD_CACHE_REDIS_URL)
-                        .unwrap_or_else(|_| "redis://127.0.0.1:6379".to_string());
+                    let connection_string = env::var(ENV_CDK_MINTD_CACHE_REDIS_URL).ok();
```
bansalayush247: why you removed the fallback?

GEET3001: The fallback wasn't removed — it was inadvertently moved to mod.rs during the refactor. I've restored it to config.rs where it belongs, but now it's conditional: single-node mode keeps the redis://127.0.0.1:6379 default, while cluster mode gets None (since it uses cluster_nodes instead). The duplicate in mod.rs has been removed

crates/cdk-axum/src/cache/config.rs
```diff
@@ -83,14 +92,22 @@ impl Config {
...
+                        .ok()
+                        .or_else(|| {
+                            if !use_cluster {
+                                Some("redis://127.0.0.1:6379".to_string())
+                            } else {
+                                None
+                            }
+                        });
+
+                    let key_prefix = env::var(ENV_CDK_MINTD_CACHE_REDIS_KEY_PREFIX).ok();
```
bansalayush247: There seems to be duplicated Redis env parsing logic between `Backend::from_env_str()` and `Config::from_env()`. The same env variables (`REDIS_URL`, `KEY_PREFIX`, `USE_CLUSTER`, `CLUSTER_NODES`) are being parsed in both places with nearly identical logic

It may be cleaner to extract this into a shared helper (e.g. `redis_config_from_env()`) and keep a single source of truth for env-based Redis configuration loading. This would also avoid repeated env lookups and make future config changes easier to maintain

GEET3001: Good catch! Extracted the shared logic into a redis_config_from_env() private helper — now there's a single source of truth for env-based Redis config loading. Both from_env_str and from_env call into it

crates/cdk-axum/src/cache/backend/redis.rs
```diff
@@ -60,46 +86,88 @@ impl HttpCacheStorage for HttpCacheRedis {
...
+                    .await
+                    .map_err(|err| {
+                        tracing::error!("Failed to get value from redis: {:?}", err);
+                        err
+                    })
+                    .ok()?
+            }
+            RedisClient::Cluster(client) => {
+                let mut conn = client
+                    .get_async_connection()
```
asmogo: I have a minor concern about this function call for every get and set operation. This will create a new connection. We should try to reuse existing connections.

https://github.com/redis-rs/redis-rs/blob/3ae0500efb4ff71e1cf58a2180d279780c530998/redis/src/cluster_handling/client.rs#L696

asmogo (commented):
Thanks for this contribution. Do you think we can reuse the redis cluster connection instead of creating a new connection for every set / get call on the cluster instance?

crates/cdk-axum/src/cache/backend/redis.rs
GEET3001: yeah asmogo , thanks , i read about it , you're right — get_async_connection() was creating a new ClusterConnection on every cache operation. Fixed by using tokio::sync::OnceCell to lazily initialize the connection on first use and reuse it for all subsequent calls via a cheap Clone on the Arc-backed handle. Applied the same pattern to the single-node variant for consistency

thesimplekid:
Please rebase and not merge https://github.com/cashubtc/cdk/blob/main/DEVELOPMENT.md#keeping-your-branch-up-to-date

crates/cdk-axum/src/cache/mod.rs
```diff
@@ -100,8 +102,28 @@ impl From<config::Config> for HttpCache {
...
             #[cfg(feature = "redis")]
             config::Backend::Redis(redis_config) => {
-                let client = redis::Client::open(redis_config.connection_string)
-                    .expect("Failed to create Redis client");
+                let client = if redis_config.use_cluster {
+                    let nodes = redis_config
+                        .cluster_nodes
+                        .expect("Redis cluster nodes not provided");
+                    let cluster_client = redis::cluster::ClusterClient::new(nodes)
+                        .expect("Failed to create Redis cluster client");
```
thesimplekid: We should not use expect in non test code.

crates/cdk-axum/src/cache/backend/redis.rs
```diff
@@ -1,41 +1,84 @@
...
+/// first cache operation. Subsequent operations clone the stored handle
+/// (an Arc-backed, cheap operation) rather than opening a new TCP connection.
+#[derive(Clone)]
+pub enum RedisClient {
+    /// Single-node Redis client.
+    Single {
+        /// The underlying redis client used to open the connection on first use.
+        client: redis::Client,
+        /// Lazily initialised, reusable multiplexed connection.
+        conn: Arc<OnceCell<redis::aio::MultiplexedConnection>>,
```
thesimplekid: I think this should be a Connection Manager not a multiplexed connection. https://docs.rs/redis/latest/redis/aio/struct.ConnectionManager.html

crates/cdk-axum/Cargo.toml
```diff
@@ -37,6 +37,8 @@ sha2 = "0.10.8"
 maud = { workspace = true, optional = true }
 redis = { version = "0.31.0", features = [
     "tokio-rustls-comp",
+    "cluster",
+    "cluster-async",
 ], optional = true }
```
thesimplekid:
```suggestion
redis = { version = "0.31.0", default-features = false, features = [
"tokio-rustls-comp",
    "cluster",
    "cluster-async",
], optional = true }
```

We should be more explicit about our deps and make sure default features are off and we only enable what we need.

crates/cdk-axum/Cargo.toml
```diff
@@ -37,6 +37,8 @@ sha2 = "0.10.8"
 maud = { workspace = true, optional = true }
 redis = { version = "0.31.0", features = [
     "tokio-rustls-comp",
+    "cluster",
```
thesimplekid:
```suggestion
```

crates/cdk-axum/src/cache/mod.rs
```diff
@@ -100,8 +102,28 @@ impl From<config::Config> for HttpCache {
...
+                    RedisClient::Cluster {
+                        client: cluster_client,
+                        conn: Arc::new(OnceCell::new()),
+                    }
+                } else {
+                    let connection_string = redis_config
+                        .connection_string
+                        .expect("Redis connection string not set for single-node mode");
+                    let single_client = redis::Client::open(connection_string)
+                        .expect("Failed to create Redis client");
```
thesimplekid: We should not use expect.

thesimplekid (changes requested):
A few comments and can you squash commits.

crates/cdk-axum/src/cache/mod.rs
GEET3001: replaced all .expect() calls using match and tracing::error!

crates/cdk-axum/src/cache/backend/redis.rs
GEET3001: switched from redis::Client + get_multiplexed_tokio_connection() to redis::aio::ConnectionManager.

crates/cdk-axum/Cargo.toml
GEET3001: Applied — the redis dependency now uses default-features = false

thesimplekid:
Code looks good. Waiting on testing and approval from @asmogo

GEET3001:
> Code looks good. Waiting on testing and approval from @asmogo

any changes ?

asmogo:
I got some panics with your current implementation on the mintd startup. 

requesting some changes here. 

https://github.com/GEET3001/cdk/pull/1

thesimplekid:
Panics stem from the use of block on. 

---
title: "Panics in HttpCache initialization due to nested block_on calls"
notify: true
status: confirmed
severity: high
target: cashubtc/cdk
pr: 1936
skills_used: ["rust-security", "cashu"]
findings_count: 1
---

## Summary
The changes introduced in PR #1936 add support for Redis clusters as a caching backend. However, the PR introduces a fatal regression where it initializes Redis clients and connections using `tokio::runtime::Handle::current().block_on(...)` synchronously inside `HttpCache::from(config)`.

`HttpCache::from` is invoked inside `configure_cache` and `start_services_with_shutdown` in `crates/cdk-mintd/src/lib.rs`. These functions are part of the application's async startup flow, originating from `cdk_mintd::run_mintd`, which itself is executed inside an outer `rt.block_on(async { ... })` in `crates/cdk-mintd/src/main.rs`. Because Tokio explicitly forbids calling `Handle::block_on` from within a thread that is already executing an active Tokio runtime, the `.block_on(...)` calls added in `crates/cdk-axum/src/cache/mod.rs` will inevitably panic and crash the `cdk-mintd` service upon startup if Redis is configured.

## Skills Used
- rust-security: Applied Rust and Tokio concurrency knowledge to identify that nesting `block_on` inside an already active Tokio thread leads to immediate panics ("Cannot start a runtime from within a runtime").
- cashu: Analyzed the cdk-mintd caching startup logic where mint settings for supported tokens dictate caching behavior during service instantiation.

## Root Cause
In `crates/cdk-axum/src/cache/mod.rs` (lines 114 and 143), the PR introduces synchronous blocks inside the `From<config::Config>` implementation for `HttpCache`:

```rust
match tokio::runtime::Handle::current().block_on(cluster_client.get_async_connection())
// ...
match tokio::runtime::Handle::current().block_on(redis::aio::ConnectionManager::new(single_client))
```

This synchronization happens via `.into()` when configuring the mint. In `crates/cdk-mintd/src/lib.rs`, `configure_cache` and `start_services_with_shutdown` invoke this exactly:

```rust
let cache: HttpCache = settings.info.http_cache.clone().into();
```

Both of these functions execute during the application startup process inside `run_mintd_with_shutdown`, which runs within the main Tokio runtime initialized in `crates/cdk-mintd/src/main.rs`:

```rust
    rt.block_on(async {
        // ...
        cdk_mintd::run_mintd(
            // ...
        ).await
    })
```

Because `Handle::current().block_on` is called on the current handle while already executing inside an active tokio worker thread (from the outer `rt.block_on(async { ... })`), Tokio will detect this and panic immediately with the error "Cannot start a runtime from within a runtime".

## Attack Steps
1. The administrator of the cashu mint updates to the version containing this PR.
2. The administrator enables Redis caching in their configuration (setting backend to `redis` and specifying a valid connection string).
3. The server is restarted, executing `cdk-mintd`.
4. `run_mintd` is called inside `rt.block_on(async { ... })`.
5. During startup, `configure_mint_builder` runs and calls `configure_cache`.
6. `configure_cache` calls `settings.info.http_cache.clone().into()`, which triggers `HttpCache::from`.
7. Inside `HttpCache::from`, Tokio's `tokio::runtime::Handle::current().block_on(...)` is executed to establish the Redis connection.
8. Tokio immediately panics because `block_on` cannot be called from an async context inside an active runtime worker thread. The application crashes.
9. The caching configuration behaves as an immediate Denial of Service killswitch, preventing the Mint service from successfully coming online.

## Impact
This introduces a critical Denial of Service (startup crash) resulting directly from legitimate configurations. Any mint operator updating to use the Redis or Redis Cluster caching backend will face an immediate, persistent crash loop. The application will unconditionally panic and terminate on startup if Redis configuration is present.

## Proof of Concept
A simple Rust test reproduces the exact Tokio panic occurring in the application:

```rust
fn main() {
    // 1. Emulate what `src/main.rs` does
    let rt = tokio::runtime::Runtime::new().unwrap();
    
    rt.block_on(async {
        // Simulates being within `run_mintd` and calling HttpCache::from()
        let handle = tokio::runtime::Handle::current();
        
        // Simulates the HttpCache::from call added in the PR
        // which naively uses `Handle::current().block_on(...)`
        handle.block_on(async {
            println!("This will not print, tokio will panic first.");
        });
    });
}
```

**Output:**
```
thread 'main' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive async tasks.'

**Comment by thesimplekid** [comment]
@GEET3001 This paniced pretty early, did you run or test this code yourself?

**Comment by GEET3001** [comment]
> @GEET3001 This paniced pretty early, did you run or test this code yourself?

You're right. I apologize for that. I tested the cache module in isolation but didn't test the full cdk-mintd startup path where the nested block_on panic occurs.

I've fixed this by replacing the sync From<Config> trait with an async fn from_config(), so Redis connections are now properly .awaited instead of using block_on. The callers in cdk-mintd/src/lib.rs have been updated to await the new async constructor. No block_on calls remain.

**Comment by GEET3001** [comment]
closes #1578

### Description
This PR adds support for Redis Cluster and Valkey Cluster to the cdk-axum HTTP caching backend.

Currently, the cache only supports a single-node Redis connection, which fails in clustered environments because it cannot handle "MOVED" redirections. This change introduces a cluster-aware client that automatically routes requests to the correct nodes and handles cluster topology changes.

### Key Changes
1. Introduced `RedisClient` enum to support both Standard and Cluster connection types.
2. Added `cluster` and `cluster-async` features to the redis dependency.
3. Updated `HttpCacheRedis` to use the appropriate connection methods for both single-node and cluster setups.
4. Added support for cluster configuration via environment variables:
   - `CDK_MINTD_CACHE_REDIS_USE_CLUSTER`: Enable cluster mode.
   - `CDK_MINTD_CACHE_REDIS_CLUSTER_NODES`: Comma-separated list of node URLs.
5. **Added `docker-compose.redis-cluster.yaml` to spin up a local 6-node Redis cluster (`grokzen/redis-cluster`) and a fully configured `mintd` service for easy local testing.**

### Notes to the reviewers
1. The `Config` struct in `redis.rs` was updated to make `connection_string` optional, as cluster users will use `cluster_nodes` instead. Backward compatibility is maintained for existing single-node setups.
2. The `HttpCache::from_config(config)` logic in `mod.rs` now intelligently selects the client type based on the provided configuration.
3. Added specific error logging for cluster-specific failures to aid in debugging distributed environments.

### Suggested CHANGELOG Updates
**ADDED**
- Redis Cluster and Valkey support for HTTP caching in `cdk-axum`.
- New environment variables for cluster configuration: `CDK_MINTD_CACHE_REDIS_USE_CLUSTER` and `CDK_MINTD_CACHE_REDIS_CLUSTER_NODES`.
- `docker-compose.redis-cluster.yaml` for testing cluster setups.

**FIXED**
- Cache failures in clustered Redis/Valkey environments due to lack of redirection handling.

### Checklist
- [x] I followed the code style guidelines
- [x] I ran `cargo check -p cdk-axum --features redis` and verified the build passes
- [x] If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`) — N/A (Mint-side cache change only)

**Comment by thesimplekid** [comment]
@codex review

**Comment by thesimplekid** [comment]
@asmogo could you give this another test when you have a chance?

File: crates/cdk-axum/src/cache/backend/redis.rs [hunk 9]
```diff
@@ -5,37 +5,68 @@ use serde::{Deserialize, Serialize};
...
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct Config {
-    /// Commong key prefix
+    /// Common key prefix
     pub key_prefix: Option<String>,
 
-    /// Connection string to the Redis server.
-    pub connection_string: String,
+    /// Connection string to the Redis server (for single node).
+    pub connection_string: Option<String>,
```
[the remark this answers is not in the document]
GEET3001: connection_string has been reverted back to a required String type with a #[serde(default)] fallback to preserve the existing public API for downstream users.

The environment parser (redis_config_from_env) now explicitly handles the fallback so it will cleanly pass "redis://127.0.0.1:6379" if no URL is provided, and the from_config logic handles empty strings

crates/cdk-mintd/example.config.toml
```diff
@@ -47,6 +47,10 @@ tti = 60
 # key_prefix = "mintd"
 # connection_string = "redis://localhost"
 
+# For redis-cluster:
+# use_cluster = true
+# cluster_nodes = ["redis://node1:6379", "redis://node2:6379"]
```
asmogo, line 52:
```suggestion
# cluster_nodes = ["redis://127.0.0.1:7001", "redis://127.0.0.1:7002", "redis://127.0.0.1:7003"]
```

.typos.toml
```diff
@@ -1,7 +1,8 @@
 [files]
 extend-exclude = [
     "target/",
-    "bindings/dart/flutter_example"
+    "bindings/dart/flutter_example",
+    "Cargo.lock.msrv"
```
asmogo, line 5: why are we adding `Cargo.lock.msrv` here? is this necessary?

asmogo (changes requested):
ACK f786d5d5bad17ad03b0182889e5222ce5d0e9a7a

.typos.toml
GEET3001: The Cargo.lock.msrv file contains auto-generated alphanumeric hashes that trigger false-positive spelling errors in the CI typos job. The spellchecker ignores standard Cargo.lock files by default, but we had to explicitly add the .msrv variant so the pipeline doesn't fail

thesimplekid:
You do not need to rebase and force push if there are no changes and conflicts, its hard to tell if there are changes since approval.

thesimplekid (approved):

asmogo (approved):

PR #1939, opened 2026-04-25
Add Kotlin bindings build config and CI publish workflow

crodas:
### Description

Add Gradle build configuration for JVM, Android, and iOS targets, along with a GitHub Actions workflow for publishing Kotlin packages.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/kotlin-publish.yml
+++ b/.github/workflows/kotlin-publish.yml
@@ -1,0 +1,399 @@
+name: "FFI - Kotlin Bindings"
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: "Tag to create for release (e.g. v0.17.0)"
+        required: true
+      cdk_version:
+        description: 'CDK dependency version (e.g. 0.17.0). Defaults to release_tag without the "v" prefix.'
+        required: false
+env:
+  CARGO_TERM_COLOR: ${{ vars.CARGO_TERM_COLOR }}
+
+jobs:
+  kotlin-sync-sources:
+    name: "kotlin: Sync sources to cdk-kotlin"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout monorepo
+        uses: actions/checkout@v4
+
+      - name: Clone cdk-kotlin
+        run: |
+          git clone https://x-access-token:${{ secrets.KOTLIN_DEPLOY_KEY }}@github.com/${{ vars.CDK_KOTLIN_REPO }}.git cdk-kotlin
+
+      - name: Copy bindings/kotlin to cdk-kotlin
+        run: |
+          rsync -av --delete \
+            --exclude='.github' \
+            --exclude='target' \
+            --exclude='.git' \
+            bindings/kotlin/ cdk-kotlin/
+          cp rust-toolchain.toml cdk-kotlin/
+
+      - name: Update versions and resolve workspace references
+        run: |
+          TAG="${{ inputs.release_tag }}"
+          CDK_VER="${{ inputs.cdk_version }}"
+          CDK_VER="${CDK_VER:-${TAG#v}}"
+          RELEASE_VER="${TAG#v}"
+          echo "Setting CDK version to ${CDK_VER}, release version to ${RELEASE_VER}"
+
+          # Helper: extract a value from [workspace.package] in root Cargo.toml
+          ws_pkg() {
+            sed -n '/^\[workspace\.package\]/,/^\[/p' Cargo.toml \
+              | grep "^$1 " | head -1 | sed 's/.*= *"\(.*\)"/\1/'
+          }
+
+          # Helper: extract a workspace dependency version
+          ws_dep_ver() {
+            local line
+            line=$(sed -n '/^\[workspace\.dependencies\]/,/^\[/p' Cargo.toml \
+              | grep "^$1 " | head -1)
+            # Handle both: dep = "version" and dep = { ..., version = "x", ... }
+            if echo "$line" | grep -q 'version *= *"'; then
+              echo "$line" | sed 's/.*version *= *"\([^"]*\)".*/\1/'
+            else
+              echo "$line" | sed 's/.*= *"\([^"]*\)".*/\1/'
+            fi
+          }
+
+          # Read workspace values dynamically
+          WS_EDITION=$(ws_pkg edition)
+          WS_RUST_VERSION=$(ws_pkg rust-version)
+          WS_LICENSE=$(ws_pkg license)
+          WS_HOMEPAGE=$(ws_pkg homepage)
+          WS_REPOSITORY=$(ws_pkg repository)
+          WS_README=$(ws_pkg readme)
+          WS_UNIFFI_VER=$(ws_dep_ver uniffi)
+
+          echo "Workspace values: edition=${WS_EDITION} rust-version=${WS_RUST_VERSION} uniffi=${WS_UNIFFI_VER}"
+
+          # Update gradle.properties version
+          sed -i "s/^VERSION_NAME=.*/VERSION_NAME=${RELEASE_VER}/" cdk-kotlin/gradle.properties
+
+          # Resolve workspace package fields in rust/Cargo.toml
+          sed -i \
+            -e "s/^edition\.workspace = true/edition = \"${WS_EDITION}\"/" \
+            -e "s/^rust-version\.workspace = true/rust-version = \"${WS_RUST_VERSION}\"/" \
+            -e "s/^license\.workspace = true/license = \"${WS_LICENSE}\"/" \
+            -e "s|^homepage\.workspace = true|homepage = \"${WS_HOMEPAGE}\"|" \
+            -e "s|^repository\.workspace = true|repository = \"${WS_REPOSITORY}\"|" \
+            -e "s/^version\.workspace = true/version = \"${CDK_VER}\"/" \
+            -e "s/^readme\.workspace = true/readme = \"${WS_README}\"/" \
+            cdk-kotlin/rust/Cargo.toml
+
+          # Resolve workspace dependency references
+          sed -i \
+            -e "s/^cdk-ffi\.workspace = true/cdk-ffi = { version = \"=${CDK_VER}\", default-features = false }/" \
+            -e "s/cdk-ffi = { workspace = true/cdk-ffi = { version = \"=${CDK_VER}\"/" \
+            -e "s/uniffi = { workspace = true/uniffi = { version = \"${WS_UNIFFI_VER}\"/" \
+            cdk-kotlin/rust/Cargo.toml
+
+          # Remove workspace lints reference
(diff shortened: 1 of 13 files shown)


bindings/kotlin/README.md
thesimplekid, line 1: Can we add what secrets are needed and any other set up like we did for the dart.

bindings/swift/generate-bindings.sh
```diff
@@ -206,6 +206,11 @@ cargo run --bin uniffi-bindgen-swift -- \
     "$SOURCES_DIR" \
     --swift-sources
 
+# Fix FFI module import to match the XCFramework module name.
+# Use perl for portable in-place editing (nix provides GNU sed where -i '' behaves differently).
+perl -pi -e 's/canImport\(cdk_ffiFFI\)/canImport(CashuDevKitFFI)/g' "$SOURCES_DIR"/*.swift
+perl -pi -e 's/import cdk_ffiFFI/import CashuDevKitFFI/g' "$SOURCES_DIR"/*.swift
+
```
thesimplekid: Can we avoid perl. We use sed everywhere else and the point of nix is to have deterministic environment so we dont have to worry about portability

bindings/kotlin/cdk-android/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+
+android {
+    namespace = "org.cashudevkit"
+    compileSdk = 34
+    defaultConfig {
+        minSdk = 24
+    }
+    publishing {
+        singleVariant("release") {
+            withSourcesJar()
```
thesimplekid, line 15: Did you verify this by publishing to Maven Central or running Central validation? I don't think all of these publications will pass without sources and javadoc artifacts.

cdk-jvm seems to have both sources and javadocs.

cdk-android has sources but is missing javadocs.

cdk-jvm-natives and cdk-ios appear to be missing both sources and javadocs.

Central requires every non-pom artifact to include matching -sources.jar and -javadoc.jar artifacts, though placeholder jars are allowed if real sources/docs are not available:

https://central.sonatype.org/publish/requirements/#supply-javadoc-and-sources

.github/workflows/kotlin-publish.yml
```diff
@@ -0,0 +1,383 @@
...
+
+jobs:
+  kotlin-sync-sources:
+    name: "kotlin: Sync sources to cdk-kotlin"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout monorepo
+        uses: actions/checkout@v4
+        with:
+          ref: main
```
thesimplekid: This should probably be the tag of what we are release right not main? Does this handle the release branches for backports?

.github/workflows/kotlin-publish.yml
```diff
@@ -0,0 +1,383 @@
...
+        run: |
+          TAG="${{ inputs.release_tag }}"
+          RELEASE_BRANCH="release/v${TAG#v}"
+          DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+          git checkout "${DEFAULT_BRANCH}"
+          git merge "${RELEASE_BRANCH}" --squash
+          git commit -m "release: ${TAG}"
+          git tag "${TAG}"
+          git push origin "${DEFAULT_BRANCH}" --tags
+          git push origin --delete "${RELEASE_BRANCH}"
```
thesimplekid: I think this should only happen after a successful publish. If publishing failed we should not modify main to make retrying easier this is how cdk-kotlin currently does it and I think its a better flow.

.github/workflows/kotlin-publish.yml
```diff
@@ -0,0 +1,383 @@
...
+            (git commit -m "prebuilt binaries before bindgen" && git push origin HEAD)
+
+      - name: Commit to release branch
+        working-directory: cdk-kotlin
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add -A
+          git diff --cached --quiet && echo "No changes to commit" && exit 0
+          git commit -m "release: ${{ inputs.release_tag }} — add bindings and prebuilt binaries"
```
thesimplekid, line 358: Doesn't the above step already committed all this so this is a duplicate?

bindings/swift/generate-bindings.sh
```diff
@@ -206,6 +206,10 @@ cargo run --bin uniffi-bindgen-swift -- \
     "$SOURCES_DIR" \
     --swift-sources
 
+# Fix FFI module import to match the XCFramework module name
+sed -i 's/canImport(cdk_ffiFFI)/canImport(CashuDevKitFFI)/g' "$SOURCES_DIR"/*.swift
+sed -i 's/import cdk_ffiFFI/import CashuDevKitFFI/g' "$SOURCES_DIR"/*.swift
```
thesimplekid, line 211: I think this is not compatiable on mac? Maybe: 

```bash
sed -i.bak 's/canImport(cdk_ffiFFI)/canImport(CashuDevKitFFI)/g' "$SOURCES_DIR"/*.swift
sed -i.bak 's/import cdk_ffiFFI/import CashuDevKitFFI/g' "$SOURCES_DIR"/*.swift
rm -f "$SOURCES_DIR"/*.swift.bak
```

.github/workflows/kotlin-publish.yml
```diff
@@ -0,0 +1,383 @@
...
+            https://x-access-token:${{ secrets.KOTLIN_DEPLOY_KEY }}@github.com/${{ vars.CDK_KOTLIN_REPO }}.git cdk-kotlin
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '17'
+
+      - name: Set up Android SDK
+        uses: android-actions/setup-android@v3
```
thesimplekid: We should use nix for these set ups, thats what we already do for cdk-kotlin.

.github/workflows/kotlin-publish.yml
```diff
@@ -0,0 +1,383 @@
...
+            https://x-access-token:${{ secrets.KOTLIN_DEPLOY_KEY }}@github.com/${{ vars.CDK_KOTLIN_REPO }}.git cdk-kotlin
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+
+      - name: Install cross (Android)
+        if: contains(matrix.target, 'android')
+        run: cargo install cross --git https://github.com/cross-rs/cross --tag v0.2.5
```
thesimplekid: I don't think we need cross cdk-kotlin doesn't use it by providing the android target in the nix env.

.github/workflows/kotlin-publish.yml
```diff
@@ -0,0 +1,383 @@
...
+
+          # macOS
+          - target: aarch64-apple-darwin
+            runner: macos-14
+            artifact: libcdk_ffi.dylib
+
+          # Windows
+          - target: x86_64-pc-windows-msvc
+            runner: windows-latest
+            artifact: cdk_ffi.dll
```
thesimplekid: Is anyone testing these? I have no way (or willingness) to test windows support. If not we should just remove it.

thesimplekid (changes requested):

bindings/swift/generate-bindings.sh
crodas: I'm testing on Linux and Mac.

.github/workflows/kotlin-publish.yml
crodas: I agree

.github/workflows/kotlin-publish.yml
crodas: Probably, but for now I need main, since the release does not have the changes yet. I can make it whenever I'm doing the release

bindings/kotlin/cdk-android/build.gradle.kts
crodas: I will test with a throw-away maven central account or I will run the central validation

.github/workflows/kotlin-publish.yml
thesimplekid: this should be changed before merge not before release.

bindings/kotlin/cdk-android/build.gradle.kts
thesimplekid: What parts of this have been tested then if the publishing has not been?

crodas: This PR creates a git-release, similar to [dart](https://github.com/cashubtc/cdk/pull/1886), that includes all generated UniFFI files and the compiled binaries for each platform. With the target repository, I pull locally and run the unit tests on Linux and OSX, ensuring no Rustc is available.

I will test now the maven publication

thesimplekid: In the future we should note on the PR if there is outstanding testing to be done.

.github/workflows/kotlin-publish.yml
```diff
@@ -0,0 +1,395 @@
...
+
+      - name: Commit to release branch
+        working-directory: cdk-kotlin
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add -A
+          git diff --cached --quiet && echo "No changes to commit" && exit 0
+          git commit -m "release: ${{ inputs.release_tag }} — add bindings and prebuilt binaries"
+          git push origin HEAD
```
thesimplekid: Should this push to the `cdk-kotlin` release branch wait until after Maven Central publish succeeds? As written, a publish failure leaves a mutated remote release branch, so retrying can require manual cleanup and line 112’s `git checkout -b` path is not idempotent. The project release guidance prefers mutating external repos only after the publish step succeeds.

.github/workflows/kotlin-publish.yml
```diff
@@ -0,0 +1,395 @@
...
+              :cdk-android:publishReleasePublicationToSonatypeRepository \
+              :cdk-ios:publishAllPublicationsToSonatypeRepository \
+              closeAndReleaseSonatypeStagingRepository
+
+      - name: Git cleanup
+        working-directory: cdk-kotlin
+        run: |
+          TAG="${{ inputs.release_tag }}"
+          git tag "${TAG}"
+          git push origin "${DEFAULT_BRANCH}" --tags
```
thesimplekid, line 396: `DEFAULT_BRANCH` is set in the previous step, but GitHub Actions runs each step in a fresh shell, so this expands to an empty string here. If Maven publishing succeeds, this cleanup step can still fail before pushing the default branch/tag. Can we recompute `DEFAULT_BRANCH` in this step or persist it through `$GITHUB_ENV`?

thesimplekid (approved):

PR #1941, opened 2026-04-27
fix: create melt quote kind index on melt table

asmogo:
### Description

-----

### Notes to the reviewers

idx_melt_quote_request_lookup_id_and_kind must be created on the melt quote table 

idx_mint_quote_request_lookup_id_and_kind is already present in the mint_quote table

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sql-common/src/mint/migrations/postgres/1_initial.sql
+++ b/crates/cdk-sql-common/src/mint/migrations/postgres/1_initial.sql
@@ -95,6 +95,6 @@ CREATE TABLE mint_quote_issued (
   FOREIGN KEY (quote_id) REFERENCES mint_quote(id)
 );
 CREATE INDEX idx_mint_quote_issued_quote_id ON mint_quote_issued(quote_id);
-CREATE INDEX idx_melt_quote_request_lookup_id_and_kind ON mint_quote(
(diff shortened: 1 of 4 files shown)


thesimplekid:
We can't do this because it is possible to have two melt quotes for the same payment request (bolt11 invoice) in the unpaid state. What is not allowed is moving multiple into pending or paid. 

We actually used to not allow this and is was often complained about as it prevented getting a second quote if you previously got one to pay an invoice but never followed through and paid and wanted a second one or even if another person got an quote but never paid it.


Is there a reason you are looking to add this did you notice an issue?

asmogo:
This is just fixing the regular (non unique) index for query performance.

`idx_melt_quote_quote_request_kind` is currently being created in the mint _quote database table.

I think you are referring to https://github.com/cashubtc/cdk/pull/1942

thesimplekid:
> This is just fixing the regular (non unique) index for query performance.
> 
> `idx_melt_quote_quote_request_kind` is currently being created in the mint _quote database table.
> 
> I think you are referring to #1942

Yep sorry, I maybe have just commented on the wrong one. Or maybe GH was showing me the wrong diff...

But this looks good thanks

thesimplekid (dismissed):

PR #1942, opened 2026-04-27
fix: melt lookup kind unique index

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


TheMhv:
Where we need to use that index?

thesimplekid:
Originally left this comment on the wrong [PR](https://github.com/cashubtc/cdk/pull/1941#issuecomment-4337588481) but it still applies here. 

We can't do this because it is possible to have two melt quotes for the same payment request (bolt11 invoice) in the unpaid state. What is not allowed is moving multiple into pending or paid.

We actually used to not allow this and is was often complained about as it prevented getting a second quote if you previously got one to pay an invoice but never followed through and paid and wanted a second one or even if another person got an quote but never paid it.

Is there a reason you are looking to add this did you notice an issue?

thesimplekid:
Closing per https://github.com/cashubtc/cdk/pull/1942#issuecomment-4359252670 please reopen if I'm wrong.

asmogo:
no i did not notice any issues. my assumption was that this index should be used by functions like https://github.com/cashubtc/cdk/blob/main/crates/cdk-sql-common/src/mint/quotes.rs#L365 

We might want to introduce a another non-unique index for that.

I'm okay with closing this pr.

Issue #1945, opened 2026-04-28
`rotate_keyset` does not validate that expiry is in the future


gudnuf:
Currently,  `rotate_keyset` accepts `final_expiry` values in the past, creating keysets that are born expired and immediately unusable. Both `Mint::rotate_keyset` (`crates/cdk/src/mint/keysets/mod.rs:74`) and the signatory implementations forward the value to storage without validating it.

  ### Reproduction

  ```rust
  mint.rotate_keyset(
      CurrencyUnit::Sat,
      amounts,
      0,
      true,
      Some(unix_time() - 1), // accepted, not rejected
  ).await.unwrap();

### Expected behavior

Attempting to set the`final_expiry` to a time in the past should return an error

TheMhv:
How this going @thesimplekid ?

j-kon:
Hi, I checked the current implementation and this still appears to be reproducible.

`Mint::rotate_keyset` forwards `final_expiry` to the signatory without validation, and `DbSignatory::rotate_keyset` currently creates and stores the keyset even when the timestamp has already passed.

I’d like to work on this by rejecting `Some(expiry)` when `expiry <= unix_time()` and adding tests for both the signatory and mint paths.

Would you prefer reusing `Error::ExpiredKeyset`, or adding a more specific error such as `InvalidKeysetExpiry`?

PR #1946, opened 2026-04-28
feat: enforce keyset expiry

### LINKED ISSUE #1895: Enforce final expiry of keysetv2 keyset

gudnuf:
closes #1895

### Description

Enforces keyset expiry by rejecting inputs and outputs for an expired keyset. The expired keyset stays active.

-----

### Notes to the reviewers

This doesn't handle the wallet side of things. Also, while working on the tests I realized that the mint allows for keysets to be rotated with an expiry in the past, I created an issue for that #1945 

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/error.rs
+++ b/crates/cdk-common/src/error.rs
@@ -207,6 +207,9 @@ pub enum Error {
     /// Inactive Keyset
     #[error("Inactive Keyset")]
     InactiveKeyset,
+    /// Keyset has expired
+    #[error("Keyset has expired")]
+    ExpiredKeyset,
     /// Transaction unbalanced
     #[error("Inputs: `{0}`, Outputs: `{1}`, Expected Fee: `{2}`")]
     TransactionUnbalanced(u64, u64, u64),
@@ -574,6 +577,7 @@ impl Error {
             | Self::UnknownKeySet
             | Self::BlindedMessageAlreadySigned
             | Self::InactiveKeyset
+            | Self::ExpiredKeyset
             | Self::TransactionUnbalanced(_, _, _)
             | Self::DuplicateInputs
             | Self::DuplicateOutputs
@@ -879,6 +883,10 @@ impl From<Error> for ErrorResponse {
                 code: ErrorCode::KeysetInactive,
                 detail: err.to_string(),
             },
+            Error::ExpiredKeyset => ErrorResponse {
+                code: ErrorCode::KeysetExpired,
+                detail: err.to_string(),
+            },
             Error::AmountLessNotAllowed => ErrorResponse {
                 code: ErrorCode::AmountlessInvoiceNotSupported,
                 detail: err.to_string(),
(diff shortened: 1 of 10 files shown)


TheMhv:
How i can test it? I try to change the `active` and `valid_to` fields directly in keyset table inside mint database but does work

gudnuf:
> How i can test it?

use the management RPC CLI to rotate 

``` bash
mintd-cli  rotate-next-keyset --final-expiry $(($(date +%s) + SECONDS_IN_THE_FUTURE))
```

Then mint ecash before SECONDS_IN_THE_FUTURE. Wait for expiry, then try to use that ecash.

>  I try to change the `active` and `valid_to` fields directly in keyset table inside mint database but does work

I'm not sure that works because the keyset ID commits to the expiry and mintd will auto-rotate on startup if you have no active keys. Also the mint might keep the active keys in memory.

TheMhv:
The error occurs after invoice has created and paid, i think that the validation needs to be before the invoice creation.

image

---

image

gudnuf:
> The error occurs after invoice has created and paid, i think that the validation needs to be before the invoice creation.

Yea this is tricky because keyset can also expire after the quote is created like I described here https://github.com/cashubtc/cdk/issues/1895#issuecomment-4307094192

So we could return an error for the request to create a mint quote if the active keyset is expired, but that only solves half the problem. Do you have any suggestion on how this should work @TheMhv?

 I think that this just needs to be handle client side by checking the keyset expiry.

TheMhv:
> Do you have any suggestion on how this should work @TheMhv?

I cannot create new mint quotes from expired keyset, the mint should return an error.

Beside that, we can has mint quotes that was created when keyset was valid, but now it's expired. For this case i have one idea, we can create a "quote renew" request that requires a payment proof (the `preimage` for lightning payments for example) for renew the paid mint quote to the new keyset.

thesimplekid:
> > The error occurs after invoice has created and paid, i think that the validation needs to be before the invoice creation.
> 
> Yea this is tricky because keyset can also expire after the quote is created like I described here [#1895 (comment)](https://github.com/cashubtc/cdk/issues/1895#issuecomment-4307094192)
> 
> So we could return an error for the request to create a mint quote if the active keyset is expired, but that only solves half the problem. Do you have any suggestion on how this should work @TheMhv?
> 
> I think that this just needs to be handle client side by checking the keyset expiry.

We could disable minting if there is no active keyset but really that doesn't always solve the problem since you could have got a mint quote before the expiry and then attempt to mint after, no check we can do can protect from that. I think its an edge case and mints generally won't be running in this state it is only if they are shutting down.

thesimplekid (approved):

PR #1948, opened 2026-04-29
Fix Prometheus in-flight metric handling

asmogo:
### Description

added a small MintMetricGuard that increments on creation, decrements on drop, and records success/error once.

Bolt11 and Bolt12 quote paths now close in-flight metrics on every normal/error path; Bolt12 uses get_melt_bolt12_quote.

swap metrics are recorded once at the top-level result boundary; removed the saga compensation double-decrement.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


asmogo:
i have noticed a configuration issue with prometheus. the toml configuration to enable prometheus won't work. It's currently required to use the env variable `CDK_MINTD_PROMETHEUS_ENABLED=true`. 

fixing this rn.

asmogo:
> i have noticed a configuration issue with prometheus

The issue is me having mixed toml and env configuration. my toml configuration does not include the postgres url. the parser is silently failing and falling back to env variables. I will address this in another PR that handles the configuration merge.

crates/cdk-common/src/payment.rs
```diff
@@ -298,6 +298,15 @@ pub enum OutgoingPaymentOptions {
...
 
 impl OutgoingPaymentOptions {
+    /// Return the low-cardinality payment method label used for metrics.
+    pub fn payment_method_label(&self) -> &str {
+        match self {
+            Self::Bolt11(_) => "bolt11",
+            Self::Bolt12(_) => "bolt12",
+            Self::Custom(options) => options.method.as_str(),
+        }
+    }
```
asmogo, line 365: is this required ?

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -411,103 +414,110 @@ impl Mint {
...
-            .get(&PaymentProcessorKey::new(
-                unit.clone(),
-                PaymentMethod::Known(KnownMethod::Bolt12),
-            ))
-            .ok_or_else(|| {
-                tracing::info!("Could not get ln backend for {}, bolt12 ", unit);
+        #[cfg(feature = "prometheus")]
+        let metrics = super::MintMetricGuard::new("get_melt_bolt12_quote");
+
+        let result = async {
```
asmogo, line 430: I don't like this result wrapper pattern.

asmogo (commented):
- updated tokio listener
- removed global metrics
- rename lightning_payment to payment.
- record lightning payment in make payment function.

We are not recording any fees.

thesimplekid:
Rebased over here https://github.com/cashubtc/cdk/pull/2017. Ill take this over

PR #1949, opened 2026-04-30
feat(mintd): allow multiple lightning backends per currency unit

### LINKED ISSUE #1465: Feature: allow multiple payment backends
Currently, cdk-mintd is designed to allow a single lightning backend. Expanding this to allow multiple concurrent backends would enable: 

- Having multiple lightning backends for redundancy. 
- Unit specific backends: sat - usd - eur in a single mint 

Solution to this could be allowing a arrays of `[ln]` backends

Micah-Shallom:
## Summary

Adds support for running multiple Lightning backends in a single `cdk-mintd`, with each backend serving a different currency unit. So one mint can hand out sat-denominated ecash via CLN and eur-denominated ecash via LNbits, all through the same mint URL.

The mint engine itself already supported this (its payment-processor map is keyed by `(unit, method)`); the only thing blocking it was `cdk-mintd`'s config and startup, which assumed a single backend. This PR removes that assumption.

Closes #1465

## What operators see

Existing configs keep working. The old single `[ln]` block is still accepted and behaves identically to today.

For a multi-backend setup, repeat `[[ln]]` instead, one entry per `(unit, method)`:

```toml
[[ln]]
ln_backend = "cln"
unit = "sat"

[[ln]]
ln_backend = "lnbits"
unit = "eur"

[cln]
rpc_path = "..."

[lnbits]
admin_api_key = "..."
# ...
```

Each entry has its own mint/melt limits. Two entries pointing to the same `(unit, method)` are rejected at startup with a clear error.


crates/cdk-mintd/src/lib.rs
```diff
@@ -476,171 +476,207 @@ async fn configure_lightning_backend(
...
+                    "Attempting to start with gRPC payment processor at {}:{}.",
+                    grpc_processor.addr,
+                    grpc_processor.port
+                );
 
-            for unit in grpc_processor.clone().supported_units {
-                tracing::debug!("Adding unit: {:?}", unit);
-                let processor = grpc_processor
-                    .setup(settings, unit.clone(), None, work_dir, None)
+                for unit in grpc_processor.clone().supported_units {
```
asmogo: I believe we should deprecate the `supported_units` variable and instead use `ln_entry` units in this context. With this setup, we may need to define the unit twice in the configuration, once for the `supported_units` variable and once for the `ln_entry` variable. 
```
[grpc_processor]
addr = "http://127.0.0.1"
port = 50051
supported_units = ["sat"]
``` 

```
[[ln]]
ln_backend = "grpcprocessor"
unit = "sat"
```

I’m not sure if we should do this in the same pull request.

asmogo (changes requested):
Thanks for your contribution. 

I have a minor change request regarding the grpc processor configuration. Apart from that, this looks good to me.

crates/cdk-mintd/src/lib.rs
Micah-Shallom: @asmogo , i pushed e4c2b2d3 in this PR since it's small. supported_units stays in the struct so legacy configs still parse, but it no longer drives registration.

thesimplekid:
@asmogo can you rebase and not merge please

asmogo:
sorry, I have accidentally clicked the update branch button in github.

thesimplekid:
Close for https://github.com/cashubtc/cdk/pull/2012

PR #1954, opened 2026-05-04
 Migrate INTEGER/BIGINT columns storing u64 values to TEXT

crodas:
### Description

Cashu amounts are u64, but neither SQLite nor PostgreSQL has an unsigned 64-bit integer type. Both INTEGER and BIGINT are signed, so values above `i64::MAX` silently overflow when cast to i64. This PR stores all amount-like columns as TEXT instead, eliminating the risk of truncation entirely.

Once columns become TEXT, PostgreSQL refuses to do arithmetic on them; expressions like total_issued + EXCLUDED.total_issued simply fail. Even with SQLite, casting large text values back to integers for math is unreliable. Rather than patching the SQL with ever more complex `CAST/COALESCE` chains, this PR moves all arithmetic into Rust. Each keyset amounts update now reads the current value, adds the delta with checked_add (which returns an explicit error on overflow instead of wrapping silently), and writes the result back. The same applies to SUM() aggregations — amounts are fetched as individual rows and summed in Rust.

A secondary fix handles PostgreSQL's `::` cast syntax (`column::TYPE`), which the SQL parser was misinterpreting as a named bind parameter.

As a next step, amounts should be streamed from storage rather than loaded as full result sets into memory, and all math should be performed in the mint or wallet, outside the storage layer.

This PR depends on #1957 and #1808 

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk-sql-common/src/mint/keyset_amounts.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/// Fetches the current `keyset_amounts` row for the given keyset, returning
+/// `None` if no row exists yet.
+async fn get_keyset_amounts<C>(conn: &C, keyset_id: &Id) -> Result<Option<KeysetAmounts>, Error>
+where
+    C: DatabaseExecutor + Send + Sync,
+{
+    let row = query(
+        r#"SELECT total_issued, total_redeemed, fee_collected FROM keyset_amounts WHERE keyset_id = :keyset_id"#,
```
thesimplekid: Don't we need to lock for update here or we will have a race condition?

crates/cdk-sql-common/src/mint/keyset_amounts.rs
```diff
@@ -0,0 +1,156 @@
...
+where
+    C: DatabaseExecutor + Send + Sync,
+{
+    let mut amounts = get_keyset_amounts(conn, keyset_id)
+        .await?
+        .unwrap_or(KeysetAmounts {
+            total_issued: 0,
+            total_redeemed: 0,
+            fee_collected: 0,
+        });
```
thesimplekid: I think if it does not exist we need to insert it first so we can lock it and avoid race conditions? https://github.com/cashubtc/cdk/pull/1954/changes#diff-07885d2738de414326d8f83768f8cf1bbf37bcf8299a9e95ca7079461caba845R24

crates/cdk-sql-common/src/mint/proofs.rs
```diff
@@ -212,20 +212,16 @@ where
...
-                    DO UPDATE SET total_redeemed = keyset_amounts.total_redeemed + EXCLUDED.total_redeemed
-                    "#,
-                )?
-                .bind_vec("ys", ys.iter().map(|y| y.to_bytes().to_vec()).collect())?
-                .execute(&self.inner)
-                .await?;
+            // Group proofs by keyset and sum amounts in Rust
+            let mut redeemed_by_keyset: HashMap<Id, u64> = HashMap::new();
+            for proof in proofs.iter() {
+                *redeemed_by_keyset.entry(proof.keyset_id).or_default() += u64::from(proof.amount);
```
thesimplekid: This can overflow, we should do check math.

crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -583,21 +583,18 @@ where
...
-                }
+                unpack_into!(let (amount)  = row);
+                let r: u64 = column_as_number!(amount);
+                Ok::<_, Error>(r)
             })
-            .transpose()?
-            .unwrap_or(0);
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
+            .sum();
```
thesimplekid: This can also overflow we should always used checked math.

crates/cdk-sql-common/src/wallet/migrations/sqlite/20260504000000_numeric_columns_to_text.sql
thesimplekid, line 1: I don't think this recreates all the indexes that we have?

crates/cdk-sql-common/src/wallet/migrations/postgres/20260504000000_numeric_columns_to_text.sql
```diff
@@ -0,0 +1,22 @@
+-- Migrate amount/fee columns to TEXT
+-- u64 amounts cannot be faithfully represented as PostgreSQL INTEGER/BIGINT (signed i64)
+-- Only amount and fee columns are migrated; timestamps and indices stay as INTEGER/BIGINT
+
+-- proof
+ALTER TABLE proof ALTER COLUMN amount TYPE TEXT USING amount::TEXT;
```
thesimplekid, line 6: For these columns we could add a check that they are numbers even as text. Though I'm also okay to just relay on rust for that.

thesimplekid (changes requested):
It seems this is still i64 should be u64, there might be others? https://github.com/crodas/cdk/blob/eecb267cff1768f00872a961afeaa5cc766992e8/crates/cdk-sql-common/src/wallet/mod.rs#L1801

crates/cdk-sql-common/src/wallet/migrations/postgres/20260504000000_numeric_columns_to_text.sql
crodas: Is there a way to enforce the SQL level? I'm not aware of that. Our rust codebase is quite flexible, though, when [reading a number off the database](https://github.com/cashubtc/cdk/blob/bdf9dc618ef963f540ef3db27f6da2f3a47856ee/crates/cdk-sql-common/src/macros.rs#L95-L116)

crates/cdk-sql-common/src/wallet/migrations/sqlite/20260504000000_numeric_columns_to_text.sql
crodas: Yes we do, for SQLite we do.

crates/cdk-sql-common/src/wallet/migrations/postgres/20260504000000_numeric_columns_to_text.sql
thesimplekid: I think it could be done with a check constraint on the column that probably uses a regex. 

> Is there a way to enforce the SQL level? I'm not aware of that. Our rust codebase is quite flexible, though, when [reading a number off the database](https://github.com/cashubtc/cdk/blob/bdf9dc618ef963f540ef3db27f6da2f3a47856ee/crates/cdk-sql-common/src/macros.rs#L95-L116)

It would fail at the row to struct level though I think even if it passes here. Which i think it probably okay, since we're also in control of what gets put in and we know thats a number?

crodas: > I think it could be done with a check constraint on the column that probably uses a regex.

I will explore this, it's a good idea.

> It would fail at the row to struct level though I think even if it passes here. Which i think it probably okay, since we're also in control of what gets put in and we know thats a number?


It is [explicit](https://github.com/cashubtc/cdk/blob/bdf9dc618ef963f540ef3db27f6da2f3a47856ee/crates/cdk-sql-common/src/wallet/mod.rs#L1804-L1809)  what kind of number should come from the storage layer, regardless of how it was stored. There are two main goals: never lose precision, and never ever do math outside of Rust.

crates/cdk-sql-common/src/wallet/mod.rs
crodas: https://github.com/cashubtc/cdk/pull/1152

crates/cdk-sql-common/src/mint/keyset_amounts.rs
crodas: This is fixed already, [the record exists and it is locked](https://github.com/cashubtc/cdk/pull/1954/changes#diff-07885d2738de414326d8f83768f8cf1bbf37bcf8299a9e95ca7079461caba845R68-R82).

thesimplekid:
Per the dev call we said it would be good to see benchmarks of performance before changing to text. Specifically https://github.com/cashubtc/cdk/pull/1152 was introduced to resolve performance bottlenecks we should keep in mind.

PR #1957, opened 2026-05-08
Add load_from_db to MintMetadataCache

crodas:
### Description

Populate the in-memory metadata cache from the database without HTTP requests. Useful for offline scenarios and fast startup. The load() method now tries load_from_db before falling back to load_from_mint.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/mint_metadata_cache.rs
+++ b/crates/cdk/src/wallet/mint_metadata_cache.rs
@@ -224,7 +224,6 @@ impl MintMetadataCache {
     /// // Force refresh from mint (ignores cache)
     /// let fresh = cache.load_from_mint(&storage, &client).await?;
     /// ```
-    #[inline(always)]
     pub async fn load_from_mint(
         &self,
         storage: &Arc<dyn WalletDatabase<database::Error> + Send + Sync>,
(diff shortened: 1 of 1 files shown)


crates/cdk/src/wallet/mint_metadata_cache.rs
```diff
@@ -245,27 +245,99 @@ impl MintMetadataCache {
...
+    ///
+    /// Metadata loaded from the database
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// // Load from database (no HTTP)
+    /// let metadata = cache.load_from_db(&storage).await?;
+    /// ```
+    #[inline(always)]
```
thesimplekid: We should not inline this.

thesimplekid (changes requested):
---
title: "Duplicate entries in active_keysets due to successive appends across cache updates"
notify: true
status: confirmed
severity: medium
target: cashubtc/cdk
pr: 1957
skills_used: ["rust-security"]
findings_count: 1
---

## Summary
The PR refactors the metadata cache logic in `MintMetadataCache` for `cashubtc/cdk`. Specifically, it introduces a pattern where `load_from_db` and `fetch_from_http` clone the existing cache object and then indiscriminately `.push()` active keyset records onto the copied `active_keysets` vector without clearing the vector first. Because the copied vector already contains existing active keysets from previous updates (either prior DB loads or earlier HTTP fetches), any newly fetched or re-loaded active keysets are appended, leading to unbounded duplication of the active keyset list. This bloats in-memory application metadata over time, particularly under scenarios where continuous connection checks or re-loads refresh the HTTP data.

## Skills Used
rust-security - Traced state buildup during ArcSwap state cloning over consecutive cache update passes. 

## Root Cause
In `crates/cdk/src/wallet/mint_metadata_cache.rs`, both `load_from_db` (lines 307-309) and `fetch_from_http` (lines 645-647) implement logic that appends to the `active_keysets` vector of a cloned state object (`new_metadata`):

```rust
        let mut new_metadata = (*self.metadata.load().clone()).clone();
...
            // Track active keysets separately for quick access
            if keyset_info.active {
                new_metadata.active_keysets.push(keyset_arc);
            }
```

Since `new_metadata` is initialized as a clone of the prior cache state, if the state already had `active_keysets` populated, those elements remain in `new_metadata.active_keysets`. When `fetch_from_http` subsequently queries the mint and loops over the response (`keysets_to_fetch`), it pushes `keyset_arc` corresponding to all returned active keysets onto `new_metadata.active_keysets`, thereby adding duplicates of any prior entries. The same applies for subsequent or competing `load_from_db` triggers.

## Attack Steps
1. The wallet performs regular application function, querying `load_from_db` explicitly or triggering loads as part of cache population checks. Let's assume there's 1 active keyset ("0xbeef") loaded. `active_keysets` length is 1.
2. A forced refresh operation, an authentication fetch, or another network routine triggers `load_from_mint` or `load_auth` leading to `fetch_from_http`.
3. `fetch_from_http` begins by cloning current cache, meaning `new_metadata.active_keysets` has the 1 entry ("0xbeef").
4. `keysets_to_fetch` retrieves the mint keysets via RPC, returning the key array including the same active key "0xbeef".
5. The `for keyset_info in keysets_to_fetch` loop notes the key is active and executes `new_metadata.active_keysets.push(keyset_arc);`
6. `new_metadata.active_keysets` vector now contains two references to the identical "0xbeef" keyset. Over multiple fetches or runtime routines, the vector grows boundlessly.

## Impact
In-memory dataset bloat due to an unbounded addition list in a frequently re-instantiated metadata state object. Components operating under assumptions about vector uniqueness (`active_keysets`) iterate multiple instances of the same ID, causing performance degradation when selecting active keys, fetching, mapping outputs during melting/minting operations, and overall larger memory surface. In the constrained execution environments like mobile applications integrating `cdk`, repeated operations balloon state causing potential Out of Memory (OOM) or DoS crashes for long-running services.

## Proof of Concept
Because this uses `ArcSwap` with standard data cloning, we can demonstrate the failure using a trivial standard library reproduction mirroring the internal logic directly:

```rust
use std::sync::Arc;
use std::collections::HashMap;

#[derive(Debug, Clone, PartialEq, Eq)]
struct KeySetInfo {
    id: String,
    active: bool,
}

#[derive(Debug, Clone)]
struct MintMetadata {
    keysets: HashMap<String, Arc<KeySetInfo>>,
    active_keysets: Vec<Arc<KeySetInfo>>,
}

fn main() {
    let mut initial = MintMetadata {
        keysets: HashMap::new(),
        active_keysets: Vec::new(),
    };
    
    let info = Arc::new(KeySetInfo {
        id: "1".into(),
        active: true,
    });
    
    // Simulate first load adding the key to vec and hashmap
    initial.keysets.insert("1".into(), info.clone());
    initial.active_keysets.push(info.clone());
    
    // Simulate subsequent HTTP fetch logic
    let info2 = Arc::new(KeySetInfo {
        id: "2".into(),
        active: true,
    });
    let http_fetched_keysets = vec![info.as_ref().clone(), info2.as_ref().clone()];
    
    // Start with current cache clone
    let mut new_metadata = initial.clone();
    
    // THE FIX: We'd normally want to handle clear here, but we don't.
    // In fact we shouldn't even clone the active_keyset if we are re-calculating it over keysets
    // Here we reproduce the exact same flow.
    for keyset_info in http_fetched_keysets {
        let keyset_arc = Arc::new(keyset_info.clone());
        // keysets gets overwritten successfully because it's a Map
        new_metadata.keysets.insert(keyset_info.id.clone(), keyset_arc.clone());
        
        // BUG: active_keysets gets blindly pushed to without clear, duplicating the array content
        if keyset_info.active {
            new_metadata.active_keysets.push(keyset_arc);
        }
    }
    
    assert_eq!(new_metadata.keysets.len(), 2); 
    assert_eq!(new_metadata.active_keysets.len(), 3); // Uniqueness invariant violated (1 becomes 3 instead of 2)
    println!("Duplication confirmed.");
}
```

crates/cdk/src/wallet/mint_metadata_cache.rs
```diff
@@ -574,6 +634,7 @@ impl MintMetadataCache {
         );
 
         // Fetch keys for each keyset
+        new_metadata.active_keysets.clear();
```
thesimplekid, line 637: This fixes the duplicate growth issue, but I think there is a follow-up edge case if the same metadata cache is used for both auth and non-auth keysets. The current call paths invoke `fetch_from_http` with either `Some(client), None` from `load_from_mint` or `None, Some(auth_client)` from `load_auth`, so this clears `active_keysets` globally and then only repopulates it for the path that was called. Some auth/non-auth setups appear to use separate caches, but there are also paths that pass `self.metadata_cache.clone()` into the `AuthWallet`.

I don't think this needs to block the duplicate-growth fix, but we should revisit whether `active_keysets` should be rebuilt from the full merged `new_metadata.keysets` map, or split by auth/non-auth state.

thesimplekid (approved):

thesimplekid:
Would also be good to increase test coverage around this. https://github.com/cashubtc/cdk/pull/1957#issuecomment-4409800710

PR #1958, opened 2026-05-10
Fix: Wallet can get mint info without specify `mint_url`

TheMhv:
### Description

Accept HTTP proxy on the CLI and ensure that the Wallet's can get mint info without specify `mint_url`.

-----

### Notes to the reviewers

- 0a88ae65c98fe392dd624f3ec20ca448f4b8fbc2:  Get mint info from wallets on `cdk-cli mint-info` command without needing to specifying `mint_url` and define `danger_accept_invalid_certs` from `Wallet Repository` as public.

-----

#### CHANGED
- `cdk-cli mint-info` command
- Mark `danger_accept_invalid_certs` as public

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/mint_metadata_cache.rs
+++ b/crates/cdk/src/wallet/mint_metadata_cache.rs
@@ -328,6 +328,7 @@ impl MintMetadataCache {
     /// // Load from database (no HTTP)
     /// let metadata = cache.load_from_db(&storage).await?;
     /// ```
+    #[inline(always)]
     async fn load_from_db(
         &self,
         storage: &Arc<dyn WalletDatabase<database::Error> + Send + Sync>,
--- a/crates/cdk-cli/src/sub_commands/mint_info.rs
+++ b/crates/cdk-cli/src/sub_commands/mint_info.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use anyhow::Result;
 use cdk::mint_url::MintUrl;
 use cdk::wallet::WalletRepository;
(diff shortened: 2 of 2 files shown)


crodas:
Can you please rebase it please @TheMhv ?

crates/cdk/src/wallet/mint_metadata_cache.rs
```diff
@@ -328,6 +328,7 @@ impl MintMetadataCache {
     /// // Load from database (no HTTP)
     /// let metadata = cache.load_from_db(&storage).await?;
     /// ```
+    #[inline(always)]
```
crodas: I think this is of little help

crodas:
I think the description of this PR should be renamed to "accepting an HTTP proxy on the CLI". Once that is done, I will approve it

crodas (dismissed):

thesimplekid (changes requested):
155dcf572fb4759ffbe5c170c5819797c1e9fc9f and 5fed26467a02330c69ce1269d19a04d7ea926ed8 look like stray commits that need to be removed. 

This comment is also correct but is not addressed https://github.com/cashubtc/cdk/pull/1958#discussion_r3564502371. There are multiple IO await points inside this fn so inline does not help but its a private fn anyway so the compiler can choose to inline it if it wants. In-lining of async fns also does not do much as it just inlines the future construction not the full fn body. 

This comment is also still outstanding https://github.com/cashubtc/cdk/pull/1958#pullrequestreview-4602227914

The title of the PR also needs to be updated to reflect what this PR actually does.

crates/cdk/src/wallet/mint_metadata_cache.rs
thesimplekid: This is not resolved

thesimplekid:
A few specific issues remain on the current head:

- [`crates/cdk-cli/src/sub_commands/mint_info.rs:19`](https://github.com/cashubtc/cdk/blob/ebd287f2b17f79bb0d5097fba1e2b6d38ee14e4e/crates/cdk-cli/src/sub_commands/mint_info.rs#L19-L26): Could we keep using `wallet_repository.fetch_mint_info(mint_url)` here? The repository is already configured with the CLI proxy, TLS setting, and optional Tor transport. Constructing an `HttpClient` directly duplicates that configuration and, in a Tor-enabled build with `--tor on` and no proxy, sends this request through the default HTTP transport instead of the repository's shared Tor transport. Reusing `fetch_mint_info` would also make the new `proxy` argument and public TLS field unnecessary.

- [`crates/cdk-cli/src/sub_commands/mint_info.rs:33`](https://github.com/cashubtc/cdk/blob/ebd287f2b17f79bb0d5097fba1e2b6d38ee14e4e/crates/cdk-cli/src/sub_commands/mint_info.rs#L33-L47): Should we preserve this fetch error and return it when no cached mint info can be loaded? If `get_wallets_for_mint` returns an empty list, or every `load_mint_info` call fails, this command currently returns `Ok(())` and exits successfully despite producing no mint info. The cache fallback is useful, but the command should still fail when both live and cached data are unavailable.

- [`crates/cdk-cli/src/sub_commands/mint_info.rs:50`](https://github.com/cashubtc/cdk/blob/ebd287f2b17f79bb0d5097fba1e2b6d38ee14e4e/crates/cdk-cli/src/sub_commands/mint_info.rs#L50-L62): Should this iterate over unique mint URLs instead of every wallet? `WalletRepository` stores one wallet per `(mint_url, currency_unit)`, so a mint supporting multiple units will print the same mint information more than once.

- [`crates/cdk/src/wallet/wallet_repository.rs:291`](https://github.com/cashubtc/cdk/blob/ebd287f2b17f79bb0d5097fba1e2b6d38ee14e4e/crates/cdk/src/wallet/wallet_repository.rs#L291): Can we avoid exposing this security-related configuration as a public mutable field? It is only needed because `mint_info` reconstructs the HTTP client. Calling `WalletRepository::fetch_mint_info` would preserve the existing encapsulation and transport behavior without expanding the public API.

I have not repeated the existing unresolved `#[inline(always)]` thread.

thesimplekid:
@cdk-bot review

TheMhv:
@cdk-bot review

TheMhv:
@cdk-bot review

TheMhv:
@cdk-bot review

TheMhv:
@cdk-bot review

thesimplekid (approved):

PR #1959, opened 2026-05-10
Make connection pool fully async to prevent tokio runtime starvation

crodas:
### Description

Replace `std::sync::Condvar` with `tokio::sync::Notify` in the connection pool. The blocking `Condvar::wait_timeout` was starving the tokio runtime when pool size=1 (in-memory SQLite), causing 5s cascading timeouts on concurrent ops.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sql-common/src/keyvalue.rs
+++ b/crates/cdk-sql-common/src/keyvalue.rs
@@ -162,7 +162,7 @@ where
     // Validate parameters according to KV store requirements
     validate_kvstore_params(primary_namespace, secondary_namespace, Some(key))?;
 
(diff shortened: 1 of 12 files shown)


thesimplekid:
Does this replace https://github.com/cashubtc/cdk/pull/1808? It seems pretty similar what are the differences that we don't just finish that PR?

crodas:
> Does this replace https://github.com/cashubtc/cdk/pull/1808? It seems pretty similar what are the differences that we don't just finish that PR?

I will finish #1808. This is what I did to pass tests on #1954, but I will close this in favor of #1808

Issue #1933, opened 2026-04-23
get_blind_signatures_for_quote returns non-deterministic order


gudnuf:
When a wallet checks a completed melt quote via GET /v1/melt/quote/bolt11/{id}, the change field contains BlindSignatures loaded from the blind_signature table without an ORDER BY clause. 

Also, it seems like maybe  `get_melt_request_and_blinded_messages` in quotes.rs fetches the unsigned blinded messages from DB during saga recovery (mint crash/restart). These B_ values are then zipped with denomination amounts and signed. If they come back out of order, the mint assigns the wrong denomination to the wrong B_, producing mismatched signatures at the source. This means even the POST response can have wrong pairings after a saga recovery.

The overall issue that when reading from the db the ordering of outputs is not guaranteed and wallets expect the outputs to be in the same order as the blinded messages sent to the mint

thesimplekid:
Nutshell fixed in https://github.com/cashubtc/nutshell/pull/991

orbitalturtle:
Since this issue is unassigned, I'm interested in tackling it.

PR #1961, opened 2026-05-12
fix(mint): preserve change promises order in paid melt quotes

a1denvalu3:
## Summary
- Added `order_index` to `blind_signature` table to preserve the insertion order of change outputs.
- Updated database queries to use `ORDER BY order_index` when retrieving blinded messages/signatures for melt quotes.
- Ensures deterministic order for change signatures in paid melt quotes, fixing non-deterministic behavior found in nutshell (https://github.com/cashubtc/nutshell/pull/991).

### DIFF
--- a/crates/cdk-common/src/database/mint/test/signatures.rs
+++ b/crates/cdk-common/src/database/mint/test/signatures.rs
@@ -119,8 +119,8 @@ where
     let quote_id1 = QuoteId::new_uuid();
     let quote_id2 = QuoteId::new_uuid();
 
-    // Create signatures for quote 1
     let blinded_message1 = SecretKey::generate().public_key();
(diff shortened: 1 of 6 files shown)


crodas (dismissed):
LGTM

thesimplekid:
Fixes https://github.com/cashubtc/cdk/issues/1933

thesimplekid:
While in we add the the message and then the signature so this fix would be enough in practice. We do allow inserting a signature if the message does not exist first, so to be complete should we add the ordering there as well https://github.com/cashubtc/cdk/blob/8ce06b1a8620e5c584f12a7733e7bc072ec45ea7/crates/cdk-sql-common/src/mint/signatures.rs#L95-L137.

thesimplekid (approved):

PR #1962, opened 2026-05-13
check and merge config from toml and env

asmogo:
### Description
-----
Adds upfront, fail-fast validation of cdk-mintd settings after TOML + env-var merge, so misconfiguration produces a clear error pointing at both the offending TOML key and env var instead of a silent default or a confusing downstream crash.
### Notes to the reviewers

-----
  - New `validate_settings()` covers listen address, signing source (mnemonic / seed length / signatory URL), per-backend lightning config, database, auth, management-RPC, and Prometheus.
  - `#[serde(default)]` added across config structs so partial TOML files merge cleanly with defaults and env vars.
  - `Settings::new` split into try_new (Result) + new (panics on any error) — removes the old silent-fallback-to-defaults footgun.
  - Replaces an .expect() panic in Prometheus server startup with a propagated error.
  - Adds a shared test-utils module (env-mutex + unique temp dirs) and a full set of negative tests for every validator branch.
#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-mintd/example.config.toml
+++ b/crates/cdk-mintd/example.config.toml
@@ -44,7 +44,7 @@ enabled = false
 #[prometheus]
 #enabled = true
 #address = "127.0.0.1"
-#port = 9090
+#port = 9000
 # 
 [info.http_cache]
 # memory or redis
--- a/crates/cdk-mintd/src/config.rs
+++ b/crates/cdk-mintd/src/config.rs
@@ -47,6 +47,7 @@ pub struct LoggingConfig {
 }
 
 #[derive(Clone, Serialize, Deserialize)]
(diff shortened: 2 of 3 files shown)


orangeshyguy21:
> Replaces an `.expect()` panic in Prometheus server startup with a propagated error.

I couldn't find that in the diff. `.expect("Invalid prometheus address")` in the
Prometheus startup still looks present. 

The new `validate_prometheus_config` seems to cover the address
up front, but in a way that doesn't quite match the desc.

asmogo:
@cdk-bot review

asmogo:
@cdk-bot please review the latest rebase.

thesimplekid (approved):

PR #1970, opened 2026-05-15
fix ffi send options condition errors

vnprc:
### Description

FFI `SendOptions.conditions` previously used `.try_into().ok()`, which silently converted malformed spending conditions into `None`. That could cause an FFI caller intending to create a locked send to accidentally create an unlocked send instead.

This now matches the existing fallible conversion pattern used by `ReceiveOptions`.

-----

### Notes to the reviewers

This removes the infallible `From<cdk_ffi::SendOptions> for cdk::wallet::SendOptions` impl and replaces it with `TryFrom`. That is intentional: invalid FFI send conditions must be observable as errors rather than being dropped. FFI method signatures are unchanged.

-----

#### CHANGED
- make FFI `SendOptions` conversion to `cdk::wallet::SendOptions` fallible
- propagate invalid send condition errors from `Wallet::prepare_send`

#### ADDED
- add regression coverage for invalid P2PK/HTLC send conditions and the FFI wallet prepare-send path

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-ffi/src/lib.rs
+++ b/crates/cdk-ffi/src/lib.rs
@@ -219,6 +219,72 @@ mod tests {
     }
 
     #[test]
+    fn test_send_options_invalid_p2pk_conditions_returns_error() {
+        let options = SendOptions {
+            conditions: Some(SpendingConditions::P2PK {
+                pubkey: "not-a-pubkey".to_string(),
+                conditions: None,
+            }),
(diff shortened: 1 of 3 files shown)


thesimplekid:
Thanks good catch but looks like I missed this and fixed it in https://github.com/cashubtc/cdk/pull/2002

PR #1972, opened 2026-05-18
Add Swift publication pipeline workflow

crodas:
### Description

Here is an [example](https://github.com/crodas/cdk/actions/runs/26013120509) that creates a mock release at [this test repo](https://github.com/crodas/cdk-swift)

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


bindings/swift/rust/Cargo.toml
```diff
@@ -18,7 +18,7 @@ name = "uniffi-bindgen-swift"
 path = "uniffi-bindgen-swift.rs"
 
 [dependencies]
-cdk-ffi.workspace = true
+cdk-ffi = { workspace = true, features = ["npubcash", "bip353"] }
```
thesimplekid: We should default features off like we do for kotlin.

.github/workflows/swift-publish.yml
```diff
@@ -0,0 +1,423 @@
...
+
+jobs:
+  swift-sync-sources:
+    name: "swift: Sync sources to cdk-swift"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout monorepo
+        uses: actions/checkout@v4
+        with:
+          ref: main
```
thesimplekid: I know this is `main` for tests, but so we do not forget we should change it before merging to match the Kotlin flow from #1939:

`ref: ${{ inputs.cdk_ref || inputs.release_tag }}`

and add the `cdk_ref` workflow input. Otherwise the Swift release can package whatever is currently on `main` instead of the exact CDK tag that was published to crates.io.

justfile
```diff
@@ -908,13 +908,6 @@ test-kotlin:
   ./gradlew :cdk-jvm:test
 
 
-# Generate Swift FFI bindings and XCFramework
-binding-swift:
```
thesimplekid, line 915: This removes `binding-swift`, but CI still calls it in `.github/workflows/ci.yml`. Should we either keep a local recipe for the existing CI/dev validation path, or update the Swift binding test job to use the replacement command? As written, the CI job will fail with an unknown just recipe.

.github/workflows/swift-publish.yml
```diff
@@ -0,0 +1,423 @@
...
+          TAG="${{ inputs.release_tag }}"
+          RELEASE_BRANCH="release/v${TAG#v}"
+          DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+          echo "DEFAULT_BRANCH=${DEFAULT_BRANCH}" >> "$GITHUB_ENV"
+          echo "RELEASE_BRANCH=${RELEASE_BRANCH}" >> "$GITHUB_ENV"
+          git checkout "${DEFAULT_BRANCH}"
+          git merge "${RELEASE_BRANCH}" --squash
+          git commit -m "release: ${TAG}"
+
+      - name: Push and tag
```
thesimplekid: Should we only push the default branch/tag after the GitHub release asset has been created? This currently pushes the package commit and tag before `gh release create` uploads `CashuDevKitFFI.xcframework.zip`. If release creation fails, the tag points at a `Package.swift` URL for an asset that does not exist, making the release harder to retry cleanly.

thesimplekid:
One more release-flow item that GitHub will not let me attach inline because the `justfile` trigger is not part of this PR diff: should `ffi-release-swift` trigger the new monorepo workflow like the Kotlin flow from #1939?

`just release` calls `ffi-release-swift`, but that recipe still invokes `"Publish Swift Package"` in `cashubtc/cdk-swift` with the old `version`/`cdk_repo`/`cdk_ref` inputs. The new workflow is `"FFI - Swift Bindings"` in `cashubtc/cdk` and expects `release_tag`/`cdk_version`/`cdk_ref`, so the standard release path will not run this workflow as written.

.github/workflows/swift-publish.yml
crodas: > and add the cdk_ref workflow input. Otherwise the Swift release can package whatever is currently on main instead of the exact CDK tag that was published to crates.io.

This is mainly just for the boilerplate (the wrapper itself) since the CI would import the cdk and cdk-ffi from cargo. But I agree with you

.github/workflows/swift-publish.yml
crodas: I addressed it. I also created a tiny cli wallet which installs from github (without any rustc) and it works as expected

crodas:
> One more release-flow item that GitHub will not let me attach inline because the `justfile` trigger is not part of this PR diff: should `ffi-release-swift` trigger the new monorepo workflow like the Kotlin flow from #1939?
> 
> `just release` calls `ffi-release-swift`, but that recipe still invokes `"Publish Swift Package"` in `cashubtc/cdk-swift` with the old `version`/`cdk_repo`/`cdk_ref` inputs. The new workflow is `"FFI - Swift Bindings"` in `cashubtc/cdk` and expects `release_tag`/`cdk_version`/`cdk_ref`, so the standard release path will not run this workflow as written.

My goal is to have a single release that would trigger dart, swift and kotlin. All builds would happen in the same way, from Github; fetching the crates from crates.io.

It can be initiated from Github's UI or our just script. That would be a follow up PR, for instance each target language have their own Github Token to write to their repo; I will make them use a single secret

thesimplekid:
> > One more release-flow item that GitHub will not let me attach inline because the `justfile` trigger is not part of this PR diff: should `ffi-release-swift` trigger the new monorepo workflow like the Kotlin flow from #1939?
> > `just release` calls `ffi-release-swift`, but that recipe still invokes `"Publish Swift Package"` in `cashubtc/cdk-swift` with the old `version`/`cdk_repo`/`cdk_ref` inputs. The new workflow is `"FFI - Swift Bindings"` in `cashubtc/cdk` and expects `release_tag`/`cdk_version`/`cdk_ref`, so the standard release path will not run this workflow as written.
> 
> My goal is to have a single release that would trigger dart, swift and kotlin. All builds would happen in the same way, from Github; fetching the crates from crates.io.

We should have a CI job that is triggered by a GH release, that starts all these jobs. This is the same trigger we currently use for our docker builds. 

> It can be initiated from Github's UI or our just script. That would be a follow up PR, for instance each target language have their own Github Token to write to their repo; I will make them use a single secret

We can use a PAT from the cdk-bot account that has access to all the repos. That's okay as a follow up but we shouldn't leave in a broken just command in this PR to be removed later.

crodas:
> We should have a CI job that is triggered by a GH release, that starts all these jobs. This is the same trigger we currently use for our docker builds.

I want to create a single CI job that would trigger all the builds in parallel; dart, swift and kotlin. That should be triggered from just. I can do that right now.

PR #1973, opened 2026-05-18
feat(payment-processor): propagate mint quote_id to backend

zeugmaster:
### Description

Adds the mint's QuoteId to the three OutgoingPaymentOptions struct variants so custom payment-method backends can deterministically correlate their `get_payment_quote` and `make_payment` calls for the same melt.

Today, the gRPC payment-processor proto carries no link between those two calls for the same melt: a backend's `get_payment_quote` returns a PaymentQuoteResponse whose `request_lookup_id` the mint stores on the quote, but the eventual `make_payment` receives an OutgoingPaymentOptions with neither the stored lookup_id nor the mint's quote_id (`from_melt_quote_with_fee` in cdk-common drops `extra_json` and has no quote_id field). bolt11/bolt12 backends are insulated by the payment_hash / offer being intrinsically derivable from the invoice in both calls. Custom backends have no such anchor, and have to fabricate a derivation — typically hashing `request` — which forces the wallet to make `request` unique per melt. That's awkward for any custom rail whose "request" doesn't naturally carry uniqueness (branch tickets, account IDs, free-text references). With `quote_id` plumbed through, the backend has the same stable key the mint uses everywhere else, so correlation is trivial.

Changes:
- cdk-common::payment: add `quote_id: QuoteId` to Bolt11/Bolt12/Custom OutgoingPaymentOptions structs. Propagate it in `from_melt_quote_with_fee` from `melt_quote.id`.
- cdk::mint::melt: pre-generate the quote_id before each `get_payment_quote` call (bolt11, bolt12, custom paths) and pass it both to the backend and to `MeltQuote::new`, replacing the previous implicit auto-generation.
- payment_processor.proto: add `optional string quote_id` to Bolt11/Bolt12/CustomOutgoingPaymentOptions and PaymentQuoteRequest. Field-numbers appended (5/6/7/6) to preserve wire compatibility with the existing fields. `optional` is used so the server can distinguish "not sent" from "sent empty" and emit a clear error for old clients.
- cdk-payment-processor server/client: serialize/deserialize quote_id. The server treats absence as a protocol error (defense-in-depth backing the version check, since the optional proto field would otherwise default silently).
- cdk-common: bump PAYMENT_PROCESSOR_PROTOCOL_VERSION 2.0.0 -> 3.0.0. The VersionInterceptor's strict equality check means any old mint/backend pair will see a clear "Protocol version mismatch" at connection time instead of misbehaving silently.
- cdk-integration-tests: extend the existing custom-payment-processor fixture to capture the received quote_id, and add a regression test asserting it matches the quote_id surfaced to the wallet.

The Rust API is intentionally non-optional (`quote_id: QuoteId`, not `Option<QuoteId>`): every legitimate caller site has the id available. The optional-ness lives at the proto/wire layer for explicit version detection.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

* [ ] ~If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`)~

### DIFF
--- a/crates/cdk-common/src/lib.rs
+++ b/crates/cdk-common/src/lib.rs
@@ -16,7 +16,7 @@ pub mod auth;
 pub const MINT_RPC_PROTOCOL_VERSION: &str = "1.0.0";
 
 /// Protocol version for gRPC Payment Processor communication
-pub const PAYMENT_PROCESSOR_PROTOCOL_VERSION: &str = "2.0.0";
+pub const PAYMENT_PROCESSOR_PROTOCOL_VERSION: &str = "3.0.0";
 
 #[cfg(feature = "grpc")]
 pub mod grpc;
--- a/crates/cdk-common/src/payment.rs
+++ b/crates/cdk-common/src/payment.rs
@@ -251,6 +251,11 @@ pub struct Bolt11OutgoingPaymentOptions {
     pub timeout_secs: Option<u64>,
     /// Melt options
     pub melt_options: Option<MeltOptions>,
+    /// The mint's quote id for this melt. Set in both `get_payment_quote`
+    /// and `make_payment` so backends can correlate the two calls. For
+    /// BOLT11 backends the payment_hash already provides correlation, so
+    /// this is informational; it is still required for protocol uniformity.
+    pub quote_id: QuoteId,
 }
 
 /// Options for BOLT12 outgoing payments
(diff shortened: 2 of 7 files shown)


thesimplekid:
Should we just make the quote id required? It would be a breaking change but I think the quote id is a better look up id then what we do now and payment processor should be at least encouraged to use that if not forced. We don't have too many processor right now so updating wouldn't be too hard, though that do not use the quote id like the existing bolt11 ones (cln, lnd, lnbits) can just ignore the extra field. 

> The optional-ness lives at the proto/wire layer for explicit version detection.

We error if the payment processor and mint are different versions anyway so I think making this required is okay?

zeugmaster:
Agreed and updated, quote id is now required

asmogo:
ACK 2e09b93cf09d22fd90cfaff96f758adaefc7f5e8

thesimplekid (approved):

PR #1979, opened 2026-05-19
Move XCFramework build to CI and streamline local dev workflow

crodas:
I'm reopening #1972 since my local repo was messed up and Github closed that ticket

Replace the monolithic generate-bindings.sh script with a CI-driven swift-publish workflow that builds the XCFramework and syncs sources to the cdk-swift distribution repo. This separates the heavy cross-compilation (iOS device, iOS simulator, macOS universal) from the local development loop.

Local changes:
- Rework `ffi-generate-swift` justfile recipe to set up an SPM-compatible package structure (Sources/Cdk, Sources/cdk_ffiFFI) using the dylib from `ffi-build`, enabling `swift test` without building XCFrameworks
- Add `DYLD_LIBRARY_PATH` to CI and `test-swift` so the dynamic library is found at runtime
- Gate bip353 FFI exports behind `feature = "bip353"` to allow the Swift FFI crate to opt out of features that pull in heavy DNS dependencies
- Pin cdk-ffi dependency in Swift crate to only `npubcash` and `bip353` features (no default features) for a leaner build
- Add Swift SPM generated artifacts to .gitignore
- Remove the now-unused `binding-swift` justfile target
- Add bindings/swift/README.md with SPM and Xcode installation docs
- Update bindings/README.md to reflect the new CI-driven build flow

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


.github/workflows/swift-publish.yml
```diff
@@ -0,0 +1,421 @@
...
+env:
+  CARGO_TERM_COLOR: ${{ vars.CARGO_TERM_COLOR }}
+
+jobs:
+  swift-sync-sources:
+    name: "swift: Sync sources to cdk-swift"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout monorepo
+        uses: actions/checkout@v4
```
thesimplekid, line 33: Should this follow the Kotlin release workflow and checkout an explicit CDK ref? In `kotlin-publish.yml`, the workflow has a `cdk_ref` input and uses it for each `actions/checkout`, so the release builds from `vX.Y.Z` instead of whatever branch dispatched the workflow.

Here, `just ffi-release-swift` passes only `release_tag` and `cdk_version`, and this checkout uses the workflow dispatch ref. If the workflow is run from `main`, we can publish `vX.Y.Z` Swift artifacts built from `main`.

Could we add the same `cdk_ref` input and use it here?

```yaml
with:
  ref: ${{ inputs.cdk_ref || inputs.release_tag }}
```

.github/workflows/swift-publish.yml
```diff
@@ -0,0 +1,421 @@
...
+          TAG="${{ inputs.release_tag }}"
+          RELEASE_BRANCH="release/v${TAG#v}"
+          DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+          echo "DEFAULT_BRANCH=${DEFAULT_BRANCH}" >> "$GITHUB_ENV"
+          echo "RELEASE_BRANCH=${RELEASE_BRANCH}" >> "$GITHUB_ENV"
+          git checkout "${DEFAULT_BRANCH}"
+          git merge "${RELEASE_BRANCH}" --squash
+          git commit -m "release: ${TAG}"
+
+      - name: Push default branch (without tag)
```
thesimplekid: Should this mirror the Kotlin publish ordering? In `kotlin-publish.yml`, the workflow publishes to Maven Central before pushing the default branch and tag, which keeps retries cleaner if the actual package publish fails.

Here the default branch is pushed before `gh release create` uploads the XCFramework. If release creation or upload fails, retrying the workflow leaves the target repo partially published and may require manual cleanup. Could we delay pushing the default branch until after the release asset is created, or make the release step idempotent for an existing commit/tag?

.github/workflows/swift-publish.yml
thesimplekid: I don't think this is resolved yet. The workflow now has a `cdk_ref` input, but the checkout still falls back to `github.ref` instead of `inputs.release_tag`:

```yaml
ref: ${{ inputs.cdk_ref || github.ref }}
```

Also `just ffi-release-swift` still does not pass `cdk_ref`, unlike the Kotlin helper. So a release dispatched from `main` can still publish artifacts for `release_tag` while building from `main`. I think this should match Kotlin and use `inputs.cdk_ref || inputs.release_tag`, with the helper passing `cdk_ref="v{{VERSION}}"`.

.github/workflows/swift-publish.yml
```diff
@@ -0,0 +1,429 @@
...
+          git checkout "${RELEASE_BRANCH}"
+          git push origin "${RELEASE_BRANCH}"
+
+      - name: Create release on cdk-swift (creates tag and uploads asset)
+        env:
+          GH_TOKEN: ${{ secrets.SWIFT_DEPLOY_KEY }}
+        run: |
+          gh release create "${{ inputs.release_tag }}" \
+            --repo ${{ vars.CDK_SWIFT_REPO }} \
+            --target "${DEFAULT_BRANCH}" \
```
thesimplekid: Should this target a ref that already contains the generated `Package.swift` and Swift sources?

At this point the local default branch has the squash commit from `git merge "${RELEASE_BRANCH}" --squash`, but it has not been pushed yet. `gh release create --target "${DEFAULT_BRANCH}"` resolves against the remote branch, so the release tag can be created from the old default-branch tip. SwiftPM would then resolve `vX.Y.Z` to package contents that do not include the generated sources/checksum for this release.

Could we either push the default branch before creating the release, or create the release tag against a pushed release ref/commit that already contains these files? The Kotlin flow avoids partially pushed refs by publishing first and pushing tags/branches during cleanup; for Swift we need both retry-safe ordering and a tag target that contains the package contents.

thesimplekid (commented):
Adding one inline release-flow comment.

crodas:
Here are the latest changes in action https://github.com/crodas/cdk/actions/runs/26258736160/job/77289141849 and here is the release it created:

https://github.com/crodas/cdk-swift

I set #1981 to Draft until this PR is merged.

.github/workflows/swift-publish.yml
crodas: > Should this target a ref that already contains the generated Package.swift and Swift sources?

**_Package.swift_** is generated in a [previous step](https://github.com/cashubtc/cdk/pull/1979/changes#diff-0bac2fb6d09e489382e66fca9491302a296abe1f144d280eec65a953e482ce03R358-R384)

.github/workflows/swift-publish.yml
crodas: I think it's fixed now. In #1981 I will set all CI files using this as the reference (same variable names and similar steps using github envs)

.github/workflows/swift-publish.yml
thesimplekid: Right, `Package.swift` is generated earlier in the job. The concern here was whether the published SwiftPM tag resolves to a commit that contains that generated `Package.swift` and the generated Swift sources, since SwiftPM reads the package contents from the tag, not from the CI workspace.

Looking at the latest version, this tag-target part now appears fixed: the workflow creates the local default-branch squash commit, checks out that commit, tags it, and pushes the default branch plus tags. That should make `vX.Y.Z` resolve to package contents with the generated files.

The remaining concern is the separate retry-safety issue from the earlier thread: the branch/tag are pushed before `gh release create` uploads the XCFramework asset, so a release upload failure can still leave a tag whose `Package.swift` points to a missing binary asset.

crodas: Good catch on the retry-safety gap. The fix is to not push the tag separately; instead let `gh release create --target "${DEFAULT_BRANCH}"` create the tag atomically with the asset upload. Since we now push the default branch before creating the release, --target resolves to the right commit. If the release upload fails, no tag is created, so it's safe to retry.

I think 9452f39de536233235ee7b608746d5bd0cc1fa7e addressed that comment. I also push the change towards all other targets in #1981

.github/workflows/swift-publish.yml
```diff
@@ -0,0 +1,423 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.SWIFT_DEPLOY_KEY }}
+        run: |
+          COMMIT_SHA=$(git rev-parse HEAD)
+          gh release create "${TAG}" \
+            --repo ${{ vars.CDK_SWIFT_REPO }} \
+            --target "${COMMIT_SHA}" \
+            --title "Release ${TAG}" \
+            --generate-notes \
+            artifacts/CashuDevKitFFI.xcframework.zip
```
thesimplekid: Did you run this exact version of the workflow after adding `working-directory: cdk-swift` to this release step? I’m not sure this asset path is correct anymore.

The XCFramework artifact is downloaded to the workspace root with `path: ./artifacts`, but this step now runs from inside `cdk-swift`, so I think `artifacts/CashuDevKitFFI.xcframework.zip` resolves to `cdk-swift/artifacts/CashuDevKitFFI.xcframework.zip`. From this working directory, shouldn’t this be `../artifacts/CashuDevKitFFI.xcframework.zip`, or should the release step run from the workspace root?

thesimplekid (commented):
Left one inline question on the Swift release asset path.

.github/workflows/swift-publish.yml
crodas: > Did you run this exact version of the workflow after adding working-directory: cdk-swift to this release step? I’m not sure this asset path is correct anymore.

This version exactly generated the output I linked, I wonder how it worked. Anyways, I just pushed a fix and here is the last build:

https://github.com/crodas/cdk/actions/runs/26376185453

https://github.com/crodas/cdk-swift

https://github.com/crodas/cdk-swift/releases/download/v99.99.1/CashuDevKitFFI.xcframework.zip

crodas: https://github.com/crodas/cdk-swift/blob/main/Package.swift

```
➜  Downloads sha256 CashuDevKitFFI.xcframework.zip
SHA256 (CashuDevKitFFI.xcframework.zip) = db328c4b0f8ca254badd442b107f14845c9da857fc886c20d62160af586ea511
```

.github/workflows/swift-publish.yml
crodas: I addressed this in cff41bc73fbc76bd548e4fbc228003694a3fa28a

Here is the latest build https://github.com/crodas/cdk/actions/runs/26413124495

thesimplekid (approved):

PR #1981, opened 2026-05-19
Feature/ffi improvements

crodas:
### Description

- Remove postgres from cdk-ffi default features
- Remove --features postgres from justfile ffi-build and ffi-generate
- Remove --features postgres from flake.nix FFI build
- Consolidate DART_DEPLOY_KEY and KOTLIN_DEPLOY_KEY into single FFI_DEPLOY_KEY

This PR depends on #1979

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/dart-publish.yml
+++ b/.github/workflows/dart-publish.yml
@@ -9,23 +9,45 @@ on:
       cdk_version:
         description: 'CDK dependency version (e.g. 0.17.0). Defaults to release_tag without the "v" prefix.'
         required: false
+      cdk_ref:
+        description: 'CDK git ref to publish from. Defaults to release_tag.'
+        required: false
+  workflow_call:
+    inputs:
+      release_tag:
+        type: string
+        required: true
+      cdk_version:
+        type: string
+        required: false
+      cdk_ref:
+        type: string
+        required: false
 env:
   CARGO_TERM_COLOR: ${{ vars.CARGO_TERM_COLOR }}
-
+  TAG: ${{ inputs.release_tag }}
+  CDK_REF: ${{ inputs.cdk_ref || inputs.release_tag }}
+  RELEASE_BRANCH: release/${{ inputs.release_tag }}
 
 jobs:
   dart-sync-sources:
     name: "dart: Sync sources to cdk-dart"
     runs-on: ubuntu-latest
     steps:
+      - name: Resolve variables
+        run: |
+          CDK_VER="${{ inputs.cdk_version }}"
+          echo "CDK_VER=${CDK_VER:-${TAG#v}}" >> "$GITHUB_ENV"
+          echo "RELEASE_VER=${TAG#v}" >> "$GITHUB_ENV"
+
       - name: Checkout monorepo
         uses: actions/checkout@v4
         with:
-          ref: main
+          ref: ${{ env.CDK_REF }}
 
       - name: Clone cdk-dart
         run: |
-          git clone https://x-access-token:${{ secrets.DART_DEPLOY_KEY }}@github.com/${{ vars.CDK_DART_REPO }}.git cdk-dart
+          git clone https://x-access-token:${{ secrets.FFI_DEPLOY_KEY }}@github.com/${{ vars.CDK_DART_REPO }}.git cdk-dart
 
       - name: Copy bindings/dart to cdk-dart
         run: |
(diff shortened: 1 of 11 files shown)


crodas:
Here it is in action: https://github.com/crodas/cdk/actions/runs/26119557864

.github/workflows/kotlin-publish.yml
```diff
@@ -23,11 +34,11 @@ jobs:
       - name: Checkout monorepo
         uses: actions/checkout@v4
         with:
-          ref: ${{ inputs.cdk_ref || inputs.release_tag }}
+          ref: ${{ inputs.cdk_ref || github.ref }}
```
thesimplekid: Should this keep the previous release-tag fallback?

This changes the checkout from `inputs.cdk_ref || inputs.release_tag` to `inputs.cdk_ref || github.ref`. For manual dispatch, or any helper that omits `cdk_ref`, that can publish `release_tag` while building from the workflow branch instead of the release tag.

Could we keep this as:

```yaml
ref: ${{ inputs.cdk_ref || inputs.release_tag }}
```

Same issue applies to the other CDK checkout steps in this workflow.

.github/workflows/dart-publish.yml
```diff
@@ -9,6 +9,14 @@ on:
       cdk_version:
         description: 'CDK dependency version (e.g. 0.17.0). Defaults to release_tag without the "v" prefix.'
         required: false
+  workflow_call:
+    inputs:
+      release_tag:
+        type: string
+        required: true
+      cdk_version:
+        type: string
```
thesimplekid, line 21: Should Dart accept the same `cdk_ref` input as Kotlin/Swift?

The unified workflow has a `cdk_ref` input and passes it to Kotlin and Swift, but Dart still checks out `main`. That means `just ffi-release-all 0.17.0` can build Dart bindings from `main` while the other bindings build from `v0.17.0`.

Could we add `cdk_ref` to Dart’s `workflow_dispatch` / `workflow_call`, pass it from `ffi-publish-all.yml`, and checkout:

```yaml
ref: ${{ inputs.cdk_ref || inputs.release_tag }}
```

.github/workflows/ffi-publish-all.yml
```diff
@@ -0,0 +1,41 @@
...
+        required: false
+      cdk_ref:
+        description: "CDK git ref to publish from. Defaults to release_tag."
+        required: false
+
+jobs:
+  dart:
+    name: "Trigger Dart bindings"
+    uses: ./.github/workflows/dart-publish.yml
+    with:
```
thesimplekid, line 20: Should this pass `cdk_ref` through to the Dart workflow too?

Kotlin and Swift receive the release ref, but Dart does not, so the “publish all” workflow does not actually guarantee all bindings are generated from the same CDK ref.

thesimplekid (commented):
Focused release-flow comments for the unified FFI workflow.

.github/workflows/ffi-publish-all.yml
```diff
@@ -0,0 +1,42 @@
...
+    uses: ./.github/workflows/kotlin-publish.yml
+    with:
+      release_tag: ${{ inputs.release_tag }}
+      cdk_version: ${{ inputs.cdk_version }}
+      cdk_ref: ${{ inputs.cdk_ref }}
+    secrets: inherit
+
+  swift:
+    name: "Trigger Swift bindings"
+    uses: ./.github/workflows/swift-publish.yml
```
thesimplekid, line 37: Should Swift be aligned with Dart/Kotlin here? Since this workflow calls all three via reusable workflow syntax, `swift-publish.yml` should also expose `workflow_call` with the same inputs. Right now Dart/Kotlin support `workflow_call`, but Swift only has `workflow_dispatch`, so the unified workflow will fail when it reaches Swift.

bindings/README.md
```diff
@@ -94,6 +94,50 @@ just test-dart       # Run tests
...
+just ffi-release-swift 0.17.0
+
+# Go (separate workflow)
+just ffi-release-go 0.17.0
+```
+
+### Prerequisites
+
+- The version tag (e.g. `v0.17.0`) must exist on the remote
+- The `FFI_DEPLOY_KEY` GitHub secret must have write access to `cdk-dart`,
```
thesimplekid, line 137: Should the deploy secret convention be aligned across the three workflows? This says `FFI_DEPLOY_KEY` needs access to `cdk-swift`, and Dart/Kotlin use that secret, but the Swift publish workflow still uses `SWIFT_DEPLOY_KEY`. Aligning the docs and workflows would make the unified release setup easier to reason about.

thesimplekid (commented):
Focused comments on aligning the unified FFI publish workflows.

bindings/README.md
crodas: Fixed in b22cdd34279965f7e2246f35e428af48057d4073

.github/workflows/ffi-publish-all.yml
crodas: Fixed in b22cdd34279965f7e2246f35e428af48057d4073

thesimplekid (approved):

PR #1982, opened 2026-05-19
Add React Native Nitro bindings for CDK

### LINKED ISSUE #1902: Publish React native nitro module for `cashu` crate
In the latest release of cashu-ts (https://github.com/cashubtc/cashu-ts/blob/main/src/model/OutputDataCreator.ts) an interface module was added so native crypto could be used in react native. This has significant performance improvements over just using TS. To support this cdk can publish a RN nitro module with just the crypto from `cashu` crate to meet the requirements of this interface. 

This has the advantage of RN users can keep using cashu-ts/coco for most parts of the wallet since they are using ts anyway and then just use cdk for the native speed improvement without needing the full cdk wallet bindings.

crodas:
### Description

Fixes #1902

Add a new `cdk-nitro` crate and React Native package that exposes CDK functionality via Nitro Modules. The bindings provide a C FFI layer with a C++ HybridObject bridge that lets React Native apps create and manipulate Cashu tokens (OutputData, proofs, keysets, etc.).

Includes:
- Rust FFI crate (`bindings/react-native/rust`) with OutputDataCreator
- C++ HybridObject bridge wrapping the FFI calls
- TypeScript Nitro spec and package entry point
- Android CMakeLists and iOS podspec for native builds
- Node.js FFI test suite using native addon loading
- CI job (nitro-binding-tests) and justfile recipes (test-nitro, test-nitro-node)
- Added nodejs to the bindings Nix devshell

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@codex review

thesimplekid:
@crodas Are the codex reviews addressed they are not marked resolved?

crodas:
> @crodas Are the codex reviews addressed they are not marked resolved?

Some of it is revisited; lemme double-check and close the addressed feedback

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

bindings/react-native/cpp/HybridOutputDataCreator.cpp
thesimplekid, line 1: I do not think we should be re-implementing this logic in cpp, this is what we are trying to avoid by using the bindings. So these should be removed in favor of calls to rust fns, we can add rust fns if we need to. There should be no logic in cpp only the interfaces.

bindings/react-native/src/specs/OutputDataCreator.nitro.ts
```diff
@@ -0,0 +1,99 @@
...
+  sigFlag?: string;
+}
+
+/**
+ * Native Cashu output data creator using Rust DHKE cryptography.
+ *
+ * Implements blinded message construction for the Cashu protocol,
+ * backed by the CDK Rust library for consistent cross-platform crypto.
+ */
+export interface HybridOutputDataCreator
```
thesimplekid: "Have we compiled the Nitrogen-generated C++ against this implementation? With Nitrogen 0.32.2, naming this interface HybridOutputDataCreator generates HybridHybridOutputDataCreatorSpec.hpp, while the handwritten implementation includes HybridOutputDataCreatorSpec.hpp. The generated namespace is also margelo::nitro::cashudevkit, not cashudevkit. This prevents native compilation. Could we align the interface name, generated class name, and namespace, then add a native compile check?"

Have you tested this with a react native build? My agent tells me there are a few other small things like this that will fail. 

Maybe we can get @kelbie to help us test it in sorvan I know he had a fork with nutpatch at one point.

thesimplekid (changes requested):

bindings/react-native/src/specs/OutputDataCreator.nitro.ts
Kelbie: Yes! I would love to test this out. Going to get my agent to implement this and see if I can run into any issues.

Kelbie: AI Summary from my Agent. It doesn't seem quite ready for my case. Pasting in case its helpful:

```
This review inspected:

- CDK PR #1982 at exact head `7d10d7d`
- Coco PR #312 and release `v2.0.0-rc.2`
- cashu-ts `v4.5.1`, the version currently pinned by Sovran

**The original compilation concern is confirmed.** Running Nitrogen 0.32.2 against the PR generates `HybridHybridOutputDataCreatorSpec.hpp` in `margelo::nitro::cashudevkit`. The handwritten bridge instead includes `HybridOutputDataCreatorSpec.hpp` and uses `cashudevkit`. A direct C++ syntax check therefore fails on the missing header. See the [Nitro specification](https://github.com/cashubtc/cdk/blob/7d10d7d199b2c4116edc719a7dd49fa986193958/bindings/react-native/src/specs/OutputDataCreator.nitro.ts#L53-L54) and [handwritten header](https://github.com/cashubtc/cdk/blob/7d10d7d199b2c4116edc719a7dd49fa986193958/bindings/react-native/cpp/HybridOutputDataCreator.hpp#L3-L10).

**HybridObject registration and native packaging are also incomplete.** `nitro.json` has an empty `autolinking` map, so the generated Android registry contains no `OutputDataCreator` registration. Android has no library `build.gradle`, manifest or JNI loader, and its CMake file does not include the generated Nitro sources. The iOS package similarly has no generated object registration. Even after correcting the class name, `createHybridObject('OutputDataCreator')` would not find an implementation.

**The lower Rust layer appears healthy.** All 22 `cdk-nitro` Rust tests pass, clippy passes, and all 35 Node FFI/API tests pass. However, those Node tests load the C ABI through Koffi and reimplement batching in JavaScript; they do not compile or exercise the Nitro C++ bridge. The current [CI job](https://github.com/cashubtc/cdk/blob/7d10d7d199b2c4116edc719a7dd49fa986193958/.github/workflows/ci.yml#L666-L685) runs code generation plus Rust/Node tests, but no Pod, Gradle or real React Native build.

**Coco’s side of the integration is complete.** [Coco PR #312](https://github.com/cashubtc/coco/pull/312) shipped in `v2.0.0-rc.2` and injects one synchronous cashu-ts `OutputDataCreator` through wallet, mint, send, receive, melt, P2PK, restore and sweep paths. Errors propagate without fallback, and persisted outputs are later reconstructed as standard cashu-ts `OutputData`.

**The current CDK object is not yet compatible with that contract.** Cashu-ts expects `AmountLike`, complete keyset objects, `Uint8Array` seeds and `OutputDataLike` results containing a standard blinded message, `bigint` blinding factor, byte secret and `toProof()`. CDK currently accepts JavaScript numbers and separate keyset fields, then returns a flatter native object. See the [cashu-ts interface](https://github.com/cashubtc/cashu-ts/blob/v4.5.1/src/model/OutputDataCreator.ts#L17-L43).

There is also a concrete behavioral incompatibility: cashu-ts splitting supports partial custom splits and explicit zero outputs. Restore can request deterministic outputs with a custom split such as `[0]`. CDK’s current C++ splitter rejects every zero denomination and requires the custom split to equal a positive amount exactly. Consequently, a superficial adapter would break Coco restore/sweep even after the native module compiles. Cashu-ts’s required behavior is defined in [`splitAmount`](https://github.com/cashubtc/cashu-ts/blob/v4.5.1/src/utils/core.ts#L27-L109).

A compatible package should therefore:

- align the generated and handwritten names/namespaces;
- register the HybridObject and complete the iOS/Android module scaffolding;
- add a real React Native native-build test;
- export a cashu-ts-compatible adapter that wraps native results in `new OutputData(...)`;
- reproduce cashu-ts amount, split ordering, zero-output and counter semantics exactly;
- preserve or explicitly reject unsupported P2PK/P2BK/HTLC options;
- test CDK results against cashu-ts’s NUT-13 vectors rather than only testing self-determinism.

The Rust crypto primitive is promising, but the current PR is not yet an end-to-end consumable React Native module. Once these packaging and contract gaps are resolved, Sovran can inject it through Coco’s existing `outputDataCreator` seam without importing Rust crates or adding a Coco plugin.
```

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@Kelbie  Here is a built trigger from my fork

https://github.com/crodas/cdk-nitro

crodas:
@cdk-bot review

Kelbie:
I finally managed to get this working in Sovran. I'm going to keep testing this today.

bindings/react-native/rust/src/lib.rs
```diff
@@ -0,0 +1,1585 @@
...
+        encode_result(
+            &self.blinded_secret.to_string(),
+            &self.blinding_factor.to_secret_hex(),
+            &self.secret.to_string(),
+        )
+    }
+}
+
+/// Blind an ephemeral random secret: B_ = hash_to_curve(secret) + r*G.
+fn build_random_parts() -> Result<BlindParts, String> {
```
thesimplekid, line 106: To clarify my [earlier comment](https://github.com/cashubtc/cdk/pull/1982#discussion_r3608432106): by “call Rust functions, and add them if needed,” I meant shared functions in the canonical cashu/CDK crates—not moving the second implementation from C++ into a binding-specific Rust crate.

This file now reimplements random, P2PK, and deterministic output construction and amount splitting using low-level Cashu primitives. Most of this already exists through PreMintSecrets and Amount::split_targeted. Where those APIs do not meet the required cashu-ts contract, we should make the necessary changes in crates/cashu and test them there.

The current duplicate-key behavior demonstrates why this must be shared: this path can construct P2PK conditions that Cashu spending validation later rejects. The splitter also differs from cashu-ts by omitting its output limit. Adding binding-local patches would leave us maintaining two implementations.

The Nitro Rust crate should only convert FFI inputs into Cashu domain types, call the shared constructor, and map the results back across the ABI. Please move the missing domain behavior into crates/cashu.

bindings/react-native/rust/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "cdk-nitro"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[lib]
+crate-type = ["staticlib", "cdylib"]
+name = "cdk_nitro"
+
+[lints.rust]
```
thesimplekid, line 11: Why is this crate opting out of the workspace-wide unsafe_code = "forbid" policy?

I understand that a handwritten C ABI requires raw-pointer operations, but bypassing the workspace lint should be an explicit architectural decision. Could we use generated bindings such as cxx or the existing UniFFI infrastructure instead?

If handwritten unsafe FFI is accepted, it should at minimum be isolated in a small audited module with documented Safety contracts. It should not be mixed with the output-construction and splitting logic.

.github/workflows/ci.yml
```diff
@@ -727,6 +728,44 @@ jobs:
...
+      - name: checkout
+        uses: actions/checkout@v5
+      - uses: cachix/cachix-action@v17
+        with:
+          name: cashudevkit
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+          useDaemon: false
+        continue-on-error: true
+      - name: Run Nitro codegen
+        run: nix develop -L .#bindings --command bash -c 'cd bindings/react-native && npx nitrogen'
```
thesimplekid, line 763: Could we install the locked dependencies before running codegen?

On a clean checkout, npx nitrogen downloads the latest Nitrogen rather than the package’s locked 0.32.x version. I reproduced the current 0.36.x release generating zero specs and still exiting successfully.

The subsequent Rust and Node tests bypass the generated Nitro code and handwritten C++ bridge, so this job can pass without validating the actual binding. Please use npm ci followed by npm run codegen.

thesimplekid:
Not much movement in a bit so going to mark as draft.

PR #1988, opened 2026-05-21
fix: support generic NUT-17 quote kinds

callebtc:
## Summary
- switch advertised and wallet-used NUT-17 quote commands to the new generic `mint_quote` and `melt_quote` strings from cashubtc/nuts#372
- keep mint-side backwards compatibility for the legacy method-specific quote kind strings and mark the temporary compatibility paths as deprecated/TODO cleanup points
- fan out mint events to both generic and legacy quote topics so old websocket clients keep working while new wallets subscribe with the generic kinds

## Testing
- `cargo test -p cashu nut17::tests --lib`
- `cargo check -p cdk-common -p cdk --lib --no-default-features --features wallet`
- `cargo check -p cdk-ffi --no-default-features --features bip353`
- `cargo test -p cdk --lib --no-default-features --features wallet subscription::tests`

## Notes
- `cargo check -p cdk-axum` / mint-feature verification is blocked in this environment because `cdk-signatory` requires a local `protoc` binary during build.

### DIFF
--- a/crates/cashu/src/nuts/nut17/mod.rs
+++ b/crates/cashu/src/nuts/nut17/mod.rs
@@ -60,8 +60,8 @@ impl SupportedMethods {
     /// Create [`SupportedMethods`] for Bolt11 with all supported commands
     pub fn default_bolt11(unit: CurrencyUnit) -> Self {
         let commands = vec![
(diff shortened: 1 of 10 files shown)


thesimplekid:
Closing since the nut change was closed.

PR #1997, opened 2026-05-23
chore: settings stream

d4rp4t:
### Description

Implements the GetSettings server-streaming RPC in the payment processor crate. The proto already declared rpc GetSettings(EmptyRequest) returns (stream SettingsResponse) but the server returned a single response and the client consumed it as such. This wires up the full streaming path so the mint can receive settings updates from the payment processor in real time without restarting.

-----

### Notes to the reviewers

The server is event-driven — it calls inner.wait_settings() and forwards whatever the backend emits, mirroring the pattern used by wait_payment_event. No polling.

wait_settings is added to MintPayment with a default implementation that emits the current settings as a single item (stream::once(get_settings())). Backends that support live updates (e.g. a chained payment processor) can override it. Backends that don't (cdk-cln, cdk-bdk, cdk-fake-wallet) get the default: one item then stream stays open until client disconnects or server shuts down.

get_settings() on the PaymentProcessorClient reads the first message from the stream — backward compatible with all existing callers. wait_settings() subscribes to the full stream for callers that want live updates.

The server uses tx.closed() to detect client disconnect without polling, so tasks don't leak when a client calls get_settings() and drops the stream after the first message.

-----

#### CHANGED
- PaymentProcessorServer::get_settings now returns a server-streaming response; the stream emits the current settings immediately and forwards any subsequent updates from the backend
#### ADDED
- MintPayment::wait_settings — subscribes to a stream of settings updates; default impl emits current settings once, backends can override for live updates
- PaymentProcessorClient::wait_settings — opens the gRPC settings stream and forwards all items
----

### Checklist

* [x ] I followed the [code style guidelines](https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md)
* [ x] If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`)


thesimplekid:
@codex review

ye0man:
@crodas

crodas:
Can you take a look at #2254 and use that for the streaming?

thesimplekid:
> Can you take a look at #2254 and use that for the streaming?

Does using the `tokio::watch` like what was done for the keysets also make sense here?

crodas:
> Does using the tokio::watch like what was done for the keysets also make sense here?

I think so, yes. Because `tokio::watch` will only delivery the latest message, which is perfect for the keysets and the settings.

From their docs

> This channel is useful for watching for changes to a value from multiple points in the code base, for example, changes to configuration values.

thesimplekid:
@crodas can you finish this?

PR #2000, opened 2026-05-23
Add golang release

crodas:
### Description

This PR depends on #1981. This PR will trigger an cdk release into the cdk-golang repository

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/go-publish.yml
+++ b/.github/workflows/go-publish.yml
@@ -1,0 +1,416 @@
+name: "FFI - Go Bindings"
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: "Tag to create for release (e.g. v0.17.0)"
+        required: true
+      cdk_version:
+        description: 'CDK dependency version (e.g. 0.17.0). Defaults to release_tag without the "v" prefix.'
+        required: false
+      cdk_ref:
+        description: "CDK git ref to publish from. Defaults to release_tag."
+        required: false
+  workflow_call:
+    inputs:
+      release_tag:
+        type: string
+        required: true
+      cdk_version:
+        type: string
+        required: false
+      cdk_ref:
+        type: string
+        required: false
+env:
+  CARGO_TERM_COLOR: ${{ vars.CARGO_TERM_COLOR }}
+  TAG: ${{ inputs.release_tag }}
+  CDK_REF: ${{ inputs.cdk_ref || inputs.release_tag }}
+  RELEASE_BRANCH: release/${{ inputs.release_tag }}
+  UNIFFI_BINDGEN_GO_TAG: v0.6.0+v0.30.0
+
+jobs:
+  go-sync-sources:
+    name: "go: Sync sources to cdk-go"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Resolve variables
+        run: |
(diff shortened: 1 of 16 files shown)


crodas:
Here is a [build triggered](https://github.com/crodas/cdk/actions/runs/26364511697) from my fork, the output is written to https://github.com/crodas/cdk-golang

I will now create a test app that would pull that and create a test app

.github/workflows/go-publish.yml
```diff
@@ -0,0 +1,422 @@
...
+      - name: Merge to default branch
+        working-directory: cdk-go
+        run: |
+          DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+          echo "DEFAULT_BRANCH=${DEFAULT_BRANCH}" >> "$GITHUB_ENV"
+          git checkout "${DEFAULT_BRANCH}"
+          git merge "${RELEASE_BRANCH}" --squash
+          git commit -m "release: ${TAG}"
+
+      - name: Push branches (without tag)
```
thesimplekid: This needs to stay aligned with the release ordering we already settled on for the other bindings workflows.

Right now this pushes the `cdk-go` default branch before `gh release create` succeeds. If release creation or asset upload fails, the target repo has already advanced and retrying can require manual cleanup.

This exact release-ordering issue was already brought up on the Kotlin workflow and fixed/approved as part of the bindings release flow:
- https://github.com/cashubtc/cdk/pull/1939#discussion_r3177922836
- https://github.com/cashubtc/cdk/pull/1939#discussion_r3253006237

bindings/go/README.md
```diff
@@ -0,0 +1,91 @@
+# cdk-go
+
+Go FFI bindings for [cashubtc/cdk](https://github.com/cashubtc/cdk) (Cashu Development Kit), generated via [uniffi-bindgen-go](https://github.com/NordSecurity/uniffi-bindgen-go).
+
+Prebuilt native libraries are included — downstream consumers only need Go.
+
+## Install
+
+```bash
+go get github.com/{{CDK_GO_REPO}}
```
thesimplekid: The generated fork output still has the test repo in user-facing install docs:

`go get github.com/crodas/cdk-golang`

For the actual release this should be `github.com/cashubtc/cdk-go`, matching `CDK_GO_REPO` and the repository users will install from. Can we make sure the published README is generated with `cashubtc/cdk-go` and not the fork/test repo value?

thesimplekid (commented):
Two release-flow comments from review.

asmogo (dismissed):
the generated bindings are working as expected.

bindings/go/README.md
crodas: Do you think the placeholder is wrong?

thesimplekid: Why do we need a place holder it should just be cashubtc/cdk-dart

crodas: You are right, that path will rarely change. I'll make it happen

thesimplekid: This same issue was also introduced for the other ones so I fixed those 1a0c7262

thesimplekid (approved):

PR #2007, opened 2026-05-26
feat: add mints_strict field to payment requests

d4rp4t:
### Description

Adds mints_strict (ms) field to PaymentRequest as specified in NUT-18 and NUT-26 PR: https://github.com/cashubtc/nuts/pull/381

When ms is absent or true (default), the recipient MUST reject payments from mints not in the mint list. When ms is false, the mint list is a suggestion for the payer.

NUT-26 (bech32m) encoding adds TLV tag 0x09 for this field.

### Notes to the reviewers
The field is serialized with skip_serializing_if = "Option::is_none" so existing encoded payment requests remain valid — absence of the tag is equivalent to true (strict mode), consistent with the proposed spec default.

-----

#### ADDED
- mints_strict field (ms) to PaymentRequest (NUT-18) with corresponding TLV tag 0x09 in NUT-26 bech32m encoding
- mints_strict() builder method on PaymentRequestBuilder

### Checklist

* [] I ran `just quick-check` before committing


d4rp4t:
Added changes so it matches: https://github.com/cashubtc/nuts/pull/381

thesimplekid:
@codex review

robwoodgate:
The nut has been updated: `ms` (mint_strict) is now flipped to `mp` (mint_preferred) to avoid the omitted=true antipattern.

thesimplekid:
@codex review

thesimplekid:
@cdk-bot review

robwoodgate:
Noted by my agent when checking compliance to current spec:

Melt check already binds to the request unit.
  - Swap `unwrap_or(CurrencyUnit::Sat)` in `Wallet::pay_request` (and the repository path) for a reject when unit is None and `sm` is present. Whether they keep the sat default for legacy unit-less requests without `sm` is their call; the spec only invalidates the `a/sm` cases. 
  - Optionally have `PaymentRequestBuilder::build()` error on `supported_methods` (or `amount`) with `unit: None` so invalid requests can't be minted in the first place.

PR #2012, opened 2026-05-28
multiple payment backends

asmogo:
### Description

rebased #1949 on main 

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-mintd/src/config.rs
+++ b/crates/cdk-mintd/src/config.rs
@@ -169,6 +169,8 @@ impl std::str::FromStr for LnBackend {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Ln {
     pub ln_backend: LnBackend,
+    #[serde(default)]
+    pub unit: CurrencyUnit,
     pub invoice_description: Option<String>,
     pub min_mint: Amount,
     pub max_mint: Amount,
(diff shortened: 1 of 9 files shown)


crates/cdk-mintd/src/lib.rs
```diff
@@ -714,22 +751,47 @@ async fn configure_onchain_backend(
...
+                    bdk.clone(),
+                )
+                .await?;
+
+                mint_builder = configure_backend_for_unit(
+                    settings,
+                    mint_builder,
+                    cdk::nuts::CurrencyUnit::Sat,
+                    mint_melt_limits,
+                    bdk,
```
thesimplekid: Is this duplicate?

crates/cdk-mintd/src/config.rs
```diff
@@ -1089,24 +1116,25 @@ impl Settings {
...
+            .any(|ln| ln.ln_backend == LnBackend::FakeWallet);
+        let has_real_ln_backend = self
+            .ln
+            .iter()
+            .any(|ln| !matches!(ln.ln_backend, LnBackend::None | LnBackend::FakeWallet));
+
+        match onchain_backend {
             #[cfg(feature = "bdk")]
-            (LnBackend::FakeWallet, OnchainBackend::Bdk) => {
+            OnchainBackend::Bdk if has_fake_wallet_ln_backend => {
```
thesimplekid: I think we should have the check for intra ln backends as well. We should not allow both a fake and real ln backend.

thesimplekid (changes requested):

crates/cdk-mintd/src/lib.rs
asmogo: yes

crates/cdk-mintd/src/config.rs
asmogo: I don't like this part either. I think this was mainly done to pass the tests. Going to rewrite this.

thesimplekid:
https://github.com/cashubtc/cdk/pull/2015

PR #2016, opened 2026-05-31
feat(cdk-mintd): add nutshell database migration subcommand

### LINKED ISSUE #1985: nutshell -> cdk migration script
https://github.com/cashubtc/nutshell/pull/976 closed as we decided it makes more sense for the script to live in cdk codebase

a1denvalu3:
This PR implements a new subcommand `migrate-nutshell` for `cdk-mintd` to automatically migrate a Nutshell mint database (SQLite or Postgres) to the configured CDK mint database.

### Key Features:
1. **Direct Parsing to CDK Structs**: Avoids custom intermediate structs by directly reading raw database query results and parsing them directly into standard, strongly-typed CDK types (`MintKeySetInfo`, `MintQuote`, `MeltQuote`, `BlindSignature`, `BlindedMessage`, `Proof`).
2. **Chunked constant-memory streaming**: Reads nutshell's `keysets`, `mint_quotes`, `melt_quotes`, `promises`, and `proofs` tables in segments of 2000 rows. Memory consumption stays flat and negligible even when migrating massive production databases (such as Minibits).
3. **Reuse Existing CDK business logic and CRUD**: The data is fully forced through CDK's transaction layers (`tx.add_mint_quote`, `tx.add_proofs`, `tx.update_proofs_state`, etc.), maintaining the integrity of all database validations, state transitions, and historic payment/issuance tracking.
4. **Pre-flight & Version Checks**:
   - Aborts if the target database is already populated.
   - Restricts nutshell software versions to <= 0.20.1 (customizable constant).
   - Skips the migration of individual keysets generated under nutshell versions < 0.15.0 (since they use different derivation path logic not supported by CDK).
5. **Warnings Summary**: If any pre-0.15 keysets, promises, or proofs are skipped, a clear summary warning is printed at the end of the migration detailing exactly how many records were skipped.
6. **Robust Column parsing**: SQLite columns are retrieved as dynamic values to make the parser completely immune to different SQLite column data types (dynamic typing of Integer vs Text in different nutshell versions).


a1denvalu3:
### Architectural Note: Standalone Fuzzer Design & Nutshell Mint Lifecycle Spawning

During my review and cleanup of the nutshell-to-CDK database migration integration test (`nutshell_migration_fuzzer.rs`), I wanted to highlight the current design approach and present an alternative for future consideration.

#### **Path A: Programmatic, Self-Contained Spawning in Rust (Current Design)**
Currently, the fuzzer test manages the entire lifecycle of both the Nutshell mint (via Python Poetry or Docker container) and the subsequent migrated CDK mint directly within the Rust process (`#[tokio::test]`).

* **How it works:**
  1. Spawns Nutshell (Docker or Python module) in the background.
  2. Funds and fuzzes the active Nutshell mint with mock wallets.
  3. Shuts down Nutshell cleanly.
  4. Runs the database migration directly against the target SQLite/Postgres.
  5. Boots up a CDK Axum mint pointing to the migrated DB on the same port in-process.
  6. Fuzzes again with the same wallets to verify exact balances and spendability.
* **Pros:** Highly self-contained and portable. Any developer or IDE can execute `cargo test -p cdk-integration-tests --test nutshell_migration_fuzzer` without needing any external bash scripting or orchestrators.

---

#### **Path B: Bash/Shell Orchestration & Multi-Phase Binaries (Alternative Design)**
An alternative would be to strip all setup, process management, and docker lifecycle handling out of the Rust codebase, delegating it to an external bash script (similar to the way `nutshell_wallet_itest.sh` coordinates tests).

* **How it would work:**
  1. Bash script starts the Nutshell container.
  2. Cargo test runs "Phase 1" fuzzing (wallet funding and operations).
  3. Bash script stops the Nutshell container.
  4. Bash script executes `cdk-mintd migrate-nutshell ...` CLI command to run the migration.
  5. Bash script starts `cdk-mintd` daemon pointing to the migrated DB.
  6. Cargo test runs "Phase 2" verification (wallet spendability).
  7. Bash script terminates everything and cleans up.
* **Pros:** Less custom process management / subprocess spawning logic in Rust.
* **Cons:** Harder to run locally (developers cannot just run `cargo test`), and more fragile under CI/CD because a failed shell command or early exit might easily orphan background processes or leave ports bound.

---

Both options are viable, but the current **Path A** keeps integration testing robust, fully portable, and extremely simple to run with a single Cargo invocation.

prusnak:
This is amazing! 🫶

.github/workflows/nutshell_itest.yml
```diff
@@ -53,3 +53,19 @@ jobs:
...
         run: docker logs nutshell-wallet || true
+
+  nutshell-migration-integration-tests:
+    name: Nutshell Migration Integration Tests
+    runs-on: self-hosted
+    timeout-minutes: 30
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+      - uses: cachix/cachix-action@v16
```
thesimplekid:
```suggestion
        uses: actions/checkout@v5
      - uses: cachix/cachix-action@v17
```

crates/cdk-sqlite/src/mint/migrate.rs
```diff
@@ -0,0 +1,1078 @@
...
+                }
+            }
+
+            let method = match PaymentMethod::from_str(&method_str) {
+                Ok(m) => m,
+                Err(_) => PaymentMethod::from("bolt11"),
+            };
+
+            let quote_res = match MeltQuote::from_db(
+                q_id.clone(),
```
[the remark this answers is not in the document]
a1denvalu3: There is no on-chain in nutshell

crates/cdk-postgres/src/migrate.rs
```diff
@@ -0,0 +1,951 @@
...
+                }
+            }
+        }
+
+        let method = match PaymentMethod::from_str(&method_str) {
+            Ok(m) => m,
+            Err(_) => PaymentMethod::from("bolt11"),
+        };
+
+        let quote_res = match MeltQuote::from_db(
```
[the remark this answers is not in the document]
a1denvalu3: There is no on-chain in nutshell

NUTSHELL_MIGRATION_PLAN.md
thesimplekid, line 1: Lets remove this before merge

docs/nutshell-migration-fuzzer.md
thesimplekid, line 1: Think this is also not meant to be left in?

crates/cdk-mintd/README.md
```diff
@@ -184,6 +184,105 @@ mint-cli rotate-next-keyset --use-keyset-v2       # Rotate to V2
 mint-cli rotate-next-keyset --use-keyset-v2=false # Rotate to V1
 ```
 
+## Migrating from a Nutshell Mint
```
thesimplekid, line 319: We should not include the full instructions here in the README but a short note and then a specefic link to a migration doc, probably worth including in the top level readme as well.

crates/cdk-integration-tests/Cargo.toml
```diff
@@ -26,6 +26,7 @@ cdk-lnd = { workspace = true }
 cdk-ldk-node = { workspace = true }
 cdk-axum = { workspace = true }
 cdk-sqlite = { workspace = true }
+rusqlite = { version = "0.31", features = ["bundled"] }
```
thesimplekid: We should make this a workspace dep since this is used in at least 2 crates

crates/cdk-sqlite/Cargo.toml
```diff
@@ -18,6 +18,7 @@ wallet = ["cdk-common/wallet", "cdk-sql-common/wallet"]
 sqlcipher = ["rusqlite/bundled-sqlcipher"]
 prometheus = ["cdk-sql-common/prometheus", "cdk-prometheus"]
 [dependencies]
+anyhow.workspace = true
```
thesimplekid:
```suggestion
```

I don't see where this is used?

crates/cdk-sqlite/Cargo.toml
```diff
@@ -31,6 +32,8 @@ serde.workspace = true
 serde_json.workspace = true
 lightning-invoice.workspace = true
 uuid.workspace = true
+chrono = { version = "0.4", features = ["serde"] }
+hex = "0.4"
```
thesimplekid:
```suggestion
```

We have our own hex util we can use instead of this dep. `use cdk_common::util::hex;`

crates/cdk-sqlite/Cargo.toml
```diff
@@ -31,6 +32,8 @@ serde.workspace = true
 serde_json.workspace = true
 lightning-invoice.workspace = true
 uuid.workspace = true
+chrono = { version = "0.4", features = ["serde"] }
```
thesimplekid:
```suggestion
chrono = { version = "0.4", default-features = false }
```

crates/cdk-sqlite/src/mint/migrate.rs
```diff
@@ -0,0 +1,1501 @@
...
+                "failed" => MeltQuoteState::Unpaid,
+                _ => MeltQuoteState::Unpaid,
+            };
+
+            if let Some(ref ref_id) = request_lookup_id {
+                if state_mapped == MeltQuoteState::Paid || state_mapped == MeltQuoteState::Pending {
+                    let id_key = ref_id.to_string();
+                    let mut borrowed = seen_paid_pending_lookup_ids.borrow_mut();
+                    if borrowed.contains(&id_key) {
+                        let dup_id = format!("{}-dup-{}", id_key, q_id);
```
thesimplekid: Should we be creating this dedup? CDK allows multiple unpaid quotes but only one pending or paid, should multiple paid quotes error?

PSGL has the same behavior.

crates/cdk-postgres/src/migrate.rs
```diff
@@ -0,0 +1,1227 @@
...
+        migrated_keyset_ids.insert(keyset_id);
+        migrated_keysets += 1;
+    }
+
+    key_tx.commit().await?;
+    tracing::info!("Migrated keysets successfully.");
+
+    // Start main database transaction after keysets are committed to avoid SQLite lock deadlock
+    let mut tx = MintDatabase::begin_transaction(&db).await?;
+
```
thesimplekid, line 1086: Why do we make a commit here instead of doing the full migration in one tx so it is atomic? This comment says sqlite deadlock but this is psgl?

crates/cdk-sqlite/src/mint/migrate.rs
```diff
@@ -0,0 +1,1501 @@
...
+    let mut migrated_promises_signed = 0;
+    let mut migrated_proofs = 0;
+
+    // Map and migrate keysets
+    let mut migrated_keyset_ids = HashSet::new();
+    for keyset in nutshell_keysets {
+        if let Some(ref version_str) = keyset.issuer_version {
+            let ver_clean = version_str.to_string().replace("nutshell/", "");
+            if let Some(keyset_v) = parse_nutshell_version(&ver_clean) {
+                if keyset_v < (0, 15, 0) {
```
thesimplekid, line 1053: Here we skip older keys but in verification we do not skip them and expect them to be migrated. So a migration with older keys we skip here will always fail verification. https://github.com/cashubtc/cdk/pull/2016/changes#diff-fce7f7fb8607d3b1471fd0209d83997819495a9d6190ab1cb6d823367c197ae1R162

crates/cdk-postgres/src/migrate.rs
```diff
@@ -0,0 +1,1227 @@
...
+    let (client, connection) = tokio_postgres::connect(url, tokio_postgres::NoTls)
+        .await
+        .map_err(|e| Error::Database(Box::new(e)))?;
+    tokio::spawn(async move {
+        if let Err(error) = connection.await {
+            tracing::error!(%error, "Postgres verification connection failed");
+        }
+    });
+    Ok(client)
+}
```
thesimplekid, line 110: CDK mintd supports both tls and no tls should we do the same here, looks like we only support migrations with notls?

crates/cdk-sqlite/src/mint/migrate.rs
```diff
@@ -0,0 +1,1501 @@
...
+                        );
+                        return Ok(None);
+                    }
+                };
+
+            let state_mapped = match state_str.to_lowercase().as_str() {
+                "paid" => MintQuoteState::Paid,
+                "issued" => MintQuoteState::Issued,
+                _ => MintQuoteState::Unpaid,
+            };
```
thesimplekid, line 495: Do we need to handle pending here?

crates/cdk-sqlite/src/mint/migrate.rs
```diff
@@ -0,0 +1,1501 @@
...
+            let version: String = row.get(5)?;
+            let unit_str: String = row.get(6)?;
+            let input_fee_ppk: i64 = row.get(7)?;
+            let amounts_str: String = row.get(8)?;
+            let final_expiry_val: Option<i64> = row.get(9)?;
+
+            let valid_from_str = val_to_string(valid_from_val);
+
+            let amounts_vec: Vec<u64> = if amounts_str.is_empty() || amounts_str == "[]" {
+                (0..32).map(|i| 2_u64.pow(i)).collect()
```
thesimplekid: Should we default to 64 here since that was nutshell's default?

crates/cdk-postgres/src/migrate.rs
```diff
@@ -0,0 +1,1227 @@
...
+        .map_err(|e| Error::Database(Box::new(e)))?;
+    let mut chunk = Vec::new();
+    for r in rows {
+        let quote: String = r.get(0);
+        let method_str: String = r.get(1);
+        let request_str: String = r.get(2);
+        let checking_id: String = r.get(3);
+        let unit_str: String = r.get(4);
+        let amount: i64 = r.get(5);
+        let fee_reserve: i32 = r.get::<_, Option<i32>>(6).unwrap_or(0);
```
thesimplekid: Does nutshell store the fee reserve as a big int I think so if so this get can panic and we should use a try get.

crates/cdk-sqlite/src/mint/migrate.rs
```diff
@@ -0,0 +1,1501 @@
...
+fn read_proofs_chunk_sqlite(
+    conn: &rusqlite::Connection,
+    limit: i64,
+    offset: i64,
+    spent: bool,
+) -> Result<Vec<MigratedProofInfo>, Error> {
+    let query_str = if spent {
+        "SELECT amount, id, c, secret, witness, melt_quote FROM proofs_used ORDER BY secret LIMIT ? OFFSET ?;"
+    } else {
+        "SELECT amount, id, c, secret, NULL, melt_quote FROM proofs_pending ORDER BY secret LIMIT ? OFFSET ?;"
```
thesimplekid: What is the reason for discarding the witness on a pending proof?

thesimplekid (changes requested):
Thanks for working on this.

crates/cdk-postgres/src/migrate.rs
a1denvalu3: I started with a sqlite only migration, ported it over to psql. This persisted.

thesimplekid:
Are the changes minibits needed captured in this pr or did they do other things?

Kelbie:
Am I right in thinking as it stands right now, this wouldn't help us on Nutshell 0.20.3 migrate to CDK?

a1denvalu3:
@thesimplekid I am not sure because misovan didn't get back to me when I asked.

a1denvalu3:
@Kelbie Why?

PR #2043, opened 2026-06-07
fix(wallet): honor TTL in auth metadata cache

thesimplekid:
load_auth compared auth_status.updated_at > Instant::now(), but updated_at is stamped at write time and is therefore always in the past. The comparison was always false, so the cache-hit branch was dead and every load_auth re-fetched auth keysets from the mint over HTTP, ignoring the configured cache TTL.

Mirror the non-auth load() path: pass an Option<Duration> TTL into load_auth and treat cached data as fresh while updated_at + ttl is in the future. Thread the wallet's metadata_cache_ttl into AuthWallet so both paths share one configurable TTL via set_metadata_cache_ttl.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/auth/auth_wallet.rs
+++ b/crates/cdk/src/wallet/auth/auth_wallet.rs
@@ -1,8 +1,10 @@
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::time::Duration;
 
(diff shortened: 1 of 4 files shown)


thesimplekid:
@codex review

PR #2048, opened 2026-06-07
feat: bls

thesimplekid:
### Description

-----

- [ ] handle keyset rotation/migration

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut01/bls.rs
+++ b/crates/cashu/src/nuts/nut01/bls.rs
@@ -189,16 +189,24 @@ fn derive_batch_weights(
     let challenge = Sha256Hash::hash(&transcript).to_byte_array();
     (0..mint_pubkeys.len())
         .map(|i| {
-            let mut counter = 0u8;
+            // Per NUT-00, derive each weight by rejection sampling in `Fr*`:
+            //   h = SHA256(challenge || u32_BE(i) || u32_BE(ctr))
+            //   x = OS2IP(h); reject if x == 0 or x >= BLS_FR_ORDER.
+            // `BlsSecretKey::from_bytes` interprets `h` big-endian and only succeeds
+            // when x < BLS_FR_ORDER (canonical), so it rejects the out-of-range case;
+            // we additionally reject the zero scalar. Plain modular reduction would
+            // bias the distribution and diverge from the spec's deterministic weights.
+            let mut counter = 0u32;
             loop {
-                let mut weight_material = Vec::with_capacity(37);
+                let mut weight_material = Vec::with_capacity(40);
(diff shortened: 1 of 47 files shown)


a1denvalu3 (approved):

crates/cashu/src/nuts/nut02.rs
```diff
@@ -666,10 +666,9 @@ impl MintKeySet {
                     secret_key: secret_key.into(),
                     public_key: public_key.into(),
                 },
-                KeySetVersion::Version02 => {
-                    let digest = Sha256::hash(&secret_key.secret_bytes()).to_byte_array();
-                    MintKeyPair::from_secret_key(SecretKey::bls_from_reduced_bytes(&digest))
-                }
+                KeySetVersion::Version02 => MintKeyPair::from_secret_key(
+                    SecretKey::bls_from_reduced_bytes(&secret_key.secret_bytes()),
```
a1denvalu3: Keys should also be rejection sampled, because the curve order of BLS is much smaller than $2^{256}$. Since $r$ is smaller than $2^{256}$, the probability that a random 256-bit integer $x$ is larger than or equal to $r$ is:
$$P(x \geq r) = 1 - \frac{r}{2^{256}} \approx 1 - 0.4528 = \mathbf{54.72\%}$$

nutshell reference: https://github.com/cashubtc/nutshell/pull/999/changes#diff-3fb6bb16c1a8d73a624c853e623ef777662162ec5378f7970ad1ec2233caffbfR235-R247

crates/cashu/src/nuts/nut01/bls.rs
```diff
@@ -0,0 +1,269 @@
...
+    (0..mint_pubkeys.len())
+        .map(|i| {
+            let mut counter = 0u8;
+            loop {
+                let mut weight_material = Vec::with_capacity(37);
+                weight_material.extend_from_slice(&challenge);
+                weight_material.extend_from_slice(&(i as u32).to_be_bytes());
+                weight_material.push(counter);
+                let weight = Sha256Hash::hash(&weight_material).to_byte_array();
+                let scalar = BlsSecretKey::from_reduced_bytes(&weight);
```
a1denvalu3: `from_reduced_bytes` reduces the sample modulo `r` instead of erroring? we should reject sample here as well. I'm pretty sure that's how the spec goes.

robwoodgate: Well spotted! Also, the counter should be u32, not u8 - it is part of the transcript, so a change in byte width means the transcript is different to spec.

thesimplekid:
close for #2194

PR #2050, opened 2026-06-08
fix: soft-error the mint time-tolerance check

### LINKED ISSUE #1194: Mint time outside of tolerance - some mints about 15 minutes behind
I sometimes get these errors about timestamps, in my chat+cashu [app](https://chat.satsandsports.cash/), where CDK generates this error: `Mint time outside of tolerance`. This was with coinos.io, but I think I've seen it with others too. With coinos, sometimes it works and sometimes it doesn't.

*__Update:__ Here's a straightforward example. I put in some logs. I'll try to find a way to contact the mint, and close this*

Image

The mint is sometimes about 20 minutes behind. I verified my wallet's time with https://time.is/Unix_time

```
Mint time does match wallet time. Mint: 1760686887, Wallet: 1760687877
TRACE /home/satsandsports/MyCode/Cashu/mdk_ecash/cdk/crates/cashu/src/nuts/nut01/secret_key.rs:165 Secret Key dropped.
Error: Failed to receive token: Failed to receive token: Mint time outside of tolerance
AbortError: The user aborted a request.
```

I removed the timecheck from the cdk and am continuing to test, so far this seems to be working:

Image

Although, of course we do need the correct time for various spending conditions, so I'm not proposing to remove this check

mvanhorn:
### Description

Soft-errors the mint time-tolerance check in the wallet so a mint whose timestamp is slightly outside the tolerance window no longer hard-fails the operation; it logs and continues instead, per the maintainer's prescribed approach in #1194. Fixes #1194.

-----

### Notes to the reviewers

The change is scoped to the non-consensus time-tolerance check at `crates/cdk/src/wallet/mod.rs`; token/proof validation logic is untouched.

-----

#### CHANGED

None.

#### ADDED

None.

#### REMOVED

None.

#### FIXED

- Mint time-tolerance check is now a soft error (log and continue) instead of a hard failure.

----

### Checklist

* [x] I followed the code style guidelines
* [x] The change compiles for the cdk wallet feature set
* [x] The Wallet API signature was not modified, so no FFI binding changes are needed

### DIFF
--- a/crates/cdk/src/wallet/mod.rs
+++ b/crates/cdk/src/wallet/mod.rs
@@ -400,13 +400,14 @@ impl Wallet {
         // If mint provides time make sure it is accurate
         if let Some(mint_unix_time) = mint_info.time {
             let current_unix_time = crate::util::unix_time();
-            if current_unix_time.abs_diff(mint_unix_time) > 30 {
+            let mint_time_delta = current_unix_time.abs_diff(mint_unix_time);
+            if mint_time_delta > 30 {
                 tracing::warn!(
-                    "Mint time does match wallet time. Mint: {}, Wallet: {}",
+                    "Mint time does not match wallet time. Mint: {}, Wallet: {}, Delta: {} seconds",
                     mint_unix_time,
-                    current_unix_time
+                    current_unix_time,
+                    mint_time_delta
                 );
-                return Err(Error::MintTimeExceedsTolerance);
             }
         }
 


thesimplekid:
I think this is a mint issue and mints should be forced to fix this there is no reason a mints time should be this far out of date and if it is scripting locks cannot be relied on so it should not be used and erroring is correct. We should not soften the wallet to work with this.

PR #2078, opened 2026-06-10
docker hard image

asmogo:
### Description
docker image without `/bin/sh` built from scratch image. 
 

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Dockerfile.hardened
+++ b/Dockerfile.hardened
@@ -15,11 +15,14 @@ COPY --from=rootfs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certific
 COPY --from=rootfs --chown=10001:10001 /data /data
 COPY --chown=0:0 --chmod=0555 cdk-mintd /usr/local/bin/cdk-mintd
 
+# Add tmpfs for temporary files (faster, ephemeral)
+RUN mkdir -p /tmp && chmod 1777 /tmp
+
 ENV CDK_MINTD_WORK_DIR=/data \
     HOME=/data \
     SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt \
-    SQLITE_TMPDIR=/data \
-    TMPDIR=/data
+    SQLITE_TMPDIR=/tmp \
+    TMPDIR=/tmp
 
 WORKDIR /data
 USER 10001:10001
--- a/Dockerfile.hardened
+++ b/Dockerfile.hardened
@@ -1,0 +1,29 @@
+# syntax=docker/dockerfile:1.7
+
+FROM alpine:3.21 AS rootfs
+
+RUN apk add --no-cache ca-certificates && \
+    mkdir -p /data && \
+    chown 10001:10001 /data
+
+FROM scratch
+
+LABEL org.opencontainers.image.title="CDK Mintd Hardened" \
+    org.opencontainers.image.description="Shell-less CDK mintd image with a static binary and scratch runtime"
+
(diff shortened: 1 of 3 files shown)


Dockerfile.hardened
```diff
@@ -0,0 +1,29 @@
...
+
+COPY --from=rootfs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=rootfs --chown=10001:10001 /data /data
+COPY --chown=0:0 --chmod=0555 cdk-mintd /usr/local/bin/cdk-mintd
+
+ENV CDK_MINTD_WORK_DIR=/data \
+    HOME=/data \
+    SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt \
+    SQLITE_TMPDIR=/data \
+    TMPDIR=/data
```
prusnak: Don't we want to create /tmp too and point SQLITE_TMPDIR and TMPDIR there?

Mixing data files and temp files together sounds bad.

asmogo: I think you are right. We probably don't have to create the directory at all. 
I'm going to update this. Thanks.

asmogo: fixed in [71a87ce](https://github.com/cashubtc/cdk/pull/2078/commits/71a87ce322a2cf3d52bb8af258319986d4d54e76)

Dockerfile.hardened
```diff
@@ -0,0 +1,32 @@
...
+
+LABEL org.opencontainers.image.title="CDK Mintd Hardened" \
+    org.opencontainers.image.description="Shell-less CDK mintd image with a static binary and scratch runtime"
+
+COPY --from=rootfs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=rootfs --chown=10001:10001 /data /data
+COPY --chown=0:0 --chmod=0555 cdk-mintd /usr/local/bin/cdk-mintd
+
+# Add tmpfs for temporary files (faster, ephemeral)
+RUN mkdir -p /tmp && chmod 1777 /tmp
```
thesimplekid: Will this work? Scratch has no shell right so can we use mdkir?

thesimplekid (changes requested):

Dockerfile.hardened
asmogo: You are right. we need to create the folder first and copy it from rootfs.

Dockerfile.hardened
```diff
@@ -0,0 +1,32 @@
...
+    chmod 1777 /tmp
+
+FROM scratch
+
+LABEL org.opencontainers.image.title="CDK Mintd Hardened" \
+    org.opencontainers.image.description="Shell-less CDK mintd image with a static binary and scratch runtime"
+
+COPY --from=rootfs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=rootfs --chown=10001:10001 /data /data
+COPY --from=rootfs /tmp /tmp
```
prusnak, line 18: do we need `--chmod=1777` here?

asmogo: Good catch. Docker will only copy the contents and preserve the file permissions, not the folder. I have tried this with  ```--chmod=1777``` as well, but it's still using default permissions when creating the target /tmp folder. 

I will come up with another solution to this issue.

PR #2079, opened 2026-06-10
Rework Docker release workflow ordering

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


.github/workflows/static-build-publish.yml
```diff
@@ -23,13 +23,15 @@ jobs:
         with:
           ref: ${{ github.event.inputs.tag || github.ref }}
 
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v14
+
```
asmogo, line 28: why is this line missing in current main ? seems like we don't need to install nix on these jobs.

PR #2082, opened 2026-06-10
Supabase fixes

thesimplekid:
### Description

-----

This finding was discovered by [Project Loupe](https://github.com/project-loupe/loupe).

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-supabase/Cargo.toml
+++ b/crates/cdk-supabase/Cargo.toml
@@ -30,7 +30,7 @@ url.workspace = true
 once_cell.workspace = true
 urlencoding = "2.1"
 aes-gcm = "0.10"
-scrypt = { version = "0.12", default-features = false }
+scrypt = { version = "0.12.0", default-features = false }
 paste.workspace = true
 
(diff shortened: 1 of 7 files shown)


crates/cdk-supabase/Cargo.toml
```diff
@@ -30,6 +30,7 @@ url.workspace = true
 once_cell.workspace = true
 urlencoding = "2.1"
 aes-gcm = "0.10"
+scrypt = { version = "0.11", default-features = false }
```
thesimplekid: It seems like the newest script is version 0.12 is there a reason to not used the newest?

asmogo: true. the lock files are pinning 0.11. i assumed that we want to keep that version.

https://github.com/cashubtc/cdk/actions/runs/27268829336/job/80667587897#step:4:438

i'll upgrade the dependency to the latest version.

asmogo (approved):

PR #2084, opened 2026-06-10
fix(wallet): stop persisting wallet seed bytes in npubcash quotes

thesimplekid:
add_npubcash_mint_quote attached seed[..32] as the quote secret key, writing raw master seed material to the mint_quote table in cleartext for every synced npubcash quote.

Derive the npubcash identity via NIP-06 (m/44'/1237'/0'/0/0) instead of using raw seed bytes, and stop persisting the key entirely: npubcash quotes are marked in the KV store and the NUT-20 signing key is re-derived from the seed at claim time, including saga resume paths.

This finding was discovered by [Project Loupe](https://github.com/project-loupe/loupe).

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/issue/mod.rs
+++ b/crates/cdk/src/wallet/issue/mod.rs
@@ -68,6 +68,26 @@ fn mint_quote_response_amount(response: &MintQuoteResponse<String>) -> Option<Am
 }
 
 impl Wallet {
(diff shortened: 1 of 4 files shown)


thesimplekid:
@codex review

PR #2096, opened 2026-06-13
Simplify keyset management to three core functions

crodas:
### Description

Replace 6+ overlapping keyset functions with a clean, minimal API:

- `keysets()`: Returns `Vec<KeySet>` (full keysets with keys included).  In automatic mode, refreshes from the mint with a graceful fallback to cached data on network failure. In manual mode (`set_keysets(Some(...))`), returns injected keysets directly without network access.

- `active_keyset()`: Returns the active `KeySet` with the lowest input fee.

- `keyset(id)`: Returns a single `KeySet` by ID.

- `set_keysets(Option<Vec<KeySet>>)`: Enables manual keyset mode, bypassing the
  metadata cache entirely. Pass `None` to return to automatic mode.

By returning full `KeySet` objects (which include `.keys`), the separate `load_keyset_keys(id)` function is no longer needed; callers access `.keys` directly from the keyset.

Removed from Wallet and WalletTrait:
  - load_keyset_keys()
  - refresh_keysets()
  - fetch_active_keyset()
  - get_active_keyset()
  - get_mint_keysets(KeysetFilter)
  - load_mint_keysets()

Also moves `metadata_cache_ttl` into `MintMetadataCache`, removing boilerplate TTL passing on every cache load call.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-ffi/src/wallet.rs
+++ b/crates/cdk-ffi/src/wallet.rs
@@ -653,15 +653,26 @@ impl Wallet {
         )))
     }
 
-    /// Refresh keysets from the mint
-    pub async fn refresh_keysets(&self) -> Result<Vec<KeySetInfo>, FfiError> {
-        let keysets = self.inner.refresh_keysets().await?;
+    /// Get all keysets for this wallet's unit
+    pub async fn keysets(
+        &self,
+        policy: Option<crate::types::wallet::KeysetLoadPolicy>,
+    ) -> Result<Vec<KeySet>, FfiError> {
+        let cdk_policy: cdk_common::wallet::KeysetLoadPolicy = policy.unwrap_or_default().into();
+        let keysets = self.inner.keysets(cdk_policy).await?;
         Ok(keysets.into_iter().map(Into::into).collect())
     }
 
-    /// Get the active keyset for the wallet's unit
-    pub async fn get_active_keyset(&self) -> Result<KeySetInfo, FfiError> {
-        let keyset = self.inner.get_active_keyset().await?;
+    /// Get the active keyset with the lowest fees
+    pub async fn active_keyset(&self) -> Result<KeySet, FfiError> {
+        let keyset = self.inner.active_keyset().await?;
+        Ok(keyset.into())
+    }
+
+    /// Get a single keyset by ID
+    pub async fn keyset(&self, keyset_id: String) -> Result<KeySet, FfiError> {
+        let id = cdk::nuts::Id::from_str(&keyset_id).map_err(FfiError::internal)?;
+        let keyset = self.inner.keyset(id).await?;
         Ok(keyset.into())
     }
 
@@ -669,34 +680,6 @@ impl Wallet {
     pub async fn get_keyset_fees_by_id(&self, keyset_id: String) -> Result<u64, FfiError> {
         let id = cdk::nuts::Id::from_str(&keyset_id).map_err(FfiError::internal)?;
         Ok(self.inner.get_keyset_fees_by_id(id).await?)
-    }
-
-    /// Load keys for a specific keyset
-    pub async fn load_keyset_keys(&self, keyset_id: String) -> Result<Keys, FfiError> {
-        let id = cdk::nuts::Id::from_str(&keyset_id).map_err(FfiError::internal)?;
-        let keys = self.inner.load_keyset_keys(id).await?;
-        Ok(keys.into())
-    }
-
-    /// Get keysets for this wallet's unit with filter
-    pub async fn get_mint_keysets(
-        &self,
-        filter: KeysetFilter,
-    ) -> Result<Vec<KeySetInfo>, FfiError> {
-        let keysets = self.inner.get_mint_keysets(filter.into()).await?;
-        Ok(keysets.into_iter().map(Into::into).collect())
-    }
-
-    /// Load active keysets
-    pub async fn load_mint_keysets(&self) -> Result<Vec<KeySetInfo>, FfiError> {
-        let keysets = self.inner.load_mint_keysets().await?;
-        Ok(keysets.into_iter().map(Into::into).collect())
-    }
-
-    /// Fetch active keyset with lowest fees
-    pub async fn fetch_active_keyset(&self) -> Result<KeySetInfo, FfiError> {
-        let keyset = self.inner.fetch_active_keyset().await?;
-        Ok(keyset.into())
     }
 
     /// Get fees and amounts for all keysets
--- a/crates/cdk-ffi/src/wallet_trait.rs
+++ b/crates/cdk-ffi/src/wallet_trait.rs
@@ -23,7 +23,7 @@ impl WalletTraitDef for Wallet {
     type MintUrl = MintUrl;
     type CurrencyUnit = CurrencyUnit;
     type MintInfo = MintInfo;
-    type KeySetInfo = KeySetInfo;
+    type KeySetInfo = KeySet;
     type MintQuote = MintQuote;
     type MeltQuote = MeltQuote;
     type PaymentMethod = PaymentMethod;
@@ -68,34 +68,24 @@ impl WalletTraitDef for Wallet {
         Ok(info.into())
     }
 
-    async fn refresh_keysets(&self) -> Result<Vec<Self::KeySetInfo>, Self::Error> {
-        let keysets = WalletTraitDef::refresh_keysets(self.inner().as_ref()).await?;
+    async fn keysets(
+        &self,
+        policy: cdk_common::wallet::KeysetLoadPolicy,
+    ) -> Result<Vec<Self::KeySetInfo>, Self::Error> {
+        let keysets = WalletTraitDef::keysets(self.inner().as_ref(), policy).await?;
         Ok(keysets.into_iter().map(Into::into).collect())
     }
 
-    async fn get_active_keyset(&self) -> Result<Self::KeySetInfo, Self::Error> {
-        let keyset = WalletTraitDef::get_active_keyset(self.inner().as_ref()).await?;
+    async fn active_keyset(&self) -> Result<Self::KeySetInfo, Self::Error> {
+        let keyset = WalletTraitDef::active_keyset(self.inner().as_ref()).await?;
         Ok(keyset.into())
     }
 
-    async fn load_keyset_keys(
+    async fn keyset(
         &self,
         keyset_id: cdk_common::nuts::Id,
-    ) -> Result<cdk_common::nuts::Keys, Self::Error> {
-        let keys = WalletTraitDef::load_keyset_keys(self.inner().as_ref(), keyset_id).await?;
-        Ok(keys)
-    }
-
-    async fn get_mint_keysets(
-        &self,
-        filter: cdk_common::wallet::KeysetFilter,
-    ) -> Result<Vec<Self::KeySetInfo>, Self::Error> {
-        let keysets = WalletTraitDef::get_mint_keysets(self.inner().as_ref(), filter).await?;
-        Ok(keysets.into_iter().map(Into::into).collect())
-    }
-
-    async fn fetch_active_keyset(&self) -> Result<Self::KeySetInfo, Self::Error> {
-        let keyset = WalletTraitDef::fetch_active_keyset(self.inner().as_ref()).await?;
+    ) -> Result<Self::KeySetInfo, Self::Error> {
+        let keyset = WalletTraitDef::keyset(self.inner().as_ref(), keyset_id).await?;
         Ok(keyset.into())
     }
 
--- a/crates/cdk/src/wallet/keysets.rs
+++ b/crates/cdk/src/wallet/keysets.rs
@@ -1,180 +1,229 @@
 use std::collections::HashMap;
 
 use cdk_common::amount::{FeeAndAmounts, KeysetFeeAndAmounts};
-use cdk_common::nut02::KeySetInfosMethods;
-pub use cdk_common::wallet::KeysetFilter;
+use cdk_common::wallet::KeysetLoadPolicy;
 use tracing::instrument;
 
-use crate::nuts::{Id, KeySetInfo, Keys, Proofs, Token};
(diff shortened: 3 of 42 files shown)


crates/cdk/src/wallet/send/saga/mod.rs
```diff
@@ -401,12 +399,6 @@ impl<'a> SendSaga<'a, Initial> {
             self.state_data.operation_id
         );
 
-        if opts.send_kind.is_online() {
-            if let Err(e) = self.wallet.refresh_keysets().await {
-                tracing::error!("Error refreshing keysets: {:?}. Using stored keysets", e);
-            }
-        }
-
         let keyset_fees = self.wallet.get_keyset_fees_and_amounts().await?;
```
thesimplekid: This can go online so this breaks the ability too choose offline send kind

crates/cdk/src/wallet/keysets.rs
```diff
@@ -1,169 +1,118 @@
...
-    #[instrument(skip(self))]
-    pub async fn refresh_keysets(&self) -> Result<Vec<KeySetInfo>, Error> {
-        tracing::debug!("Refreshing keysets from mint");
 
-        let keysets = self
+        let metadata = match self
             .metadata_cache
             .load_from_mint(&self.localstore, &self.client)
-            .await?
+            .await
```
thesimplekid: We never check the ttl so I think on hot paths we check with the mint every time?

crates/cdk/src/wallet/melt/mod.rs
```diff
@@ -739,7 +739,18 @@ impl Wallet {
...
+            .into_iter()
+            .map(|ks| KeySetInfo {
+                id: ks.id,
+                unit: ks.unit,
+                active: ks.active.unwrap_or(true),
+                input_fee_ppk: ks.input_fee_ppk,
+                final_expiry: ks.final_expiry,
+            })
+            .collect();
         println!("{:?}", keysets_info);
```
thesimplekid:
```suggestion
```

thesimplekid:
> ### Description
> 
> Replace 6+ overlapping keyset functions with a clean, minimal API:
> 
>     * `keysets()`: Returns `Vec<KeySet>` (full keysets with keys included).  In automatic mode, refreshes from the mint with a graceful fallback to cached data on network failure. In manual mode (`set_keysets(Some(...))`), returns injected keysets directly without network access.
> 
>     * `active_keyset()`: Returns the active `KeySet` with the lowest input fee.
> 
>     * `keyset(id)`: Returns a single `KeySet` by ID.
> 
>     * `set_keysets(Option<Vec<KeySet>>)`: Enables manual keyset mode, bypassing the
>       metadata cache entirely. Pass `None` to return to automatic mode.
> 
> 
> By returning full `KeySet` objects (which include `.keys`), the separate `load_keyset_keys(id)` function is no longer needed; callers access `.keys` directly from the keyset.
> 
> Removed from Wallet and WalletTrait:
> 
>     * load_keyset_keys()
> 
>     * refresh_keysets()
> 
>     * fetch_active_keyset()
> 
>     * get_active_keyset()
> 
>     * get_mint_keysets(KeysetFilter)
> 
>     * load_mint_keysets()
> 
> 


I think I would change `keysets()` to accept an `online` parameter this way for the `SendOffline` we can make sure it does not send a request to the mint. I would also make `keysets()` respect the ttl so if the data it has is newer then ttl it does not send a network request, this would address the hot path issue. 

I would also keep `refresh_keysets()` this would refresh the keysets regardless of cache ttl and could be used by wallet at start or via some action where the wallet wants to force a refresh but cdk would not use it internally we would use the `keysets()`. 

What is the use case of the `set_keysets()` I am not sure why a wallet would want to set this manually?

thesimplekid:
Or better we could avoid the `refresh_keysets` fn by making the `keysets()` accept an enum: 

```rust
enum KeysetLoadPolicy {
      CacheOnly,
      CacheThenNetwork,
      Refresh,
  }
```

With the default CacheThenNetwork, `Cache` used for the offline send. Cache here also includes the wallets local DB.

crates/cdk-ffi/src/wallet.rs
```diff
@@ -603,50 +603,47 @@ impl Wallet {
...
+    pub fn set_keysets(&self, keysets: Option<Vec<KeySet>>) -> Result<(), FfiError> {
+        let cdk_keysets = keysets
+            .map(|ks| {
+                ks.into_iter()
+                    .map(cdk::nuts::KeySet::try_from)
+                    .collect::<Result<Vec<_>, _>>()
+            })
+            .transpose()?;
+        self.inner.set_keysets(cdk_keysets);
+        Ok(())
```
thesimplekid: I thought on the call we said we would remove this as we were unsure of the use case for it and could add it later?

crates/cdk-ffi/src/wallet_trait.rs
```diff
@@ -68,34 +68,28 @@ impl WalletTraitDef for Wallet {
...
     }
 
-    async fn load_keyset_keys(
-        &self,
-        keyset_id: cdk_common::nuts::Id,
-    ) -> Result<cdk_common::nuts::Keys, Self::Error> {
-        let keys = WalletTraitDef::load_keyset_keys(self.inner().as_ref(), keyset_id).await?;
-        Ok(keys)
+    fn set_keysets(&self, keysets: Option<Vec<cdk_common::KeySet>>) {
+        WalletTraitDef::set_keysets(self.inner().as_ref(), keysets);
```
thesimplekid: https://github.com/cashubtc/cdk/pull/2096/changes#r3440863456

crates/cdk/src/wallet/keysets.rs
```diff
@@ -1,180 +1,176 @@
...
 use crate::{Error, Wallet};
 
 impl Wallet {
-    /// Load keys for mint keyset
+    /// Get all keysets for this wallet's unit.
     ///
-    /// Returns keys from metadata cache if available.
-    /// If keys are not cached, fetches from mint server.
+    /// - **Manual mode** (`set_keysets(Some(...))`): returns the injected keysets
+    ///   directly — no network or cache access.
```
thesimplekid: I thought on call we said we would delete this manual mode?

crates/cdk/src/wallet/keysets.rs
```diff
@@ -1,180 +1,176 @@
...
-    /// and available amounts. Uses cached data if available, fetches from mint if not.
-    pub async fn get_keyset_fees_and_amounts(&self) -> Result<KeysetFeeAndAmounts, Error> {
-        let metadata = self
-            .metadata_cache
-            .load(&self.localstore, &self.client, {
-                let ttl = self.metadata_cache_ttl.read();
-                *ttl
+    /// Tokens may contain proofs from inactive keysets. Inactive keysets no
+    /// longer issue new signatures, but mints can still accept their proofs for
+    /// redemption.
```
thesimplekid: I think this comment is a bit outdated not that its not using the `KeysetFilter::Active` filter.

crates/cdk/src/wallet/keysets.rs
```diff
@@ -1,180 +1,176 @@
...
+        F: Fn() -> Fut,
+        Fut: std::future::Future<Output = Result<T, Error>>,
+    {
+        match f().await {
+            Err(Error::InactiveKeyset) => {
+                tracing::info!(
+                    "Mint rejected outputs with inactive keyset, refreshing keysets and retrying"
+                );
+                let _ = self.keysets(KeysetLoadPolicy::Refresh).await;
+                f().await
```
thesimplekid: Should we check if we actually got new keysets and log if so? If we do not I don't think the operation should be retried.

crates/cdk/src/wallet/keysets.rs
```diff
@@ -1,180 +1,176 @@
...
-    pub async fn refresh_keysets(&self) -> Result<Vec<KeySetInfo>, Error> {
-        tracing::debug!("Refreshing keysets from mint");
+        let metadata = match policy {
+            KeysetLoadPolicy::CacheOnly => {
+                self.metadata_cache.load_cached(&self.localstore).await?
+            }
+            KeysetLoadPolicy::CacheThenNetwork => {
+                self.metadata_cache
+                    .load(&self.localstore, &self.client)
+                    .await?
```
thesimplekid, line 25: Marker

crates/cdk/src/wallet/mint_metadata_cache.rs
thesimplekid, line 450: We document `KeysetLoadPolicy::CacheThenNetwork` as refreshing the keysets if they are older then the ttl however when that is the policy we call this `load()` ([here](https://github.com/cashubtc/cdk/pull/2096/changes#diff-8cdfa94edfa1ce249ac67609c0cf9ded9d9913060169369428f5e0b3dd9af4aaR38-R41)) but this seem to return before refreshing with the mint if there is data in the db, so when does the refresh with the mint happen?

thesimplekid: Should we update the time here? Shouldn't the `updated_at` be the time we got it from the mint so we know when we should refresh with the mint?

thesimplekid (changes requested):
Can we add a test for how a wallet handle a receive that includes proofs from inactive keysets? I have seen issues with this and hope this PR fixes it or at least does not regress after https://github.com/cashubtc/cdk/pull/2114

crates/cdk-ffi/src/wallet.rs
crodas: Yes we did. I'll make it happen now

crates/cdk/src/wallet/mint_metadata_cache.rs
crodas: Good catch, I fixed in c8797339

crates/cdk/src/wallet/keysets.rs
crodas: Fixed in 9b87dc50509899978c9d187119cae878d89bc47b

crates/cdk/src/wallet/send/saga/mod.rs
```diff
@@ -401,13 +399,48 @@ impl<'a> SendSaga<'a, Initial> {
...
-        if opts.send_kind.is_online() {
-            if let Err(e) = self.wallet.refresh_keysets().await {
-                tracing::error!("Error refreshing keysets: {:?}. Using stored keysets", e);
-            }
-        }
+        let keyset_policy = if opts.send_kind.is_offline() {
+            KeysetLoadPolicy::CacheOnly
+        } else {
+            KeysetLoadPolicy::default()
+        };
```
thesimplekid: Here we set the policy to be Cache only if offline but later we call [prepare_internal](https://github.com/cashubtc/cdk/pull/2096/changes#diff-34d47b41368a93fb08ff57d4450f983d5b1ab0ccdf4f893e80369762c803f22cL559) that uses `keyset_active()` which uses the default policy regardless of what is set here so it can go online again violating the offline send.

crates/cdk/src/wallet/send/saga/mod.rs
```diff
@@ -566,7 +585,7 @@ impl<'a> SendSaga<'a, Initial> {
         proofs: Proofs,
         force_swap: bool,
     ) -> Result<SendSaga<'a, Prepared>, Error> {
-        let active_keyset_id = self.wallet.get_active_keyset().await?.id;
+        let active_keyset_id = self.wallet.active_keyset().await?.id;
```
thesimplekid: Marker

crates/cdk/src/wallet/keysets.rs
```diff
@@ -1,180 +1,189 @@
...
+    /// Filters the output of [`keysets()`](Self::keysets) for active keysets
+    /// and returns the one with the minimum `input_fee_ppk`.
+    #[instrument(skip(self))]
+    pub async fn active_keyset(&self) -> Result<KeySet, Error> {
+        self.keysets(Default::default())
+            .await?
+            .into_iter()
+            .filter(|k| k.active.unwrap_or(false))
+            .min_by_key(|k| k.input_fee_ppk)
+            .ok_or(Error::NoActiveKeyset)
```
thesimplekid, line 95: Marker

crates/cdk/src/wallet/keysets.rs
```diff
@@ -1,180 +1,189 @@
...
+
+                let new_active =
+                    self.keysets(KeysetLoadPolicy::Refresh)
+                        .await
+                        .ok()
+                        .and_then(|ks| {
+                            ks.into_iter()
+                                .find(|k| k.active.unwrap_or(false))
+                                .map(|k| k.id)
+                        });
```
thesimplekid, line 135: The selection here should account for the fee like we do https://github.com/cashubtc/cdk/pull/2096/changes#r3459368988. In spec the fee can change so we need to account for this

crates/cdk/src/wallet/keysets.rs
```diff
@@ -1,180 +1,189 @@
...
-    #[instrument(skip(self))]
-    pub async fn refresh_keysets(&self) -> Result<Vec<KeySetInfo>, Error> {
-        tracing::debug!("Refreshing keysets from mint");
+    /// Use this around any code path that selects the active keyset, builds
+    /// outputs, and posts them to the mint.
+    pub(crate) async fn retry_on_inactive_keyset<F, Fut, T>(&self, f: F) -> Result<T, Error>
+    where
+        F: Fn() -> Fut,
+        Fut: std::future::Future<Output = Result<T, Error>>,
+    {
```
thesimplekid, line 108: I do not think we have tests where the wallet tries a swap and the mint returns an inactive keyset where this retry path gets used. It would be good to handle this.

thesimplekid (changes requested):

crates/cdk/src/wallet/keysets.rs
crodas: ff091ba7c3234770b4b904e6b8c9f91cc0680708

thesimplekid (approved):

PR #2100, opened 2026-06-14
fix(payment-processor): gate plaintext gRPC behind explicit opt-in

prusnak:
Require CDK_PAYMENT_PROCESSOR_ALLOW_INSECURE=true before starting the payment processor gRPC server without a TLS directory. This removes the silent fallback to plaintext when CDK_PAYMENT_PROCESSOR_TLS_DIR is unset, while still allowing operators to intentionally run insecure plaintext gRPC.


thesimplekid:
Duplicate of https://github.com/cashubtc/cdk/pull/2102

prusnak:
Superseded by https://github.com/cashubtc/cdk/pull/2138

PR #2101, opened 2026-06-14
fix(signatory): gate plaintext gRPC behind explicit opt-in

prusnak:
Require CDK_SIGNATORY_ALLOW_INSECURE=true before using signatory gRPC without TLS. This prevents the signatory server and remote client from silently falling back to unauthenticated plaintext transport when TLS configuration is absent.

Apply the same opt-in to the incoming-stream server helper so all plaintext signatory RPC entry points are explicitly acknowledged by the operator.

### DIFF
--- a/crates/cdk-signatory/src/proto/mod.rs
+++ b/crates/cdk-signatory/src/proto/mod.rs
@@ -2,5 +2,20 @@ mod convert;
 
 tonic::include_proto!("signatory");
 
+pub(crate) const ENV_SIGNATORY_ALLOW_INSECURE: &str = "CDK_SIGNATORY_ALLOW_INSECURE";
+
+pub(crate) fn allow_insecure_signatory_rpc() -> std::io::Result<bool> {
+    match std::env::var(ENV_SIGNATORY_ALLOW_INSECURE) {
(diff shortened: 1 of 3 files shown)


crates/cdk-signatory/src/proto/mod.rs
```diff
@@ -2,5 +2,20 @@ mod convert;
...
+            std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                format!("{ENV_SIGNATORY_ALLOW_INSECURE} must be true or false: {err}"),
+            )
+        }),
+        Err(std::env::VarError::NotPresent) => Ok(false),
+        Err(err) => Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, err)),
+    }
+}
+
```
thesimplekid, line 19: I think we should move this check up a level to within the cdk-mintd config and allow it to be set both within the config.toml and ENV VARS. This pattern of reading env vars from within a sub crate is not one we have and don't think we should introduce it.

thesimplekid (changes requested):

prusnak:
Superseded by https://github.com/cashubtc/cdk/pull/2141

PR #2102, opened 2026-06-14
fix(mint-rpc): gate plaintext management RPC behind explicit opt-in

prusnak:
Require CDK_MINTD_MANAGEMENT_ALLOW_INSECURE=true before starting the management RPC server without TLS. Keep mTLS as the default posture, surface invalid opt-in values as startup errors, and document the plaintext development override.

### DIFF
--- a/crates/cdk-mint-rpc/src/proto/server.rs
+++ b/crates/cdk-mint-rpc/src/proto/server.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::str::FromStr;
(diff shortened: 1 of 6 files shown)


prusnak:
Updated second commit to include fallbacks for old variables

crates/cdk-mint-rpc/src/proto/server.rs
```diff
@@ -188,6 +202,19 @@ impl MintRPCServer {
...
+            std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                format!("{ENV_MINT_MANAGEMENT_ALLOW_INSECURE} must be true or false: {err}"),
+            )
+        }),
+        Err(env::VarError::NotPresent) => Ok(false),
+        Err(err) => Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, err)),
+    }
+}
+
```
thesimplekid, line 217: Same comment as https://github.com/cashubtc/cdk/pull/2101#discussion_r3437740865 where we shouldn't read the env vars within these crates and should push the config up to mint.

thesimplekid (changes requested):
I like both this and https://github.com/cashubtc/cdk/pull/2101. It is a good idea to be explicit about allowing the insecure. Since they are closely related If it is easier for you to combine them that is fine with me or keep them as separate PRs, up to you.

prusnak:
Superseded by https://github.com/cashubtc/cdk/pull/2138

PR #2108, opened 2026-06-15
fix(cdk-axum): validate batch mint quote methods before status checks

prusnak:
Prevent `/v1/mint/quote/{method}/check` from returning quote status for quotes that belong to a different payment method than the method in the URL.

The batch custom mint quote check path authenticated the requested route, but then passed all supplied quote IDs directly to `check_mint_quotes`. For custom methods, the response mapping only checked that each returned quote was custom, not that it matched the URL method. This allowed a caller authorized for one custom method route to query known quote IDs from another custom method.

Reuse the existing mint quote method validation before loading quote statuses, matching the method-scoped behavior already used by mint requests.

Add a regression test that creates a `venmo` custom mint quote and verifies it cannot be checked through the `paypal` batch quote check route.

### DIFF
--- a/crates/cdk-axum/src/custom_handlers.rs
+++ b/crates/cdk-axum/src/custom_handlers.rs
@@ -356,6 +356,10 @@ pub async fn post_batch_check_mint_quote(
         .await
         .map_err(into_response)?;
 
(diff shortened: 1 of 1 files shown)


thesimplekid (approved):
Thank you

PR #2109, opened 2026-06-15
fix(cdk-axum): require swap auth before serving cached responses

prusnak:
Move `/v1/swap` auth verification ahead of the NUT-19 cache lookup.

The generated cache wrapper for `post_swap` calculated the cache key and returned cached `SwapResponse` values before calling the wrapped handler. Since `post_swap` performed `verify_auth` internally, a caller who replayed an identical cached swap body could receive the cached response without presenting or consuming auth on protected swap routes.

Replace the generated swap cache wrapper with a swap-specific implementation that verifies route auth first, then checks the cache. On cache misses, the wrapper processes the swap and stores the response as before. The swap path is split into small helpers so auth is not checked twice on misses and blind auth tokens are consumed exactly once per request.

Add a regression test that preloads a cached swap response for a clear-auth protected swap route and verifies an unauthenticated request is rejected before the cached response can be returned.

### DIFF
--- a/crates/cdk-axum/src/router_handlers.rs
+++ b/crates/cdk-axum/src/router_handlers.rs
@@ -10,7 +10,6 @@ use cdk::nuts::{
     RestoreRequest, RestoreResponse, SwapRequest, SwapResponse,
 };
 use cdk::util::unix_time;
-use paste::paste;
 use tracing::instrument;
 
 use crate::auth::AuthHeader;
@@ -84,8 +83,6 @@ macro_rules! post_cache_wrapper_with_prefer {
         }
     };
 }
-
-post_cache_wrapper!(post_swap, SwapRequest, SwapResponse);
 
 /// Get the public keys of the newest mint keyset
 ///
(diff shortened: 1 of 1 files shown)


thesimplekid:
I think this is intended. A user should not have to burn a second auth to get the request from cache. For example they only have one auth token for some reason they attempt an operation but get an http error they should be able to hit the cache to complete this without a second token.

prusnak:
I understand the retry concern, especially for blind auth, but I don't think the only options are “consume a second BAT” or “serve the cached response without authentication.”

Today the swap cache key contains only the request payload, and the cache is consulted before `post_swap`. Consequently, possession of an exact previously successful swap body is sufficient to retrieve the cached `SwapResponse`, even when the route is configured as protected. The practical fund-theft risk may be limited because the outputs are blinded, but this still differs from the expected semantics of a protected endpoint.

The regression test in this PR covers clear auth, where revalidating the CAT does not consume it. Blind auth is harder: CDK's automatic NUT-19 retries reuse the same BAT, so calling the current `verify_auth` again would reject that BAT as already spent. I agree that an exact replay must not consume a second BAT.

However, the swap saga is not a substitute for the cache. Once the first swap has finalized, replaying it through `process_swap_request` fails on the already-spent inputs; recovery comes from the NUT-19 cache or `/restore`, not from saga-level idempotency.

There is also an implementation inconsistency: the cached mint and melt handlers validate auth before cache lookup, while swap's generic wrapper does not. `/restore` validates auth but is not itself cached.

Could we distinguish initial authorization from idempotent replay? For clear auth, verify the CAT before serving a cache hit. For blind auth, the cache could record a fingerprint of the BAT that authorized the original request and allow that same BAT to retrieve the cached response without attempting to consume it again. If possession of the exact request body is intentionally meant to serve as the replay credential, I think that interaction between NUT-19 and NUT-21/22 should be explicitly documented and tested.

PR #2113, opened 2026-06-16
fix: treat mint limit errors as definitive during issue recovery

thesimplekid:
Issue saga recovery now stops on definitive replay failures such as max output/input rejections instead of falling back to restore. Decode those mint error responses into structured variants so recovery preserves and surfaces the original failure.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@codex review

PR #2119, opened 2026-06-17
feat: align mint quote accounting with NUT-04

thesimplekid:
Implement the mint quote accounting changes from cashubtc/nuts#377 by exposing amount_paid, amount_issued, and updated_at on mint quote responses.

Keep the deprecated state field only for BOLT11 compatibility while deriving local quote state from the canonical counters for all mint methods. Persist updated_at in wallet quote storage so stale responses cannot move amount_paid or amount_issued backwards.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk/src/wallet/streams/payment.rs
```diff
@@ -207,17 +241,18 @@ impl<'a> PaymentStream<'a> {
...
                 // the future will be created in subscription_receiver_future.
                 self.active_subscription = Some(subscription);
                 self.cancellation_future = None; // resets timeout
                 match notification {
-                    None => {
+                    None if is_complete => {
                         self.is_finalized = true;
                         Poll::Ready(None)
                     }
+                    None => self.poll_event(cx),
```
asmogo: We should also implement a recursion limit and a sleep interval between events to reduce the load.

crates/cdk-common/src/mint_quote.rs
```diff
@@ -146,7 +146,13 @@ impl<Q> MintQuoteResponse<Q> {
     /// Returns the quote state derived from the response data.
     pub fn state(&self) -> Option<QuoteState> {
         match self {
-            Self::Bolt11(r) => Some(r.state),
+            Self::Bolt11(r) => {
+                if r.amount_paid > Amount::ZERO || r.amount_issued > Amount::ZERO {
+                    Some(quote_state_from_amounts(r.amount_paid, r.amount_issued))
+                } else {
+                    Some(r.state)
+                }
```
asmogo, line 172: Why are we using the state here ? I think we want to deprecate this. 

```suggestion
                    Some(quote_state_from_amounts(r.amount_paid, r.amount_issued))
```

crates/cdk-common/src/mint_quote.rs
asmogo, line 208:
```suggestion
if amount_paid == Amount::ZERO {
  return QuoteState::Unpaid;
}
if amount_issued == Amount::ZERO {
  return QuoteState::Paid;
}

```

asmogo (changes requested):

crates/cdk/src/wallet/streams/payment.rs
crodas: If I remember correctly, this is fine because the next future would exit the call stack with a [Pending](https://github.com/thesimplekid/cdk/blob/d5257c10293c2f1cdcc2c3f5d8674b2a6dca6d35/crates/cdk/src/wallet/streams/payment.rs#L300-L303)

thesimplekid: It looks like the recursion limit issue you bring up is valid but as @crodas said its "mostly" okay based on the next future. However, I still think we should clean it up but as a separate focused PR as its an existing pattern not introduced by this pr.

crates/cdk-common/src/mint_quote.rs
thesimplekid: The bolt11 portion highlighted in the comments we need for backwards compatibility with mints that have not updated. We could move this logic of computing the amount_paid and amount_issued to the deserlizer and that would remove some of this logic. But I don't love computing fields in the serialize/deserialized but maybe for this it makes sense?

crates/cdk-common/src/mint_quote.rs
thesimplekid: This is not a state that should happen but I cleaned it up in [a830f779](https://github.com/cashubtc/cdk/pull/2119/commits/841b53ce3739816769bfd2a2c4930762a733a1c6).

thesimplekid:
@cdk-bot review

crates/cashu/src/nuts/nut17/mod.rs
```diff
@@ -188,6 +188,10 @@ where
 #[serde(bound(serialize = "T: Serialize + DeserializeOwned"))]
```
j-kon, line 188: Looks good, verified Bolt11/Bolt12/Onchain quote structs.

j-kon (approved):

crates/cdk-mint-rpc/src/proto/server.rs
```diff
@@ -758,6 +758,7 @@ impl CdkMint for MintRPCServer {
                     mint_quote.amount_paid(),             // amount_paid
                     mint_quote.payment_method.clone(),    // method
                     0,                                    // created_at
+                    0,                                    // updated_at
```
j-kon, line 761: Is this update necessary? Since it is always 0 in the test.

thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

thesimplekid:
@asmogo I believe your requested changes are addressed could you re-review.

thesimplekid:
@cdk-bot review

crodas (dismissed):

crates/cdk/src/wallet/streams/payment.rs
asmogo: should we still use this state for bolt11, considering it’s deprecated?

asmogo, line 347: should we still use this state for bolt11, considering it’s deprecated?

crates/cdk/src/wallet/streams/payment.rs
```diff
@@ -29,6 +32,83 @@ struct NextPayment {
...
+            let quote_id = info.quote.clone();
+            if let Ok(Some(mut quote)) = localstore.get_mint_quote(&quote_id).await {
+                let applied = apply_mint_quote_response(
+                    &mut quote,
+                    &cdk_common::MintQuoteResponse::Bolt11(info.clone()),
+                );
+                if applied {
+                    if let Err(e) = localstore.add_mint_quote(quote).await {
+                        tracing::warn!("Failed to update quote state: {}", e);
+                    }
```
asmogo: why are we just logging a warning here instead of returning false ?

crates/cdk/src/wallet/streams/payment.rs
thesimplekid: This is melt quote where we still use the state

thesimplekid: Yes addressed int 6dcd475f

crates/cdk/src/wallet/streams/payment.rs
thesimplekid: Addressed in e2c0d0a4

thesimplekid:
@cdk-bot review

asmogo (approved):

crodas (approved):

PR #2124, opened 2026-06-17
fix(cdk-integration-tests): migrate regtest LND payments to SendPaymentV2

prusnak:
builds on top of https://github.com/cashubtc/cdk/pull/2127 - closing until that PR is merged

cc @lescuer97 

-----

Replace the integration-test LND client's use of the deprecated Lightning.SendPaymentSync RPC with Router.SendPaymentV2.

LND v0.21 removes SendPaymentSync and the corresponding legacy payment endpoint, so keeping the regtest helper on the old RPC would break local regtest and integration tests once the Nix-provided LND package advances past 0.20.x. The production cdk-lnd backend already uses SendPaymentV2; this change brings the test-side payer in line with that path.

The helper now builds a routerrpc::SendPaymentRequest, starts the SendPaymentV2 stream through the router client, and consumes updates until a terminal payment status is observed. It returns the payment preimage on SUCCEEDED, continues through IN_FLIGHT and INITIATED updates, and fails explicitly for FAILED, UNKNOWN, unrecognized statuses, or an exhausted stream.

Bump fedimint-tonic-lnd from 0.2.0 to 0.4.0 so the generated LND proto includes the newer INITIATED payment status. That lets the helper convert the raw protobuf status with PaymentStatus::try_from(...) and match named enum variants instead of carrying a hard-coded status value. The newer generated request structs also get ..Default::default() where appropriate so added proto fields do not make the test client brittle.

Set no_inflight_updates=true to match the simple synchronous behavior the test helper needs, and set timeout_seconds=60 to make the request explicit. Use fee_limit_msat=i64::MAX so the regtest payer preserves the old broad payment behavior instead of accidentally limiting itself to zero-fee routes, which is SendPaymentV2's default when no fee limit is provided.

This affects only the regtest/integration-test LND client; the production mint LND backend was already migrated in https://github.com/cashubtc/cdk/pull/1656.


prusnak:
will reopen after https://github.com/cashubtc/cdk/pull/2127 is merged

PR #2126, opened 2026-06-18
Copy Cargo.lock to binding target repos for reproducible builds

### LINKED ISSUE #2097: Build bindings release artifacts from the CDK lockfile
During the latest bindings release, CI passed but the publish workflows failed because a bad compatible release of the time crate was published and then quickly yanked. Rerunning after the yank fixed the release.

This exposed a reproducibility issue: the bindings release workflows can build from a different dependency graph than the one CI tested.

Currently, some publish workflows copy bindings/<lang>/ into the downstream bindings repo, rewrite rust Cargo.toml to depend on cdk-ffi = "=<version>" from crates.io, and then run release builds there. Because those builds are not tied to the CDK monorepo Cargo.lock, Cargo can resolve newer semver-compatible transitive dependencies than CI used.

That means a semver-compatible dependency published between CI and release can break the release, or produce native artifacts that were not built from the CI-tested dependency graph.

We should change the bindings publish workflows so native Rust artifacts are built from the matching cashubtc/cdk tag using the root Cargo.lock and --locked, then copy the generated bindings and native libraries into the downstream repos for publishing.

This would make releases match the CI-tested dependency graph and likely reduce the amount of Rust Cargo.toml rewriting in the workflows.

Standalone downstream builds can still be supported, but they should not be the source of truth for release artifacts. We can document that published binaries are built from the matching CDK tag and lockfile, while downstream rust/ wrappers are for local development.

crodas:
### Description

Fixes #2097

### Problem

Binding releases resolved their dependencies from scratch on every run and on
every matrix leg. Nothing tied the crates a release compiled against to the
versions the monorepo had pinned, so the same tag rebuilt later could ship
different transitive versions, and afterward there was no record of what it
had shipped.

Go was in worse shape than the rest: it compiled `cdk-ffi` straight from the
monorepo checkout, so `cdk-go` published a wrapper crate that nothing built,
and the release recorded no binding lockfile at all.

### What this does

Every release now seeds the downstream binding crate's `Cargo.lock` from a
committed reference lockfile, builds each leg with `--locked`, and installs the
compiler from `rust-toolchain.toml` instead of whatever stable is current.

`.github/scripts/seed-binding-lockfile.sh` copies the reference lock into the
binding crate and resolves it minimally with `cargo fetch`, so only `cdk-ffi`
moves from a path dependency to a registry or git source. If the binding
resolves anything the reference lock did not pin, the release stops there
rather than shipping it.

### Which lockfile is the reference

Not the release checkout. A release pins `cdk-ffi` to a published version, and
that artifact's requirements were frozen when it was published. The reference
is therefore the lockfile at the tag that published it:

```mermaid
flowchart LR
  A["release tag<br/>binding sources, versions"] --> C["downstream release branch"]
  B["tag that published cdk-ffi<br/>Cargo.lock"] --> C
```

Both are the same tag for an ordinary release. They differ when only the
bindings change and the release pins an older published `cdk-ffi`, which is
exactly the case that used to fail: seeding from the newer tree downgraded the
nostr subtree to satisfy the older `cdk-ffi` and every downgrade was reported
as drift.

The workflows derive that tag from `cdk_version`, check it out alongside the
release checkout, and refuse to run if it does not exist or does not declare
the version it claims. Nightlies are unchanged: they build `cdk-ffi` from the
checked-out commit, so that checkout's own lock is already the right reference.

### Provenance

Each release ships a `build-manifest.json` recording the source commit, the
`cdk-ffi` version it pins, the ref its lockfile was seeded from, the pinned
Rust channel, and hashes of the lockfiles. `FFI - Verify Published Bindings`
re-checks published releases against that record weekly, so drift surfaces
without waiting for the next release.

### Go

`cdk-ffi-go` is now built from the seeded lockfile on the release branch like
the other bindings, which also makes the released library the one CI already
exercises through Nix. That costs a rename to `libcdk_ffi_go`.

## Also in here

- `cdk-sql-common`: `read_dir` order leaked into the generated `MIGRATIONS`
  array whenever two migrations shared a version prefix. Name parsing moved to
  `migration_name.rs` and the ordering is now explicit.
- `flake.nix`: the Go build passes `--locked`.
- `just ffi-verify-lock <language> <tag>` and `ffi-verify-lock-all <tag>` to run
  the verification locally.

## Verifying

```bash
git checkout <source commit from the manifest>
just ffi-verify-lock dart v0.18.0
just ffi-verify-lock-all v0.18.0
```

A dart release run has already exercised the two-tree path end to end:
`release_tag v0.19.0` with `cdk_version 0.18.0` seeded from `refs/tags/v0.18.0`
and pushed a release branch whose lock carries the `cdk-ffi 0.18.0` dependency
graph.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/scripts/verify-binding-lockfile.sh
+++ b/.github/scripts/verify-binding-lockfile.sh
@@ -1,0 +1,90 @@
+#!/usr/bin/env bash
+#
+# Checks, after the fact, that a published binding release resolved only
+# dependency versions that the monorepo's Cargo.lock pinned at the commit the
+# release claims to come from.
+#
+# Usage: verify-binding-lockfile.sh <language> <release-tag>
+#   run from the monorepo checkout root. Needs `gh` and a checkout at the
+#   source commit recorded in the release's build-manifest.json.
+
+set -euo pipefail
+
+LANGUAGE="${1:?usage: verify-binding-lockfile.sh <language> <release-tag>}"
+TAG="${2:?missing release tag}"
+
+case "${LANGUAGE}" in
+  dart|swift|kotlin|go) ;;
+  *) echo "unknown language: ${LANGUAGE}" >&2; exit 2 ;;
+esac
+
+REPO="cashubtc/cdk-${LANGUAGE}"
+workdir="$(mktemp -d)"
+trap 'rm -rf "${workdir}"' EXIT
+
+git clone --quiet --depth 1 --branch "${TAG}" \
+  "https://github.com/${REPO}.git" "${workdir}/downstream"
+
+MANIFEST="${workdir}/downstream/build-manifest.json"
+if [[ ! -f "${MANIFEST}" ]]; then
+  echo "::error::${REPO}@${TAG} has no build-manifest.json; it predates provenance recording"
+  exit 1
+fi
+
+CLAIMED_COMMIT="$(jq -r '.source.commit' "${MANIFEST}")"
+HEAD_COMMIT="$(git rev-parse HEAD)"
+if [[ "${CLAIMED_COMMIT}" != "${HEAD_COMMIT}" ]]; then
+  echo "::error::checkout mismatch: release was built from ${CLAIMED_COMMIT}, this tree is ${HEAD_COMMIT}"
+  echo "Run: git checkout ${CLAIMED_COMMIT}"
+  exit 1
+fi
+
+CLAIMED_WS_LOCK="$(jq -r '.lockfiles.workspace_cargo_lock_sha256' "${MANIFEST}")"
+ACTUAL_WS_LOCK="$(shasum -a 256 Cargo.lock 2>/dev/null | cut -d' ' -f1)"
+if [[ "${CLAIMED_WS_LOCK}" != "${ACTUAL_WS_LOCK}" ]]; then
+  echo "::error::workspace Cargo.lock hash mismatch"
+  echo "  manifest: ${CLAIMED_WS_LOCK}"
+  echo "  checkout: ${ACTUAL_WS_LOCK}"
+  exit 1
+fi
+
+BINDING_LOCK="${workdir}/downstream/rust/Cargo.lock"
+if [[ ! -f "${BINDING_LOCK}" ]]; then
+  # Releases cut before a language started seeding its lockfile have none, and
+  # the workspace lock hash checked above is then the whole guarantee.
+  echo "${LANGUAGE}: no downstream lockfile; verified against the workspace lock."
+  exit 0
+fi
+
+lock_versions() {
+  awk -F'"' '/^name = /{n=$2} /^version = /{print n" "$2}' "$1" \
+    | grep -v '^cdk-ffi' \
+    | sort -u
+}
+
+# Releases that pin an older cdk-ffi than their own tag were seeded from that
+# cdk-ffi's tag, so that is what the binding lock has to be compared against.
+LOCK_REF="$(jq -r '.lockfiles.binding_cargo_lock_source_ref // empty' "${MANIFEST}")"
+REFERENCE_LOCK="${workdir}/reference.lock"
+if [[ -n "${LOCK_REF}" && "${LOCK_REF}" != "${HEAD_COMMIT}" ]]; then
+  if ! git fetch --quiet --depth 1 origin "${LOCK_REF}"; then
+    echo "::error::cannot fetch ${LOCK_REF}, the ref the binding lockfile was seeded from"
+    exit 1
+  fi
+  git show FETCH_HEAD:Cargo.lock > "${REFERENCE_LOCK}"
+else
+  LOCK_REF="${CLAIMED_COMMIT}"
+  cp Cargo.lock "${REFERENCE_LOCK}"
+fi
+
+lock_versions "${REFERENCE_LOCK}" > "${workdir}/workspace.txt"
+lock_versions "${BINDING_LOCK}" > "${workdir}/binding.txt"
+
+drift="$(comm -13 "${workdir}/workspace.txt" "${workdir}/binding.txt")"
+if [[ -n "${drift}" ]]; then
+  echo "::error::${REPO}@${TAG} used dependencies the lockfile at ${LOCK_REF} did not pin"
+  echo "${drift}" | sed 's/^/  /'
+  exit 1
+fi
+
+echo "${LANGUAGE} ${TAG}: all $(wc -l < "${workdir}/binding.txt" | tr -d ' ') dependencies match the lockfile at ${LOCK_REF}."
--- a/.github/scripts/write-build-manifest.sh
+++ b/.github/scripts/write-build-manifest.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+#
+# Records what a binding release was built from, so the dependency graph and
+# toolchain can be checked after the fact without re-running the release.
+#
+# Usage: write-build-manifest.sh <language> <downstream-dir> <output-path>
+#   run from the monorepo checkout root. Reads TAG, CDK_REF, CDK_VER, NIGHTLY
+#   and GITHUB_REPOSITORY from the environment.
+
+set -euo pipefail
+
+LANGUAGE="${1:?usage: write-build-manifest.sh <language> <downstream-dir> <output-path>}"
+DOWNSTREAM_DIR="${2:?missing downstream dir}"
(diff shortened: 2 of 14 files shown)


thesimplekid:
This copies the lockfile, but it does not make the release fail if Cargo has to update dependency resolution. For reproducible release builds we should also use `--locked` on the project `cargo build`, `cargo run`, and `cross build` steps.

thesimplekid (changes requested):
This currently copies the monorepo Cargo.lock before rewriting rust/Cargo.toml from workspace/path dependencies to registry dependencies. After that rewrite, the copied lockfile no longer matches the detached package manifest, so the later `cargo build --release --locked` steps will fail because Cargo wants to update Cargo.lock

crodas:
[Here](https://github.com/crodas/cdk/actions/runs/27931629353) is a test release created by this pull request, creating the [v0.17.0](https://github.com/crodas/cdk-swift/tree/v0.17.0) release, which was built using [this Cargo.lock file](https://github.com/crodas/cdk-swift/blob/v0.17.0/rust/Cargo.lock)

.github/workflows/dart-publish.yml
```diff
@@ -161,12 +161,19 @@ jobs:
...
           cat cdk-dart/pubspec.yaml
           echo "---"
           cat cdk-dart/rust/Cargo.toml
 
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Generate Cargo.lock for standalone crate
+        working-directory: cdk-dart/rust
+        run: cargo generate-lockfile
```
thesimplekid, line 175: Doesn't generating the lock file new defeat the purpose of what we are trying to solve here? The motivation for this was we had the CI pass with a good version of a dep, then a new version of that dep was released and since we did not built from the lock that new version was pull in for the release. This exact scenario still would happen here since we regenerate the lock file. 

Doing it this way is an improvement as at least we have the lockfile it was built with but does not address what we set off to address.

thesimplekid (changes requested):

.github/workflows/dart-publish.yml
crodas: I naively compared both locks, and they are remarkably similar. I checked a few checksums from a few packages; they all matched. I think it was luck. I'll revisit my solution now.

thesimplekid:
Not much movement so going to mark as draft.

crodas:
@cdk-bot review

crodas:
@thesimplekid This is the simpler way I found to pin versions in the target repos. The alternative is to build from this repo, so anyone wanting a custom build would have to clone it, which isn't that bad either.

Here [is a test build that I run](https://github.com/crodas/cdk/actions/runs/34292780381); and [here is the output](https://github.com/crodas/cdk-dart)

crates/cdk-sql-common/build.rs
thesimplekid, line 1: Why are these here? left from other work?

crodas: Yeah sorry, will fix it shortly

crodas:
@cdk-bot review

crodas:
@cdk-bot review

.github/workflows/go-publish.yml
```diff
@@ -212,64 +254,82 @@ jobs:
...
+          RUSTC_RELEASE=$(awk '/^release:/ { print $2 }' <<< "${RUSTC_VV}")
+          if [[ "${RUSTC_RELEASE}" != "${CHANNEL}" ]]; then
+            echo "::error::expected rustc ${CHANNEL} per rust-toolchain.toml; got ${RUSTC_RELEASE}"
+            exit 1
+          fi
+
+      - name: Build cdk-ffi-go
+        working-directory: cdk-go/rust
+        shell: bash
+        run: cargo build --locked --release --target ${{ matrix.target }}
```
thesimplekid: Building the downstream wrapper now activates its `strip = true` release profile. UniFFI 0.30’s ELF reader uses the symbol table that this removes, so the subsequent Go generation step cannot discover the interface. Could we use `strip = "debuginfo"` until bindings have been generated, then strip the distributed library afterward, as Kotlin does?

thesimplekid (commented):
I think a much simpler approach would be to require building from the monorepo and use the downstream repositories for distribution, as originally proposed in #2097.

We can build the workspace binding wrappers using the root `Cargo.lock` and `--locked`, then publish the generated bindings and native libraries downstream. That avoids the manifest rewriting, lockfile reconciliation, and extra verification machinery, while letting CI and releases share the same build process.

I’m comfortable requiring a monorepo checkout for custom builds. I think that tradeoff is worth the simpler release process.

.github/scripts/write-build-manifest.sh
```diff
@@ -0,0 +1,69 @@
...
+BINDING_LOCK="$(sha256 "${DOWNSTREAM_DIR}/rust/Cargo.lock")"
+FLAKE_LOCK="$(sha256 flake.lock)"
+
+# binding_cargo_lock_sha256 stays null for a binding with no downstream Rust
+# crate to seed. Every language currently released has one.
+jq -n \
+  --arg language "${LANGUAGE}" \
+  --arg tag "${TAG:-}" \
+  --arg repo "${GITHUB_REPOSITORY:-cashubtc/cdk}" \
+  --arg commit "${CDK_REF:-}" \
```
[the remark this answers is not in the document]
thesimplekid: This appears to remain at the current head despite being marked resolved. For normal releases, `CDK_REF` is a tag such as `v0.19.0`, but verification compares it directly with the SHA returned by `git rev-parse HEAD`. That comparison always fails. Could we record the checked-out commit SHA when writing the manifest?

.github/workflows/ffi-verify.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          REPO="cashubtc/cdk-${{ matrix.language }}"
+          TAG="${{ inputs.release_tag }}"
+          if [[ -z "${TAG}" ]]; then
+            TAG=$(gh release view --repo "${REPO}" --json tagName -q .tagName)
+            echo "No tag given; using latest release ${TAG} of ${REPO}"
+          fi
+          echo "tag=${TAG}" >> "$GITHUB_OUTPUT"
+
+          MANIFEST=$(gh release download "${TAG}" --repo "${REPO}" \
```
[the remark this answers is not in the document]
thesimplekid: This also remains at the current head. `2>/dev/null || true` makes authentication, network, and API failures indistinguishable from a release without a manifest. The workflow then succeeds without verifying anything. Could we explicitly check whether the asset exists and let download failures fail the job?

.github/scripts/verify-binding-lockfile.sh
```diff
@@ -0,0 +1,90 @@
...
+
+BINDING_LOCK="${workdir}/downstream/rust/Cargo.lock"
+if [[ ! -f "${BINDING_LOCK}" ]]; then
+  # Releases cut before a language started seeding its lockfile have none, and
+  # the workspace lock hash checked above is then the whole guarantee.
+  echo "${LANGUAGE}: no downstream lockfile; verified against the workspace lock."
+  exit 0
+fi
+
+lock_versions() {
```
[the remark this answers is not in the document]
thesimplekid: The verifier still ignores the recorded `binding_cargo_lock_sha256` and compares only package names and versions, excluding `cdk-ffi` entries. I confirmed that a mismatched recorded hash passes. Could we verify the downstream lockfile hash, require the file when the manifest records it, and include dependency sources and checksums in the comparison? The expected reference should also be tied to the intended release rather than accepted solely from the manifest being checked.

.github/workflows/ffi-verify.yml
```diff
@@ -0,0 +1,83 @@
...
+          persist-credentials: false
+
+      - name: Resolve the release and its source commit
+        if: steps.gate.outputs.run == 'true'
+        id: resolve
+        env:
+          GH_TOKEN: ${{ secrets.FFI_DEPLOY_KEY }}
+        run: |
+          REPO="cashubtc/cdk-${{ matrix.language }}"
+          TAG="${{ inputs.release_tag }}"
```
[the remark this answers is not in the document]
thesimplekid: The downloaded commit and tag values are still inserted directly into `run:` scripts through Actions expressions. Could we pass them through environment variables and validate their expected formats before use? Shell quoting around an Actions expression does not prevent its contents from becoming shell syntax.

PR #2138, opened 2026-06-22
fix(cdk-mintd): require explicit opt-in for insecure mintd RPC connections

prusnak:
Add allow_insecure config gates for management RPC startup and gRPC payment processor connections when TLS is not configured.

Keep defaults secure and document the opt-in in example.config.toml.

This change also unifies the TLS_DIR vs TLS_DIR_PATH usage.

Supersedes https://github.com/cashubtc/cdk/pull/2100 + https://github.com/cashubtc/cdk/pull/2102


prusnak:
> - **Preserve the previous management RPC TLS config/env names** (medium) - Existing cdk-mintd deployments using the documented previous management RPC TLS config/env name can lose their custom TLS directory on upgrade, causing startup failure or unintended certificate selection.

I am very well aware of this.

@thesimplekid do we want to bloat the code with keeping legacy code to read old variable names, or management rpc via tls is deployed only by small minority?

The option so far has not been documented in example config.toml (this PR adds this). And if deployment already had tls_dir_path set, this change stops the server with telling user to set tls_dir anyway. 

I am ok with adding code to handle legacy names, if you insist.

thesimplekid:
No, I think the way you did it is good and its just the bot being a bit over eager. 

This looks good to me just noticed the ci failed since we're not allowing the insecure there so its hitting the error block as expected if we could just address that `Error: gRPC payment processor TLS is not configured. Set [grpc_processor].tls_dir or [grpc_processor].allow_insecure = true to connect without TLS`

prusnak:
> ci failed since we're not allowing the insecure there so its hitting the error block

addressed in 0fa98276d53a0b8e58fc655a924426c2d19123e5

thesimplekid (approved):
Thanks for this. There is a CI failure but it's a flaky test not caused by this PR, I ran it locally and it passed.

PR #2145, opened 2026-06-23
feat: add cdk-bark payment backend for Ark/bark

Codepocketdev:
Implements MintPayment on top of bark-wallet (Second's Ark implementation), routing payments over Ark (zero fee) when the destination supports it and falling back to the Ark-to-Lightning gateway otherwise.

- Arkoor (Ark-native) sends: confirmed 0 fee against ark.second.tech mainnet
- Lightning receives: confirmed working with async settlement model
- Lightning sends: fire-and-forget with check_outgoing_payment polling
- Fresh wallet seeding requirement documented (vtxopool edge case)

Depends on bark-wallet 0.2.5 and ark-lib 0.2.5.

### Description

Adds `cdk-bark`, a new crate under `crates/` that implements the `MintPayment` trait using `bark-wallet` — Second's implementation of the Ark protocol. This allows a Cashu mint to settle payments over Ark instead of a traditional Lightning node, with zero-fee Ark-native (arkoor) transfers and Lightning support via the Ark-to-Lightning gateway.

### Notes to the reviewers

This is a self-contained new crate — no existing crates were modified. Reviewers only need to check `crates/cdk-bark/`.

The async Lightning settlement model (`make_payment` returns `Pending`, `check_outgoing_payment` polls for terminal state) mirrors how other backends handle async settlement and is required by how the mainnet gateway operates.

The fresh wallet seeding requirement (seed with at least one Ark-native deposit before relying on Lightning receives) is a known limitation of `bark` 0.2.5 and is documented in the README.

#### ADDED
- `cdk-bark`: new `MintPayment` backend backed by `bark-wallet` (Second's Ark protocol implementation), supporting zero-fee Ark-native transfers and Lightning via Ark-to-Lightning gateway

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -233,6 +233,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-lib"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b9e9c6da70b2d2ac23a80b3095b47cc38e956b1d55fa87cf378e1581015cb353"
+dependencies = [
+ "async-trait",
+ "bark-bitcoin-ext",
+ "bitcoin 0.32.100",
+ "chrono",
+ "getrandom 0.3.4",
+ "hex-conservative 0.3.2",
+ "lazy_static",
+ "lightning",
+ "lightning-invoice",
+ "rand 0.9.4",
+ "secp256k1 0.32.0-beta.2",
+ "serde",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "arraydeque"
 version = "0.5.1"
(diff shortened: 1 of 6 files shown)


thesimplekid:
Thanks, we already have bark implemented in an external payment processor in this repo. https://github.com/thesimplekid/cdk-bark-payment-procssor

It looks like you also added some intra-ark payments that we didn't add there which could be interesting but that would also be added to the external processor not here.

Codepocketdev:
Thanks for the pointer, Didn't catch the external processor  my bad.

The arkoor (intra-Ark, zero-fee) path is the main addition beyond what's
there. Happy to port that over to cdk-bark-payment-processor as a PR
instead if that's the right place for it.

PR #2150, opened 2026-06-24
feat: nwc

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-nwc/src/service.rs
+++ b/crates/cdk-nwc/src/service.rs
@@ -1,0 +1,579 @@
+//! Nostr Wallet Connect (NIP-47) wallet **service**.
+//!
+//! This is the side that holds the funds and answers commands. It owns the
+//! Nostr relay connection, publishes the kind `13194` info event advertising
+//! capabilities, listens for kind `23194` requests from the authorized client,
+//! decrypts and validates them, dispatches to a [`NwcRequestHandler`], and
+//! publishes the encrypted kind `23195` response.
+//!
+//! Security properties enforced here (monetary software — defense in depth):
+//! - **Authorization**: only events authored by the single client public key
+//!   issued in the connection URI are processed (enforced both by the relay
+//!   subscription filter and an explicit author check).
+//! - **Replay / idempotency**: each request event id is processed at most once,
+//!   so relay re-delivery cannot trigger a duplicate `pay_invoice`.
+//! - **Freshness**: requests carrying an expired NIP-40 `expiration` tag are
+//!   dropped, and only events created after the service started are considered.
+//! - **No information leak / no panics**: every handler error is mapped to a
+//!   NIP-47 error response; the relay loop never aborts on a single bad request.
+
+use std::collections::{HashSet, VecDeque};
+use std::sync::Arc;
+
+use nostr_sdk::nips::nip47::{
+    ErrorCode, NIP47Error, NostrWalletConnectURI, Request, RequestParams, Response, ResponseResult,
+};
+use nostr_sdk::nips::{nip04, nip44};
+use nostr_sdk::prelude::*;
+use nostr_sdk::{Client as NostrClient, Keys, PublicKey, RelayUrl, SecretKey};
+use tokio::sync::broadcast::error::RecvError;
+use tokio_util::sync::CancellationToken;
+
+use crate::error::{Error, Result};
+use crate::handler::NwcRequestHandler;
+
+/// Commands advertised in the info event and reported by `get_info`.
+///
+/// Order is the canonical advertisement order; it is also the set of commands
+/// this service will actually dispatch — anything else returns
+/// [`ErrorCode::NotImplemented`].
+pub const SUPPORTED_METHODS: [&str; 6] = [
+    "pay_invoice",
+    "make_invoice",
+    "lookup_invoice",
+    "list_transactions",
+    "get_balance",
+    "get_info",
+];
+
+const SUPPORTED_ENCRYPTION_SCHEMES: &str = "nip44_v2 nip04";
+
+/// Maximum number of recently-seen request event ids kept for replay
+/// protection. Bounded to keep memory usage flat on long-running services.
+const DEDUP_CAPACITY: usize = 10_000;
+
+/// Encryption scheme negotiated for a single request/response exchange.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Encryption {
+    /// NIP-44 v2 (preferred).
+    Nip44,
+    /// NIP-04 (legacy, still widely used by NWC clients).
+    Nip04,
+}
+
+/// Configuration for an [`NwcService`].
+#[derive(Debug, Clone)]
+pub struct NwcServiceConfig {
+    /// Keys of the wallet service (the "signer"). Its public key is the one
+    /// advertised in the connection URI.
+    pub service_keys: Keys,
+    /// The client secret embedded in the connection URI. The public key derived
+    /// from it is the only key authorized to send requests.
+    pub client_secret: SecretKey,
+    /// Relays the service connects to and listens on. Must be non-empty.
+    pub relays: Vec<RelayUrl>,
+    /// Optional lightning address advertised in the connection URI (`lud16`).
+    pub lud16: Option<String>,
+}
+
+/// A NIP-47 wallet service bound to a single client connection.
+#[derive(Debug, Clone)]
+pub struct NwcService {
+    service_keys: Keys,
+    client_secret: SecretKey,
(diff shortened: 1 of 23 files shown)


asmogo:
closes #2106

asmogo:
I have tested this with https://nwc-playground.vercel.app/

crates/cdk-nwc/src/service.rs
```diff
@@ -0,0 +1,548 @@
...
+                        return Ok(false);
+                    }
+
+                    // Replay protection: process each request id at most once.
+                    if !dedup.lock().await.insert(event.id) {
+                        tracing::debug!("Dropping duplicate NWC request {}", event.id);
+                        return Ok(false);
+                    }
+
+                    handle_request(&service_keys, &client, handler.as_ref(), &event).await;
```
thesimplekid: Right now `handle_request(...).await` runs inside the relay notification callback, so requests are processed serially. For a `pay_invoice` request this can take a while: we need to get a melt quote, pay it, and wait for the LN invoice/payment flow to complete. While that is happening, other requests from the client will not be processed.

I wonder if we eventually want to spawn request handling here, probably with a bounded concurrency limit. That said, I think it is fine to leave the implementation serial for now since this only becomes noticeable if a client sends many zaps/requests at once and payments are slow.

crates/cdk/src/wallet/nwc.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// request (in millisatoshis); pass `None` for no cap.
+    pub fn nwc_handler(&self, budget_msat: Option<u64>) -> WalletNwcHandler {
+        WalletNwcHandler::new(self.clone(), budget_msat)
+    }
+}
+
+/// A [`cdk_nwc::NwcRequestHandler`] backed by a Cashu [`Wallet`].
+#[derive(Debug, Clone)]
+pub struct WalletNwcHandler {
+    wallet: Wallet,
```
thesimplekid:
```suggestion
    wallet: Arc<Wallet>,
```

We should make this an Arc<> as we eventually want to remove clone from the Wallet. And I think the FFI aligns with this already

crates/cdk/src/wallet/nwc.rs
```diff
@@ -0,0 +1,554 @@
...
+
+        Ok(GetBalanceResponse {
+            balance: amount_to_msat(balance, &self.wallet.unit)?,
+        })
+    }
+
+    #[instrument(skip(self))]
+    async fn make_invoice(
+        &self,
+        request: MakeInvoiceRequest,
```
thesimplekid, line 361: `MakeInvoiceRequest` accepts a `description_hash` but we can't support this in cashu so I think we should error if this is set.

crates/cdk/src/wallet/nwc.rs
```diff
@@ -0,0 +1,554 @@
...
+    ) -> Result<MakeInvoiceResponse, NIP47Error> {
+        let amount = msat_to_amount(request.amount, &self.wallet.unit)?;
+
+        let quote = self
+            .wallet
+            .mint_quote(
+                PaymentMethod::Known(KnownMethod::Bolt11),
+                Some(amount),
+                request.description.clone(),
```
thesimplekid, line 384: Not all mints support setting the description as its optional https://github.com/cashubtc/nuts/blob/main/23.md#mint-quote. So to avoid silently attempting to create an invoice without the correct description I think we should check the mint info before getting the quote if the description is set in the request. 

This is important for ZAPs as the invoice description has to be set to the zap receipt I believe.

crates/cdk/src/wallet/nwc.rs
```diff
@@ -0,0 +1,554 @@
...
+    pub fn nwc_handler(&self, budget_msat: Option<u64>) -> WalletNwcHandler {
+        WalletNwcHandler::new(self.clone(), budget_msat)
+    }
+}
+
+/// A [`cdk_nwc::NwcRequestHandler`] backed by a Cashu [`Wallet`].
+#[derive(Debug, Clone)]
+pub struct WalletNwcHandler {
+    wallet: Wallet,
+    budget_msat: Option<u64>,
```
thesimplekid: I would rename this to `max_payment_msat`. budget made me think it was some sort of cumulatively daily budget, which would maybe something useful we could add later, I think many nostr apps allow you to set things like hourly, daily zap budgets?

crates/cdk/src/wallet/nwc.rs
```diff
@@ -0,0 +1,554 @@
...
+        }),
+        CurrencyUnit::Msat => Ok(value),
+        other => Err(nip47_err(
+            ErrorCode::Other,
+            format!("unsupported wallet unit: {other}"),
+        )),
+    }
+}
+
+/// Convert millisatoshis to a wallet [`Amount`] for the given unit.
```
thesimplekid, line 146: Could we use the cdk Amount type we have `to_sat` and `to_msat` fns instead of having the logic here as well? https://github.com/cashubtc/cdk/blob/04d39a621005aa77b7768eb55985709ee57e9e6e/crates/cashu/src/amount.rs#L569-L583.

So I think we could use that internally and convert at the boundary?

crates/cdk/src/wallet/nwc.rs
```diff
@@ -0,0 +1,554 @@
...
+                    state: Some(cdk_nwc::nip47::TransactionState::Pending),
+                    invoice: Some(quote.request.clone()),
+                    description: None,
+                    description_hash: None,
+                    preimage: None,
+                    payment_hash: target_hash,
+                    amount,
+                    fees_paid: 0,
+                    created_at: Timestamp::from(quote.expiry),
+                    expires_at: Some(Timestamp::from(quote.expiry)),
```
thesimplekid: Should these both be the expiry?

thesimplekid (changes requested):
I see there is a nwc crate https://github.com/rust-nostr/nostr/tree/master/crates/nwc for the client side of this. I think it would be good to use that to add some end to end tests in for this or maybe just an example would suffice.

crates/cdk/src/wallet/nwc.rs
asmogo: fixed

crates/cdk/src/wallet/nwc.rs
asmogo: Amount conversions now use CDK Amount::to_msat / to_sat,

crates/cdk/src/wallet/nwc.rs
asmogo: done

crates/cdk/src/wallet/nwc.rs
asmogo: done

crates/cdk/src/wallet/nwc.rs
asmogo: done

thesimplekid:
@cdk-bot review

crates/cdk-nwc/src/service.rs
```diff
@@ -0,0 +1,566 @@
...
+
+    /// Publish the kind `13194` info event advertising supported commands and
+    /// encryption schemes.
+    async fn publish_info_event(&self, client: &NostrClient) -> Result<()> {
+        let content = SUPPORTED_METHODS.join(" ");
+        let encryption_tag = Tag::custom(
+            TagKind::Custom(std::borrow::Cow::Borrowed("encryption")),
+            ["nip44_v2".to_string(), "nip04".to_string()],
+        );
+
```
thesimplekid, line 245: I think this is not to the nip47 spec. https://github.com/nostr-protocol/nips/blob/master/47.md#info-event. 

It should be ["encryption", "nip44_v2 nip04"] not ["encryption", "nip44_v2", "nip04"]

crates/cdk/src/wallet/nwc.rs
```diff
@@ -0,0 +1,751 @@
...
+            .total_balance()
+            .await
+            .map_err(|e| nip47_err(ErrorCode::Internal, e.to_string()))?;
+
+        Ok(GetBalanceResponse {
+            balance: amount_to_msat(balance, &self.wallet.unit)?,
+        })
+    }
+
+    #[instrument(skip(self))]
```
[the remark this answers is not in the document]
thesimplekid: I think we should address this and do the same as we do for description_hash

thesimplekid (changes requested):
Two small things we should address before merge. 

I think there is a bit of clean up we can do and combine our nostr stack (nwc and npubcash) but we can do that in a future PR and release.

thesimplekid (approved):

PR #2161, opened 2026-06-27
fix: improve mint/melt limit ux

orangeshyguy21:
`[ln]`/`[onchain]` mint/melt bounds stay required, but a missing config now reports explicitly instead of an opaque error. Example config also sets values to more appropriate default values to align with network policy.

### Description

Since 0.17.0 a single `[ln]`/`[onchain]` block missing the mint/melt bounds fails with:

```
data did not match any variant of untagged enum LnOneOrMany for key `ln`
```

The `#[serde(untagged)]` deserializer (from the multi-backend `Vec<Ln>` change) masks the real cause. This swaps it for a visitor that deserializes the single table directly, yielding `missing field "ln.min_mint"`.

-----

### Notes to the reviewers

- Fixes the diagnostic; bounds remain mandatory by design.
- Improves example values; onchain mirrors bdk configs: `min_receive_amount_sat`,  `min_send_amount_sat`

-----

#### FIXED
- `cdk-mintd`: `[ln]`/`[onchain]` missing `min_mint`/`max_mint`/`min_melt`/`max_melt` now fails with a clear "missing field" error instead of the untagged-enum message (regressed in 0.17.0).

----

### Checklist

* [x] If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`) — N/A

### DIFF
--- a/crates/cdk-mintd/example.config.toml
+++ b/crates/cdk-mintd/example.config.toml
@@ -120,10 +120,10 @@ connection_timeout_seconds = 10
 # NOTE: fakewallet is isolated testing mode and cannot be mixed with real payment backends.
 ln_backend = "fakewallet"
 # unit = "sat"          # Optional, defaults to "sat"
-# min_mint = 1          # Optional, defaults to 1
-# max_mint = 500000     # Optional, defaults to 500000
-# min_melt = 1          # Optional, defaults to 1
-# max_melt = 500000     # Optional, defaults to 500000
+min_mint = 1
+max_mint = 500000
+min_melt = 1
+max_melt = 500000
 
 [onchain]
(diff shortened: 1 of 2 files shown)


prusnak:
I am OK with setting min to 1 by default.

Not sure how I feel about having some default max value. Maybe it is fine that the mint does not boot until an admin sets the proper maximum values?

thesimplekid:
A min of one can't really be used for onchain as it would be dust the mint could never move. I would prefer we force this being set by the operator and fix the example to show this must be set and recommend something higher.

orangeshyguy21:
Tend to agree here that they should be required configs, with more sensible defaults per backend. These values have increased implications with onchain.

Will rework with new intent and sensible defaults.

PR #2164, opened 2026-06-27
Add LDK Server mint backend

vincenzopalazzo:
## Summary
- add a new cdk-ldk-server crate implementing MintPayment for an external lightningdevkit/ldk-server daemon
- wire cdk-mintd feature/config/env setup for ln_backend = "ldk-server"
- support BOLT11 and BOLT12 mint/melt flows via ldk-server-client with bounded payment scans for incoming status lookup
- document the backend in README files and example.config.toml

## Test plan
- cargo fmt --all -- --check
- CARGO_BUILD_JOBS=1 cargo test -p cdk-ldk-server
- CARGO_BUILD_JOBS=1 cargo clippy -p cdk-ldk-server -- -D warnings
- CARGO_BUILD_JOBS=1 cargo clippy -p cdk-mintd --no-default-features --features ldk-server,sqlite -- -D warnings
- CARGO_BUILD_JOBS=1 cargo test -p cdk-mintd --no-default-features --features ldk-server,sqlite test_env_var_only_config_all_backends -- --nocapture

## Notes
- ldk-server is still moving quickly, so the client dependency is pinned to the deployed-compatible upstream revision.
- Live integration testing requires a running ldk-server daemon and is not included in this PR.


Cargo.toml
```diff
@@ -82,6 +83,7 @@ futures = { version = "0.3.28", default-features = false, features = ["async-awa
 lightning-invoice = { version = "0.34.0", features = ["serde", "std"] }
 lightning = { version = "0.2.0", default-features = false, features = ["std"]}
 ldk-node = "0.7.0"
+ldk-server-client = { git = "https://github.com/lightningdevkit/ldk-server", rev = "50fe7523be3529d86bfee0dfc35df9a52aca7310", default-features = false }
```
thesimplekid, line 86: Need to wait for a ldk-server on crates as this would break our publishing.  

One option to avoid needing to use this as a fork right now is to build it as a payment processor until ldk-server publishes. https://github.com/thesimplekid/cdk-template-payment-processor

thesimplekid (commented):
Nice thanks for this. Been thinking about how cdk should support ldk-server for a bit. I opened an issue on it here and would be curious your thoughts. https://github.com/cashubtc/cdk/issues/2170

crates/cdk-ldk-server/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+                Some(payment) => return Ok(payment),
+                None if start.elapsed() >= timeout => return Err(Error::PaymentNotFound),
+                None => tokio::time::sleep(Duration::from_millis(100)).await,
+            }
+        }
+    }
+
+    fn make_payment_response_from_payment(
+        unit: &CurrencyUnit,
+        payment_lookup_id: PaymentIdentifier,
```
mubarak23, line 192: I noticed `payment_lookup_id` is only forwarded into the returned MakePaymentResponse and isn't otherwise used in this function. Is this just to satisfy the response shape, or is this function responsible for propagating the identifier back to the caller?

vincenzopalazzo:
Thanks for the reviews! Following @thesimplekid's suggestion, we've taken the **payment-processor route to production** while this PR waits on ldk-server's crates.io release:

**[vincenzopalazzo/cdk-ldk-server-processor](https://github.com/vincenzopalazzo/cdk-ldk-server-processor)** — a standalone gRPC payment-processor binary that wraps this PR's `MintPayment` implementation (git-pinned to this branch, so no code is duplicated) and serves it through the published `cdk-payment-processor` protocol. A **stock** `cdk-mintd` (feature `grpc-processor`, no fork) connects via `ln_backend = "grpcprocessor"`.

This resolves the publishing blocker: the processor binary is never published to crates.io, so the git-pinned `ldk-server-client` is fine there, and cdk-mintd only needs the already-published client crate.

Status:
- The backend (this branch) is running in production at the [Hedwig mint](https://mint.hedwig.sh) — BOLT11 and BOLT12 mint/melt flows working.
- The processor wrapper is built and smoke-tested against the same production ldk-server; full mint switchover is next.
- Production debugging surfaced an LDK bug that made BOLT12 offers silently unpayable (compact introduction node referencing a closed channel): root cause + fix upstreamed at [rust-lightning#4828](https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/pulls/4828) (issue [#4826](https://git.rust-bitcoin.org/lightningdevkit/rust-lightning/issues/4826)), requested for backport to 0.1/0.2/0.3.

Plan for this PR: keep it as the in-tree backend for when ldk-server lands on crates.io. I'll also address the codex review findings here (P1: BOLT11 melt lookup keyed by payment id instead of payment hash; P2: scan-limit check before returning collected matches) — happy to hear preferences on #2170 for the long-term direction.

vincenzopalazzo:
Closing this PR in favor of the external payment processor approach, which is now running in production on the Hedwig mint.

Instead of merging an LDK Server backend into CDK (with the crates.io publishing problem that a git-pinned `ldk-server-client` creates), the same stack works today with **zero changes to CDK** using the existing `grpcprocessor` backend in the official v0.17.3 release:

```
stock cdk-mintd (v0.17.3, grpcprocessor backend)
  -> gRPC :50071  cdk-ldk-server-processor
  -> gRPC+TLS :3536  ldk-server
```

The processor lives here: https://github.com/vincenzopalazzo/cdk-ldk-server-processor

It wraps the `CdkLdkServer` backend from this branch in the published `cdk-payment-processor` server, so the git-pinned dependency only ever lives inside a binary that is never published to crates.io.

Thanks @thesimplekid for the review feedback — the payment-processor route turned out to be the right call. The branch stays around for reference; happy to reopen if a native in-tree backend becomes desirable again.

PR #2172, opened 2026-06-29
fix(wallet): record recovered melt transactions

thesimplekid:
When a pending melt is recovered as paid, persist the outgoing transaction, update the stored quote with the payment proof, and release the quote reservation before deleting the saga.

Also persist melt quote payment_proof in the SQL wallet store so restored quotes keep the preimage or on-chain outpoint.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@codex review

PR #2173, opened 2026-06-30
ADR for append-only change log in mint database

crodas:
### Description

Proposal solution for append-only storage: here is the rendered. Feedback and inline comments are more than welcome

[Here is the rendered document](https://github.com/crodas/cdk/blob/proposal/append-only/docs/adr/0001-append-only-change-log.md)

Once this proposal is approved and merged I will open a PR.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/docs/adr/0001-append-only-change-log.md
+++ b/docs/adr/0001-append-only-change-log.md
@@ -19,33 +19,39 @@ The mint database uses mutable tables for current state. After an update the pre
 
 ## Considered Options
 
-* **Single generic `event_log` table** -- one shared table with `table_name`, `entity_id`, `change_type`, and a JSON `data` blob. Simple to set up, but the untyped JSON blob prevents column-level queries and schema validation, and mixing heterogeneous entities in one table makes indexing awkward.
+* **Single append-only delta table with a typed delta enum** -- one shared insert-only table keyed by a global record id (`table_name:pk`). Each change is stored as a `delta` value, a typed enum with one variant per writable field, serialized to JSON or a compact binary form. The variants are derived from the mutation method signatures: every parameter a method writes becomes a variant. The enum is forward compatible: a new writable field is a new variant, and existing rows never need rewriting. One migration and one insert path, at the cost of losing column-level SQL queries over the delta payload.
 
-* **Per-table log tables with typed columns** -- each tracked entity gets its own `_log` table containing only identity columns and the columns that actually change. More tables to maintain, but each one is small, focused, schema-validated, and directly queryable. Insert-only entities need no log table at all.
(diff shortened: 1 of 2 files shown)


docs/adr/0001-append-only-change-log.md
```diff
@@ -0,0 +1,169 @@
...
+
+### `change_id`
+
+Application-generated `i64`:
+
+```
+Bit 63:      always 0 (positive signed i64)
+Bits 62..22: 41 bits of millisecond timestamp since custom epoch
+Bits 21..0:  22 bits of hash(changeset)
+```
```
asmogo: can we use an auto-increment or sequence here? It would simplify the logic and fix the duplicate key conflicts.

docs/adr/0001-append-only-change-log.md
```diff
@@ -0,0 +1,169 @@
...
+Triggered by `update_melt_quote_state` and `update_melt_quote_request_lookup_id`.
+
+```sql
+CREATE TABLE melt_quote_log (
+    change_id           BIGINT PRIMARY KEY,
+    quote_id            TEXT NOT NULL,
+    state               TEXT,
+    request_lookup_id   TEXT,
+    payment_proof       TEXT,
+    paid_time           BIGINT
```
asmogo: i think this is missing some member variables (`request_lookup_id_kind` gets updated in `update_melt_quote_request_lookup_id` or `estimated_blocks`, `fee_reserve` in `update_melt_quote_state` )

docs/adr/0001-append-only-change-log.md
```diff
@@ -0,0 +1,169 @@
...
+| **melt_quote** | `state`, `request_lookup_id`, `payment_proof`, `paid_time` | `melt_quote_log` |
+| **proof** | `state` (+ deletions on compensation) | `proof_log` |
+| **keyset** | `active` | `keyset_log` |
+
+### Entities that do NOT need log tables
+
+| Entity | Reason |
+|--------|--------|
+| **mint_quote** | Already append-only. `update_mint_quote` writes to `mint_quote_payments` and `mint_quote_issued` detail tables. The `amount_paid`/`amount_issued` on the quote row are materialized sums. |
+| **blind_signature** | Insert-only, never updated or deleted. |
```
asmogo: there are certain conditions in `add_blind_signatures` that result in a update statement for the blind_signature.

docs/adr/0001-append-only-change-log.md
crodas: If we do, the database service assigning the autoincrement number would serialize all our operations. That is not a problem with SQLite, but it would be with Postgres.

docs/adr/0001-append-only-change-log.md
crodas: Good catch, anything that is updated should be reflected here.

thesimplekid:
Closing for now at least as we're not working on this at the moment

PR #2176, opened 2026-06-30
fix: update NUT-20/29

### LINKED ISSUE #2149: NUT-29: Batch minting signature
Implement https://github.com/cashubtc/nuts/pull/375 for cdk

asmogo:
### Description

closes #2149 

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut20.rs
+++ b/crates/cashu/src/nuts/nut20.rs
@@ -119,12 +119,21 @@ where
 
         let signature = Signature::from_str(signature).map_err(|_| Error::InvalidSignature)?;
 
-        let msg_to_sign = self.msg_to_sign();
+        let quote_id = self.quote.to_string();
+        let msg_to_sign = mint_quote_msg_to_sign(&quote_id, &self.outputs);
 
         match pubkey.verify(&msg_to_sign, &signature) {
-            Ok(()) => Ok(()),
-            Err(_) => pubkey.verify(&self.legacy_msg_to_sign(), &signature),
-        }?;
+            Ok(()) => return Ok(()),
+            Err(_) => {
+                let legacy_msg = legacy_mint_quote_msg_to_sign(&quote_id, &self.outputs);
+                pubkey.verify(&legacy_msg, &signature)?;
+                tracing::warn!(
+                    quote_id = %quote_id,
+                    output_count = self.outputs.len(),
+                    "Accepted legacy mint quote signature format"
+                );
+            }
+        }
 
         Ok(())
     }
--- a/crates/cashu/src/nuts/nut20.rs
+++ b/crates/cashu/src/nuts/nut20.rs
@@ -99,6 +99,20 @@ where
         Ok(())
     }
 
+    /// Sign [`MintRequest`] using the legacy NUT-20 message format.
+    ///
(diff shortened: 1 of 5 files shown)


crates/cashu/src/nuts/nut20.rs
```diff
@@ -21,29 +21,71 @@ pub enum Error {
...
-            // to_hex() creates a hex string, into_bytes() converts it to UTF-8 bytes
-            msg.append(&mut output.blinded_secret.to_hex().into_bytes());
-        }
-        msg
+        mint_quote_msg_to_sign(&quote_id, &self.outputs)
+    }
+
+    fn legacy_msg_to_sign(&self) -> Vec<u8> {
+        let quote_id = self.quote.to_string();
+        legacy_mint_quote_msg_to_sign(&quote_id, &self.outputs)
```
thesimplekid, line 88: Do we plan to remove this at some point in the future should maybe include a TODO or something?

thesimplekid (commented):
I think the code here is fine. My concern is the roll out and I'm not sure the best way to handle this tbh.

 If we merge this as is the wallet will start using the new format so will break will all mints that have not updated (so all mints). I think @robwoodgate said in cashu-ts they are first trying the new msg to sign and if that fails retrying the request with the old msg. We could do that or separate the wallet and mint and roll out mint first and wallet later if we know the version that support the new sig scheme can check that first. This would be a new pattern in cdk and also probably not ideal.

robwoodgate:
> I think the code here is fine. My concern is the roll out and I'm not sure the best way to handle this tbh.
> 
>  If we merge this as is the wallet will start using the new format so will break will all mints that have not updated (so all mints). I think @robwoodgate said in cashu-ts they are first trying the new msg to sign and if that fails retrying the request with the old msg. We could do that or separate the wallet and mint and roll out mint first and wallet later if we know the version that support the new sig scheme can check that first. This would be a new pattern in cdk and also probably not ideal. 

Yes, in Cashu-TS I originally went with a versioned gate, but decided to switch to try and fallback because it would remove any version spoofing vector and handle all mints.

thesimplekid:
> > I think the code here is fine. My concern is the roll out and I'm not sure the best way to handle this tbh.
> > If we merge this as is the wallet will start using the new format so will break will all mints that have not updated (so all mints). I think @robwoodgate said in cashu-ts they are first trying the new msg to sign and if that fails retrying the request with the old msg. We could do that or separate the wallet and mint and roll out mint first and wallet later if we know the version that support the new sig scheme can check that first. This would be a new pattern in cdk and also probably not ideal.
> 
> Yes, in Cashu-TS I originally went with a versioned gate, but decided to switch to try and fallback because it would remove any version spoofing vector and handle all mints.

Offline (but online) we said we would follow this model as well. 

- [x] follow cashu-ts model of a retry

crodas (dismissed):

thesimplekid:
@cdk-bot review

thesimplekid:
@asmogo I pushed one commit to this fixing a test and logging when we use the legacy vs not if you want to take a look.

crodas (approved):

thesimplekid (approved):

PR #2177, opened 2026-06-30
feat(ffi): add waitable async melt confirmations

thesimplekid:
Expose async-preferred melt confirmation through UniFFI as a paid-or-pending outcome, with a PendingMelt handle that can wait for finalization.

Keep Rust-native PreparedMelt confirmation on the in-memory saga while the FFI path reconstructs from cached prepared data and persisted pending state.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -513,6 +513,8 @@ jobs:
         continue-on-error: true
       - name: Run FFI tests
         run: nix build -L .#checks.x86_64-linux.ffi-tests
+      - name: Run live Python FFI tests
+        run: nix develop -i -L .#ffi --command just ffi-test-live-python
 
   coverage:
     name: "Code Coverage"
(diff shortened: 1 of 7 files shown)


crodas:
LGTM, but it would be much better if it could leverage the subscription and be part of the wallet and not of the FFI. I can approve it now and improve it later.

crodas (approved):

thesimplekid:
> LGTM, but it would be much better if it could leverage the subscription and be part of the wallet and not of the FFI. I can approve it now and improve it later.

Yeah it's an issue with the lifetimes over the ffi boundary, saga might have to be reworked a bit.

crodas:
I think it's fine to merge it as is and revisit it later.

crodas (approved):

PR #2184, opened 2026-07-02
fix(cashu): accept unknown fields in onchain quote responses

thesimplekid:
The attribute existed only to make serde(untagged) discrimination of NotificationPayload reject non-onchain payloads. Replace that mechanism with a hand-written Deserialize that classifies payloads by their discriminating fields (Y, fee_options, fee_reserve, state, amount), so websocket notification dispatch no longer depends on strict structs.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas (approved):
LGTM

PR #2185, opened 2026-07-02
feat: expose tor in ffi

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-ffi/Cargo.toml
+++ b/crates/cdk-ffi/Cargo.toml
@@ -43,8 +43,10 @@ log = "0.4"
 
 
 [features]
-default = ["npubcash", "bip353"]
+default = ["npubcash", "bip353", "tor"]
 bip353 = ["cdk/bip353"]
+# Enable Tor transport for mint connections (arti-based)
+tor = ["cdk/tor"]
 # Enable Postgres-backed wallet database support in FFI
 postgres = ["cdk-postgres"]
 # Enable Supabase-backed wallet database support in FFI
--- a/crates/cdk-ffi/src/wallet_repository.rs
+++ b/crates/cdk-ffi/src/wallet_repository.rs
@@ -51,44 +51,6 @@ impl WalletRepository {
         })
     }
 
-    /// Create a new WalletRepository that routes all mint connections over Tor
-    ///
-    /// Uses an embedded Tor client (arti). Any Tor bootstrap triggered during
-    /// construction runs on the process-wide shared runtime, so the client's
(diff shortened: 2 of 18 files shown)


asmogo:
@cdk-bot review this pull request.

asmogo:
@cdk-bot review

crates/cdk-ffi/Cargo.toml
```diff
@@ -43,8 +43,10 @@ log = "0.4"
 
 
 [features]
```
j-kon, line 45: Should `tor` be part of the default `cdk-ffi` feature set here, or should it stay opt-in for bindings that explicitly want Tor support?

Adding it to defaults means any wrapper crate that does not disable default features may now pull in arti/Tor and its extra build/runtime dependencies automatically. That might be fine if the intent is for Tor-backed FFI to be enabled everywhere by default, but it would be good to make that decision explicit since it affects binding consumers beyond Swift.

crates/cdk-ffi/src/wallet_repository.rs
```diff
@@ -51,6 +51,44 @@ impl WalletRepository {
...
     }
 
+    /// Create a new WalletRepository that routes all mint connections over Tor
+    ///
+    /// Uses an embedded Tor client (arti). Any Tor bootstrap triggered during
+    /// construction runs on the process-wide shared runtime, so the client's
+    /// background tasks survive after the constructor returns. If the database
+    /// already contains mints, construction blocks until the initial Tor
+    /// bootstrap completes and may take tens of seconds.
+    #[cfg(feature = "tor")]
```
j-kon: Since this is a new UniFFI constructor, should the generated binding outputs be regenerated for the affected languages in this PR?

The feature is being exposed at the FFI layer, but I only see the Swift wrapper Cargo feature being updated directly. If Dart/Go also build with Tor enabled through `cdk-ffi` defaults, it would be good to confirm whether their generated APIs are expected to include `new_with_tor` now, or whether binding regeneration happens separately during release.

j-kon (commented):
I reviewed the changes and left a few inline comments on features and runtime usage.

crates/cdk-ffi/Cargo.toml
asmogo: We will probably ship tor by default with the bindings.

thesimplekid:
Going to close for now but would be good to pick up at some point

PR #2193, opened 2026-07-03
append only database with signed merkle tree and checkpoint anchoring 

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/docs/adr/0001-append-only-transparency-log.md
+++ b/docs/adr/0001-append-only-transparency-log.md
@@ -3,7 +3,7 @@
 * Status: proposed
 * Authors: @asmogo
 * Date: 2026-07-01
-* Targeted modules: `cashu`, `cdk-common`, `cdk-sql-common`, `cdk` (mint), `cdk-axum`, `cdk-mintd`
+* Targeted modules: `cashu`, `cdk-common`, `cdk-sql-common`, `cdk` (mint), `cdk-axum`, `cdk-mintd`, `cdk-sigsum` (new)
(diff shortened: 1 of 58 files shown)


docs/adr/0001-append-only-transparency-log.md
```diff
@@ -0,0 +1,296 @@
...
+
+CREATE TABLE mint_event_log (
+    seq         INTEGER PRIMARY KEY AUTOINCREMENT, -- IDENTITY on Postgres; commit-order hint only
+    entity_type TEXT   NOT NULL,      -- 'melt_quote' | 'proof' | 'keyset' | 'blind_signature'
+    entity_id   TEXT   NOT NULL,      -- quote id / Y hex / keyset id / blinded message hex
+    op          SMALLINT NOT NULL,    -- 0 = Insert, 1 = Update, 2 = Delete
+    payload     BLOB   NOT NULL,      -- canonical JSON of the new field values (see below)
+    leaf_hash   BLOB   NOT NULL,      -- RFC 6962 leaf hash, precomputed at insert time
+    created_time BIGINT NOT NULL,
+    leaf_index  BIGINT                -- zero-based Merkle tree position; NULL until the
```
crodas, line 103: To avoid serializing entries, I think it would be best to produce the Merkle tree every few minutes (or even seconds). As long as the beginning/end of the snapshot is clear enough, I think that would be best.

thesimplekid:
Closing as we're not actively working on this

PR #2199, opened 2026-07-04
fix(mint): return null expiry for no-expiry Bolt12 quotes

### LINKED ISSUE #2195: Bolt12 mint quote serializes expiry: 0 instead of null for no-expiry offers
Bolt12 mint quotes are created with no expiry (unix_expiry: None in crates/cdk/src/mint/issue/mod.rs), stored internally as the 0 sentinel (expiry.unwrap_or(0)), and then unconditionally re-wrapped as Some(0) in the TryFrom<MintQuote> for MintQuoteBolt12Response conversions in crates/cdk-common/src/mint.rs. The wire response is therefore "expiry": 0.

NUT-25 types the field as <int|null> — "the Unix timestamp until which the mint quote is valid" — so 0 reads as "expired since 1970-01-01". Wallets taking the spec literally drop the quote as expired and stop watching the offer, so payments to it are never claimed (observed with coco against an LDK-backed mint).

The project already has a convention for this: the onchain conversion maps the sentinel back to null with expiry: (quote.expiry != 0).then_some(quote.expiry) (added in #1870, with a regression test). Bolt12 should do the same in both conversion sites so clients parse the offer as no-expiry instead of 1970.

> AI Summary above

I found this when working with coco v2. I was testing with one of @thesimplekid's bolt 12 mints which have this `expiry: 0` and the intention behind that seems to be "never expire" but in practice it instantly expires as far as wallets are concerned so they stop listening immediately as per the spec.

My clanker just recommends doing `expiry: (quote.expiry != 0).then_some(quote.expiry),` which is what seems to already be done for Onchain.

thesimplekid:
Map the internal 0 expiry sentinel back to None when converting Bolt12 mint quotes to API responses, matching the onchain conversion behavior.

Add regression coverage for direct and generic Bolt12 quote response paths.

fixes: https://github.com/cashubtc/cdk/issues/2195

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/mint.rs
+++ b/crates/cdk-common/src/mint.rs
@@ -1267,7 +1267,7 @@ impl TryFrom<MintQuote> for MintQuoteBolt12Response<QuoteId> {
         Ok(MintQuoteBolt12Response {
             quote: mint_quote.id.clone(),
             request: mint_quote.request,
-            expiry: Some(mint_quote.expiry),
+            expiry: (mint_quote.expiry != 0).then_some(mint_quote.expiry),
             amount_paid: mint_quote.amount_paid.into(),
             amount_issued: mint_quote.amount_issued.into(),
             pubkey: mint_quote.pubkey.ok_or(Error::PubkeyRequired)?,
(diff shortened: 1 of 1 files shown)


thesimplekid:
@cdk-bot review

j-kon (commented):
I reviewed the change and it looks good to me.

The fix maps the internal `0` expiry sentinel back to `None` in both Bolt12 response conversion paths, which matches the expected no-expiry behavior and the on-chain quote convention.

The regression tests also cover the main cases well: direct Bolt12 zero-expiry, direct Bolt12 non-zero expiry, and the generic Bolt12 response path for zero-expiry.

One small optional note: Codecov is showing one changed line missing coverage, which may be the generic Bolt12 response path for non-zero expiry. If needed, adding a small check for `MintQuoteResponse::Bolt12` preserving a non-zero expiry would likely cover that too.

Nice fix overall.

PR #2201, opened 2026-07-04
feat(ffi): expose mint_unissued_quotes

j-kon:
## Summary

This PR exposes the existing CDK wallet `mint_unissued_quotes` method through the FFI `Wallet`.

The core Rust wallet already supports minting paid but unissued mint quotes as part of startup/recovery flows. Exposing this method through `cdk-ffi` makes the same capability available to generated bindings such as Dart.

## Changes

- Add `Wallet::mint_unissued_quotes` to `crates/cdk-ffi`
- Convert the core return value into the correct FFI-compatible type
- Generated binding outputs are not committed in this PR because the monorepo does not currently track the generated Dart file.

## Testing

- `cargo fmt`
- `cargo test -p cdk-ffi`

### DIFF
--- a/crates/cdk-common/src/wallet/mod.rs
+++ b/crates/cdk-common/src/wallet/mod.rs
@@ -995,6 +995,9 @@ pub trait Wallet: Send + Sync {
         spending_conditions: Option<SpendingConditions>,
     ) -> Result<Proofs, Self::Error>;
 
+    /// Check and mint any paid but unissued mint quotes
+    async fn mint_unissued_quotes(&self) -> Result<Self::Amount, Self::Error>;
+
     /// Check mint quote status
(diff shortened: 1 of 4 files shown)


j-kon:
@cdk-bot review

thesimplekid:
This is missing from the trait I guess if it wasn't forced to be exported on the FFI so should add it there.

j-kon:
Addressed this by adding `mint_unissued_quotes` to `cdk_common::wallet::Wallet`, then forwarding it in both the core `Wallet` trait implementation and the FFI `Wallet` trait implementation. I also added a small FFI test that calls both the concrete wallet method and the trait method against an empty in-memory store.

crodas (dismissed):

thesimplekid (dismissed):

Issue #1767, opened 2026-03-24
Here is a list of missing methods from the trait and the FFI.


crodas:
# Here is a list of missing methods from the trait and the FFI.

All the methods in the FFI are part of the Wallet Trait and the CDK Wallet.

Methods NOT present in all 3 (trait, CDK wallet, FFI wallet). 48 shared methods excluded.

| Method | Trait | CDK | FFI |
|---|---|---|---|
| `get_proofs_fee` | | x | |
| `get_proofs_fee_by_count` | | x | |
| `update_mint_url` | | x | |
| `verify_token_p2pk` | | x | |
| `set_client` | | x | |
| `mint_connector` | | x | |
| `set_target_proof_count` | | x | |
| `set_auth_client` | | x | |
| `confirm_send` | | x | |
| `cancel_send` | | x | |
| `receive_raw` | | x | |
| `check_all_mint_quotes` | | x | |
| `mint_unissued_quotes` | | x | |
| `get_active_mint_quotes` | | x | |
| `get_unissued_mint_quotes` | | x | |
| `batch_check_mint_quote_status` | | x | |
| `batch_mint` | | x | |
| `finalize_pending_melts` | | x | |
| `confirm_prepared_melt` | | x | |
| `confirm_prepared_melt_with_options` | | x | |
| `cancel_prepared_melt` | | x | |
| `get_active_melt_quotes` | | x | |
| `get_pending_melt_quotes` | | x | |
| `check_melt_quote_status` | | x | |
| `get_unspent_proofs` | | x | |
| `get_pending_proofs` | | x | |
| `get_reserved_proofs` | | x | |
| `get_pending_spent_proofs` | | x | |
| `get_proofs_with` | | x | |
| `unreserve_proofs` | | x | |
| `select_exact_proofs` | | x | |
| `select_proofs` | | x | |
| `recover_incomplete_sagas` | | x | |
| `sync_proofs_state` | | x | |
| `proof_stream` | | x | |
| `mints_proof_stream` | | x | |
| `payment_stream` | | x | |
| `wait_and_mint_quote` | | x | |
| `wait_for_payment` | | x | |
| `create_request` | | x | |
| `wait_for_nostr_payment` | | x | |
| `enable_npubcash` | | x | |
| `get_npubcash_keys` | | x | |
| `sync_npubcash_quotes` | | x | |
| `sync_npubcash_quotes_since` | | x | |
| `npubcash_proof_stream` | | x | |
[...]

PR #2205, opened 2026-07-07
feat(ffi): expose check_all_mint_quotes

j-kon:
## Summary

This PR exposes the existing CDK wallet `check_all_mint_quotes` method through the FFI `Wallet`.

The FFI wallet already exposes individual mint quote operations. Exposing `check_all_mint_quotes` makes the broader quote status sync flow available through generated bindings.

## Changes

- Add `Wallet::check_all_mint_quotes` to `crates/cdk-ffi`
- Convert the core return value into the existing FFI-compatible type
- Keep generated binding outputs out of this PR

## Testing

- `cargo fmt`
- `cargo test -p cdk-ffi`

Partially addresses #1767

### DIFF
--- a/crates/cdk-ffi/src/wallet.rs
+++ b/crates/cdk-ffi/src/wallet.rs
@@ -300,6 +300,16 @@ impl Wallet {
         Ok(quote.into())
     }
 
+    /// Check all unissued mint quote states from the mint.
+    ///
+    /// This updates local quote state from the mint and attempts recovery for
+    /// any quote with an in-progress mint saga. It does not mint tokens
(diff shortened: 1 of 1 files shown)


thesimplekid:
I think we don't need this if we have https://github.com/cashubtc/cdk/pull/2201, we should try and reduce the number of fns exported thus making the api easier to use. There isn't really a reason to check and not mint I think.

j-kon:
Closing this in favor of #2201. I agree that exposing mint_unissued_quotes gives the useful startup/recovery flow without adding another lower-level FFI method for checking quotes without minting.

PR #2210, opened 2026-07-07
fix(nut17): decode notification payloads with subscription kind

thesimplekid:
NUT-17 notification payloads are ambiguous without the subscription kind. Quote response shapes overlap and tolerate unknown fields, so shape-based decoding can select the wrong payload variant.

Add a kind-aware payload decoder and use it in the wallet subscription client while preserving the existing generic payload deserialization surface for backports.

This prevents onchain and custom quote notifications from being decoded as the wrong payload variant.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

PR #2224, opened 2026-07-13
perf(wallet): avoid network calls during repository startup

asmogo:
Restore wallet units from persisted mint metadata instead of fetching mint info sequentially for every configured mint during repository construction. This keeps startup offline-first and preserves the sat fallback for legacy records without cached metadata.

Document the FFI behavior and add regression coverage for multi-unit restoration, missing metadata, and zero network connections during startup.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/wallet_repository.rs
+++ b/crates/cdk/src/wallet/wallet_repository.rs
@@ -205,7 +205,9 @@ impl WalletRepositoryBuilder {
         self
     }
 
-    /// Build the WalletRepository and load existing wallets from database
+    /// Build the WalletRepository and load existing wallets from the database.
+    ///
+    /// This only uses persisted mint metadata and does not make network requests.
     pub async fn build(self) -> Result<WalletRepository, Error> {
         let localstore = self
             .localstore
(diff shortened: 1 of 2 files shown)


crates/cdk/src/wallet/wallet_repository.rs
```diff
@@ -1021,6 +1042,77 @@ mod tests {
...
+                .expect("Failed to create in-memory database"),
+        );
+        let (mint_url, direct_connections, listener_handle) =
+            local_mint_url_with_connection_counter().await;
+        localstore
+            .add_mint(
+                mint_url.clone(),
+                Some(mint_info_with_units(vec![
+                    CurrencyUnit::Sat,
+                    CurrencyUnit::Usd,
```
mubarak23, line 1059: is their a need to add  `CurrencyUnit::Usd` since the base unit is `CurrencyUnit::Sat`

asmogo: This will create a mint info object with different units for testing purposes. I think we should test non base units as well.

asmogo:
@cdk-bot review

crodas (approved):

thesimplekid (approved):

PR #2225, opened 2026-07-14
Push keyset updates to the mint on rotation and reconnect

crodas:
### Description

The mint pulled keysets once at construction and refreshed them only when it initiated a rotation itself. A rotation performed on the signatory side, out of band from a mint instance, never reached that mint, and the gRPC client had no reconnect that would re-sync after a dropped connection.

Add a keyset subscription to the Signatory trait, backed by a watch channel. DbSignatory publishes a snapshot whenever its in-memory keysets change. A new server-streaming SubscribeKeysets RPC sends the current snapshot on connect and again on every rotation; the gRPC client reconnects with exponential backoff, so a restored connection re-injects the latest keysets by construction. The mint drains the subscription into its existing ArcSwap and signals in-process consumers through a keyset-change broadcast exposed as Mint::subscribe_keyset_changes.

Document the design in two ADRs: the existing key segregation between signatory and mint, and this push path. Cover the drain with a mock signatory that injects snapshots directly through the subscription, alongside the rotation-driven tests.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-signatory/src/proto/signatory.proto
+++ b/crates/cdk-signatory/src/proto/signatory.proto
@@ -7,6 +7,9 @@ service Signatory {
   rpc VerifyProofs(Proofs) returns (BooleanResponse);
   // returns all the keysets for the mint
   rpc Keysets(EmptyRequest) returns (KeysResponse);
+  // streams the keysets: the current set is sent on subscribe and again on
+  // every rotation
+  rpc SubscribeKeysets(EmptyRequest) returns (stream KeysResponse);
   // rotates the keysets
   rpc RotateKeyset(RotationRequest) returns (KeyRotationResponse);
 }
@@ -14,7 +17,7 @@ service Signatory {
 enum Constants {
   CONSTANTS_UNSPECIFIED = 0;
   // this constant should be sent on the header of the grpc code for verification
-  CONSTANTS_SCHEMA_VERSION = 1;
+  CONSTANTS_SCHEMA_VERSION = 2;
 }
 
 message BlindSignResponse {
--- a/crates/cdk/src/mint/keysets/mod.rs
+++ b/crates/cdk/src/mint/keysets/mod.rs
@@ -94,6 +94,11 @@ impl Mint {
             })
             .await?;
 
+        // Read the freshest snapshot and store it under the shared lock. The
+        // subscription drain task writes the same ArcSwap; serializing the
+        // read-then-store here keeps the last write the newest one, so a
+        // concurrent rotation's snapshot can never be clobbered by a stale read.
+        let _store = self.keyset_store_lock.lock().await;
         let new_keyset = self.signatory.keysets().await?;
         self.keysets.store(new_keyset.keysets.into());
 
--- a/crates/cdk/src/mint/mod.rs
+++ b/crates/cdk/src/mint/mod.rs
@@ -99,6 +99,8 @@ struct TaskState {
     shutdown_notify: Option<Arc<Notify>>,
     /// Handle to the main supervisor task
     supervisor_handle: Option<JoinHandle<Result<(), Error>>>,
+    /// Handle to the keyset drain task
+    keyset_drain_handle: Option<JoinHandle<()>>,
     /// Keyset subscription retained from construction, drained once by the first
     /// `start()`. `None` after it has been taken; a restart re-subscribes.
     keyset_updates: Option<watch::Receiver<SignatoryKeysets>>,
@@ -388,14 +390,24 @@ impl Mint {
             },
         };
 
-        if let Some(mut keyset_updates) = keyset_updates {
+        let keyset_drain_handle = if let Some(mut keyset_updates) = keyset_updates {
             let keysets = self.keysets.clone();
             let keyset_store_lock = Arc::clone(&self.keyset_store_lock);
             let shutdown = shutdown_notify.clone();
-            tokio::spawn(async move {
+            Some(tokio::spawn(async move {
+                // Register the shutdown waiter once and hold it across
+                // iterations. A fresh `shutdown.notified()` per loop would be
+                // dropped whenever the `changed` branch wins the select,
+                // deregistering the waiter for the duration of the handler. A
+                // `notify_waiters()` from `stop()` in that window buffers
+                // nothing, so it would be lost and the next `notified()` would
+                // wait forever, hanging `stop()` on the drain handle. A single
+                // pinned future stays registered, so shutdown is never missed.
+                let shutdown_wait = shutdown.notified();
+                tokio::pin!(shutdown_wait);
                 loop {
                     tokio::select! {
-                        _ = shutdown.notified() => break,
+                        _ = &mut shutdown_wait => break,
                         changed = keyset_updates.changed() => {
                             if changed.is_err() {
                                 // Signatory dropped the sender; stop draining.
@@ -415,12 +427,15 @@ impl Mint {
                         }
                     }
                 }
-            });
-        }
+            }))
+        } else {
+            None
+        };
 
         // Store the handles
         task_state.shutdown_notify = Some(shutdown_notify);
         task_state.supervisor_handle = Some(supervisor_handle);
+        task_state.keyset_drain_handle = keyset_drain_handle;
 
         // Give the background task a tiny bit of time to start waiting
         tokio::time::sleep(std::time::Duration::from_millis(10)).await;
@@ -444,6 +459,7 @@ impl Mint {
         // Take the handles out of the state
         let shutdown_notify = task_state.shutdown_notify.take();
         let supervisor_handle = task_state.supervisor_handle.take();
+        let keyset_drain_handle = task_state.keyset_drain_handle.take();
 
         // If nothing to stop, return early
         let (shutdown_notify, supervisor_handle) = match (shutdown_notify, supervisor_handle) {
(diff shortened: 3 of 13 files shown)


crates/cdk-signatory/src/proto/signatory.proto
asmogo: I think we need to bump this version.

asmogo (changes requested):

crates/cdk/src/mint/mod.rs
```diff
@@ -339,6 +339,40 @@ impl Mint {
...
 
+        // Keyset refresh: drain signatory keyset updates into the in-memory
+        // keysets and notify subscribers. A signatory-side rotation reaches the
+        // mint here without a restart or a mint-initiated rotate.
+        match self.signatory.subscribe_keysets().await {
+            Ok(mut keyset_updates) => {
+                let keysets = self.keysets.clone();
+                let pubsub_manager = Arc::clone(&self.pubsub_manager);
+                let shutdown = shutdown_notify.clone();
+                tokio::spawn(async move {
```
thesimplekid: Above on 331 where we spawn the wait payments we also keep the handle and then await it in shutdown to ensure it completes. Should we do the same here to be consistent about how these processes are handled.

crates/cdk/src/mint/subscription.rs
thesimplekid, line 1: I am wondering if these are needed? The notifications from the pubsub are not consumed anywhere in this PR the mint is updated of the change from the signatory via the watch snapshot, so I am not sure who the consumer of the pubsub notification is? If we had ws notifications for keyset rotations that would be a use case for this but we do not.

docs/adr/0002-signatory-keyset-subscription.md
```diff
@@ -0,0 +1,187 @@
...
+  gap from ADR 0001.
+* In-process consumers are notified of a keyset change instead of polling, and
+  the seam for a future wallet-facing WS notification is in place.
+* The private-key boundary from ADR 0001 is preserved: only public
+  `SignatoryKeysets` cross the trait, and the mint's storage type stays
+  private.
+
+### Negative Consequences
+
+* A new streaming RPC and a schema-version bump on the proto. Older clients
```
thesimplekid, line 187: Older clients will not keep working as we bump the version and we currently require an exact match for these (but there are no older clients so its fine).

crates/cdk/src/mint/mod.rs
```diff
@@ -339,6 +339,40 @@ impl Mint {
             .await
         });
 
+        // Keyset refresh: drain signatory keyset updates into the in-memory
+        // keysets and notify subscribers. A signatory-side rotation reaches the
+        // mint here without a restart or a mint-initiated rotate.
+        match self.signatory.subscribe_keysets().await {
```
thesimplekid: Could we use subscribe_keysets() as the bootstrap source as well as the update source? Currently Mint::new() fetches keysets separately, then start() subscribes and waits for changed(). If a rotation happens between those operations, the receiver is created with the new snapshot already marked as seen, so the mint keeps the stale snapshot until another rotation occurs. Ideally, subscribe during construction, initialize ArcSwap from receiver.borrow_and_update(), retain the receiver, and drain it after start(). At minimum, start() should apply borrow_and_update() immediately before waiting for changed().

crates/cdk/src/mint/keysets/mod.rs
thesimplekid, line 103: I think this PR introduces a race condition here, `signatory.rotate_keyset` requests a rotation and then fetches and writes to cache but the watcher also writes to cache so if the response and notification of two rotations arrive out of order the cache could be invalid.

thesimplekid (changes requested):

docs/adr/0002-signatory-keyset-subscription.md
```diff
@@ -0,0 +1,187 @@
...
+
+```rust
+/// Latest-snapshot stream of the signatory's keysets. Yields the full
+/// current set on subscribe and again on every rotation.
+async fn subscribe_keysets(&self)
+    -> Result<tokio::sync::watch::Receiver<SignatoryKeysets>, Error>;
+```
+
+The existing `keysets()` method stays; the mint uses it for the synchronous
+bootstrap at construction, and `subscribe_keysets()` handles every update
```
thesimplekid: We still have Mint::rotate_keyset that does it own update so the every here is not correct. https://github.com/cashubtc/cdk/pull/2225#discussion_r3597904310

docs/adr/0002-signatory-keyset-subscription.md
```diff
@@ -0,0 +1,187 @@
...
+* Good, because `watch` retains only the latest value (like `ArcSwap`) so a
+  slow consumer never accumulates stale versions.
+* Good, because it wakes the consumer on change (like `mpsc`) so the mint can
+  react: notify wallets, log, refresh derived state.
+* Good, because the mint keeps its storage type private to itself.
+
+**Cons:**
+
+* Bad, because it costs one small drain task on the mint side.
+
```
thesimplekid, line 78: Maybe you considered it and just didn't include it here but another simple option is the mint polling the signatory. Though I am not suggesting we do that I think notifications are better.

docs/adr/0002-signatory-keyset-subscription.md
```diff
@@ -0,0 +1,187 @@
...
+**Pros:**
+
+* Good, because the mint is woken on change and can react.
+
+**Cons:**
+
+* Bad, because an `mpsc` can back up with stale full snapshots under a slow
+  consumer, which contradicts the current-set semantics.
+
+#### `watch` receiver drained into the mint's `ArcSwap`
```
thesimplekid, line 62: Did you consider a generic stream like we already use for payment notifications? I think watch is better but stream does have the advantage of being more generic and not tokio tied and more closly mirrors the underlying transport in the case of grpc.

crodas: I will check that and reuse as much as possible

docs/adr/0002-signatory-keyset-subscription.md
crodas: I did consider it in previous iterations, but it would need yet another payload. Some sort of fingerprint to avoid exchanging a lot of data. Since it would require changes to the signatory trait (and therefore an upgrade to the protofiles), I decided to just support the event stream.

thesimplekid: Yes I think event stream is better.

docs/adr/0002-signatory-keyset-subscription.md
thesimplekid: I think the watch probably works bests is the semantics we want as we don't need every even just what is current.

crodas: I think `watch` is the natural fit; keysets are state, and payments are events.

crodas:
Maybe the last two commits can be its own PR, as a follow up @thesimplekid, let me know and I can fix git history

docs/adr/0002-signatory-keyset-subscription.md
```diff
@@ -0,0 +1,192 @@
...
+
+```rust
+/// Latest-snapshot stream of the signatory's keysets. Yields the full
+/// current set on subscribe and again on every rotation.
+async fn subscribe_keysets(&self)
+    -> Result<tokio::sync::watch::Receiver<SignatoryKeysets>, Error>;
+```
+
+The existing `keysets()` method stays. The mint uses it for the synchronous
+bootstrap at construction and for the read-back inside a mint-initiated
```
thesimplekid: we bootstrap through `subscribe_keysets` now not `keysets`.

docs/adr/0002-signatory-keyset-subscription.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+Because the server sends the current snapshot first on every connect, a
+dropped-then-restored stream re-injects the latest keysets automatically, with
+no client-side cache or diff.
+
+**Mint integration.** Bootstrap is unchanged (`keysets().await` at
+construction). The mint then subscribes and spawns a drain task under the
+existing supervisor:
```
thesimplekid: https://github.com/cashubtc/cdk/pull/2225/changes#r3624944087

crates/cdk/src/mint/mod.rs
```diff
@@ -339,9 +362,71 @@ impl Mint {
...
+
+        let keyset_drain_handle = if let Some(mut keyset_updates) = keyset_updates {
+            let keysets = self.keysets.clone();
+            let keyset_store_lock = Arc::clone(&self.keyset_store_lock);
+            let shutdown = shutdown_notify.clone();
+            Some(tokio::spawn(async move {
+                loop {
+                    tokio::select! {
+                        _ = shutdown.notified() => break,
+                        changed = keyset_updates.changed() => {
```
thesimplekid, line 411: Using the `Notify` in the select like this introduces a case where the notification could be missed. If the change wins the select the notified() future is dropped while the update branch runs. If stop() calls notify_waiters() during that window, the drain misses shutdown, starts a new waiter on the next iteration, and stop() can hang awaiting keyset_drain_handle.

I think we can let this in to get this PR merged and then do a quick follow up to switch the `Notify` use for a Cancellation token that can be shared for all tasks.

thesimplekid (changes requested):
> Maybe the last two commits can be its own PR, as a follow up @thesimplekid, let me know and I can fix git history

Yes lets do this. I think the rest is ready just a few nits on the ADR. So we have two close follow ups for those 2 commits and https://github.com/cashubtc/cdk/pull/2225/changes#diff-f5ceb86f843038da7e04913cad4489e37052c36fb340f3f948ecb5e6fe17d7aeR402

crates/cdk/src/mint/mod.rs
thesimplekid: This fix https://github.com/cashubtc/cdk/pull/2225/commits/8fa730862070aa07f4eea837713b5e10602a43a6 improves this PR and resolves the issue, but I still think a cancel token is worth a follow up PR and the better long term solution.

thesimplekid (approved):

PR #2230, opened 2026-07-17
feat(http-client): support custom and preconfigured transports

prusnak:
needed for https://github.com/cashubtc/cdk/pull/2231

----

Improve `cdk-http-client` extensibility for transports that maintain their own state or establish connections outside the standard HTTP/WebSocket flow.

This PR:

- Removes the global `Default` requirement from the `Transport` trait. Constructors that instantiate a transport still require `T: Default`, while callers can provide preconfigured transports that do not have a meaningful default configuration.
- Adds `from_websocket_stream`, allowing an already-established `WebSocketStream` to be adapted into CDK’s `WsSender` and `WsReceiver` types. Existing direct and Tor connections now reuse this adapter.
- Makes `resolve_dns_txt` an unconditional trait method with a default unsupported implementation. Feature flags therefore no longer change the trait’s required methods or cause compilation failures through Cargo feature unification.

Together, these changes enable stateful transports, encrypted tunnels, and transports that perform their own WebSocket upgrade without complicating the standard client path.

-----

### Notes to the reviewers

The public API changes are additive or relax existing constraints:

- Existing transports implementing `Default` continue to work unchanged.
- `Default` is required only where CDK constructs the transport itself.
- DNS-capable transports can override `resolve_dns_txt`; other transports receive a clear runtime error if it is called.
- The WebSocket adapter accepts any compatible, established `WebSocketStream`.

### DIFF
--- a/crates/cdk-http-client/src/transport/mod.rs
+++ b/crates/cdk-http-client/src/transport/mod.rs
@@ -12,10 +12,14 @@ use url::Url;
 use crate::{HttpClient, HttpClientBuilder};
 use crate::{HttpError, RawResponse};
 
-/// Expected HTTP transport
+/// Expected HTTP transport.
+///
+/// Callers that construct a transport implicitly may add a [`Default`] bound,
+/// while configured transports can be supplied directly without implementing
+/// a meaningless default configuration.
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-pub trait Transport: Default + Send + Sync + Debug + Clone {
+pub trait Transport: Send + Sync + Debug + Clone {
     /// Connect to a WebSocket endpoint using this transport.
     async fn ws_connect(
         &self,
--- a/crates/cdk-http-client/src/transport/mod.rs
+++ b/crates/cdk-http-client/src/transport/mod.rs
@@ -41,9 +41,16 @@ pub trait Transport: Send + Sync + Debug + Clone {
         accept_invalid_certs: bool,
     ) -> Result<(), HttpError>;
 
-    #[cfg(all(feature = "bip353", not(target_arch = "wasm32")))]
     /// DNS resolver to get TXT records from a domain name.
-    async fn resolve_dns_txt(&self, domain: &str) -> Result<Vec<String>, HttpError>;
+    ///
+    /// Transports that support DNS resolution should override this method. The
+    /// default implementation keeps the trait API stable when the `bip353`
+    /// feature is disabled.
+    async fn resolve_dns_txt(&self, _domain: &str) -> Result<Vec<String>, HttpError> {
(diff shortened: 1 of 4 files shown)


prusnak:
More info about 5a71888 for @crodas :

# Why Remove `Default` From `Transport`?

Commit `5a7188827481a0f9433065618de7cf586b6a934f` does not remove `Default` from existing transports. It removes it as a mandatory supertrait of `Transport`.

`Default` is not required to perform HTTP requests. It is only needed by constructors that create a transport internally:

- `HttpClient::new`
- `HttpClient::with_proxy`
- Their `AuthHttpClient` equivalents

Those methods now carry the narrower `T: Default` bound. [`with_transport`](crates/cdk/src/wallet/mint_connector/http_client.rs) accepts an already-constructed transport and therefore does not need it.

This matters for [`EnclaviaTransport`](crates/cdk-enclavia/src/transport.rs). A valid instance requires a mint origin, an enclave endpoint, pinned PCRs, and an asynchronously established and attested client. There is no meaningful synchronous `Default` value. Inventing one would either create an unusable transport or weaken the guarantee that the transport is fully configured and attested before use.

The resulting API expresses the actual requirements:

- Default-constructible transports retain the convenience constructors.
- Configured transports use `with_transport`.
- Transports with mandatory configuration are not forced to implement a misleading `Default`.

## Alternative: Retain `Transport: Default`

We could keep the original bound and make `EnclaviaTransport` represent an unconfigured state:

```rust
enum EnclaviaTransport {
    Unconfigured,
    Ready {
        target: MintTarget,
        client: Client,
        fallback: Async,
    },
}

impl Default for EnclaviaTransport {
    fn default() -> Self {
        Self::Unconfigured
    }
}
```

Every transport operation would return an error when called on `Unconfigured`.

### Advantages

- Preserves the existing public trait contract.
- Downstream generic code can continue assuming that `T: Transport` also means `T: Default`.
- Avoids adding method-specific bounds to the client constructors.
- Reduces semantic-versioning risk: removing a supertrait is a relaxation for implementors, but can still break generic callers that relied on the implied `Default` bound.

### Disadvantages

- `Default::default()` produces a transport that cannot transport anything.
- `BaseHttpClient::<EnclaviaTransport>::new(...)` compiles but fails only when used.
- It makes an unattested, unusable state representable in a security-sensitive type.
- Every request path needs additional state checking and testing.
- A panicking `Default`, global configuration, or silent ordinary-HTTP fallback would be even less desirable.

A cleaner but more invasive alternative would introduce separate traits for configured transports and default-constructible transports. That avoids invalid states while retaining the old `Transport: Default` contract, but adds
public API surface and migration complexity. The method-level bounds in `5a7188827481a0f9433065618de7cf586b6a934f` achieve the same capability separation with a much smaller design.

crodas (approved):
I like this, now I understand why you moved Default from the Transport to the functions taht would need it.

thesimplekid (approved):

PR #2231, opened 2026-07-17
feat(enclavia): add attested wallet transport

prusnak:
### Description

Add `cdk-enclavia`, a wallet transport for connecting to CDK mints running inside [Enclavia](https://enclavia.io) enclaves.

The transport establishes an encrypted Noise channel and verifies the enclave’s AWS Nitro attestation against pinned PCR0, PCR1, and PCR2 measurements before constructing the wallet client.

Key functionality:

- Adds `EnclaviaClientBuilder` for configuring PCR pinning, attestation mode, upgrade trust, authentication, headers, and fallback HTTP transport.
- Routes requests for the configured mint origin exclusively through the attested channel.
- Fails closed for mint requests: errors from the enclave are never retried over ordinary HTTP.
- Continues routing unrelated requests, such as LNURL callbacks and external OIDC requests, through the regular HTTP transport.
- Carries native NUT-17 WebSocket subscriptions through the same attested channel.
- Rejects WebSocket connections that do not match the configured mint origin.
- Adds `cdk-enclavia-cli` for verifying attestation and requesting `/v1/info` through the encrypted channel.

-----

### Notes to the reviewers

This is stacked on the accompanying `cdk-http-client` transport extensibility changes (https://github.com/cashubtc/cdk/pull/2230).

Security-sensitive behavior:

- Mint routing compares the full URL origin, including scheme, host, and effective port.
- WebSocket routing additionally requires the corresponding `ws` or `wss` scheme and rejects credentials or mismatched origins.
- There is no ordinary HTTP fallback for requests addressed to the configured mint.
- NUT-17 WebSocket support is currently native-only; WASM callers receive an explicit error.
- `--debug` is intended only for QEMU/development enclaves. It skips AWS Nitro certificate-chain and attestation-signature verification, while retaining Noise-session nonce and PCR checks.

The upstream `enclavia` 0.1.0 crate requires Rust 1.88, so `cdk-enclavia` has a crate-specific MSRV of 1.88. The rest of the workspace retains its Rust 1.85 MSRV.


crates/cdk-enclavia/src/transport.rs
```diff
@@ -0,0 +1,344 @@
...
+            .upgrade(Method::Get, &target, &upgrade_headers)
+            .await
+            .map_err(|error| cdk_http_client::ws::WsError::Connection(error.to_string()))?;
+        let ws_stream = WebSocketStream::from_raw_socket(stream, Role::Client, None).await;
+
+        Ok(cdk_http_client::ws::from_websocket_stream(ws_stream))
+    }
+
+    #[cfg(target_arch = "wasm32")]
+    async fn ws_connect(
```
crodas, line 294: This is fine; subscriptions would fall back to [polling](https://github.com/cashubtc/cdk/blob/23fa905d661fa651d5d6a9f65f95f722cd124898/crates/cdk/src/wallet/subscription.rs#L250-L477) whenever the WebSocket [fails](https://github.com/cashubtc/cdk/blob/23fa905d661fa651d5d6a9f65f95f722cd124898/crates/cdk/src/wallet/subscription.rs#L240-L247)

crates/cdk-http-client/src/transport/mod.rs
```diff
@@ -37,9 +41,16 @@ pub trait Transport: Default + Send + Sync + Debug + Clone {
...
     ) -> Result<(), HttpError>;
 
-    #[cfg(all(feature = "bip353", not(target_arch = "wasm32")))]
     /// DNS resolver to get TXT records from a domain name.
-    async fn resolve_dns_txt(&self, domain: &str) -> Result<Vec<String>, HttpError>;
+    ///
+    /// Transports that support DNS resolution should override this method. The
+    /// default implementation keeps the trait API stable when the `bip353`
+    /// feature is disabled.
+    async fn resolve_dns_txt(&self, _domain: &str) -> Result<Vec<String>, HttpError> {
```
crodas, line 49: Unrelated to the PR, but this can be ported independently to main.

crates/cdk-http-client/src/transport/mod.rs
```diff
@@ -12,10 +12,14 @@ use url::Url;
...
 
-/// Expected HTTP transport
+/// Expected HTTP transport.
+///
+/// Callers that construct a transport implicitly may add a [`Default`] bound,
+/// while configured transports can be supplied directly without implementing
+/// a meaningless default configuration.
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-pub trait Transport: Default + Send + Sync + Debug + Clone {
```
crodas: Why are you dropping the Default?

prusnak:
@crodas lets discuss changes in http-client in #2230, not here - the description of the PR provides answers for what you ask

thesimplekid:
I think this is really cool but might be a bit before we merge it as we don't have mints running in this yet (or maybe we do?) so would like to see that closer before truly supporting this. Though since this is pretty self contained in its own crate the other option is to merge and just mark it as more experimental. BDK does something like this https://github.com/bitcoindevkit/#experimental though at the org level since they are not a mono repo but this could still be okay, open to discussion. 

This is a good opportunity to review how easy it is to use a transport cdk does not "officially" support as this should be possible with the trait but I don't think as really been done and maybe https://github.com/cashubtc/cdk/pull/2230 is a step towards that.

crates/cdk-http-client/src/transport/mod.rs
prusnak: see https://github.com/cashubtc/cdk/pull/2230 for more details

crates/cdk-http-client/src/transport/mod.rs
prusnak: see https://github.com/cashubtc/cdk/pull/2230 for more details

prusnak:
> I think this is really cool but might be a bit before we merge it

Probably so, I just wanted to introduce changes to http-client earlier (in #2230) and this PR is providing context to justify these changes.

> as we don't have mints running in this yet (or maybe we do?) so would like to see that closer before truly supporting this.

I tested it with the cdk-mintd instance running in Enclavia, but it's true we are far from this being common.

> Though since this is pretty self contained in its own crate the other option is to merge and just mark it as more experimental. BDK does something like this https://github.com/bitcoindevkit/#experimental though at the org level since they are not a mono repo but this could still be okay, open to discussion.

We could also set cdk-enclavia to `publish = false` for the time being.

> This is a good opportunity to review how easy it is to use a transport cdk does not "officially" support as this should be possible with the trait but I don't think as really been done and maybe #2230 is a step towards that.

+1 that was my main goal

prusnak:
Rebased on top of current master since https://github.com/cashubtc/cdk/pull/2230 has been merged

afilini (commented):
I can't speak for the cdk-specific parts but the Enclavia integration looks good to me!

I was going to comment that earlier today I released 0.2.0 with the auto-reconnect but it looks like you've beaten me to it :)

PR #2232, opened 2026-07-17
feat: verify and sign the length-framed NUT-11 SIG_ALL message

robwoodgate:
### Description

Implements the length-framed NUT-11 SIG_ALL message aggregation (`Cashu_SigAllSig_v1`) proposed in https://github.com/cashubtc/nuts/pull/404. The framing matches NUT-20/29 mint quote signatures and reuses their helpers; the message is signed over its BIP-340 tagged hash, the same construction NUT-06 signed mint info (cashubtc/nuts#416) uses.

-----

### Notes to the reviewers

- The mint accepts the v1 and current message formats as digests (`sig_all_digests_to_verify`: the v1 tagged hash and the SHA-256 of the current string, verified and signed via `verify_digest`/`sign_digest`), and wallet signing produces both, so requests verify on mints on either side of the upgrade. The pre-0.14 format is neither signed nor accepted: it does not commit to C values or output amounts, and `test_sig_all_should_reject_if_the_output_amounts_are_swapped` covers this.
- `valid_signatures_any_digest` counts each pubkey at most once across formats, so a multi-format witness cannot trip `DuplicateSignature`.
- The witness length cap is split from the secret cap (`MAX_PROOF_WITNESS_LEN`, 8192 bytes) so multisig witnesses carrying one signature per format per signer still fit.
- Canonical vectors from the spec's `tests/11-test.md` are pinned in the nut11 test module.
- Ran on stable 1.96 in a container: `cargo test -p cashu --features mint,wallet,nostr`, `cargo test -p cdk --lib`, `cargo clippy -p cashu --all-targets`, `cargo fmt --check`.

-----

#### CHANGED

- SIG_ALL verification accepts the length-framed v1 message alongside the current format, counting unique pubkeys with valid signatures and ignoring signatures that do not verify
- Proof witness length is capped separately from secret length (8192 bytes)

#### ADDED

- NUT-11 v1 length-framed SIG_ALL message; `sign_sig_all` signs all accepted message formats

-----

### Checklist

* [ ] I ran `just quick-check` before committing (ran the equivalent commands listed above on stable in a container)
* [x] If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`) — not modified

### DIFF
--- a/crates/cashu/src/nuts/nut01/public_key.rs
+++ b/crates/cashu/src/nuts/nut01/public_key.rs
@@ -94,7 +94,12 @@ impl PublicKey {
     /// Verify schnorr signature
     pub fn verify(&self, msg: &[u8], sig: &Signature) -> Result<(), Error> {
         let hash: Sha256Hash = Sha256Hash::hash(msg);
-        let msg = Message::from_digest_slice(hash.as_ref())?;
+        self.verify_digest(hash.to_byte_array(), sig)
+    }
+
+    /// Verify a schnorr signature over an already hashed 32-byte message.
+    pub(crate) fn verify_digest(&self, digest: [u8; 32], sig: &Signature) -> Result<(), Error> {
+        let msg = Message::from_digest(digest);
         SECP256K1.verify_schnorr(sig, &msg, &self.inner.x_only_public_key().0)?;
         Ok(())
     }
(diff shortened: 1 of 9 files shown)


asmogo (dismissed):

asmogo:
@cdk-bot review

thesimplekid:
> @cdk-bot review

Seem like the bot took a nap last night. Should be back up and reviewing now

robwoodgate:
Removed the legacy signing format this PR had re-added.

robwoodgate:
UPDATED: https://github.com/cashubtc/cdk/pull/2232/changes/8064df40fb5f110717619e874dc4fd763488c96c

This brings CDK in line with the latest version of the spec, which replaces the flat DST with a tagged hash

PR #2239, opened 2026-07-20
feat(cdk-mint-rpc): add KeysetService alongside legacy CdkMint

orangeshyguy21:
### Description

This PR begins splitting the monolithic `CdkMint` mint-management gRPC service into focused, per-domain services. It introduces the first one, `KeysetService` (package `cdk_mint_keyset_v1`), which exposes `RotateNextKeyset`.

The new service is registered **alongside** the existing `CdkMint` service on the same endpoint (additive dual-serve), so existing RPC clients are unaffected and the mint RPC protocol version is unchanged (`1.0.0`). The `cdk-mint-cli rotate-next-keyset` command now targets the new service.

-----

### Notes to the reviewers

- **First of a planned per-service split.** Follow-up PRs move the remaining domains (quotes, payment methods, mint info) onto their own services the same way. Doing it additively means every intermediate release keeps the legacy `CdkMint` service working; a later PR removes the legacy service and bumps the protocol version — this PR is not a breaking change.
- **No behavior change to the legacy path.** The rotation logic was extracted into an inherent `MintRPCServer::rotate_keyset` method that both the legacy `CdkMint` impl and the new `KeysetService` impl delegate to.
- **Naming/layout follows existing house style.** Package `cdk_mint_keyset_v1` mirrors the existing `cdk_mint_management_v1`, and the new proto lives beside the legacy one in `src/proto/`. Proto lint / breaking-change CI (buf) is intentionally deferred to a follow-up PR to keep this one focused.
- The version-check interceptor is applied to the new service as well.

-----

#### CHANGED

- `cdk-mint-cli rotate-next-keyset` now calls the new `KeysetService` instead of `CdkMint`.

#### ADDED

- `KeysetService` gRPC service (`cdk_mint_keyset_v1`, `RotateNextKeyset`), served alongside the existing `CdkMint` management service as the first step of a per-domain split.

#### REMOVED

- None.

#### FIXED

- None.

----

### Checklist

* [x] If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`) — N/A, no Wallet API change

### DIFF
--- a/crates/cdk-mint-rpc/src/proto/keyset.proto
+++ b/crates/cdk-mint-rpc/src/proto/keyset.proto
@@ -1,0 +1,24 @@
+syntax = "proto3";
+
+package cdk_mint_keyset_v1;
+
+// Keyset administration for a mint.
+service KeysetService {
+    // Rotates to the next keyset for the specified currency unit.
+    rpc RotateNextKeyset(RotateNextKeysetRequest) returns (RotateNextKeysetResponse) {}
+}
+
+message RotateNextKeysetRequest {
+    string unit = 1;
+    repeated uint64 amounts = 2;
+    optional uint64 input_fee_ppk = 3;
+    optional bool use_keyset_v2 = 4;
+    optional uint64 final_expiry = 5;
+}
+
+message RotateNextKeysetResponse {
+    string id = 1;
+    string unit = 2;
+    repeated uint64 amounts = 3;
+    uint64 input_fee_ppk = 4;
+}
--- a/crates/cdk-mint-rpc/src/proto/keyset.proto
+++ b/crates/cdk-mint-rpc/src/proto/keyset.proto
@@ -9,16 +9,36 @@ service KeysetService {
 }
 
 message RotateNextKeysetRequest {
+    // Currency unit the new keyset signs, e.g. "sat". Matched case-
+    // insensitively. A unit the mint cannot derive a keyset for fails with
+    // gRPC status INVALID_ARGUMENT.
     string unit = 1;
+    // Token denominations the keyset can sign, e.g. [1, 2, 4, 8, 16]. If
+    // empty, the amounts of the currently active keyset for this unit are
+    // reused; rotating a unit that has no active keyset requires a non-empty
+    // list.
     repeated uint64 amounts = 2;
+    // Input fee charged per input, in parts per thousand. Defaults to 0 when
+    // omitted.
     optional uint64 input_fee_ppk = 3;
+    // When true (the default), emit a NUT-02 Version01 keyset; when false,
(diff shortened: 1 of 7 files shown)


thesimplekid:
@cdk-bot review

crates/cdk-mint-rpc/src/proto/keyset.proto
```diff
@@ -0,0 +1,44 @@
...
+service KeysetService {
+    // Rotates to the next keyset for the specified currency unit.
+    rpc RotateNextKeyset(RotateNextKeysetRequest) returns (RotateNextKeysetResponse) {}
+}
+
+message RotateNextKeysetRequest {
+    // Currency unit the new keyset signs, e.g. "sat". Matched case-
+    // insensitively. A unit the mint cannot derive a keyset for fails with
+    // gRPC status INVALID_ARGUMENT.
+    string unit = 1;
```
thesimplekid: cdk is only case insensitive for known units right now I think, unknown units we keep what is set maybe we should change this?

thesimplekid (commented):
Nice, I think this looks good. 

We'll have to think about how we want to handle versions as right now there is just a global version that we check but I think we can resolve that latter, and just follow better grpc practices.

crates/cdk-mint-rpc/src/proto/keyset.proto
orangeshyguy21: Good catch.

This asymmetry seems worth addressing in core as a dedicated effort. Uppercase custom units would create issues since the mint API always serves the lowercased values. The mint would likely throw unsupported unit errors for its own advertised unit.

I'll fix this comment to say known units only.

orangeshyguy21:
On versioning, yeah the global header won't play nicely with modular services that are independently versioned.

We could drop it once the swap is complete in favor of the standard package-versioning pattern

* unknown versions get `UNIMPLEMENTED` errors
* breaking changes become a `_v2` served alongside
* A buf breaking CI job (after the split) keeps `_v1` additive-only.

crates/cdk-mint-rpc/src/proto/keyset.proto
thesimplekid: I think I intentionally kept custom case sensitive because someone was using mixed case custom units or depend on one case or the other, but maybe we should reevaluate this and just say everything is lower case.

orangeshyguy21: I ran an uppercased custom unit through the blender and documented the issue in [2243](https://github.com/cashubtc/cdk/issues/2243)

thesimplekid (approved):

PR #2242, opened 2026-07-21
move mintd config to db

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


asmogo:
@cdk-bot review

crates/cdk-mintd/src/lib.rs
```diff
@@ -1581,6 +1591,48 @@ async fn build_mint(
...
+    if !preserve_database_values {
+        tracing::info!(
+            "Applying mint info and quote TTL from the database-backed configuration document."
+        );
+        if let Ok(stored_mint_info) = mint.mint_info().await {
+            if configured_mint_info.pubkey.is_none() {
+                configured_mint_info.pubkey = stored_mint_info.pubkey;
+            }
+        }
+        mint.set_mint_info(configured_mint_info).await?;
```
orangeshyguy21: `set_mint_info(configured_mint_info)` overwrites mint_info wholesale, so any RPC-managed fields (name, MOTD, description, icon, urls, contact, tos) that were changed between the apply are discarded on boot when `applied == false`.

It's an edge case, as most operators would swiftly reboot after. However it it's also a known state so the rpc command could warn there is a pending config apply or similar.

crates/cdk-mintd/src/config_service.rs
```diff
@@ -0,0 +1,943 @@
...
+    }
+
+    /// Parses, resolves, and validates an import document.
+    pub fn validate_document(
+        document: &str,
+    ) -> Result<ResolvedConfiguration, ConfigurationServiceError> {
+        let mut settings = Settings::try_from_toml(document)?;
+        prune_inactive_configuration(&mut settings);
+        resolve_secrets(&mut settings)?;
+        crate::validate_settings(&settings)
```
orangeshyguy21, line 175: `validate_document` runs `validate_settings`, but that doesn't include the backend-construction checks that startup performs, so `config validate` (and `config init`) report success for configs the daemon then refuses to boot. 

Repro on this branch: a document selecting `ln_backend = "fakewallet"` with no `[fake_wallet]` section returns "Configuration is valid." and `config init` succeeds, but `cdk-mintd` then exits with `Fake wallet backend selected but [fake_wallet] config section is missing`.

crates/cdk-mintd/src/config_service.rs
```diff
@@ -0,0 +1,943 @@
...
+        return false;
+    }
+    if configured.engine != DatabaseEngine::Postgres {
+        return true;
+    }
+    match (&configured.postgres, &bootstrap.postgres) {
+        (Some(configured), Some(bootstrap)) => {
+            configured.url == bootstrap.url
+                && configured.tls_mode == bootstrap.tls_mode
+                && configured.max_connections == bootstrap.max_connections
```
orangeshyguy21, line 406: Including `max_connections`, `connection_timeout_seconds`, and `tls_mode` in the primary-database identity check means these pool-tuning values can't be changed via `config apply` 

The attempt gets rejected as `PrimaryDatabaseChange` ("primary database settings do not match the bootstrap database settings")

Is this the intent? or would `engine` + `url` be enough of a gate?

crates/cdk-mintd/src/config_service.rs
```diff
@@ -0,0 +1,943 @@
...
+        }
+        std::fs::read_to_string(path).map_err(|source| ConfigurationServiceError::FileSecret {
+            field,
+            path: path.to_path_buf(),
+            source,
+        })?
+    } else {
+        return Err(ConfigurationServiceError::LiteralSecret { field });
+    };
+    let resolved = resolved.trim().to_owned();
```
orangeshyguy21: This alters any secret whose value contains leading/trailing whitespace.

Could this be an issue for API keys or DB passwords?

crates/cdk-mintd/src/main.rs
```diff
@@ -15,23 +15,103 @@ fn main() -> Result<()> {
...
+                    print!("{document}");
+                    Ok(())
+                }
+                ConfigCommands::Export(file) => {
+                    let work_dir = work_dir
+                        .as_deref()
+                        .expect("database commands have a work directory");
+                    let document = cdk_mintd::stored_configuration_document(work_dir, password)
+                        .await?;
+                    std::fs::write(&file.file, document).with_context(|| {
```
orangeshyguy21: `config export` overwrites the target path with no check here. 

Worth the convenience? or better to error if the file already exists (unless a `--force`/`--overwrite` flag is passed)

orangeshyguy21 (commented):
## Review summary

Robust change to make the database the single source of truth. It's a tricky problem with lots of tradeoffs and this appears to thread the needle fairly well. Were the methods here rolled fresh or are there analogs in other projects that drew inspiration?

I left some comments inline with questions on implementation and noted an unhanded edge case.

One general question: Readme updates are good in conjunction with cli --help, but should this breaking change come with some dedicated migration material to cover the upgrade path?

crates/cdk-mintd/src/config_service.rs
```diff
@@ -0,0 +1,943 @@
...
+        .onchain
+        .as_ref()
+        .is_some_and(|onchain| onchain.onchain_backend == crate::config::OnchainBackend::Bdk)
+    {
+        settings.bdk = None;
+    }
+}
+
+fn resolve_secrets(settings: &mut Settings) -> Result<(), ConfigurationServiceError> {
+    resolve_optional_secret(&mut settings.info.seed, "info.seed")?;
```
crodas: Keys are suppossed to be managed by the signatory, which may run from a different place than the mint

crates/cdk-mintd/src/config_service.rs
```diff
@@ -0,0 +1,943 @@
...
+        })
+    }
+
+    /// Loads and validates the document selected for startup.
+    pub(crate) async fn startup(&self) -> Result<StartupConfiguration, ConfigurationServiceError> {
+        let envelope = self.repository.active().await?;
+        let resolved = Self::validate_document(&envelope.toml)?;
+        self.require_primary_database(&resolved.settings.database)?;
+        let signing_identity = discover_signing_identity_async(&resolved.settings).await?;
+        validate_authored_mint_pubkey(&resolved.settings, &signing_identity)?;
```
crodas: Why are we signing the configuration? The signing and then verification is for adversarial scenarios, but the database is trusted by the mint. 

I cannot find any reason for this; is it documented somewhere?

asmogo: This is not signing the configuration. The TOML configuration contains a secret reference `env:CDK_MINTD_MNEMONIC` or `file:/path/to/file`. 
We verify that the resolved seed / mnemonic still matches the discovered signatory identity.

crates/cdk-mintd/src/lib.rs
asmogo: fixed in [63dd55b](https://github.com/cashubtc/cdk/pull/2242/commits/63dd55be8ad153c0bacc8efc24e0bf98ce2dfefd)

crates/cdk-mintd/src/config_service.rs
asmogo: fixed in https://github.com/cashubtc/cdk/pull/2242/commits/b91bec30deb401088e1f952c4e2b71895b81863c

crates/cdk-mintd/src/config_service.rs
asmogo: fixed in [635707b](https://github.com/cashubtc/cdk/pull/2242/commits/635707b5450b49797eb7ac403e070a2ce9196434)

crates/cdk-mintd/src/config_service.rs
asmogo: fixed in [9f08ecc](https://github.com/cashubtc/cdk/pull/2242/commits/9f08ecc0add5ebe373b35c5733628926840c41ee).

if they are managed by a remote signatory, we do not require any keys in the mint configuration.

crates/cdk-mintd/src/main.rs
asmogo: i have added a `--force` parameter https://github.com/cashubtc/cdk/pull/2242/commits/1683d8b12bb1afa7129875d577717d626945ceef

asmogo:
enforcing remote or embedded signatory in configuration https://github.com/cashubtc/cdk/pull/2242/commits/9f08ecc0add5ebe373b35c5733628926840c41ee

asmogo:
@cdk-bot review

asmogo:
@cdk-bot review

crates/cdk-mintd/src/lib.rs
```diff
@@ -1642,41 +1817,28 @@ async fn start_services_with_shutdown(
...
+        rpc_enabled && !force_configuration,
+    )
+    .await?;
 
-        if let Ok(mint_info) = mint.mint_info().await {
-            if mint_builder_info.pubkey.is_none() {
-                mint_builder_info.pubkey = mint_info.pubkey;
+    if let Some(activation) = activation {
+        if let Some(expected_revision) = activation.expected_revision {
+            if !activation.service.mark_applied(expected_revision).await? {
```
thesimplekid: We mark applied before the management rpc or the mint itself starts, so an invalid new config could be marked applied and then fail later do you think this is a problem or this just requires manual intervention and fixing the config?

crates/cdk-mintd/src/lib.rs
```diff
@@ -1581,6 +1697,59 @@ async fn build_mint(
...
+        mint.set_quote_ttl(configured_quote_ttl).await?;
+        return Ok(());
+    }
+
+    if mint.mint_info().await.is_err() {
+        tracing::info!("Mint info not set on mint, setting.");
+        mint.set_mint_info(configured_mint_info).await?;
+        mint.set_quote_ttl(configured_quote_ttl).await?;
+        return Ok(());
+    }
```
thesimplekid: I wonder if these should be atomic, if `set_quote` succeeds but `set_mint_info` fails we are left with a partially updated state?

crates/cdk-mintd/src/config_migration.rs
```diff
@@ -0,0 +1,1505 @@
...
+                    &mut secrets,
+                )?;
+            }
+        }
+    }
+
+    crate::validate_settings(&resolved).context("legacy effective configuration is invalid")?;
+    let migrated_document =
+        toml::to_string_pretty(&migrated).context("could not serialize migrated configuration")?;
+    Settings::try_from_toml(&migrated_document)
```
thesimplekid, line 348: Should forced migration outputs be staged in temporary files and renamed only after every write succeeds? With --force, secret files are truncated sequentially, and this path deliberately skips cleanup if a later secret or output write fails. That can leave an old document alongside partially replaced secrets, including the mint mnemonic.

thesimplekid (commented):
I think there are a few things we can do as follow ups to this but can keep this PR like this for now. For example:

1. centralize activation into an explicit Pending → Activating → Applied state machine? Lifecycle policy is currently spread across ConfigRepository, ConfigurationService, ConfigurationActivation, force_configuration, and preserve_database_values. Consolidating those transitions would eliminate several ambiguous boolean/Option combinations and make failure and rollback behavior easier to verify. 

2. start_services_with_shutdown() currently builds resources, mutates databases, starts tasks, binds listeners, commits configuration state, and serves requests.

  Split it into:
```rust
  let prepared = PreparedMintd::prepare(...).await?;
  let running = prepared.activate(...).await?;
  running.serve(shutdown).await
```
3.  Add a secret type that resolves either the file or env currently this is done manually in a few places.
```rust
enum SecretRef {
      Environment(String),
      File(PathBuf),
  }
```

crates/cdk-mintd/src/lib.rs
asmogo: Yes I think this is a real issue. I will move the mark_applied invocation to be after the mint binds successfully the tcp listener.

crates/cdk-mintd/src/lib.rs
asmogo: This is a real concern. I will move the KV writes into a single transaction.

crates/cdk-mintd/src/config_migration.rs
asmogo: Yes. I am going to update this migration.

asmogo:
@cdk-bot review

crates/cdk-mintd/src/lib.rs
asmogo: fixed in https://github.com/cashubtc/cdk/pull/2242/commits/e1332e2f9f3ba96bfc82b6f146eb5d2f7c2ebc63

crates/cdk-mintd/src/config_migration.rs
asmogo: fixed in https://github.com/cashubtc/cdk/pull/2242/commits/02fe1b5f176b08abbecdb52dc0468f132035f797

crates/cdk-mintd/src/lib.rs
asmogo: fixed in https://github.com/cashubtc/cdk/pull/2242/commits/3ab32fcb7087200e2ce328f6e85008cf8f1fd5f3

misc/interactive_regtest_mprocs.sh
```diff
@@ -166,139 +166,224 @@ export CDK_TEST_MINT_URL="http://$CDK_ITESTS_MINT_ADDR:$CDK_ITESTS_MINT_PORT_0"
...
+console_level = "debug"
+file_level = "debug"
+
+[database]
+engine = "$MINTD_DATABASE_ENGINE"
+
+[database.postgres]
+url = "env:CDK_MINTD_DATABASE_URL"
+
+[[ln]]
```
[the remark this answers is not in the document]
asmogo: fixed in https://github.com/cashubtc/cdk/pull/2242/commits/796808555eb289e638f74b08d723c917478b047c

asmogo:
@cdk-bot review

asmogo:
@cdk-bot please review the latest changes

thesimplekid (dismissed):

PR #2251, opened 2026-07-23
Rate-limit mint requests with a persistent budget

crodas:
> This PR supersedes #1769, as it is a generic implementation with persistence in mind. Thanks @Micah-Shallom for your contribution

Mints cap how many requests a client may send per minute, and a busy wallet can trip that cap and start getting its requests rejected. Pace outbound HTTP so the wallet stays under the limit without the caller having to manage it.

A GCRA token bucket wraps the HTTP transport and is enabled by default; it can be reconfigured or turned off on the wallet builder. The main client and the blind-auth client share one bucket, so NUT-22 auth traffic and normal mint traffic draw down a single budget instead of two independent streams. Sharing is why auth-client construction moved from set_auth_cat to build().

The budget is persisted per mint host in the wallet key-value store, so a wallet that is built, used, and dropped hands its remaining budget to the next wallet built for the same mint instead of starting full and bursting again. The write stays on the request path so the value is durable as soon as a request returns, but it is bounded by a short timeout so a slow or hung storage backend can never stall outbound requests. Persistence errors are logged and never propagated into the request path.

The limiter lives in cdk-common so any crate can reuse it, gated behind the wallet and http features; the wallet consumes it through the same transport path as before, so existing call sites and type aliases are unchanged. cdk-common gains a cross-platform sleep (tokio time on native, gloo-timers on wasm) and keeps its bip353 feature in lockstep with the http client's so the resolve_dns_txt gate lines up. Timing uses web-time so the algorithm also works under wasm.

Testing covers the bucket (burst, pacing, refill, concurrency, try_acquire, shared clones), the rate-limited transport decorator (it paces HTTP, passes proxy setup through untouched, and shares one budget across clones), the builder wiring down both the rate-limited and plain paths, and a persistence round-trip that proves a rebuilt wallet inherits a drained budget.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/Cargo.toml
+++ b/crates/cdk-common/Cargo.toml
@@ -23,7 +23,10 @@ prometheus = ["cdk-prometheus/default"]
 # precedence; the backends are additive).
 http = ["dep:cdk-http-client", "cdk-http-client/bitreq"]
 grpc = ["dep:tonic"]
-bip353 = []
+# Keep cdk-common's bip353 in lockstep with the http client's, so the
+# `Transport::resolve_dns_txt` gate (and the rate limiter's decorator of it)
+# lines up whenever the HTTP backend is present.
+bip353 = ["cdk-http-client?/bip353"]
 
 [dependencies]
 async-trait.workspace = true
@@ -56,9 +59,10 @@ tokio = { version = "1", default-features = false, features = ["rt", "rt-multi-t
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 uuid = { workspace = true, features = ["js"], optional = true }
 getrandom = { version = "0.2", features = ["js"] }
-tokio = { workspace = true, default-features = false }
+tokio = { workspace = true, default-features = false, features = ["sync"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
+gloo-timers = { workspace = true }
 
 [dev-dependencies]
 rand.workspace = true
--- a/crates/cdk-common/Cargo.toml
+++ b/crates/cdk-common/Cargo.toml
@@ -23,7 +23,10 @@ prometheus = ["cdk-prometheus/default"]
 # precedence; the backends are additive).
 http = ["dep:cdk-http-client", "cdk-http-client/bitreq"]
 grpc = ["dep:tonic"]
-bip353 = []
+# Keep cdk-common's bip353 in lockstep with the http client's, so the
+# `Transport::resolve_dns_txt` gate (and the rate limiter's decorator of it)
+# lines up whenever the HTTP backend is present.
+bip353 = ["cdk-http-client?/bip353"]
 
 [dependencies]
 async-trait.workspace = true
@@ -51,14 +54,15 @@ tonic = { workspace = true, optional = true }
 cdk-http-client = { workspace = true, default-features = false, optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio = { version = "1", default-features = false, features = ["rt", "rt-multi-thread", "macros", "test-util", "sync"] }
+tokio = { version = "1", default-features = false, features = ["rt", "rt-multi-thread", "macros", "test-util", "sync", "time"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 uuid = { workspace = true, features = ["js"], optional = true }
 getrandom = { version = "0.2", features = ["js"] }
-tokio = { workspace = true, default-features = false }
+tokio = { workspace = true, default-features = false, features = ["sync"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
+gloo-timers = { workspace = true }
 
 [dev-dependencies]
 rand.workspace = true
--- a/crates/cdk-common/src/rate_limit.rs
+++ b/crates/cdk-common/src/rate_limit.rs
@@ -1,0 +1,1135 @@
+//! Client-side request rate limiting for the wallet.
+//!
+//! The wallet paces its outbound HTTP requests to a mint so it stays under the
+//! mint's server-side request cap without the caller having to think about it.
+//! Pacing uses the Generic Cell Rate Algorithm (GCRA): the bucket tracks a
+//! single theoretical arrival time (TAT) rather than a refilling token count.
+//!
+//! The budget is persisted per mint host in the wallet key-value store, so a
+//! wallet that is built, used, and dropped hands its remaining budget to the
+//! next wallet built for the same mint instead of starting full and bursting
+//! again.
+
+use std::future::Future;
+use std::num::NonZeroU32;
+use std::sync::{Arc, Mutex};
+
+use async_trait::async_trait;
+use cdk_http_client::Transport;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use tokio::sync::{watch, OnceCell};
+use url::Url;
+use web_time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+
+use crate::database::{self, WalletDatabase, KVSTORE_NAMESPACE_KEY_MAX_LEN};
+use crate::mint_url::MintUrl;
+use crate::{AuthToken, HttpError, RawResponse};
+
+/// Namespace under which per-host rate-limit budgets are stored.
+const KV_NAMESPACE: &str = "rate_limiter";
+
+/// Configuration for a [`TokenBucket`].
+///
+/// Both fields are `NonZeroU32`, so a zero or otherwise invalid configuration
+/// cannot be constructed; there is no separate runtime validation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct RateLimitConfig {
+    /// Maximum burst: how many requests may go out back-to-back before pacing
+    /// kicks in.
+    pub capacity: NonZeroU32,
+    /// Sustained rate: how many requests are earned back per minute.
+    pub refill_per_minute: NonZeroU32,
+}
+
+impl RateLimitConfig {
+    /// Create a new configuration.
+    pub fn new(capacity: NonZeroU32, refill_per_minute: NonZeroU32) -> Self {
+        Self {
+            capacity,
+            refill_per_minute,
+        }
+    }
+
+    /// Create a configuration from plain integers, returning `None` if either is
+    /// zero. A convenience over building the `NonZeroU32`s by hand.
+    pub fn try_new(capacity: u32, refill_per_minute: u32) -> Option<Self> {
+        Some(Self::new(
+            NonZeroU32::new(capacity)?,
+            NonZeroU32::new(refill_per_minute)?,
+        ))
+    }
+}
+
+impl Default for RateLimitConfig {
+    fn default() -> Self {
+        // Chosen to stay under a mint's per-minute cap (nutshell defaults to 60)
+        // and to match coco's effective wallet defaults so both wallets pace a
+        // shared mint identically. Over any 60s window a GCRA bucket admits at
+        // most `capacity + refill` requests: one full burst plus the steady
+        // trickle. So the sum, not the sustained rate alone, is what must stay
+        // under the cap. 20 + 20 = 40 leaves headroom for a larger burst while
+        // sustaining 20 requests/minute. The `expect`s are on compile-time
+        // constant literals that are trivially non-zero.
+        Self {
+            capacity: NonZeroU32::new(20).expect("20 is non-zero"),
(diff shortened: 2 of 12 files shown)


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk/src/wallet/builder.rs
```diff
@@ -253,6 +254,57 @@ impl WalletBuilder {
...
+            .take()
+            .map(|config| TokenBucket::for_mint(config, &mint_url, localstore.clone()));
+        let shared_transport = rate_limit_bucket
+            .clone()
+            .map(|bucket| Arc::new(RateLimitedTransport::with_bucket(Async::default(), bucket)));
+
+        // The auth wallet comes either from a CAT set on the builder (built here
+        // so it can share the transport) or from a pre-built wallet supplied
+        // directly, which is used verbatim.
+        let auth_wallet = match self.auth_cat.take() {
```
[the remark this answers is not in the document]
crodas: I guess a custom wallet should come with its own transport.

CHANGELOG.md
thesimplekid, line 1: I'll do edits to the change log before release can you remove this.

crates/cdk-ffi/src/wallet.rs
```diff
@@ -913,6 +943,42 @@ impl Wallet {
...
+///
+/// // Allow a burst of 20 requests, refilling 60 per minute.
+/// wallet.set_rate_limit(RateLimit::Custom {
+///     capacity: 20,
+///     refill_per_minute: 60,
+/// })?;
+/// ```
+#[derive(Debug, Clone, uniffi::Enum)]
+pub enum RateLimit {
+    /// Built-in default pacing (capacity 10, refill 45/min).
```
thesimplekid: Do you know if this rate limit is inline with what coco does?

crates/cdk/src/wallet/builder.rs
```diff
@@ -253,6 +254,57 @@ impl WalletBuilder {
...
+            None => match shared_transport {
+                Some(transport) => Arc::new(RateLimitedHttpClient::with_shared_transport(
+                    mint_url.clone(),
+                    transport,
+                    auth_wallet.clone(),
+                )) as Arc<dyn MintConnector + Send + Sync>,
+                None => Arc::new(HttpClient::new(mint_url.clone(), auth_wallet.clone()))
+                    as Arc<dyn MintConnector + Send + Sync>,
+            },
+        };
```
thesimplekid: I think using the tor client or proxy will bypass the rate limit config, those should use it as well if it is configured.

crates/cdk-common/src/rate_limit.rs
```diff
@@ -0,0 +1,1050 @@
...
+    /// the store in the background. Runs exactly once via the `OnceCell`.
+    async fn ensure_started(&self) {
+        self.inner
+            .started
+            .get_or_init(|| async {
+                let store = self.inner.persistence.clone()?;
+
+                // Read the first: seed the cache from the persisted value.
+                let loaded = match store.load().await {
+                    Some(bytes) => <[u8; 8]>::try_from(bytes.as_slice())
```
thesimplekid: Should this have a bounded time out and use a fresh bucket if it times out? Other wise it will block requests, probably only an issue with a slow db like a remote supabase

Though I guess in the case of a slow db we'll probably just have issue later in the cycle so maybe its not too much of an issue.

crates/cdk-common/src/rate_limit.rs
```diff
@@ -0,0 +1,1050 @@
...
+    /// requests in memory but persists nothing.
+    pub fn for_mint(
+        config: RateLimitConfig,
+        mint_url: &MintUrl,
+        db: Arc<dyn WalletDatabase<database::Error> + Send + Sync>,
+    ) -> Self {
+        let persistence = kv_key_for(mint_url)
+            .map(|key| Arc::new(KvBudgetStore { db, key }) as Arc<dyn BudgetStore>);
+        Self::build(config, persistence)
+    }
```
thesimplekid: For each wallet this gives a separate bucket however mint A Unit A and Mint A Unit B should really share a bucket and rate limit as its the same mint, but currently they do not.

crates/cdk-common/Cargo.toml
```diff
@@ -51,14 +54,15 @@ tonic = { workspace = true, optional = true }
...
+tokio = { version = "1", default-features = false, features = ["rt", "rt-multi-thread", "macros", "test-util", "sync", "time"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 uuid = { workspace = true, features = ["js"], optional = true }
 getrandom = { version = "0.2", features = ["js"] }
-tokio = { workspace = true, default-features = false }
+tokio = { workspace = true, default-features = false, features = ["sync"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
+gloo-timers = { version = "0.3", features = ["futures"] }
```
thesimplekid: If we really need to add a dep we should at least disable default features.

crates/cdk/src/wallet/builder.rs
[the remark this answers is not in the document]
thesimplekid: I think it is a bit weird we change this in this PR. Im not really sure why we have both set auth wallet and set cat maybe it should be just one fn and accept an enum or remove one.

thesimplekid (changes requested):
I think adding a db write per http request might be a bit too aggressive especially on remote DB's. This cost is also not really used as we only read it on start up and it is not used for coordination between wallets, so I think we could write less often and flush it to the db on some longer interval then every http call and on wallet shutdown.

crates/cdk-ffi/src/wallet.rs
crodas: I will make sure they are in sync

crates/cdk-common/Cargo.toml
crodas: It is needed for wasm, but it is quite used https://crates.io/crates/gloo-timers

crodas: and it is already used in cdk, I'll promote it to a global dependency

crodas:
> I think it is a bit weird we change this in this PR. Im not really sure why we have both set auth wallet and set cat maybe it should be just one fn and accept an enum or remove one.

Because the auth wallet and the wallet should share the same HTTP client, in `main` its transport is created right there. It seems it would have been a problem with Tor as well; for instance, it would have been [completely ignored](https://github.com/crodas/cdk/blob/main/crates/cdk/src/wallet/auth/auth_wallet.rs#L65). I will address it in a separate PR.

So I delayed creating the client, but as I said, I will address this in a separate PR.

crates/cdk-common/src/rate_limit.rs
crodas: It gives a bucket per Mint URL. I guess the server limits would be enforced before decoding the payload; therefore, the unit would not be taken into account to separate limits. Am I missing something?

crates/cdk-ffi/src/wallet.rs
crodas: They are in sync now 32cdf2921b4debfa000d855d65a4202015c17b22

crates/cdk/src/wallet/builder.rs
crodas: Not exactly. If the user passes a RatedLimit transport, it will be used ([Tor](https://github.com/cashubtc/cdk/pull/2251/changes#diff-6ad6cff5510c47e445c55e8a2b74a22493446db21903d7be6fe2fec250367738R44-R51) or [NoTor](https://github.com/cashubtc/cdk/pull/2251/changes#diff-6ad6cff5510c47e445c55e8a2b74a22493446db21903d7be6fe2fec250367738R34-R36)), that happens in line 296. If `shared_transport` is set, it would be rated as limited.

crates/cdk-common/src/rate_limit.rs
crodas: Yeah, I will give it a fair timeout and fall back to the default on failure. It is a best-effort thing.

crodas: 326090af56b3f96283f1e97ca0ff3e6748e88968

crates/cdk/src/wallet/builder.rs
crodas: I added a follow-up ticket #2262

crodas:
> I think adding a db write per http request might be a bit too aggressive especially on remote DB's. This cost is also not really used as we only read it on start up and it is not used for coordination between wallets, so I think we could write less often and flush it to the db on some longer interval then every http call and on wallet shutdown.

That statement is not strictly true. We are persisting to the database, true, but that's not part of the [hot path](https://github.com/cashubtc/cdk/pull/2251/changes#diff-14b4b696d8d4583dd6889c4dad85ed141ff41ee162b81256782685f07d1b842dR371-R375), instead it happens in its [own process](https://github.com/cashubtc/cdk/pull/2251/changes#diff-14b4b696d8d4583dd6889c4dad85ed141ff41ee162b81256782685f07d1b842dR359), and it would not be concurrent and it would write the latest number.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-common/src/rate_limit.rs
```diff
@@ -0,0 +1,1071 @@
...
+    let _ = progress.send(millis);
+}
+
+/// Upper bound on how long [`TokenBucket::flush`] waits before giving up, so a
+/// store call that never returns cannot hang graceful shutdown.
+const FLUSH_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Upper bound on the one-time budget load at first use, so a hung store cannot
+/// stall the first request. On timeout the bucket starts full.
+const LOAD_TIMEOUT: Duration = Duration::from_millis(10);
```
asmogo: This LOAD_TIMEOUT may be too aggressive for postgres / supabase.

crates/cdk-common/src/rate_limit.rs
```diff
@@ -0,0 +1,1071 @@
...
+/// make two distinct hosts collide onto one key. That is safe here: a collision
+/// only makes two mints share (and so more conservatively enforce) one budget,
+/// it never leaks budget between them.
+fn kv_key_for(mint_url: &MintUrl) -> Option<String> {
+    let parsed = Url::parse(&mint_url.to_string()).ok()?;
+    let host = parsed.host_str()?;
+    let authority = match parsed.port() {
+        Some(port) => format!("{host}:{port}"),
+        None => host.to_string(),
+    };
```
asmogo: I think we should not check for the port and only use the hostname. The comments for this  function are confusing and don't match the implementation.

crates/cdk-common/src/rate_limit.rs
```diff
@@ -0,0 +1,1071 @@
...
+        .map(|d| d.as_millis() as u64)
+        .unwrap_or(0)
+}
+
+/// Derive a KV key from a mint URL's host (and port), sanitized to the KV
+/// alphabet. Returns `None` when the URL has no host.
+///
+/// The key is host (plus explicit port); scheme and default ports are ignored,
+/// so `http`/`https` on the same host, or `:443` vs an implicit default, share
+/// one budget. That is intended: one mint host is one budget.
```
asmogo:
```suggestion
/// The key is host; one hostname is one budget.
```

asmogo (changes requested):
Great PR. We need to be sure that the db writer is async (and maybe buffered with flush). If this gets too heavy for a DB, we could switch to redis / valkey. 

Some minor change requests here.

crodas:
> Great PR. We need to be sure that the db writer is async (and maybe buffered with flush). If this gets too heavy for a DB, we could switch to redis / valkey.

We don't have redis/ valkey in the wallet

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-common/src/rate_limit.rs
thesimplekid: It looks like the persistence key is the mint url but the live bucket is the mint url unit pair. So a wallet for mint A with unit Sat and a wallet for mint A with unit B will not share the bucket when they should because the rate limit is likely enforced before decoding the payload as you say.

thesimplekid (approved):
I think there is some follow up to do. But I will open issues for that

thesimplekid:
This was merged. GH just marked its closed

PR #2141, opened 2026-06-23
feat(cdk-mintd): move remote signatory config into its own section

prusnak:
Replace the old [info] signatory URL/certs fields with a dedicated [signatory] section using addr, port, tls_dir, and allow_insecure - copying the config layout for mint_management_rpc / grpc_processor.

Require an explicit allow_insecure=true when connecting to a remote signatory without TLS, and update the signatory client constructor to build its endpoint from addr/port. Update env vars, example config, and integration fixtures to use the new shape.

Add README.md to cdk-signatory.

PR #2253, opened 2026-07-23
Auto-rotate keysets on an age interval

crodas:
### Description

This PR is heavily inspired by https://github.com/cashubtc/nutshell/pull/1058 and depends #2270

Mints need to move to fresh signing keys periodically without an operator rotating them by hand. The signatory owns the private keys and each keyset's valid_from timestamp, and that timestamp does not cross the Signatory trait, so the rotation decision belongs in the signatory rather than the mint. Rotated keysets already reach mints through the keyset subscription, so no restart or mint-initiated rotate is needed.

Add a background task to the embedded signatory that rotates every active keyset once it has been valid for at least a configured interval. The replacement keeps the previous amounts, input fee and id version, and carries final_expiry forward by the keyset's active age (saturating) so it stays valid at least as long as the keyset it replaces. The task holds a weak reference to the signatory and skips the immediate first tick, so it stops on its own once the signatory is dropped and never rotates a freshly created keyset before it has aged.

Auto-rotation runs directly against the signatory's lock-guarded state rather than through the embedded actor, because the age check needs valid_from. A per-unit recheck skips a keyset whose active id changed since it was snapshotted, so a mint-initiated rotation racing the task does not issue a redundant keyset.

Bind the task to shutdown. The embedded service owns the task handle and aborts it on drop, and the mint stores an abort handle and aborts it in stop(), before the early return so rotation halts even when no other services were started. The task is spawned once at build time because it needs the concrete embedded signatory, so a later start() does not bring it back; this is documented on the builder setter and on stop().

Expose the interval where each signatory is configured: a flag on the standalone signatory binary, a builder method for the embedded signatory, and a mintd config key with a matching environment variable. Auto-rotation is disabled unless an interval is set, and a remote signatory manages its own schedule.

Cover the behavior with tests for the age threshold, metadata carry-over across both keyset versions, the final_expiry push and its absence, multi-unit rotation, the no-op and skip-already-rotated cases, task teardown when the signatory is dropped, and that stop() halts rotation while a later start() does not resume it. Document the interval in the mintd README and example config.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-mintd/src/config.rs
```diff
@@ -59,6 +59,12 @@ pub struct Info {
     /// Use keyset v2
     pub use_keyset_v2: Option<bool>,
 
+    /// Automatically rotate active keysets once they reach this age, in seconds.
+    ///
+    /// Only applies when the mint runs an embedded signatory. `None` or `0`
+    /// disables auto-rotation. A remote signatory manages its own schedule.
+    pub keyset_rotation_interval_seconds: Option<u64>,
+
```
thesimplekid: Should this be in the signatory section then? https://github.com/cashubtc/cdk/pull/2141. Configuring a remote and embedded signatory should be the same.

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -127,6 +155,140 @@ impl DbSignatory {
...
+                }
+                _ = ticker.tick() => {}
+            }
+
+            let Some(signatory) = weak.upgrade() else {
+                break;
+            };
+
+            if let Err(err) = signatory.rotate_aged_keysets(interval).await {
+                tracing::error!("Automatic keyset rotation failed: {}", err);
```
thesimplekid: I think we probably want to retry before the next tick? For example I think here if it fails and its a long tick like a monthly rotation it will not retry until the next tick so another month?

crates/cdk/src/mint/builder.rs
```diff
@@ -134,6 +145,23 @@ impl MintBuilder {
...
+    /// Only applies to the embedded signatory built through
+    /// [`MintBuilder::build_with_seed`]. A `None` value, or an interval of zero,
+    /// leaves auto-rotation disabled. A remote signatory manages its own
+    /// rotation schedule.
+    ///
+    /// The rotation task is spawned once, at build time, because it needs the
+    /// concrete embedded signatory (only reachable here, not through the
+    /// `Signatory` trait the mint holds). [`Mint::stop`] halts it, but a later
+    /// [`Mint::start`] does not respawn it; rebuild the mint to re-enable
+    /// rotation. This differs from the drain task, which resumes on restart.
```
thesimplekid: I think this could be problematic, for our use in mintd I think its okay, but if cdk is used as a lib I would expect the start to restart the rotation logic.

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -127,6 +155,140 @@ impl DbSignatory {
...
+    /// This runs directly against the signatory's own `RwLock`-guarded state
+    /// rather than through the embedded actor, because the age check needs each
+    /// keyset's `valid_from`, which does not cross the `Signatory` trait. The
+    /// locks make concurrent signing requests safe; the only real contention is
+    /// a mint-initiated `rotate_keyset` for the same unit, which the per-unit
+    /// recheck below guards against.
+    async fn rotate_aged_keysets(&self, max_age: Duration) -> Result<(), Error> {
+        let now = unix_time();
+        let max_age = max_age.as_secs();
+
```
thesimplekid: We should check this is not zero? The api accepts Duration that can be sub second and we would later panic if this is Zero. I see no reason to support sub second roations so just returning an error on zero to be safe.

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -127,6 +155,140 @@ impl DbSignatory {
...
+        };
+
+        for info in due {
+            // Another rotation may have landed between snapshotting `due` and
+            // reaching this unit (for example a mint-initiated rotate). If the
+            // unit's active keyset is no longer the one judged due, it was
+            // already advanced; skip to avoid issuing a redundant keyset. A tiny
+            // race remains before `rotate_keyset` re-reads the active id, but its
+            // worst case is one extra keyset, never inconsistent state.
+            let still_active = self.active_keysets.read().await.get(&info.unit).copied();
```
thesimplekid: I think there is a TOCTOU race condition here we check it is still active but there is still time between the check and the commit and if a manual rotation is triggered there this check could be stale. I think we should ensure the manual and auto rotations share the same path so they are forced to be serialized.

thesimplekid (changes requested):

crates/cdk-mintd/src/config.rs
crodas: True, I'll make this happen

crates/cdk-signatory/src/db_signatory.rs
crodas: I have a local version where the ticker is fixed, like every 30s. So it would not be deadly to wait for the next one.

thesimplekid: But I don't think 30s is likely in practice is more likely to be a month or something so waiting a full extra month is a big difference

crodas: I think that's fine; `rotate_aged_keysets` should be cheap enough to run and exit.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-signatory/src/db_signatory.rs
thesimplekid: We need to check that the `as_secs()` is zero not that the `Duration` is non zero to address the sub second issue.

crodas: [I think I already did that](https://github.com/cashubtc/cdk/pull/2253/changes#diff-829214cd88f152e8258dee4bfd490e0baaf3cb027ede39e00746ac2c71401327R187-R189)

crates/cdk-signatory/src/db_signatory.rs
thesimplekid, line 715: This PR does not handle custom derivation path rotation on every tick the same keys are generated

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -127,6 +166,164 @@ impl DbSignatory {
...
+                if i > 0 {
+                    self.reload_keys_from_db().await?;
+                }
+                return Err(err);
+            }
+        }
+
+        // Reload once for the whole pass, after every rotation has been
+        // committed, so subscribers see a single updated snapshot.
+        self.reload_keys_from_db().await?;
```
thesimplekid: Should we track a local needs_reload state here? At this point one or more rotations may already be committed. If this reload fails, the in-memory active keysets remain stale, and the next tick can rotate the newer database keyset again.

if we replace `rotation_lock: Mutex<()>` with a mutex-protected rotation state containing needs_reload. Set it before committing, clear it only after a successful reload, and require any subsequent rotation to reload first while the state is dirty.

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -211,6 +408,34 @@ impl Signatory for DbSignatory {
...
+    /// `rotation_lock` and reloads once at the end; keeping this helper
+    /// lock-free lets that batch reuse a single snapshot instead of re-reading
+    /// (and re-locking) `keysets` per rotation.
+    async fn commit_rotated_keyset(
+        &self,
+        args: RotateKeyArguments,
+        current_keysets: &[SignatoryKeySet],
+    ) -> Result<(MintKeySetInfo, MintKeySet), Error> {
         let (path_index, amounts) = if let Some(current_keyset_id) =
             self.localstore.get_active_keyset_id(&args.unit).await?
```
thesimplekid: Should this active-keyset transition use a database-level compare-and-set instead of relying only on rotation_lock? The mutex protects one DbSignatory, but two instances sharing a key store can both act on the same stale keyset.

thesimplekid:
On the commit description for https://github.com/cashubtc/cdk/pull/2253/commits/6ff93abf08644e581ae18423f5bdd82fc5f584bd I think we should default to rotating keysets, the linked nutshell pr rotates every 90 days for example.

crates/cdk-signatory/src/db_signatory.rs
crodas: We don't support multiple instances of the signatory through a database. Should we support it? 

We support multiple mints (after #2263), but all of them are connected to a single external signatory.

thesimplekid:
I think everything this depends on is merge and we can pick it back up for the 0.19 release now that 0.18 is merged?

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-mintd/src/lib.rs
```diff
@@ -1204,8 +1205,25 @@ fn configure_basic_info(settings: &config::Settings, mint_builder: MintBuilder)
...
+
     builder
 }
+
+/// Embedded signatory keyset auto-rotation interval, or `None` when rotation is
+/// disabled.
+///
+/// A missing `[signatory]` section falls back to the default, so an embedded
+/// mint auto-rotates without explicit config; `0` is the documented off switch.
+fn keyset_rotation_interval(settings: &config::Settings) -> Option<Duration> {
```
thesimplekid, line 1203: Should `prune_inactive_configuration()` preserve the new embedded rotation setting? It currently removes the entire `[signatory]` section when `enabled = false`, which is exactly when this setting applies. Startup therefore replaces a configured `0` or custom interval with the 90-day default. Could we update the pruning and add a regression test through startup configuration validation? The current helper tests bypass that path.

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -334,6 +340,214 @@ impl DbSignatory {
...
+                )
+                .await
+                .map(|_| ());
+
+            if staged.is_err() {
+                break;
+            }
+        }
+
+        self.finish_rotation(tx, staged).await
```
thesimplekid: Non-blocking: Could we log successful completion of the automatic rotation sweep after the transaction commits and the keysets are refreshed? The current “Auto-rotating keyset…” message is emitted before staging finishes, so the sweep can subsequently roll back. A completion message would let operators distinguish attempted rotations from successful ones.

crates/cdk-signatory/src/db_signatory.rs
thesimplekid: Confirmed this still occurs at 9eceed0839edea591381d339e9cce3f1bda4d97c with a focused test: configure a fixed custom derivation path, create a Version01 keyset with no final expiry, age it beyond the rotation interval, and run the automatic sweep. The keyset ID remains unchanged. The database upsert resets `valid_from`, so unchanged keys appear freshly rotated indefinitely. Could we derive distinct paths for subsequent rotations or explicitly reject automatic rotation for fixed custom paths?

thesimplekid (changes requested):

crates/cdk-mintd/src/lib.rs
crodas: Good catch, and it was a real bug. Instead of special-casing pruning, the setting moved out of [signatory] entirely: it is now [info].keyset_rotation_interval_seconds (env CDK_MINTD_KEYSET_ROTATION_INTERVAL_SECONDS), a plain u64 with 0 as the off switch. A disabled [signatory] section is still pruned, which no longer affects rotation.
 

The regression test goes through startup rather than the helpers: 

keyset_rotation_interval_survives_startup_validation in config_service.rs writes a document with [signatory] enabled = false, runs initialize() and startup(), and asserts both that the section is pruned and that a configured 0 or 3600 reaches keyset_rotation_interval() intact.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

thesimplekid:
This was marked resolved in the [earlier discussion](https://github.com/cashubtc/cdk/pull/2253#discussion_r3962491726), but I think it is still an issue. I reproduced it again at `c8047f4179c5ae0a74bb1248e1ee3af0842b979c` with the regression test below.

The test below uses the existing `age_active_keyset` and `active_keyset` helpers in `crates/cdk-signatory/src/db_signatory.rs`'s `mod test`. It creates a V2 keyset with a fixed custom derivation path and no final expiry, backdates it by 120 seconds, and runs a sweep with a 60-second threshold. There are no sleeps or background-task timing dependencies.

Run after adding the test:

```sh
nix develop --command cargo test -p cdk-signatory --lib auto_rotation_replaces_keys_with_custom_derivation_path --locked -- --nocapture
```

Actual result:

```text
assertion `left != right` failed: auto-rotation must create a distinct keyset, not reset the age of the same keys
  left: 0110d3bf0b2bcccabb32410a0c444c4c2c03ebd24456fb17e225b2b47c47c3115b
 right: 0110d3bf0b2bcccabb32410a0c444c4c2c03ebd24456fb17e225b2b47c47c3115b
```

`stage_rotation` still uses `Some(path) => path.clone()`, so the newly allocated derivation index does not change the actual custom path. The same keyset is upserted with a fresh `valid_from`. `check_unit_collision` allows an existing currency unit and does not check that the replacement has different keys.

Could we either make custom-path rotation produce distinct keys and pass this test, or explicitly reject automatic rotation with fixed custom paths and test that rejection? Either outcome should prevent unchanged keys from being reported as successfully rotated.

<details>
<summary>Regression test — paste inside the existing <code>mod test</code></summary>

```rust
#[tokio::test]
async fn auto_rotation_replaces_keys_with_custom_derivation_path() {
    let store = Arc::new(cdk_sqlite::mint::memory::empty().await.unwrap());
    let path: DerivationPath = "m/129372'/0'/1'".parse().unwrap();
    let sig = Arc::new(
        DbSignatory::new(
            store,
            b"test-seed-custom-auto-rotation",
            HashMap::new(),
            HashMap::from([(CurrencyUnit::Sat, path)]),
        )
        .await
        .unwrap(),
    );
    let original = sig
        .rotate_keyset(RotateKeyArguments {
            unit: CurrencyUnit::Sat,
            amounts: vec![1, 2, 4, 8],
            input_fee_ppk: 0,
            keyset_id_type: cdk_common::nut02::KeySetVersion::Version01,
            final_expiry: None,
        })
        .await
        .unwrap();

    age_active_keyset(&sig, &CurrencyUnit::Sat, 120).await;
    sig.rotate_aged_keysets(Duration::from_secs(60)).await;

    let replacement = active_keyset(&sig, &CurrencyUnit::Sat).await;
    assert_ne!(
        replacement.id, original.id,
        "auto-rotation must create a distinct keyset, not reset the age of the same keys"
    );
    let snapshot = sig.keysets().await.unwrap();
    assert!(
        snapshot.keysets.iter().any(|k| k.id == original.id && !k.active),
        "the original keyset must remain available as inactive"
    );
}
```

</details>

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-mintd/src/config.rs
```diff
@@ -61,6 +61,15 @@ pub struct Info {
...
     pub use_keyset_v2: Option<bool>,
 
+    /// Automatically rotate active keysets once they reach this age, in
+    /// seconds.
+    ///
+    /// Defaults to `0`, which keeps auto-rotation off so an upgrade never
+    /// changes an existing mint's keysets on its own. Applies to the embedded
+    /// signatory the mint runs for itself; a remote `[signatory]` manages its
+    /// own rotation schedule and ignores this value.
+    pub keyset_rotation_interval_seconds: u64,
```
orangeshyguy21, line 71: This accepts any u64 with no floor. Footgun in theory.

Is it worth warning on intervals below that theoretical ~floor~ or being more opinionated with some enforced minimum?

orangeshyguy21 (commented):
Looks good.

This sharpens the gap in CDK on a sanctioned route for unit retire. 
Only really possible today through a manual database edit, but an improvement for that is outside scope.

I left comment on a small nit related to `keyset_rotation_interval_seconds`

thesimplekid (commented):
One correctness concern in the build-time rotation deduplication.

crates/cdk/src/mint/builder.rs
```diff
@@ -60,6 +62,199 @@ pub struct KeysetRotation {
...
+}
+
+impl From<&RotateKeyArguments> for RotationShape {
+    fn from(args: &RotateKeyArguments) -> Self {
+        Self {
+            unit: args.unit.clone(),
+            amounts: args.amounts.clone(),
+            input_fee_ppk: args.input_fee_ppk,
+            use_keyset_v2: args.keyset_id_type == KeySetVersion::Version01,
+            expired: args.final_expiry.map(|expiry| expiry < unix_time()),
```
thesimplekid, line 129: Should future `final_expiry` values be compared as exact timestamps rather than reduced to `Some(false)`? For example, if the active keyset expires tomorrow and `with_keyset_rotation` requests an otherwise identical keyset expiring next month, both produce the same `RotationShape`. `plan_rotations` then considers the requested rotation already applied and skips it, leaving tomorrow's expiry in effect. This also occurs with automatic rotation disabled.

Could we retain the exact timestamp for future expiries, limiting any normalization to the expired test fixtures that need it? A regression test should verify that changing a future expiry creates the requested replacement, while rebuilding with the same expiry remains idempotent.

PR #2254, opened 2026-07-24
Extract a reusable stream supervisor for long-lived subscriptions

crodas:
This is a follow-up for #2225

Two consumers of long-lived streams, the signatory keyset subscription and the mint payment-event stream, each carried their own reconnect, backoff, and shutdown loop. The copies had already drifted: one used capped exponential backoff, the other a flat sleep; one keyed shutdown off a dropped watch receiver, the other off a Notify token. That duplication is where the transport supervision should be factored out from the per-payload decoding.

Add cdk_common::stream::supervise_stream, which owns only the reconnect, backoff, and shutdown machinery and is transport-agnostic: the caller supplies a connect closure that opens a fresh stream and an async handler run for each item. It selects the shutdown signal against the connect step, the pending item, and the backoff sleep, so a dead consumer stops the loop promptly instead of lingering on an open stream.  The backoff is capped exponential and resets once a message is delivered, so a healthy connection that later drops reconnects quickly while a failing endpoint is not hammered. BackoffPolicy is clamped defensively (initial forced non-zero, max raised to at least initial) so a degenerate policy cannot busy-loop or invert the cap in any build.

Migrate both callers onto it. The signatory adapter supplies the policy, uses the watch sender's closed() as the shutdown signal, opens the subscription, and republishes each decoded snapshot into the watch. The mint payment loop supplies its Notify-based shutdown, wraps its already-decoded event stream, dispatches each payment event, and runs its stream teardown on the line after the supervisor returns.

One behavior change falls out of the mint migration: the payment loop now waits a short initial delay before every reconnect, including a clean close, where it previously reconnected immediately. The floor stops a backend that accepts, delivers, and instantly drops from spinning in a hot reconnect loop, and it is kept short so payment-detection latency stays low.

Enable tokio's time feature on cdk-common, since the supervisor sleeps between reconnects, and cover the supervisor with tests: shutdown at each await point, item forwarding across reconnects, mid-stream error handling, backoff growth and reset on delivery, and policy clamping.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/stream.rs
+++ b/crates/cdk-common/src/stream.rs
@@ -1,0 +1,452 @@
+//! Supervision for long-lived streaming subscriptions.
+//!
+//! [`SupervisedStream`] owns the reconnect, backoff, and shutdown loop that
+//! every consumer of a long-lived stream (a gRPC server-stream, a payment
+//! backend's event stream) would otherwise hand-roll. It is transport-agnostic:
+//! the implementor supplies how to open a fresh stream and how to handle each
+//! item, holding its connection state as fields rather than cloning it into
+//! per-call closures.
+
+use std::fmt;
+use std::future::Future;
+
+use futures::{pin_mut, Stream, StreamExt};
+
+/// Capped exponential backoff for [`SupervisedStream`] reconnect attempts.
+///
+/// Only consecutive connect failures grow the delay; a successful connect resets
+/// it. So an endpoint that keeps refusing is backed off exponentially, while one
+/// that recovers is not held to a delay earned by earlier failures.
+#[derive(Debug, Clone, Copy)]
+pub struct BackoffPolicy {
+    /// Delay before the first reconnect, and the floor a successful connect
+    /// resets the growing backoff to. Must be non-zero, else the doubling
+    /// (`0 * 2 == 0`) never grows and reconnects busy-loop.
+    pub initial_connect_backoff: std::time::Duration,
+    /// Cap on the delay while backing off. Must be at least `initial`.
+    pub max_connect_backoff: std::time::Duration,
+}
+
+/// A supervised, self-reconnecting streaming subscription.
+///
+/// Implementors own the connection state (clients, channels, publishers) as
+/// fields, so the reconnect/backoff/shutdown loop can hand each item to
+/// [`on_message`](Self::on_message) without the per-item state-cloning a
+/// closure-based supervisor forces. The provided [`supervise`](Self::supervise)
+/// method owns that loop; an implementor supplies only how to connect, how to
+/// handle an item, and (optionally) how to tear down.
+#[async_trait::async_trait]
+pub trait SupervisedStream: Send {
+    /// Item the stream yields and [`on_message`](Self::on_message) consumes.
+    type Item: Send;
+    /// Error a failed connect attempt yields. Logged, then retried.
+    type ConnectError: fmt::Display + Send;
+    /// Error the stream may yield per item. Logged, then reconnected.
+    type StreamError: fmt::Display + Send;
+    /// The stream a successful [`connect`](Self::connect) opens.
+    type Stream: Stream<Item = Result<Self::Item, Self::StreamError>> + Send;
+
+    /// Names this subscription in the supervisor's reconnect/close logs.
+    fn name(&self) -> &str;
+
+    /// Backoff policy for reconnect attempts.
+    fn backoff_policy(&self) -> BackoffPolicy;
+
+    /// Open a fresh stream of items.
+    async fn connect(&mut self) -> Result<Self::Stream, Self::ConnectError>;
+
+    /// Handle one delivered item. Awaited to completion, so a slow handler
+    /// holds the read loop; offload work that must not block it.
+    async fn on_message(&mut self, item: Self::Item);
+
+    /// Teardown run once before [`supervise`](Self::supervise) returns, on every
+    /// exit path. Cancel tokens or release resources here.
+    async fn on_shutdown(&mut self) {}
+
+    /// Keep the subscription alive across reconnects until `shutdown` resolves.
+    ///
+    /// Every item [`connect`](Self::connect) yields is handed to
+    /// [`on_message`](Self::on_message). Reconnect timing follows
+    /// [`BackoffPolicy`]: an opened stream that later closes or errors reconnects
+    /// at the floor, since the connection itself was healthy.
+    ///
+    /// `shutdown` stops the supervisor promptly whenever it is waiting: to
+    /// connect, for the next item, or during a backoff. It does not interrupt an
+    /// in-flight `on_message`. [`on_shutdown`](Self::on_shutdown) runs on every
+    /// exit path.
+    async fn supervise<S>(&mut self, shutdown: S)
+    where
+        S: Future<Output = ()> + Send,
+    {
+        // Clamp a degenerate policy rather than trusting the implementor: a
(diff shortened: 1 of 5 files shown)


crodas:
@cdk-bot review

crates/cdk-common/src/stream.rs
```diff
@@ -0,0 +1,413 @@
...
+) where
+    C: FnMut() -> CFut,
+    CFut: std::future::Future<Output = Result<St, E1>>,
+    St: Stream<Item = Result<T, E2>>,
+    E1: fmt::Display,
+    E2: fmt::Display,
+    F: FnMut(T) -> Fut,
+    Fut: std::future::Future<Output = ()>,
+    S: std::future::Future<Output = ()>,
+{
```
thesimplekid: We should simplify this I do not think this is needed. Maybe its trying to be too generic or maybe over optimizing avoiding an allocation? One allocation per long-lived connection seems preferable to carrying nine generic parameters in the public API.

crates/cdk-common/src/stream.rs
```diff
@@ -0,0 +1,413 @@
...
+                        _ = &mut shutdown => return,
+                        next = stream.next() => next,
+                    };
+
+                    match next {
+                        Some(Ok(item)) => {
+                            // A delivered message proves the connection is
+                            // productive, so a later drop reconnects quickly.
+                            backoff = initial;
+                            on_message(item).await;
```
thesimplekid: Could we simplify the retry behavior by increasing the backoff only when `connect()` fails? Payment-event streams can be healthy while producing no messages, so previous connection failures can leave the backoff at 30 seconds even after the connection has recovered. If that idle stream later disconnects, payment detection may remain unavailable for another 30 seconds.

What about resetting the backoff as soon as `connect()` succeeds, waiting the fixed initial delay after an opened stream closes or errors, and doubling only in the connection-error branch? This would also remove the per-message reset. A test covering failures → successful empty stream → disconnect would verify the behavior.

crates/cdk-common/src/stream.rs
```diff
@@ -0,0 +1,413 @@
...
+                        }
+                        Some(Err(err)) => {
+                            tracing::warn!("Stream error: {err}");
```
thesimplekid: It would be good if we knew what stream was reconnecting in the logs.

crates/cdk-common/src/stream.rs
```diff
@@ -0,0 +1,413 @@
...
+        }
+
+        // Wait before reconnecting so a persistently failing endpoint is not
+        // hammered. Shutdown during the wait ends the loop immediately.
+        tokio::select! {
+            biased;
+            _ = &mut shutdown => return,
+            _ = tokio::time::sleep(backoff) => {}
+        }
+        backoff = (backoff * 2).min(max);
```
thesimplekid: Could we use saturating mul here. I think we should remove all unchecked math from cdk.

crates/cdk-common/src/stream.rs
```diff
@@ -0,0 +1,413 @@
...
+    fn never_yields() -> Result<ItemStream, TestErr> {
+        Ok(stream_of(vec![]))
+    }
+
+    // `connect` closures increment their counters inside the returned future,
+    // not in the closure body. `tokio::select!` evaluates the `connect()` branch
+    // expression eagerly every loop iteration, even on the pass where `shutdown`
+    // wins, so only the poll of the future marks a real connection attempt.
+
+    #[tokio::test(start_paused = true)]
```
thesimplekid: Nice I didn't know you could do this.

thesimplekid (changes requested):

crodas:
@thesimplekid I added a [Trait](https://github.com/cashubtc/cdk/pull/2254/changes/5c22f3e0e95a9c807285d629b87887d70eb8758f#diff-f7a547063823c1c78e21bdb69def6d08eb52c710615f0d6180325484e2d81802R50) that would keep things cleaner and less intimidating. I think the usage is [much simpler](https://github.com/cashubtc/cdk/pull/2254/changes/5c22f3e0e95a9c807285d629b87887d70eb8758f#diff-f4f4a9e9c2b6872ed4dc3fc93b1a05ab0e543f9017e5074f6bc9d31ed6897658R162-R197) and [elegant](https://github.com/cashubtc/cdk/pull/2254/changes/5c22f3e0e95a9c807285d629b87887d70eb8758f#diff-f5ceb86f843038da7e04913cad4489e37052c36fb340f3f948ecb5e6fe17d7aeR125-R220).

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-common/src/stream.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    /// Handle one delivered item. Awaited to completion (item processing is
+    /// sequential, and a slow handler holds the loop, so a handler that must
+    /// not block the read should offload its work).
+    async fn on_message(&mut self, item: Self::Item);
+
+    /// Teardown run once, after [`supervise`](Self::supervise) observes shutdown
+    /// and before it returns. Default: nothing. Cancelling a stream token or
+    /// releasing a resource belongs here rather than after the `supervise` call.
+    async fn on_shutdown(&mut self) {}
```
thesimplekid: Should we remove this in favor of just using `Drop`? This will not run if the task containing supervise() is aborted—which the mint’s JoinSet::shutdown() can do.

crates/cdk-common/src/stream.rs
```diff
@@ -0,0 +1,479 @@
...
+/// the delays predictable.
+#[derive(Debug, Clone, Copy)]
+pub struct BackoffPolicy {
+    /// Delay before the first reconnect, the fixed delay after an opened stream
+    /// closes or errors, and the floor a successful connect resets the growing
+    /// backoff to. Must be non-zero.
+    pub initial: std::time::Duration,
+    /// Upper bound the delay is capped at while backing off across connect
+    /// failures. Must be at least `initial`.
+    pub max: std::time::Duration,
```
thesimplekid: I think we could rename these, initial_connect_backoff and max_connect_backoff.

crates/cdk-common/src/stream.rs
```diff
@@ -0,0 +1,479 @@
...
+/// closes or errors, the supervisor waits exactly `initial` before reconnecting:
+/// the connection was working, so the reconnect should be prompt, and this floor
+/// still stops a server that accepts then instantly drops from spinning in a hot
+/// loop. A successful connect resets the growth, so an endpoint that recovers is
+/// not held to a delay earned by earlier failures.
+///
+/// `initial` must be non-zero (a zero `initial` never grows, `0 * 2 == 0`, so
+/// the backoff would degenerate into a busy reconnect loop) and `max` must be at
+/// least `initial` (a smaller cap would clamp the delay below the floor).
+/// [`SupervisedStream::supervise`] also clamps both at runtime, so a violated
```
thesimplekid: I think these comments are too verbose and we need tighten them up. Comments should be why do we need this (ie the trait) not what every line does that should be clear from the code.

crates/cdk-common/src/stream.rs
crodas: 1. `Drop` is synchronous
2.  `on_shutdown` runs at a [defined point ](https://github.com/cashubtc/cdk/pull/2254/changes#diff-f7a547063823c1c78e21bdb69def6d08eb52c710615f0d6180325484e2d81802R175); `Drop` depends on when it goes out of scope, outside of `supervise`

thesimplekid (approved):

PR #2263, opened 2026-07-26
Distribute NUT-17 notifications across mint instances

crodas:
### Description

A `Bus` trait is inserted between `Pubsub::publish` and the in-memory subscriber
fan-out (`crates/cdk-common/src/pub_sub/bus.rs`). `publish` stays sync
fire-and-forget, so the ~15 `PubSubManager` publish helpers, the mint spec, and
the WS handlers are untouched. A distributed bus is opt-in through
`Pubsub::new_with_bus` / `PubSubManager::new_with_bus`; both local and peer
events flow through one `LocalDelivery` path, so remote events are
indistinguishable downstream. Correctness never depends on the transport: it is
a latency optimization for the live push, and the database backfill remains the
source of truth.

## Transports

| transport | where it runs | notes |
|---|---|---|
| `in-memory` (default) | single instance | in-process, no dependency, unchanged behavior |
| `sql` | SQLite and Postgres, including behind PgBouncer | polling outbox using plain `INSERT`/`SELECT`; reuses the mint's connection pool |
| `postgres-listen-notify` | directly-connected Postgres | `LISTEN`/`NOTIFY`; lowest latency; needs a session-pinned connection |

- **Postgres `LISTEN`/`NOTIFY`** (`crates/cdk-postgres/src/bus.rs`): a dedicated
  connection with a reconnect/backoff loop; JSON wire format; self-echo dropped
  by a per-instance origin id; events over the 8000-byte `NOTIFY` cap are
  delivered locally and skipped for peers.
- **SQL polling** (`crates/cdk-sql-common/src/mint/bus.rs`): appends each event
  to a `pubsub_outbox` table and polls `WHERE id > cursor AND origin <> self`.
  Works on any SQL backend and through transaction-pooling proxies where
  `LISTEN` cannot. No payload cap, and it resumes from its cursor after a missed
  poll or restart (within the retention window). New per-backend migration adds
  the outbox table.

## Configuration

Selected in `cdk-mintd` under `[database.pubsub]` (all fields also available as
`CDK_MINTD_PUBSUB_*` env vars):

```toml
[database.pubsub]
transport = "in-memory"        # in-memory | sql | postgres-listen-notify
# channel = "cdk_mint_pubsub"  # postgres-listen-notify only
# poll_interval_ms = 200       # sql only
# retention_seconds = 3600     # sql only
```

cdk-mintd installs the chosen bus at startup (reusing the mint DB pool for the
SQL transport). Selecting postgres-listen-notify on a non-Postgres engine is a
startup error.

### Testing

- A shared cross-instance bus suite lives in cdk-sql-common behind a test
feature and is run from both cdk-sqlite and cdk-postgres (mirroring the
existing mint_db_test! pattern), covering cross-instance delivery, origin
self-filtering, and the cursor-starts-at-now guarantee.
- The generic local-delivery suite (bus_test!) runs against every bus.
- The Postgres LISTEN/NOTIFY bus has its own cross-instance and unit tests.
- Verified locally on SQLite and against a live Postgres.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

thesimplekid:
@cdk-bot you there buddy?

thesimplekid:
@crodas I think cdk-bot error-ed out trying to review this on the first tag and then won't re review and things it already did, I'll try and harden this bad state but since this needs to be rebased anyway can you just do that and the new commit should force a review on the next tag.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-postgres/src/bus.rs
```diff
@@ -0,0 +1,797 @@
...
+    /// Undecodable payload; ignore it.
+    Malformed,
+}
+
+/// Decode an inbound payload and decide what to do with it.
+///
+/// Pure, so the self-echo and malformed-payload handling are unit-testable
+/// without a live connection.
+fn classify<E: DeserializeOwned>(payload: &str, our_origin: &str) -> Inbound<E> {
+    match serde_json::from_str::<InEnvelope<E>>(payload) {
```
thesimplekid, line 96: This fails for the actual `MintEvent` type: it wraps `NotificationPayload`, whose deserialization requires subscription-kind context. I reproduced this with a proof-state event: serialization succeeds, but deserialization returns “notification payloads require subscription kind context.”

Consequently, every incoming mint event is classified as malformed. The SQL bus has the same issue. Could we preserve the event kind in the internal wire format and add cross-instance tests using real mint events? The current synthetic `Message` tests don’t catch this.

crates/cdk-sql-common/src/mint/bus.rs
```diff
@@ -0,0 +1,456 @@
...
+    cursor: i64,
+) -> Result<i64, Error>
+where
+    S: Spec + 'static,
+    RM: DatabasePool + 'static,
+{
+    let conn = pool.get().await.map_err(|e| Error::Database(Box::new(e)))?;
+    let rows = query(
+        "SELECT id, payload FROM pubsub_outbox \
+         WHERE id > :cursor AND origin <> :origin \
```
thesimplekid: Can we make cursor advancement safe for concurrent commits? If one insert allocates ID 1 but another commits ID 2 first, a poll can advance to 2 before ID 1 becomes visible. ID 1 is then permanently excluded from subsequent polls. I reproduced this against PostgreSQL.

The module documents this limitation, but backfill only runs on a new subscription. An existing WebSocket subscriber can remain connected and never receive its payment update. Could we account for late commits and add a regression test for this ordering?

crates/cdk-mintd/src/env_vars/database.rs
```diff
@@ -45,6 +51,34 @@ impl PostgresConfig {
     }
 }
 
+impl PubSubConfig {
+    pub fn from_env(mut self) -> Self {
+        if let Ok(transport) = env::var(ENV_PUBSUB_TRANSPORT) {
+            if let Ok(parsed) = PubSubTransport::from_str(&transport) {
```
thesimplekid: Could invalid values return a configuration error identifying the environment variable? A typo in `CDK_MINTD_PUBSUB_TRANSPORT` currently leaves the default in-memory transport enabled, so a multi-instance deployment starts successfully without distributing notifications. Invalid numeric values below are also silently ignored instead of reaching the timing validation.

thesimplekid (changes requested):
Do we need both transports in this PR? Polling adds quite a bit of code and complexity around the outbox, cursors, retention, and pruning. Could we start with LISTEN/NOTIFY and add polling later if there’s a concrete need for transaction-pooling or multi-process SQLite support?

crodas:
> Do we need both transports in this PR? Polling adds quite a bit of code and complexity around the outbox, cursors, retention, and pruning. Could we start with LISTEN/NOTIFY and add polling later if there’s a concrete need for transaction-pooling or multi-process SQLite support?

I can split this into 2 PR. The trait + LISTEN/Notify; another PR for the SQL as you suggested.

thesimplekid:
> another PR for the SQL as you suggested.

Do we need the second at all?

crodas:
> > another PR for the SQL as you suggested.
> 
> Do we need the second at all?

Yes, otherwise multi-mint instances with SQLite will miss the other nodes' notifications.

thesimplekid:
> Yes, otherwise multi-mint instances with SQLite will miss the other nodes' notifications.

Yes, I know but is this something we even want to support? If you want multi instance use psgl?

crodas:
I like that; should we document that instead? That would trim a lot of code.

crates/cdk-postgres/src/bus.rs
crodas: Addressed in 3cffc07b25c5767d4354b95e1faaf40ca71a976c

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-postgres/src/bus.rs
```diff
@@ -0,0 +1,971 @@
...
+    /// Undecodable payload; ignore it. Carries the decode error so the dispatcher
+    /// can say why a peer's event was dropped.
+    Malformed(serde_json::Error),
+}
+
+/// Decode an inbound payload and decide what to do with it.
+///
+/// Pure, so the self-echo and malformed-payload handling are unit-testable
+/// without a live connection.
+fn classify<E: DeserializeOwned>(payload: &str, our_origin: &str) -> Inbound<E> {
```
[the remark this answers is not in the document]
crodas: I think this is not a real threat, if anyone has access to postgres a lot more damage can be done other than some notification

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-postgres/src/bus.rs
```diff
@@ -0,0 +1,985 @@
...
+{
+    local: LocalDelivery<S>,
+    client: SharedClient,
+    channel: Arc<str>,
+    origin: Arc<str>,
+    /// Dropped when the bus is dropped, which stops the background tasks.
+    _shutdown: watch::Sender<()>,
+}
+
+impl<S> Bus<S> for PostgresBus<S>
```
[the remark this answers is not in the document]
crodas: This will be addressed later, when #2478 gets implemented, and LISTEN/NOTIFY could stream the event id.

crodas:
@cdk-bot review

crates/cdk-postgres/src/bus.rs
```diff
@@ -0,0 +1,985 @@
...
+where
+    S: Spec + 'static,
+{
+    /// Serialization and local fan-out run on the spawned task, like
+    /// [`LocalBus`](cdk_common::pub_sub::LocalBus), so a publishing request
+    /// handler never pays for them or contends with a concurrent `subscribe`.
+    fn publish(&self, event: S::Event) {
```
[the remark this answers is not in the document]
crodas: If postgres is down, everything else is down

crates/cdk-postgres/src/bus.rs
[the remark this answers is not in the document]
thesimplekid: I don’t think we should merge a new resource-exhaustion vector and defer its fix to #2478. Previously, publish tasks completed after local delivery; enabling this transport adds an unbounded wait on Postgres for every event. If the bus connection stalls while requests continue generating notifications, pending tasks and payloads accumulate without a limit.

Sending event IDs later would reduce payload size, but it wouldn’t bound pending operations. Could we add a bounded outbound queue and write deadlines in this PR, with a regression test for a stalled connection?

crates/cdk-postgres/src/bus.rs
```diff
@@ -0,0 +1,985 @@
...
+
+    fn backoff_policy(&self) -> BackoffPolicy {
+        BackoffPolicy {
+            initial_connect_backoff: INITIAL_BACKOFF,
+            max_connect_backoff: MAX_BACKOFF,
+        }
+    }
+
+    async fn connect(&mut self) -> Result<Self::Stream, Self::ConnectError> {
+        let result = self.open().await;
```
thesimplekid: Could we apply the configured timeout to every `open()` attempt, including `LISTEN`? Currently it only bounds the initial startup wait. A subsequent attempt can hang indefinitely and prevent further retries.

This differs from the earlier resolved `LISTEN` error finding: that covered an operation returning an error; this covers one never returning. Cancellation should also clean up any spawned connection driver.

crates/cdk-mintd/example.config.toml
```diff
@@ -105,6 +105,39 @@ max_connections = 20
...
+#                              needs a direct connection carved out for it.
+#
+# Before enabling "postgres-listen-notify", note that LISTEN and NOTIFY need no
+# privileges beyond connecting: pg_notify is executable by PUBLIC and neither
+# statement touches a table. Any role that can open a session to this database
+# can therefore read the channel (quote ids, invoices, amounts and payment
+# preimages) and publish on it, and what it publishes reaches wallets over
+# NUT-17 as if the mint had sent it. Enable it only on a database the mint has
+# to itself: give the mint its own role, keep other principals off the database,
+# run REVOKE EXECUTE ON FUNCTION pg_notify(text, text) FROM PUBLIC, and set a
```
thesimplekid: Could we include an explicit `GRANT EXECUTE ON FUNCTION pg_notify(text, text) TO <mint_role>`? Revoking execution from `PUBLIC` also removes a normal mint role’s publishing permission. The cross-instance test failed after applying these instructions and passed after adding the grant.

Also, the ADR should avoid claiming this prevents other roles from publishing: direct SQL `NOTIFY` still succeeds after the revoke.

crates/cdk-postgres/src/bus.rs
```diff
@@ -0,0 +1,985 @@
...
+        Ok(ConnectionLiveness {
+            driver: Some(driver),
+            client: self.client.clone(),
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl SupervisedStream for PgListen {
+    type Item = Infallible;
```
thesimplekid: Could the supervised stream carry actual notifications, following the existing `KeysetSubscription` and `PaymentWaiter` implementations? Currently it only tracks connection liveness, while messages travel through a separate dispatcher.

Consolidating those paths looks like an opportunity to remove adapter layers and simplify ownership and shutdown while retaining the existing reconnect machinery.

crates/cdk-mintd/src/config.rs
```diff
@@ -962,6 +962,59 @@ impl std::str::FromStr for DatabaseEngine {
...
+                    .to_string(),
+            ),
+            _ => Err(format!("Unknown pubsub transport: {s}")),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
+#[serde(default)]
+pub struct PubSubConfig {
```
thesimplekid: Do we need a separate transport selector and channel setting? The same Postgres bus works for both single- and multi-instance deployments. Could setup automatically select local delivery for SQLite and local delivery plus `LISTEN/NOTIFY` for Postgres, using a fixed internal channel?

I’d retain one Postgres opt-out if we need compatibility with transaction-pooling proxies. Bus connection failures should preserve local delivery and trigger retries, so automatic selection wouldn’t reintroduce the earlier startup regression. This would also remove most of the pub/sub configuration plumbing and its precedence questions.

thesimplekid (commented):
Two correctness fixes and two design questions below. I’ve also replied to the existing outbound-publishing thread.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

PR #2267, opened 2026-07-27
feat(tx): status to wallet transaction list

### LINKED ISSUE #2248: Feature Request: Add pending transactions to the wallet tx list and interface
Currently cdk's transaction list is only completed tx's so if a user wallet wants to show something like a pending melt they have to track this themselves. CDK does track this internally in the sagas but only appends to the tx list once it completes. If we add a state to the tx items we can add the tx earlier in the lifecycle so it can be consumed and cdk can update its state once it completes/fails.

asmogo:
### Description

closes #2248 

* transaction status is controlled by the saga
* changed transaction primary key to saga_id

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/nwc.rs
+++ b/crates/cdk/src/wallet/nwc.rs
@@ -20,7 +20,7 @@
 //! be represented exactly are rejected rather than silently rounded. Only `Sat`
 //! and `Msat` wallets are supported.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::str::FromStr;
 use std::sync::Arc;
 
@@ -28,7 +28,9 @@ use async_trait::async_trait;
 use bitcoin::bip32::{ChildNumber, DerivationPath, Xpriv};
 use bitcoin::Network;
 use cdk_common::nut00::KnownMethod;
-use cdk_common::wallet::{MeltQuote, MintQuote, Transaction, TransactionDirection};
+use cdk_common::wallet::{
+    MeltQuote, MintQuote, Transaction, TransactionDirection, TransactionStatus,
+};
 use cdk_common::{PaymentMethod, SECP256K1};
 use cdk_nwc::nip47::{
     ErrorCode, GetBalanceResponse, GetInfoResponse, ListTransactionsRequest, LookupInvoiceRequest,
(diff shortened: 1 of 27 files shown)


crates/cdk/src/wallet/nwc.rs
```diff
@@ -221,9 +223,20 @@ fn transaction_to_nip47(
...
+    let (state, settled_at) = match tx.status {
+        TransactionStatus::Pending => (cdk_nwc::nip47::TransactionState::Pending, None),
+        TransactionStatus::Completed => (
+            cdk_nwc::nip47::TransactionState::Settled,
+            tx.saga_id
+                .is_none()
+                .then_some(Timestamp::from(tx.timestamp)),
+        ),
+        TransactionStatus::Failed => (cdk_nwc::nip47::TransactionState::Failed, None),
+    };
```
thesimplekid: This actually might be the most correct for what we want to express but its worth noting. That on the mint we start our saga after the invoice has been paid, so by deriving the state from the saga here we can show the wallet_tx as pending/fail when the actual ln payment is paid. 

One option would be to compute the state from the mint quote instead of the saga. However, maybe this is really what we want to express in the wallet_tx the status of minting the ecash and not the ln payment?

crates/cdk/src/wallet/nwc.rs
```diff
@@ -582,14 +584,17 @@ impl cdk_nwc::NwcRequestHandler for WalletNwcHandler {
             }
         }
```
thesimplekid, line 703: Do we need to filter unpaid ones here depending on the unpaid flag set on line 565?

crates/cdk/src/wallet/nwc.rs
```diff
@@ -549,30 +562,19 @@ impl cdk_nwc::NwcRequestHandler for WalletNwcHandler {
...
-                    }
-                }
-                out.push(transaction_to_nip47(&tx, unit)?);
-            }
-
-            return Ok(out);
-        }
+        let pending_quote_ids = transactions
+            .iter()
+            .filter(|transaction| transaction.status == TransactionStatus::Pending)
```
thesimplekid: I think we need to include failed in this deduplication or do we want to show the failed as a retry?

thesimplekid (commented):
Few small things but otherwise looks good. 

Nice commit message 7b4656bf8c261072837d5c46682398607d937dbf

crates/cdk/src/wallet/nwc.rs
asmogo: I think the transactions should represent completion of ecash / saga operations. I will make a nwc mapping that is quote state aware and retain the saga status for the wallet transaction api.

crates/cdk/src/wallet/nwc.rs
asmogo: fixed in https://github.com/cashubtc/cdk/pull/2267/commits/a36d2aa5807f83a358a05d10335e95c40f8e43dc

crates/cdk/src/wallet/nwc.rs
asmogo: fixed in https://github.com/cashubtc/cdk/pull/2267/commits/a36d2aa5807f83a358a05d10335e95c40f8e43dc

thesimplekid:
@cdk-bot review

thesimplekid (dismissed):

PR #2268, opened 2026-07-27
fix(signatory): reject non-future keyset expiries

### LINKED ISSUE #1945: `rotate_keyset` does not validate that expiry is in the future

thesimplekid:
Validate keyset rotation requests before they reach storage or cross embedded and RPC signatory boundaries. Return a structured error for expiries at or before the current time and cover mint and signatory paths with regression tests.

### Description

-----

closes https://github.com/cashubtc/cdk/issues/1945#issuecomment-5093754155

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-signatory/src/db_signatory.rs
+++ b/crates/cdk-signatory/src/db_signatory.rs
@@ -251,7 +251,7 @@ impl Signatory for DbSignatory {
             args.unit.clone(),
             &amounts,
             args.input_fee_ppk,
-            args.final_expiry,
+            args.final_expiry.map(|expiry| expiry.to_u64()),
             args.keyset_id_type,
         );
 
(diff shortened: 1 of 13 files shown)


crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -211,6 +211,8 @@ impl Signatory for DbSignatory {
     /// Generate new keyset
     #[tracing::instrument(skip(self))]
     async fn rotate_keyset(&self, args: RotateKeyArguments) -> Result<SignatoryKeySet, Error> {
+        args.validate()?;
```
crodas: Instead of validating in multiple places, the type should not represent an incorrect state. It should fail in another layer when it is parsed

crodas (changes requested):
If `RotateKeyArguments` is unable to represent invalid state the codebase is much simpler, and there would be no need to call `validate()?`.

This would also allow fewer bugs in the future for other implementations where the keys are managed differently.

crodas (dismissed):

crodas (approved):

PR #2269, opened 2026-07-27
fix: normalize custom currency units to lowercase

### LINKED ISSUE #2243: bug: custom units with uppercase chars mishandled
# Custom units with uppercase characters are not handled consistently

## Summary

`CurrencyUnit::from_str` preserves the case of **custom** units, but `Display`/`Serialize` (and therefore every wire, log, and DB representation) lowercase them. Because `CurrencyUnit` equality is case-sensitive, a custom unit that contains any uppercase character (e.g. `BADCOIN`) becomes two distinct values that render identically: the case-preserved value held in memory from config, and the lowercased value produced everywhere else. The two never compare equal, so the mint effectively rejects its own unit.

Known units (`sat`, `msat`, `usd`, `eur`, `auth`) are unaffected because `from_str` matches them case-insensitively. All-lowercase custom units are unaffected because they have no case to lose.

## Impact

**Any mint configured with a custom unit containing ≥1 uppercase character is unusable for that unit.** The failure surfaces at two different points depending on how the mint is run:

- **`cdk-mintd` (daemon):** the mint **fails to start**. Startup validation compares the config unit against the backend's reported unit and aborts. The error message is itself a symptom of the bug — both sides print identically:
  ```
  Error: Payment backend reports unit badcoin but config registers unit badcoin;
         only matching units or sat/msat conversions are supported
  ```
- **Embedded `cdk::mint::Mint` (library use, e.g. tests):** the mint builds and starts, advertises the unit lowercased via `GET /v1/info`, `/v1/keysets`, `/v1/keys`, then rejects every `POST /v1/mint/quote/{method}` and `POST /v1/melt/quote/{method}` with `UnsupportedUnit` — the wallet is told to use `badcoin` and then refused when it does.

Severity is bounded: mints using known units or lowercase custom units are not affected.

thesimplekid:
### Description

-----

Canonicalize custom units during parsing, construction, and protocol conversions so case variants compare equally and round-trip consistently.

Cover JSON, TLV, gRPC, and FFI boundaries with regression tests.

closes https://github.com/cashubtc/cdk/issues/2243

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut00/mod.rs
+++ b/crates/cashu/src/nuts/nut00/mod.rs
@@ -611,6 +611,13 @@ pub enum CurrencyUnit {
     Custom(String),
 }
 
+impl CurrencyUnit {
+    /// Construct a custom unit, normalizing to lowercase and trimming whitespace.
+    pub fn custom<S: AsRef<str>>(value: S) -> Self {
+        Self::Custom(normalize_custom_unit(value.as_ref()))
+    }
+}
+
 #[cfg(feature = "mint")]
 impl CurrencyUnit {
     /// Derivation index mint will use for unit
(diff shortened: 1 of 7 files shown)


thesimplekid:
@cdk-bot review

crodas (approved):

crates/cashu/src/nuts/nut00/mod.rs
```diff
@@ -611,6 +611,13 @@ pub enum CurrencyUnit {
     Custom(String),
```
crodas, line 611: I think it can be `Arc<str>` as a follow-up, so cloning becomes cheap.

PR #2270, opened 2026-07-27
Signatory: serve keys from memory

crodas:
### Description

The signatory round-tripped the database on the hot path: rotate_keyset re-read the active keyset to compute the next derivation index and then reloaded the whole keyset set to refresh memory. The database's role in steady-state operation was never pinned down.

Treat the database as a persistence layer only. It is read once on boot and written only inside the rotate_keyset transaction; every other operation is served from and mutates in-memory state. Rotation reads the next index from memory and updates the in-memory maps directly after
committing, mirroring what a fresh boot would compute, so it never reads the database back.

Load strictly at boot. new runs the load once and propagates any error,so a failed load fails construction rather than leaving a signatorywithout keys; both the embedded mint and the standalone signatory server want a bad database to surface at startup. The in-memory state lives in DbSignatory's own fields, with no wrapper indirection.

Serialize rotations. rotate_keyset holds a mutex across the whole operation, so two concurrent rotations of the same unit cannot read the same derivation index and derive duplicate keysets. This guards the
standalone gRPC server, which calls rotate_keyset directly rather than through the embedded single-runner service.

Relax the embedded mint boot: a signatory reporting no active keyset is a warning rather than a hard error, so an unconfigured mint still starts and its endpoints return keyset errors until it is configured.

This is groundwork to simplify cashubtc/cdk#2253 (auto-rotate keysets on an age interval): with memory as the source of truth and rotation updating it in place, a periodic rotator only has to call rotate_keyset.

Document the persistence model, the strict boot, and rotation serialization in ADR-0003.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-signatory/src/db_signatory.rs
+++ b/crates/cdk-signatory/src/db_signatory.rs
@@ -1,6 +1,14 @@
 //! Main Signatory implementation
 //!
-//! It is named db_signatory because it uses a database to maintain state.
+//! It is named db_signatory because it uses a database to persist state. The
+//! database is a persistence layer only: it is read once on boot to hydrate the
+//! in-memory keysets and written only when rotating keys. Every other operation
+//! is served from and mutates in-memory state. See ADR-0003.
+//!
+//! Boot is strict: `new` attempts the initial keyset load from the database
+//! once and bubbles up any error, so a failed load fails construction rather
+//! than leaving a signatory without keys. On success the returned signatory is
+//! loaded and serving.
 use std::collections::HashMap;
 use std::sync::Arc;
 
@@ -10,7 +18,7 @@ use cdk_common::dhke::{sign_message, verify_message};
 use cdk_common::mint::MintKeySetInfo;
 use cdk_common::nuts::{BlindSignature, BlindedMessage, CurrencyUnit, Id, MintKeySet, Proof};
 use cdk_common::{database, Error, PublicKey};
-use tokio::sync::{watch, RwLock};
+use tokio::sync::{watch, Mutex, RwLock};
 use tracing::instrument;
 
 use crate::common::{
@@ -28,9 +36,16 @@ use crate::signatory::{RotateKeyArguments, Signatory, SignatoryKeySet, Signatory
 pub struct DbSignatory {
     keysets: RwLock<HashMap<Id, (MintKeySetInfo, MintKeySet)>>,
     active_keysets: RwLock<HashMap<CurrencyUnit, Id>>,
+    /// Serializes keyset rotations. The standalone signatory gRPC server calls
+    /// rotate_keyset directly (no embedded single-runner), so two concurrent
+    /// rotations of the same unit could otherwise read the same path index and
+    /// derive duplicate keysets.
+    rotation_lock: Mutex<()>,
     localstore: Arc<dyn database::MintKeysDatabase<Err = database::Error> + Send + Sync>,
     secp_ctx: Secp256k1<secp256k1::All>,
     custom_paths: HashMap<CurrencyUnit, DerivationPath>,
+    /// Units to initialize on boot, as `init_keysets` expects them.
+    supported_units: HashMap<CurrencyUnit, (u64, Vec<u64>)>,
     xpriv: Xpriv,
     xpub: PublicKey,
     /// Latest keyset snapshot, published on every reload (initial load and each
@@ -39,7 +54,11 @@ pub struct DbSignatory {
 }
 
 impl DbSignatory {
-    /// Creates a new MemorySignatory instance
+    /// Creates a new signatory, loading its keysets from the database.
+    ///
+    /// The load is attempted once and any error is bubbled up: a failed load
+    /// fails construction rather than returning a signatory without keys. On
+    /// success the returned signatory is loaded and serving.
     ///
     /// # Panics
     ///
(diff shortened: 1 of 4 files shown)


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

docs/adr/0003-signatory-database-persistence-only.md
```diff
@@ -0,0 +1,202 @@
...
+  succeeds, then sets `loaded`. The task holds a `Weak` reference so it stops
+  when the signatory is dropped.
+
+Until the first successful load, every key-using operation returns the new
+transient error `Error::KeysetsNotLoaded`: `blind_sign`, `verify_proofs`,
+`keysets`, and `rotate_keyset` are gated on `loaded`. `subscribe_keysets` and
+`name` are not gated: `subscribe_keysets` is the watch channel through which
+consumers receive the keysets once the background load publishes them, so it
+must keep working while loading. `Error::KeysetsNotLoaded` is classified as a
+non-definitive (retryable) failure.
```
thesimplekid: What is the reason for starting in this state? If the mint cannot connect to it's db isn't it correct to error as its unusable instead of having these two paths?

crodas: I don't have a strong opinion; it is only used on the standalone signatory. The embedded will try once and fail on the stop.

Do we want the standalone to halt? My reasoning was simple; I thought it was too harsh to fail completely since the DB now has to read on boot or fail, since all interactions going forward would be from memory. But as I stated above, the embedded would try once and bubble up the error.

crodas: Fixed in c6a739274de7ee4db75959b88cf5b3dcd2c46a0a

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -211,21 +254,29 @@ impl Signatory for DbSignatory {
...
-                keyset_info.amounts,
-            )
-        } else {
-            (1, vec![])
+        // Serialize rotations. The standalone signatory gRPC server invokes this
+        // directly (no embedded single-runner), so without this two concurrent
+        // rotations of the same unit could read the same path index below and
+        // derive duplicate keysets. Held for the whole method.
+        let _rotation = self.rotation_lock.lock().await;
+
```
thesimplekid, line 262: I think this comment is still valid here https://github.com/cashubtc/cdk/pull/2253#discussion_r3658783540. 

If we are running multiple mint instances that each have there own db_signatory but are the same mint as they share seed and psgl db this rotation lock will not protect from duplicate rotations.

crates/cdk/src/mint/mod.rs
```diff
@@ -168,24 +168,28 @@ impl Mint {
...
-            .any(|keyset| keyset.active && keyset.unit != CurrencyUnit::Auth)
-        {
-            return Err(Error::NoActiveKeyset);
+            .filter(|keyset| keyset.active && keyset.unit != CurrencyUnit::Auth)
+            .count();
+        // The signatory may not report an active keyset yet (for example a
+        // remote signatory that is still connecting). Start anyway and let the
+        // keyset drain task install keysets if they arrive over the
+        // subscription; endpoints return keyset errors until then.
+        if active_keys == 0 {
```
thesimplekid, line 289: I still think I prefer a fail fast if the mint does not have keysets?

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -255,18 +306,39 @@ impl Signatory for DbSignatory {
...
+        info.active = true;
+        let signatory_keyset: SignatoryKeySet = (&(info.clone(), keyset.clone())).into();
+
+        let mut keysets = self.keysets.write().await;
+        let mut active_keysets = self.active_keysets.write().await;
+
+        if let Some(prev_id) = active_keysets.insert(args.unit, id) {
+            if let Some((prev_info, _)) = keysets.get_mut(&prev_id) {
+                prev_info.active = false;
+            }
```
thesimplekid, line 335: Maybe an edge case but this is not cancel safe as acquiring the memory locks introduce await points after the db commit. So if the task is dropped between these two point the db will have the new keyset but memory will have the old one and I don't think this is resolved till a restart or maybe next rotation? 

Related: https://github.com/cashubtc/cdk/pull/2253#discussion_r3658783540

docs/adr/0003-signatory-database-persistence-only.md
```diff
@@ -0,0 +1,193 @@
...
+Making memory the single source of truth and having rotation update it in place,
+rather than reloading from the database, means a periodic auto-rotation only has
+to call `rotate_keyset`: no extra reads and no reload.
+
+## Decision Drivers
+
+* Memory is already the source of truth for every signing, verification, and
+  query path; only rotation still round-tripped the database.
+* The database is owned by this single process. The code assumes no other
+  writer, so there is no external invalidation to reconcile against.
```
thesimplekid, line 31: This may not be true, for example the support use case of two or more mint replicas sharing one db and keys

thesimplekid (changes requested):

docs/adr/0003-signatory-database-persistence-only.md
crodas: This is taken care of in #2273

crates/cdk-signatory/src/db_signatory.rs
crodas: I think this is handled in #2273

crates/cdk-signatory/src/db_signatory.rs
crodas: #2273

thesimplekid (approved):
Ill merge this now and move a few things to make sure we clean up in the follow up pr.

PR #2271, opened 2026-07-28
fix(bdk): avoid genesis scans for new Bitcoin Core wallets

thesimplekid:
### Description

-----

Checkpoint newly created wallets at the current Bitcoin Core tip while allowing operators to configure a birthday height for seed recovery. Reject future heights and abort creation when tip lookup fails so a fresh wallet is never accidentally pinned to genesis.

Keep persisted wallet checkpoints intact and expose the recovery height through mintd configuration and environment variables.

Same idea as https://github.com/lightningdevkit/ldk-node/pull/884

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-bdk/src/chain/bitcoin_rpc.rs
+++ b/crates/cdk-bdk/src/chain/bitcoin_rpc.rs
@@ -4,6 +4,7 @@ use std::time::Instant;
 use bdk_bitcoind_rpc::bitcoincore_rpc::{Auth, Client, Error as BitcoinRpcError, RawTx, RpcApi};
 use bdk_bitcoind_rpc::{BlockEvent, Emitter, NO_EXPECTED_MEMPOOL_TXS};
 use bdk_wallet::bitcoin::{Block, Transaction};
+use bdk_wallet::chain::BlockId;
 use tokio::sync::Mutex;
 use tokio::time::{interval, Duration};
 use tokio_util::sync::CancellationToken;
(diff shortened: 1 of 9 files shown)


thesimplekid:
@cdk-bot review

crates/cdk-bdk/src/chain/bitcoin_rpc.rs
```diff
@@ -12,6 +13,38 @@ use crate::chain::{BitcoinRpcConfig, BroadcastErrorKind, BroadcastFailure, Broad
...
+    let height = config.wallet_rescan_from_height.unwrap_or(tip);
+
+    if height > tip {
+        return Err(Error::WalletRescanHeightTooHigh {
+            requested: height,
+            tip,
+        });
+    }
+
+    let hash = match height == tip {
```
crodas: this can be if / else.

crates/cdk-bdk/src/lib.rs
```diff
@@ -269,12 +269,36 @@ impl CdkBdk {
...
+        // than scanning from genesis. An explicit rescan height overrides the
+        // tip for seed recovery. Fetch this before creating the wallet so an
+        // unreachable or misconfigured node cannot persist a genesis-pinned
+        // wallet by accident.
+        let initial_checkpoint = match wallet_opt.is_none() {
+            true => chain_source.initial_checkpoint()?,
+            false => None,
+        };
+
         let mut wallet = match wallet_opt {
```
crodas, line 282: This can be `wallet_opt.unwrap_or_else`

thesimplekid: I don't think so there are fail-able options within what would be be else.

crates/cdk-bdk/src/chain/bitcoin_rpc.rs
```diff
@@ -12,6 +13,39 @@ use crate::chain::{BitcoinRpcConfig, BroadcastErrorKind, BroadcastFailure, Broad
...
+        return Err(Error::WalletRescanHeightTooHigh {
+            requested: height,
+            tip,
+        });
+    }
+
+    let hash = if height == tip {
+        chain_info.best_block_hash
+    } else {
+        client.get_block_hash(u64::from(height))?
```
crodas, line 43:
```suggestion
        client.get_block_hash(height.into())?
```

crates/cdk-bdk/src/lib.rs
```diff
@@ -269,12 +269,36 @@ impl CdkBdk {
             .load_wallet(&mut db)
             .map_err(|e| Error::Wallet(e.to_string()))?;
 
+        // A fresh Bitcoin Core wallet should start at the current tip rather
+        // than scanning from genesis. An explicit rescan height overrides the
+        // tip for seed recovery. Fetch this before creating the wallet so an
+        // unreachable or misconfigured node cannot persist a genesis-pinned
+        // wallet by accident.
+        let initial_checkpoint = match wallet_opt.is_none() {
```
crodas, line 277: I think this can be map or at least a if/else.

crodas (approved):
It is a very straightforward change; I am happy with this solution. Most of my feedback is cosmetic and a nice to have not a blocker.

PR #2272, opened 2026-07-28
fix(wallet): harden WebSocket reconnects and NUT-19 replay

prusnak:
While working on enclavia auto-reconnect feature of https://github.com/cashubtc/cdk/pull/2231 (not yet added to PR), I stumbled upon these improvements:

  ### Description

  Hardens wallet recovery from unreliable mint connections while preventing retry storms and unsafe request replays.

  - Classifies WebSocket failures as transient, unsupported, or terminal across native, WASM, and Tor transports.
  - Retries transient WebSocket failures with capped exponential backoff.
  - Falls back to polling when subscriptions are unsupported and stops retrying terminal failures.
  - Restricts NUT-19 replay to ambiguous connection failures and timeouts on routes explicitly advertised by the mint.
  - Enforces the NUT-19 cache window as an end-to-end deadline and never replays HTTP responses or terminal errors.
  - Adds coverage for failure classification, reconnect backoff, replay eligibility, route support, and deadline enforcement.

  -----

  ### Notes to the reviewers

  Browsers do not expose WebSocket handshake status codes. An initial WASM connection failure is therefore treated as unsupported, allowing the wallet to fall back to polling instead of reconnecting indefinitely.

  HTTP status responses—including 408, 429, and 5xx—are intentionally not replayed. NUT-19 only guarantees caching of successful responses, so replaying a side-effecting request after receiving an HTTP response may be unsafe.

### DIFF
--- a/crates/cdk-common/src/pub_sub/remote_consumer.rs
+++ b/crates/cdk-common/src/pub_sub/remote_consumer.rs
@@ -225,12 +225,11 @@ where
                     )
                     .await
                 {
-                    retry_at = Some(Instant::now() + backoff);
-                    backoff =
-                        (backoff + STREAM_CONNECTION_BACKOFF).min(STREAM_CONNECTION_MAX_BACKOFF);
-
-                    if matches!(err, Error::NotSupported) {
+                    if matches!(&err, Error::NotSupported | Error::Terminal(_)) {
                         stream_supported = false;
+                    } else {
+                        retry_at = Some(Instant::now() + backoff);
+                        backoff = backoff.saturating_mul(2).min(STREAM_CONNECTION_MAX_BACKOFF);
                     }
                     tracing::error!("Long connection failed with error {:?}", err);
                 } else {
@@ -263,7 +262,7 @@ where
                     )
                     .await
                 {
-                    if matches!(err, Error::NotSupported) {
+                    if matches!(&err, Error::NotSupported | Error::Terminal(_)) {
                         poll_supported = false;
                     }
                     tracing::error!("Polling failed with error {:?}", err);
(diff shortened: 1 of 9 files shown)


crates/cdk-common/src/pub_sub/remote_consumer.rs
```diff
@@ -225,12 +225,11 @@ where
                     )
                     .await
                 {
-                    retry_at = Some(Instant::now() + backoff);
```
crodas, line 228: I think this can be abstracted with [SupervisedStream](https://github.com/cashubtc/cdk/commit/6217df76dfa7da9f07863c8f80b6e7fa83d92fd1#diff-f7a547063823c1c78e21bdb69def6d08eb52c710615f0d6180325484e2d81802R39-R150)

prusnak: Not cleanly. Transport::stream() returns a long-running Future<Result<(), Error>>, not a Stream, and the surrounding loop must continue polling while the stream is backing off. A SupervisedStream that sleeps internally would delay that polling fallback.

  It also needs special semantics:

  - NotSupported/Terminal disable only streaming.
  - Transient errors schedule a retry without blocking polling.
  - Each attempt recreates and exposes the control channel.
  - Successful completion resets the backoff.

  I’d keep this logic inline. If reuse becomes necessary, abstract only the retry state into something like ReconnectBackoff, rather than wrapping the whole operation in SupervisedStream. There is currently no SupervisedStream implementation in this workspace. See crates/cdk-common/src/pub_sub/remote_consumer.rs:173.

crodas:
I think https://github.com/cashubtc/cdk/pull/2298 will address this issue elegantly, allowing non-HTTP transports to integrate better. I will make sure this fix is in that branch

But since #2298 will not be merged yet, I will rebase this and approve it. Thanks @prusnak

crodas (approved):

thesimplekid (approved):

PR #2273, opened 2026-07-29
Signatory: share keysets across instances via periodic reload

crodas:
### Description

This depends on #2270

Let multiple signatories (or embedded mints) run active/active on one keys database without putting it on the signing path. The derivation index is allocated in the database under a per-unit lock and a version token gates reloads, so peers stay consistent; keys are served from an ArcSwap so a reload never blocks signing. The reload is opt-in, off by default.

See docs/adr/0004-signatory-multi-instance-sharing.md.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-signatory/Cargo.toml
+++ b/crates/cdk-signatory/Cargo.toml
@@ -16,6 +16,7 @@ sqlcipher = ["cdk-sqlite/sqlcipher"]
 grpc = ["dep:tonic", "tokio/full", "dep:tonic-prost", "dep:tonic-prost-build", "dep:prost"]
 
 [dependencies]
+arc-swap.workspace = true
 async-trait.workspace = true
 bitcoin.workspace = true
 cdk-common = { workspace = true, default-features = false, features = [
--- a/crates/cdk-signatory/src/common.rs
+++ b/crates/cdk-signatory/src/common.rs
@@ -18,55 +18,58 @@ pub async fn init_keysets(
     localstore: &Arc<dyn database::MintKeysDatabase<Err = database::Error> + Send + Sync>,
     supported_units: &HashMap<CurrencyUnit, (u64, Vec<u64>)>,
 ) -> Result<(), Error> {
-    let keysets_infos = localstore.get_keyset_infos().await?;
     let mut tx = localstore.begin_transaction().await?;
 
-    let keysets_by_unit: HashMap<CurrencyUnit, Vec<MintKeySetInfo>> =
-        keysets_infos.iter().fold(HashMap::new(), |mut acc, ks| {
-            acc.entry(ks.unit.clone()).or_default().push(ks.clone());
-            acc
-        });
+    // The transaction holds the global keyset advisory lock, so reading each
+    // unit's keysets and reassigning the active pointer is atomic against any
+    // concurrent rotation. A pre-transaction read would reopen that race.
+    //
+    // Iterate in a deterministic order for reproducibility; HashMap iteration
+    // order is randomized per process.
+    let mut units: Vec<_> = supported_units.iter().collect();
+    units.sort_by(|a, b| a.0.cmp(b.0));
 
-    for (unit, keysets) in keysets_by_unit {
-        // We only care about units that are supported
-        if let Some((input_fee_ppk, amounts)) = supported_units.get(&unit) {
-            let mut keysets = keysets;
-            keysets.sort_by_key(|b| std::cmp::Reverse(b.derivation_path_index));
+    for (unit, (input_fee_ppk, amounts)) in units {
+        let mut keysets = tx.get_keyset_infos_by_unit(unit).await?;
+        keysets.sort_by_key(|b| std::cmp::Reverse(b.derivation_path_index));
 
-            if let Some(highest_index_keyset) = keysets.first() {
-                if highest_index_keyset.is_expired() {
-                    tracing::info!(
-                        "Highest index keyset for unit {} has expired, skipping reactivation",
-                        unit
-                    );
-                    continue;
-                }
+        let Some(highest_index_keyset) = keysets.first() else {
+            continue;
+        };
 
-                // Check if it matches our criteria
-                if highest_index_keyset.input_fee_ppk == *input_fee_ppk
-                    && highest_index_keyset.amounts == *amounts
-                {
-                    tracing::debug!("Current highest index keyset matches expect fee and amounts. Setting active");
-                    let id = highest_index_keyset.id;
+        if highest_index_keyset.is_expired() {
+            tracing::info!(
+                "Highest index keyset for unit {} has expired, skipping reactivation",
+                unit
+            );
+            continue;
+        }
 
-                    // Validate we can generate it (sanity check)
-                    let _ = MintKeySet::generate_from_xpriv(
-                        secp_ctx,
-                        xpriv,
-                        &highest_index_keyset.amounts,
-                        highest_index_keyset.unit.clone(),
-                        highest_index_keyset.derivation_path.clone(),
-                        highest_index_keyset.input_fee_ppk,
-                        highest_index_keyset.final_expiry,
-                        highest_index_keyset.id.get_version(),
-                    );
+        // Check if it matches our criteria
+        if highest_index_keyset.input_fee_ppk == *input_fee_ppk
+            && highest_index_keyset.amounts == *amounts
+        {
+            tracing::debug!(
+                "Current highest index keyset matches expect fee and amounts. Setting active"
+            );
+            let id = highest_index_keyset.id;
 
-                    let mut keyset_info = highest_index_keyset.clone();
-                    keyset_info.active = true;
-                    tx.add_keyset_info(keyset_info).await?;
-                    tx.set_active_keyset(unit.clone(), id).await?;
-                }
-            }
+            // Validate we can generate it (sanity check)
+            let _ = MintKeySet::generate_from_xpriv(
+                secp_ctx,
+                xpriv,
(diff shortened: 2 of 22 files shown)


crodas:
@cdk-bot review

thesimplekid:
Things for last pr that should be resolved here:
- [ ] https://github.com/cashubtc/cdk/pull/2270#discussion_r3664640248
- [ ] https://github.com/cashubtc/cdk/pull/2270#discussion_r3664655318
- [ ] https://github.com/cashubtc/cdk/pull/2270#discussion_r3664557567

thesimplekid:
@crodas if you could rebase now that #2270 is merged.

thesimplekid:
@cdk-bot review

crates/cdk-sql-common/src/mint/keys.rs
```diff
@@ -125,6 +127,76 @@ where
             .execute(&self.inner)
             .await?;
 
+        self.bump_keyset_version().await?;
+
+        Ok(())
+    }
+
+    async fn next_derivation_index(&mut self, unit: &CurrencyUnit) -> Result<u32, Error> {
+        // Take the per-unit advisory lock (if the backend needs one) so two
```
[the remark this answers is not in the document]
crodas: I don't think the collision is reachable. Genuine NULL-index rows only exist for keysets created before the column was added, and those were derived at `m/0'`, a different scheme from today's derivation_path_from_unit → `m/129372'/<unit>'/<index>'`. Rotating such a unit to index 1 derives a different path, so different keys, different id, a plain INSERT rather than the ON CONFLICT(id) upsert. No silent reactivation.

Every keyset that actually sits at a `derivation_path_from_unit` index is written with Some(index), never `NULL`, so there's no row that's both NULL-indexed and on the unit path at index 1.

crodas:
@cdk-bot review

crates/cdk-signatory/Cargo.toml
```diff
@@ -16,6 +16,7 @@ sqlcipher = ["cdk-sqlite/sqlcipher"]
 grpc = ["dep:tonic", "tokio/full", "dep:tonic-prost", "dep:tonic-prost-build", "dep:prost"]
 
 [dependencies]
+arc-swap = "1.7.1"
```
thesimplekid: Can we make default features false as we do for other deps. Also I think we use arc-swap else where so we can make it a workspace dep?

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -112,44 +192,96 @@ impl DbSignatory {
...
         }
 
-        self.publish_snapshot(&keysets);
+        let snapshot = Arc::new(KeysetSnapshot {
+            by_id,
+            active_by_unit,
+            version: Some(version),
+        });
+        self.publish_snapshot(&snapshot);
+        self.keysets.store(snapshot);
```
thesimplekid: Could we serialize load_keys_from_db with a dedicated mutex? ArcSwap makes each publication atomic, but concurrent reloads can finish out of order and allow an older snapshot to overwrite a newer one. A writer-only mutex would preserve lock-free signing while preventing stale publications.

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -112,44 +192,96 @@ impl DbSignatory {
         Ok(())
     }
```
thesimplekid, line 187: Should boot-time reactivation participate in the same per-unit database locking? init_keysets reads the highest keyset before opening its transaction, so a concurrent rotation can commit first and then have the starting instance reactivate the older keyset. Could the selection and reassignment happen transactionally under the per-unit lock?

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -440,6 +632,66 @@ mod test {
...
+            }
+        }
+
+        assert!(
+            seen,
+            "instance B should reconcile the peer rotation without a restart"
+        );
+    }
+
     #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
```
thesimplekid, line 837: Should we create two DbSignatory signatory instances to actually test the db locks?

crates/cdk/src/mint/builder.rs
```diff
@@ -127,6 +133,17 @@ impl MintBuilder {
...
+    /// Set how often the built signatory reloads keysets from the shared
+    /// database. `None` (the default) disables the reload: a single mint owns
+    /// its database and needs no polling. Set an interval only to run several
+    /// mints/signatories against one shared database, so each picks up peers'
+    /// rotations without a restart. [`cdk_signatory::db_signatory::DEFAULT_KEYSET_REFRESH`]
+    /// is the standard cadence.
+    pub fn with_keyset_refresh_interval(mut self, interval: Option<std::time::Duration>) -> Self {
+        self.keyset_refresh_interval = interval;
+        self
+    }
```
thesimplekid: I guess we need to expose this to mintd? But we currently have a big pr changing the config so maybe create an issue and we can add it after that?

crodas: Can this be addressed in @asmogo ? I don't have enough context and this is a signatory setting not a mint.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-sql-common/src/mint/migrations/postgres/20260728000000_add_keyset_version.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE keyset_version (
+    id INTEGER PRIMARY KEY,
+    version BIGINT NOT NULL
+);
```
thesimplekid: I think we should rename this. Cashu has a `KeysetVersion` already so this is easy to confuse with that.

thesimplekid (commented):
One comment otherwise I think this is good

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk/src/mint/builder.rs
thesimplekid: Sure we can expose it later can you open an issue?

crates/cdk-signatory/src/common.rs
```diff
@@ -18,55 +18,61 @@ pub async fn init_keysets(
...
+        if highest_index_keyset.is_expired() {
+            tracing::info!(
+                "Highest index keyset for unit {} has expired, skipping reactivation",
+                unit
+            );
+            continue;
+        }
+
+        // Check if it matches our criteria
+        if highest_index_keyset.input_fee_ppk == *input_fee_ppk
```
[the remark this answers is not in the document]
crodas: This is a minor issue

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -306,37 +470,26 @@ impl Signatory for DbSignatory {
             args.keyset_id_type,
         );
 
-        let keysets = self.keysets_snapshot().await;
+        let keysets = self.keysets_snapshot();
```
thesimplekid, line 477: Should this collision check be moved into the database transaction and serialized by derivation hash? `keysets_snapshot()` can lag a peer until the next refresh, while `next_derivation_index` locks only by unit name. Two instances can therefore concurrently introduce different custom units with the same 31-bit `hashed_derivation_index`; both checks pass and both derive keys from the same BIP32 branch. With Version00 and matching denominations they can also produce the same keyset ID, allowing the `ON CONFLICT` upsert to overwrite one unit’s metadata. Querying authoritative database state while holding a lock keyed by the derivation hash would close both the stale-read and concurrent-insert cases.

crates/cdk-signatory/src/db_signatory.rs
```diff
@@ -254,42 +415,45 @@ impl Signatory for DbSignatory {
...
-                    info.amounts.clone(),
-                )
-            } else {
-                (1, vec![])
-            }
+        // Default amounts come from the in-memory active keyset and are only
+        // used when the caller does not specify any. The authoritative
+        // derivation index is allocated from the database below, not memory, so
+        // concurrent rotations across instances cannot pick the same index.
+        let default_amounts = {
```
thesimplekid, line 440: Should empty `amounts` be resolved from the authoritative database after acquiring the per-unit transaction lock? In the active/active setup this local snapshot can be stale for the refresh interval. If a peer has just rotated the unit to a new denomination set, an empty-amount rotation on this instance will reuse the older local amounts and make that old set active again, rather than reusing the currently active keyset’s amounts as the API promises.

crates/cdk-signatory/src/db_signatory.rs
crodas: I think `next_derivation_index` is serialized in the database (thanks for the [advisory_lock](https://github.com/cashubtc/cdk/pull/2273/changes/BASE..bac4a9be374353979ce53847b94ed887d9f0d915#diff-3f9414ef43318f5deeda78597a33c856517dac90a95ef068c74caa626ac1a860R201-R210)). 

I think that lock solves all race conditions, unless I am missing something

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@thesimplekid I simplified a lot, and all reading and writing from the database happens in a transaction, serialized by the database. That is OK to simplify complexity since all operations are happening off-memory and waiting for a database

crodas:
@cdk-bot review

thesimplekid (approved):

PR #2274, opened 2026-07-29
Publish only the Android artifact to the Kotlin Maven Central deployment

crodas:
### Description

Maven Central is introducing per-organization monthly publishing limits on release size, file count, and release count, enforced from October 2026. The Kotlin deployment exceeds the size limit and is close on file count, driven by desktop JVM native libraries and a per-module set of checksum and signature sidecars.

The realistic Maven audience for the Kotlin bindings is Android. Desktop JVM users can build from source, so their native libraries do not need to ship on Central at all.

Make cdk-android self-contained and the only published module. It now compiles the generated uniffi Kotlin directly into its AAR (previously an empty classes jar that re-exported cdk-jvm) and declares its own coroutines and JNA dependencies, so it resolves without a cdk-jvm coordinate. cdk-jvm stays in the repo for desktop development and tests but is no longer published, and cdk-jvm-natives is removed.

Trim the publish workflow to the Android targets, move Kotlin binding generation onto an Android build (uniffi reads the interface from the library statically, so a cross-target library works), and drop the desktop native distribution step. Update the README to document the Android-only publication and point desktop users at a source build.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

asmogo (approved):

thesimplekid:
I think this is good any reason it is still in draft? We can reevaluate adding more if https://github.com/cashubtc/cdk/pull/2289 works out.

crodas:
@thesimplekid It is ready and I'm now reviewing #2289

thesimplekid (approved):

PR #2280, opened 2026-08-01
Authenticate protected websocket connections in-band

crodas:
### Description

This depends on https://github.com/cashubtc/nuts/pull/413

The browser WebSocket API cannot set the Blind-auth header, so a browser wallet could not authenticate a protected NUT-17 /v1/ws connection. NUT-22 solves this with an in-band JSON-RPC authenticate command sent after the connection is established.

Mint: add the authenticate command, gate subscribe/unsubscribe with error 31001 until a connection is authenticated, verify and spend the BAT via the existing verify_auth path, and close never-authenticated connections after a timeout. The upgrade handler no longer rejects a header-less upgrade to a protected endpoint; it defers to the in-band command.

Wallet: for blind auth on the websocket, send the authenticate command right after connecting instead of a Blind-auth header, so one code path serves native and browser targets. Only sent when the mint protects /v1/ws.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1545,6 +1545,7 @@ version = "0.18.0-rc.1"
 dependencies = [
  "anyhow",
  "async-trait",
+ "axum",
  "bip39",
  "bitcoin",
  "bitcoincore-rpc",
--- a/crates/cashu/src/nuts/nut17/ws.rs
+++ b/crates/cashu/src/nuts/nut17/ws.rs
@@ -69,6 +69,14 @@ pub struct RawNotificationInner<I> {
     pub payload: serde_json::Value,
 }
 
+/// The response to an authenticate request (NUT-22)
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "status", rename_all = "UPPERCASE")]
+pub enum WsAuthenticateResponse {
+    /// Authentication succeeded
+    Ok,
+}
+
 /// Responses from the web socket server
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(bound = "I: Serialize + DeserializeOwned")]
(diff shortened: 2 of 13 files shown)


crodas:
@cdk-bot review

asmogo:
Two findings that might need some clearification: 

1. Deferral applies to clear-auth endpoints, which can never authenticate in-band (`crates/cdk-axum/src/router_handlers.rs:145`)
The Some(_) => arm defers all header-less upgrades to the in-band command, but authenticate::handle only parses a BlindAuthToken — there is no in-band path for a CAT. Since mintd allows websocket_auth = "clear", a header-less connection to such an endpoint can never authenticate: subscribe gets 31001 ("Endpoint requires blind auth" — the wrong error for that config), authenticate gets 31002, and the socket idles until the 30s timeout. Previously it was rejected immediately at upgrade with 401 ClearAuthRequired.
Should we match on the requirement and only defer when it's AuthRequired::Blind, keeping the upgrade-time rejection for Clear?

```rust 
Some(AuthRequired::Blind) => match token { ... None => false },
Some(AuthRequired::Clear) => {
    // as before: verify_auth errors on a missing token -> 401 at upgrade
}
```

(If the intent is to keep the door open for a future in-band clear-auth command, fine — but the 31001 message is still misleading for that config.)


2. The wallet never waits for the authenticate response (`crates/cdk/src/wallet/subscription.rs:543`)
ensure_authenticated marks the connection authenticated as soon as the command is on the wire. If the BAT is rejected (31002 — e.g. stale keyset after a mint rotation), the failure only surfaces indirectly via the subsequent subscribe's 31001, the stream errors out, and each reconnect burns another BAT (backoff-bounded, so it self-limits, but a full BAT pool still drains over ~tens of minutes). Since the request id is already assigned, should we wait for the authenticate result before sending the first subscribe? If fire-and-forget is deliberate, a short comment acknowledging the tradeoff would help.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

PR #2281, opened 2026-08-01
fix(bdk): convert msat melts before creating sends

thesimplekid:
Convert typed melt amounts and fee caps to satoshis at the BDK boundary. Reject unsupported or mismatched units and fractional-satoshi sends, while returning quote fees in the caller unit.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2282, opened 2026-08-01
perf(cashu): make CurrencyUnit::Custom hold Arc<str>

### LINKED ISSUE #2276: Make custom unit `Arc<str>` to make them cheaper to clone
> I think it can be `Arc<str>` as a follow-up, so cloning becomes cheap. 

 _Originally posted by @crodas in [#2269](https://github.com/cashubtc/cdk/pull/2269/changes#r3675028480)_

crodas:
### Description

Fixes #2276

CurrencyUnit is passed by value and cloned constantly across the workspace. Its Custom variant held a String, so every clone of a custom-unit value allocated and copied. Switching the payload to Arc<str> makes cloning a refcount bump while keeping the value immutable, which it already was in practice.

The change stays inside the enum's module plus the direct construction sites; every external boundary (SQL, redb, proto, uniffi FFI, TLV, serde) already goes through the string form and is unaffected.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/amount.rs
+++ b/crates/cashu/src/amount.rs
@@ -1694,7 +1694,7 @@ mod tests {
 
     #[test]
     fn test_amount_new_with_custom_unit() {
-        let custom_unit = CurrencyUnit::Custom("BTC".to_string());
+        let custom_unit = CurrencyUnit::Custom("BTC".into());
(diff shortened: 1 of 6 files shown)


crodas:
@cdk-bot review

thesimplekid (approved):

PR #2285, opened 2026-08-02
feat: add Iroh wallet-mint transport

callebtc:
## Summary

- add an optional `cdk-iroh` HTTP/WebSocket transport with one pooled Iroh 1.0 connection per mint, bounded retry backoff, request and idle deadlines, QUIC stream limits, admission limits, and supervised shutdown
- separate outgoing wallet endpoints from listener endpoints: clients retain relay reachability without publishing their own address
- add native wallet and CLI dispatch for `iroh://` while leaving HTTP(S) unchanged when the feature is disabled
- add mintd persistent Iroh identity startup and HTTP-only, Iroh-only, or dual listeners serving the same Axum router; `iroh://auto` derives the stable public URL from the protected endpoint key
- raise the workspace MSRV to Rust 1.91, required by Iroh 1.0.3 and `iroh-tickets` 1.0.0
- keep mintd's operator surface small by using safe library defaults for timeouts and resource limits

This PR is transport-only and contains no federation, threshold-signing, or DKG code. Mobile/FFI and browser-WASM Iroh support remain out of scope.

## Testing

- `cargo test -p cdk-iroh --no-fail-fast`
- `cargo test -p cdk --features iroh --lib wallet::mint_connector::iroh_tests`
- `cargo test -p cdk-mintd --features iroh iroh`
- `cargo clippy -p cdk-iroh --all-targets -- -D warnings`
- `cargo clippy -p cdk --features iroh --all-targets -- -D warnings`
- `cargo clippy -p cdk-mintd --features iroh --all-targets -- -D warnings`
- `cargo clippy -p cdk-mintd --all-targets -- -D warnings`
- `cargo check -p cdk --no-default-features --features wallet,iroh`
- Rust 1.91 `cargo check -p cdk-iroh` through the Nix MSRV environment
- local fakewallet mint info, issuance, and BOLT11 melt over an Iroh-only N0-relayed mint


crates/cdk/src/wallet/mint_connector/iroh_tests.rs
```diff
@@ -0,0 +1,116 @@
...
+                    break;
+                }
+            }
+            Message::Close(_) => break,
+            Message::Ping(_) | Message::Pong(_) => {}
+        }
+    }
+}
+
+#[tokio::test]
```
crodas, line 42: We need generic tests for transports (similar to the [databases](https://github.com/cashubtc/cdk/blob/main/crates/cdk-common/src/database/mint/test/mod.rs#L147-L244)), so all future transports can be [tested independently from the wallet](https://github.com/cashubtc/cdk/blob/26d68d94f7033f5cb5b35169519b701af26c966f/crates/cdk-sqlite/src/mint/mod.rs#L28-L32)

crodas:
I think it would be best to deploy Iroh and others (such as OHTTP) as a separate proxy that communicates with mintd through HTTP (like a reverse proxy). That would allow for better integrations and ease of management. I can also suggest a trait to implement to reverse-proxy mintd.

PR #2287, opened 2026-08-02
Back off between retriable HTTP retries

crodas:
### Description

The NUT-19 retry loop reissued the request immediately on failure, so a mint returning 5xx was hammered in a tight loop for the whole replay window. Retries now use exponential backoff, clamped so a sleep never extends past the end of the window.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

thesimplekid (approved):

PR #2289, opened 2026-08-02
kotlin jitpack

### LINKED ISSUE #2121: Maven central introducing limits that we exceed
It looks like maven central is adding limits that it looks like cdk will exceed by at least one metric (size) and are close on another (file count). These limits will be enforced starting in August so we will need to solve this by then.  

https://central.sonatype.org/publish/maven-central-publishing-limits/

Image

asmogo:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/.github/workflows/kotlin-publish.yml
+++ b/.github/workflows/kotlin-publish.yml
@@ -500,6 +500,36 @@ jobs:
             --title "Release ${TAG}" \
             "${RELEASE_OPTIONS[@]}"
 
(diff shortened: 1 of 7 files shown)


asmogo:
@cdk-bot review

crodas (approved):

thesimplekid:
Going to close this as we can stick to maven for now. Thanks for finding this though and could be useful in the future

PR #2290, opened 2026-08-03
Make proof reservation all-or-nothing

crodas:
### Description

Reserving a set of proofs ran one UPDATE per proof with no enclosing transaction, so hitting a proof that was no longer unspent left every proof processed before it stuck in the reserved state with no way to release it. The Supabase backend had the same defect over HTTP.

The SQL backends now reserve in a single statement inside a transaction; Supabase issues one PATCH and compensates via release_proofs when the read-back is short. Covered by a shared test all backends run.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/database/wallet/test/mod.rs
+++ b/crates/cdk-common/src/database/wallet/test/mod.rs
@@ -1526,6 +1526,47 @@ where
     assert!(result.is_err());
 }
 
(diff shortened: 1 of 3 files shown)


crodas:
@cdk-bot review

thesimplekid (approved):

PR #2291, opened 2026-08-03
Make wallet creation atomic per mint and unit

crodas:
### Description

Adding a mint did check-then-act across three separate lock acquisitions, so two concurrent callers for the same mint could both pass the existence check, both build a wallet, and the second overwrite the first.

get_or_create_wallet now holds the write lock across the lookup, the build and the insert, and never replaces an existing wallet. It is public and exposed through the FFI so bindings get the same guarantee without reimplementing the check from the foreign side. create_wallet keeps its replacing behavior, which set_mint_config relies on.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-ffi/src/wallet_repository.rs
+++ b/crates/cdk-ffi/src/wallet_repository.rs
@@ -162,6 +162,33 @@ impl WalletRepository {
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


crodas:
@cdk-bot review

thesimplekid (approved):

PR #2294, opened 2026-08-03
feat(cdk-mint-rpc): add QuoteService alongside legacy CdkMint

orangeshyguy21:
### Description

Second PR in the per-domain split of the mint management RPC, following `KeysetService` (#2239).

Adds `cdk_mint_quote_v1.QuoteService` with `GetQuoteTtl`, `UpdateQuoteTtl`, and `UpdateMintQuoteState`, registered alongside the legacy `CdkMint` service on the same port. Existing clients are unaffected as the legacy RPCs remain unchanged.

There are some notable updates to the new service which is accessible via the CLI

- **Quote state is a proto enum** (`MintQuoteState`) rather than a free-form string.
- **`UpdateMintQuoteState` returns a real response type**, where the legacy `UpdateNut04Quote` returned its own *request* type.
- **`UpdateQuoteTtl` returns the resulting TTLs** instead of an empty response.
- `UpdateMintQuoteState` accepts `MINT_QUOTE_STATE_PAID` only. `UNPAID` and `ISSUED` are rejected with `INVALID_ARGUMENT` and a reason. 
- `update-nut04-quote-state` is renamed `update-mint-quote-state`, with the old name kept as a clap alias.

-----

### Notes to the reviewers

**Rejecting `UNPAID` and `ISSUED`.** The legacy handler accepts all three states, but only `PAID` works in practice: the non-PAID branch calls `tx.add_mint_quote`, an `INSERT` against a row that already exists, so it always fails with `UNIQUE constraint failed: mint_quote.request` behind an opaque `INTERNAL: Could not update quote`. The new service returns a reason instead.

**Shared logic.** `GetQuoteTtl`/`UpdateQuoteTtl`/force-paid are extracted into inherent methods on `MintRPCServer` and called by both trait impls.. The legacy dead branch is left untouched. Planned for removal along with the legacy service.

**Known limitation, pre-existing.** The endpoint infers the payment amount from
the quote rather than accepting one. Fine for bolt11 but not so for
methods taking more than one payment (bolt12, onchain), where an operator needs
to record a specific missed payment. Legacy `UpdateNut04Quote` has the same
shape; an optional `amount` field is the natural follow-up and is additive
within `_v1` as a followup.

-----

#### CHANGED

- cdk-mint-rpc: CLI `update-nut04-quote-state` renamed to `update-mint-quote-state`; old name retained as an alias ([orangeshyguy21]).
- cdk-mint-rpc: CLI quote TTL and quote state subcommands now use `QuoteService` ([orangeshyguy21]).

#### ADDED

- cdk-mint-rpc: `cdk_mint_quote_v1.QuoteService` with `GetQuoteTtl`, `UpdateQuoteTtl`, and `UpdateMintQuoteState`, served alongside the legacy `CdkMint` service ([orangeshyguy21]).
- cdk-mint-rpc: `UpdateQuoteTtl` returns the TTL settings in effect after the update ([orangeshyguy21]).

#### FIXED

- cdk-mint-rpc: Setting a mint quote to `UNPAID` or `ISSUED` returned an opaque internal error; the new `QuoteService` rejects both with `INVALID_ARGUMENT` and a reason ([orangeshyguy21]).

----

### Checklist


thesimplekid:
@cdk-bot review

thesimplekid:
I reviewed the commits as one combined change. I do not think this is ready to merge yet.

### Findings

**critical — `crates/cdk-mint-rpc/src/proto/server.rs:302`**

`set_mint_quote_paid` credits the quote's full requested amount regardless of prior payments. For example, a multi-payment quote for 100 sats that already received 40 sats gets another 100 credited, leaving 140 sats mintable. An already-issued Bolt12/on-chain quote can similarly be reopened for additional issuance. The BOLT11-only tests do not exercise this behavior.

Could this credit only the unpaid remainder, no-op when the quote is already satisfied, or reject partially paid/issued quotes? Documenting possible double-credit in the proto does not make the accounting safe.

**warning — `crates/cdk-mint-rpc/src/bin/mint_rpc_cli.rs:234`**

The CLI unconditionally moves the existing TTL and quote-state commands to `QuoteService`. A new CLI used against an older server now receives `UNIMPLEMENTED`, even though that server still offers these operations through `CdkMint` and both advertise protocol version `1.0.0`.

Could the CLI fall back to the legacy RPC on `Code::Unimplemented`, or retain the legacy calls until the compatibility window ends?

**warning — commit `bcbd415e`**

The tip is an empty merge joining two commit-for-commit duplicate four-change stacks. The final trees are identical, but this unnecessarily doubles the feature history. Please move the bookmark to either four-commit line and drop the empty merge before publishing.

### Verification

- `cargo test -p cdk-mint-rpc` — 11 tests passed
- `cargo clippy -p cdk-mint-rpc --all-targets -- -D warnings` — passed
- `cargo fmt --all -- --check` — passed

orangeshyguy21:
Thanks for the review. I fixed the commit history. The other two I have some follow up Q's

### Force-paid accounting

This is current behavior for the mint rpc, mirrored to the new service. So the issue isn't net new.

One way to address this is to add an amount parameter so operators can be explicit about marking received payments to reusable quotes. Initial plan was to add this behavior, and expand bolt12/onchain coverage in a later PR. However this is an issue with the current rpc where a fix could be introduced here, if we'd prefer to bundle that behavioral fix.

### CLI compatibility

Yeah, this is an issue during the migration period, as the version header remains unaddressed. The same problem exists with the KeysetService today. 

I'd propose this fix be a dedicated effort outside the scope of this PR, but let me know if its a blocker and we can bundle this too.

thesimplekid:
Okay lets just keep this as is for this PR I'll merge it. Could you open an issue to capture, that we want to make this fix. Also maybe a broader tracking issue on what needs to be removed after the migration period.

thesimplekid (approved):

orangeshyguy21:
Opened both:

- #2325 : to be addressed in a follow-up PR
- #2326 : tracking issue for cleanup after the migration (legacy removal, the version header, modularizing the server file, buf CI)

PR #1870, opened 2026-04-09
Onchain bdk

thesimplekid:
### Description

-----
- [x] depends on https://github.com/cashubtc/cdk/pull/1873
- [x] depends on https://github.com/cashubtc/cdk/pull/1896
- [x] depends on https://github.com/cashubtc/cdk/pull/1875
- [x] depends on https://github.com/cashubtc/cdk/pull/1885

closes: https://github.com/cashubtc/cdk/issues/1803
### Notes to the reviewers

-----

#### FIXED

----

### Checklist

* [ ] I followed the [code style guidelines](https://github.com/cashubtc/cdk/blob/main/CODE_STYLE.md)
* [ ] I ran `just final-check` before committing

PR #2295, opened 2026-08-04
feat(payment-processor): propagate quote_id and pubkey on custom mint quotes

zeugmaster:
### Description

Adds the mint's `quote_id` (required) and the quote's NUT-20 locking `pubkey` (optional) to `CustomIncomingPaymentOptions`, so custom payment-method backends learn both at quote-creation time.

This completes the asymmetry closed for melt by #1973 and for onchain by #1870: custom incoming quotes are the only quote-creating path where the backend receives no mint-side identifier. A backend whose rail is keyed by the quote id the wallet displays (e.g. person-present settlement) has nothing to correlate on — reading it from wallet-supplied `extra_json` is spoofable — and no backend can currently see whether a quote is NUT-20-locked, so a locked-quotes-only policy is unenforceable. Both values are already in scope at the call site; the change mirrors the `Onchain` arm.

`PAYMENT_PROCESSOR_PROTOCOL_VERSION` is bumped 3.0.0 → 4.0.0 since the field is required, per the #1973 convention. Existing backends are unaffected; a regression test asserts both values match what the wallet sees.

-----

### Notes to the reviewers

* Proto fields are appended as 6/7, leaving field 5 free for the in-flight method-name addition in #2275 so the two compose without renumbering. Whichever lands second needs a trivial rebase (both touch `CustomIncomingPaymentOptions`).
* The server rejects an empty `quote_id` explicitly, since `QuoteId::from_str` accepts `""` as a base64 id.

-----

#### CHANGED

- Payment processor protocol version 3.0.0 → 4.0.0.

#### ADDED

- cdk-common: `quote_id` and NUT-20 `pubkey` on `CustomIncomingPaymentOptions`, populated by the mint and carried over the payment processor proto.

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/lib.rs
+++ b/crates/cdk-common/src/lib.rs
@@ -19,7 +19,7 @@ pub mod auth;
 pub const MINT_RPC_PROTOCOL_VERSION: &str = "1.0.0";
 
 /// Protocol version for gRPC Payment Processor communication
-pub const PAYMENT_PROCESSOR_PROTOCOL_VERSION: &str = "3.0.0";
+pub const PAYMENT_PROCESSOR_PROTOCOL_VERSION: &str = "4.0.0";
 
 #[cfg(feature = "grpc")]
 pub mod grpc;
--- a/crates/cdk-payment-processor/src/proto/client.rs
+++ b/crates/cdk-payment-processor/src/proto/client.rs
@@ -186,6 +186,8 @@ impl MintPayment for PaymentProcessorClient {
                         amount: opts.amount.map(Into::into),
                         unix_expiry: opts.unix_expiry,
                         extra_json: opts.extra_json.clone(),
+                        quote_id: opts.quote_id.to_string(),
+                        pubkey: opts.pubkey.map(|p| p.to_hex()),
                     },
                 )),
             },
--- a/crates/cdk-payment-processor/src/proto/client.rs
+++ b/crates/cdk-payment-processor/src/proto/client.rs
@@ -185,7 +185,7 @@ impl MintPayment for PaymentProcessorClient {
                         description: opts.description,
                         amount: opts.amount.map(Into::into),
                         unix_expiry: opts.unix_expiry,
-                        extra_json: opts.extra_json.clone(),
+                        extra_json: opts.extra_json,
                         quote_id: opts.quote_id.to_string(),
                         pubkey: opts.pubkey.map(|p| p.to_hex()),
                     },
--- a/crates/cdk-payment-processor/src/proto/payment_processor.proto
+++ b/crates/cdk-payment-processor/src/proto/payment_processor.proto
@@ -56,6 +56,14 @@ message CustomIncomingPaymentOptions {
   // Extra payment-method-specific fields as JSON string
   // These fields are flattened into the JSON representation on the client side
   optional string extra_json = 4;
+  // (field 5 is left free for the in-flight method-name addition, PR #2275)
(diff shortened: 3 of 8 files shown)


zeugmaster:
@cdk-bot

crates/cdk-common/src/lib.rs
```diff
@@ -19,7 +19,7 @@ pub mod auth;
 pub const MINT_RPC_PROTOCOL_VERSION: &str = "1.0.0";
 
 /// Protocol version for gRPC Payment Processor communication
-pub const PAYMENT_PROCESSOR_PROTOCOL_VERSION: &str = "3.0.0";
```
crodas, line 22: Going forward, can we just increment as a whole number? I don't see [any reason](https://github.com/cashubtc/cdk/blob/main/crates/cdk-common/src/grpc.rs#L56-L61) why it should be semver.  /cc @thesimplekid

crates/cdk-payment-processor/src/proto/payment_processor.proto
```diff
@@ -56,6 +56,14 @@ message CustomIncomingPaymentOptions {
...
   // These fields are flattened into the JSON representation on the client side
   optional string extra_json = 4;
+  // (field 5 is left free for the in-flight method-name addition, PR #2275)
+  // The mint's quote_id for this mint quote. Required. Custom backends that
+  // need a mint-side correlation key at quote-creation time should key their
+  // records on this value, as the onchain backend does.
+  string quote_id = 6;
+  // NUT-20 locking pubkey of the quote (hex), present when the wallet
+  // requested a locked quote.
+  optional string pubkey = 7;
```
crodas, line 66: Does pubkey need to be a hex string instead of just binary?

crates/cdk-payment-processor/src/proto/server.rs
```diff
@@ -231,13 +231,29 @@ impl CdkPaymentProcessor for PaymentProcessorServer {
...
+                }
+                let quote_id = parse_quote_id(&opts.quote_id)?;
+                let pubkey = opts
+                    .pubkey
+                    .map(|p| {
+                        PublicKey::from_str(&p).map_err(|_| {
+                            Status::invalid_argument("Invalid pubkey in Custom options")
+                        })
+                    })
+                    .transpose()?;
```
crodas:
```suggestion
                let pubkey = opts
                    .pubkey
                    .as_deref()
                    .map(PublicKey::from_str)
                    .transpose()
                    .map_err(|_| Status::invalid_argument("Invalid pubkey in Custom options"))?;
```

crates/cdk-payment-processor/src/proto/client.rs
```diff
@@ -186,6 +186,8 @@ impl MintPayment for PaymentProcessorClient {
                         amount: opts.amount.map(Into::into),
                         unix_expiry: opts.unix_expiry,
                         extra_json: opts.extra_json.clone(),
```
crodas:
```suggestion
                        extra_json: opts.extra_json,
```

crates/cdk-payment-processor/src/proto/payment_processor.proto
zeugmaster: The pubkey was kept a string for consistency with all other fields in this file and also the NUT-20 spec representing the pubkey as a string in the JSON. 
Both formats would work.

zeugmaster:
Applied @crodas's suggestions for `client.rs` and `server.rs` proto files

PR #2298, opened 2026-08-05
Server-side MintServer trait, a connector suite, and a stream seam

### LINKED ISSUE #2370: feat: server-side counterpart of MintConnector

crodas:
### Description

> TLDR: After this abstraction, adding other transport protocols, other than HTTP and HTTP-like, will be easier, and they can be added in external crates and be injected at runtime (#2231 and #2285 are perfect examples)
### Why

The wallet reaches a mint through `MintConnector`, a client-side trait whose
methods are the mint's RPCs. There was no matching seam on the server. The code
that turns a wire request into a `Mint` call and back existed in two hand-written
copies: the `DirectMintConnection` test helper and the `cdk-axum` HTTP handlers.
Both repeated the same `String` <-> `QuoteId` conversions and the same
payment-method dispatch, and they drifted independently. HTTP was also the only
way to reach a mint out of process.

This introduces the server-side counterpart of `MintConnector` so both sides
speak the same request/response types, and so a new transport (gRPC, Iroh, an
in-process channel) is a thin adapter instead of another copy of the
conversions.

### What

`MintServer` (`crates/cdk/src/mint/server.rs`) mirrors `MintConnector`: one
method per mint RPC, all speaking the wire types, all returning `cdk::Error`.
`impl MintServer for Mint` is the single home for the `String` <-> `QuoteId`
conversions. `cdk-axum` and `DirectMintConnection` now dispatch through it.

Streaming is a separate seam. Both traits vend a raw duplex of opaque `String`
messages via `open_stream` (`cdk_common::stream_channel`), and the NUT-17
subscription protocol is layered above it by a transport-agnostic runner
(`crates/cdk/src/mint/stream.rs`). `cdk-axum` bridges an accepted WebSocket onto
that runner, so the subscription loop is no longer duplicated.

`crates/cdk/src/rpc.rs` adds a transport-neutral RPC envelope
(`MintRpcRequest`/`dispatch`): the analog of the axum routes for a non-HTTP
transport, with `protected_endpoint()` replacing the per-route auth check.

```mermaid
graph LR
  subgraph client [client, implements MintConnector]
    W[Wallet] --> C[transport client]
  end
  subgraph server [server adapter]
    A[decode -> dispatch -> encode] --> MS[MintServer]
    MS -.impl.-> Mint
  end
  C -->|wire request| A
  A -->|wire response| C
```

The seam is deliberately the unary request/response surface plus a raw stream.
Auth verification, NUT-19 caching, the async-melt branch, and the HTTP keys
envelope stay in the transport that needs them; the ADR
(`docs/adr/0004-mint-server-transport-seam.md`) records why.

### Validation

A `mint_connector_test!` suite
(`crates/cdk-integration-tests/src/mint_connector_test/`) runs the same bodies
over three transports: in-process (`DirectMintConnection`), HTTP (`HttpClient` +
a live `cdk-axum` server), and the RPC envelope. Each backend drives a real
mint, so the bodies assert the full contract including the mint -> swap -> melt
write path and a NUT-17 subscribe handshake over `open_stream`. This is the
check that a transport and the mint agree on the wire.

### Notes for reviewers

- The RPC connector in the test suite gates auth with `protected_endpoint` +
  `verify_auth` before `dispatch`, the same order `cdk-axum` uses. It runs
  against a no-auth test mint, so the gate is exercised as a pass-through; a
  behavioral auth test (reject/accept) would need the OIDC/CAT flow and is left
  as a follow-up.
- `MintServer::post_melt` settles synchronously. A transport that wants the
  async flow calls `Mint::melt` directly, as the axum melt handler does.
- The in-process `open_stream` has no stall timeout and relies on the client
  draining its `StreamRx`; the axum bridge bounds sends with `STALL_TIMEOUT`.
- Notification delivery over a subscription is best-effort: the per-connection
  publisher channel is bounded and drops under backpressure, with the wallet's
  HTTP poll as the fallback.

### Out of scope

- No production non-HTTP transport ships here; the RPC envelope is the shared
  vocabulary plus a test that drives it.
- `cdk-axum` is not made generic over `MintServer`; its handlers still hold
  `Arc<Mint>` for auth, the subscription bridge, and the keys endpoints.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
# Adding an Iroh (QUIC p2p) transport

If this PR gets merged this is how easy it would be to add [Iroh](https://iroh.computer) instead of HTTP (compared to https://github.com/cashubtc/cdk/pull/2285)

## The seam

Read [ADR](https://github.com/crodas/cdk/blob/a2f6850c926c00c9e3c2fcc82d3cb3f91763ca0c/docs/adr/0004-mint-server-transport-seam.md) first. A transport is a byte channel plus a codec; the mint RPCs already exist as two mirror traits:

- client: `cdk::wallet::MintConnector` (`crates/cdk/src/wallet/mint_connector/mod.rs`)
- server: `cdk::mint::MintServer` (`crates/cdk/src/mint/server.rs`), implemented
  by `Mint`

Both speak the NUT request/response structs (`SwapRequest`, `MintResponse`, ...), which are already `serde`. So Iroh only needs to carry those structs as JSON and dispatch them; no HTTP.

## Wire framing

Unary RPC uses one length-delimited envelope per request, one QUIC bidi stream per call (QUIC streams are cheap):

```jsonc
// request envelope (transport-defined wrapper)
{ "auth": "<NUT-22 token, omitted if none>", "request": <MintRpcRequest> }
// reply
{ "ok": <value> }               // or
{ "err": <ErrorResponse> }      // cdk::error::ErrorResponse (NUT-00)
```

`request` is `cdk::rpc::MintRpcRequest` (an adjacently-tagged enum, one variant per `MintServer` method, serialized `{"method":..,"params":..}`) and the reply is `cdk::rpc::MintRpcResponse`. The `method` tag is the analog of the HTTP route path; the server dispatches with `cdk::rpc::dispatch` and gates auth with
`MintRpcRequest::protected_endpoint()`.

`cdk::rpc::MintRpcRequest` deliberately carries no auth, so the NUT-22 token rides in a thin transport wrapper alongside it (`auth` is a `cdk::nuts::AuthToken`; each protected request needs a fresh one):

```rust
#[derive(Serialize, Deserialize)]
struct Envelope { auth: Option<AuthToken>, request: MintRpcRequest }
```

NUT-17 subscriptions use `open_stream`: a dedicated bidi stream whose frames are the raw `String` NUT-17 messages, fed straight into `StreamTx`/`StreamRx` (`cdk::stream_channel`).

## Client: `impl MintConnector`

```rust
pub struct IrohClient { conn: iroh::endpoint::Connection }

impl IrohClient {
    // Pin the mint's NodeId; the QUIC handshake authenticates it (no MITM).
    pub async fn connect(endpoint: &iroh::Endpoint, node: iroh::NodeAddr) -> Result<Self> {
        let conn = endpoint.connect(node, b"cashu/mint/1").await?;
        Ok(Self { conn })
    }
    // Byte roundtrip over one QUIC stream; JSON in, MintRpcResponse out.
    // The per-request NUT-22 token is minted from the auth wallet keyed on the
    // request's endpoint, exactly like HttpClient does.
    async fn rpc(&self, req: MintRpcRequest) -> Result<MintRpcResponse, Error> {
        let auth = match (req.protected_endpoint(), self.get_auth_wallet().await) {
            (Some(ep), Some(w)) => w.get_auth_for_request(&ep).await?,
            _ => None,
        };
        let (mut send, mut recv) = self.conn.open_bi().await?;
        send.write_all(&serde_json::to_vec(&Envelope { auth, request: req })?).await?;
        send.finish()?;
        Ok(serde_json::from_slice(&recv.read_to_end(1 << 20).await?)?)
    }
}

#[async_trait]
impl MintConnector for IrohClient {
    async fn post_swap(&self, r: SwapRequest) -> Result<SwapResponse, Error> {
        self.rpc(MintRpcRequest::PostSwap(r)).await?.decode()
    }
    // ... one line per method ...

    async fn open_stream(&self) -> Result<(StreamTx, StreamRx), Error> {
        let (send, recv) = self.conn.open_bi().await?;
        // LengthDelimitedCodec over the QUIC stream; map frames Bytes <-> String.
        let framed_tx = FramedWrite::new(send, LengthDelimitedCodec::new());
        let framed_rx = FramedRead::new(recv, LengthDelimitedCodec::new());
        Ok((
            StreamTx::new(framed_tx.with(|s: String| ready(Ok(Bytes::from(s))))),
            StreamRx::new(framed_rx.map(|f| Ok(String::from_utf8_lossy(&f?).into_owned()))),
        ))
    }
}
```

Wire it into a wallet with `WalletBuilder::shared_client(Arc::new(iroh_client))`. Everything above `MintConnector` (`Wallet`, sagas, `SubscriptionManager`) is unchanged.

## Server adapter

```rust
pub async fn serve(mint: Arc<Mint>, endpoint: iroh::Endpoint) {
    while let Some(incoming) = endpoint.accept().await {
        let mint = mint.clone();
        tokio::spawn(async move {
            let conn = incoming.await?;
            loop {
                let (mut send, mut recv) = conn.accept_bi().await?;   // one per request
                let Envelope { auth, request } =
                    serde_json::from_slice(&recv.read_to_end(1 << 20).await?)?;
                // Auth: dispatch does NO auth. Gate here, like cdk-axum's handlers.
                if let Some(ep) = request.protected_endpoint() {
                    mint.verify_auth(auth, &ep).await?;   // auth: Option<AuthToken>
                }
                let reply = cdk::rpc::dispatch(&*mint, request).await;
                send.write_all(&serde_json::to_vec(&reply)?).await?;
                send.finish()?;
            }
        });
    }
}
```

The NUT-17 stream ALPN (or a `subscribe` method) wraps the accepted bidi stream into `StreamTx`/`StreamRx` (`cdk::stream_channel`) and hands it to `mint.serve_stream(tx, rx).await`, which runs the subscription protocol until the stream closes.

## Addressing, discovery, security

- Address = `NodeId` + a `NodeAddr` (relay url and/or direct socket addrs). Publish it out of band, in the mint's `/v1/info`, in a pkarr/DNS record, or via the n0 discovery service. No hostname or certificate authority.
- Confidentiality and mint identity come from QUIC: the client pins the expected  `NodeId`, so it cannot be MITM'd. NUT-22 blind-auth rides in the transport  envelope (the `auth` field alongside the request) and is checked by  `verify_auth` server-side; QUIC authenticates the mint, auth authenticates the
  client.

## Crate layout

`cdk-iroh`: `IrohClient` (`MintConnector`) and `serve(Arc<Mint>, Endpoint)`. Deps: `iroh`, `tokio-util` (`LengthDelimitedCodec`), `serde_json`, and `cdk` (for `cdk::rpc`, `cdk::stream_channel`, and `cdk::nuts::AuthToken`). The server mirrors `cdk_axum::create_mint_router(Arc<Mint>, ...)`: it only needs an `Arc<Mint>`.

## Prerequisites in `cdk` (state of play)

Client wrap and server *unary* wrap need no `cdk` change (`MintConnector` + `shared_client`; `MintServer` + `Mint::verify_auth` on an embedded `Arc<Mint>`). For server streaming:

- **G1 (done): `Mint::serve_stream(&self, StreamTx, StreamRx)`** runs the NUT-17 protocol on caller-owned halves. The adapter wraps its bidi stream and calls it.
- **G2 (done): `cdk::stream_channel`** re-exports `StreamTx`/`StreamRx`/`StreamError`/`from_ws`/`in_memory_pair`, so a transport crate embeds `cdk`  alone.
- **G3 (done): shared dispatch table.** `cdk::rpc` provides `MintRpcRequest`, `MintRpcResponse` (+ `decode`), `MintRpcRequest::protected_endpoint`, and  `cdk::rpc::dispatch`, so Iroh, Enclavia, and gRPC share one method<->tag mapping and one server dispatch.

Not a gap: blind-auth (NUT-22) stays off HTTP by overriding the default`auth_connector` with a transport-native `AuthMintConnector`; OIDC and LNURL are inherently HTTP and stay on HTTP.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk/src/wallet/mint_connector/http_client.rs
```diff
@@ -961,6 +966,44 @@ where
         *self.auth_wallet.write().await = wallet;
     }
 
+    async fn open_stream(&self) -> Result<(StreamTx, StreamRx), Error> {
+        let mut url = self.mint_url.join_paths(&["v1", "ws"])?;
+        if url.scheme() == "https" {
+            let _ = url.set_scheme("wss");
+        } else {
+            let _ = url.set_scheme("ws");
+        }
```
asmogo: Do you think we should keep the expectation here? 

```suggestion
        if url.scheme() == "https" {
            let _ = url.set_scheme("wss").expect("Could not set scheme");
        } else {
            let _ = url.set_scheme("ws").expect("Could not set scheme");
        }
```

crodas: I rewrote this bit to not use expect but instead to bubble up the error

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

Issue #2301, opened 2026-08-06
Feature Request: Add more custom configs to regtest environment


TheMhv:
I think that we can make regtest environment more customizable, adding options to set some parameters in justfile or using environment variables.

PR #2302, opened 2026-08-06
feat(regtest): Add custom path to regtest dir

TheMhv:
### Description

Make the regtest environment directory path configured by user.

- Can be a part of: https://github.com/cashubtc/cdk/issues/2301

-----

### Notes to the reviewers

Added a new option to set a custom `CDK_ITESTS_DIR` from justfile command.

-----

#### CHANGED
- `justfile`
- `misc/interactive_regtest_mprocs.sh`

#### FIXED

----

### Checklist

### DIFF
--- a/justfile
+++ b/justfile
@@ -579,10 +579,10 @@ nutshell-wallet-itest:
   bash ./misc/nutshell_wallet_itest.sh
 
 # Start interactive regtest environment (Bitcoin + 4 LN nodes + 2 CDK mints)
-regtest db="sqlite" host="127.0.0.1":
+regtest db="sqlite" host="127.0.0.1" path="":
   #!/usr/bin/env bash
   set -euo pipefail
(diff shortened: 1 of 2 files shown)


thesimplekid:
@cdk-bot review

TheMhv:
@cdk-bot review

thesimplekid:
This is useful for keeping regtest data outside `/tmp`, but could we clarify whether the custom path is intended to be persistent or merely select the parent directory?

As written, keeping the directory and running the command again will fail because `progress_pipe` already exists when `mkfifo` runs. I also don’t think a caller-supplied directory should ever become an `rm -rf` target, even behind a prompt.

I suggest either:

- Treat the path as persistent: never delete it and remove/recreate only transient files such as `progress_pipe`; or
- Treat it as a parent directory: create an owned directory with `mktemp -d -p "$3"` and delete only that directory.

With one of those lifecycle models, a small reuse check, documentation, and a rebase onto `main`, I think this would be worth merging.

PR #2306, opened 2026-08-09
fix(mint): harden melt saga recovery against unsafe compensation

thesimplekid:
### What changed

- Fail closed when internal-settlement or backend status cannot be determined.
- Preserve `Unknown` as an indeterminate payment state instead of converting it to `Failed`.
- Prevent stale compensation from rolling back a saga that was deleted or advanced to `Finalizing`.
- Persist and recover finalization data through the common finalization path.
- Make saga updates fail when their recovery record is missing.
- Keep per-saga startup recovery failures from aborting recovery of all remaining sagas.

### Why

Melt recovery could return payer proofs or roll back state after settlement had already progressed. Depending on the race or crash point, that could leave a paid Lightning or internally settled melt compensated, creating a double-credit or inflation risk.

The recovery paths now compensate only after fresh checks establish that payment did not succeed and no concurrent finalizer owns the saga.

### PR split

The LDK-specific BOLT12 quote-to-payment lookup work was moved to draft PR #2320. This PR now contains only the backend-independent recovery hardening and retains its existing review threads.

### Validation

The PR includes regression coverage for internal settlement, stale failure observations, missing lookup IDs, concurrent finalization, and missing saga records. CI will rerun against the rewritten branch.


thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -176,6 +176,55 @@ impl Mint {
...
+                        "payment status unverifiable for quote {quote_id} (saga {}): \
+                         no request lookup id; manual intervention required",
+                        saga.operation_id
+                    );
+                    tracing::error!("{}", err_string);
+                    return Ok(Some(err_string));
+                }
+                Err(err) => {
+                    let err_string = format!(
+                        "failed to determine internal settlement for quote {quote_id}: {err}"
```
crodas: Instead of String, can we return an enum?

crates/cdk-common/src/database/mint/test/saga.rs
```diff
@@ -430,7 +435,42 @@ where
     let mut tx = Database::begin_transaction(&db).await.unwrap();
     let saga = tx.get_saga(&operation_id).await.unwrap();
     assert!(saga.is_none(), "Non-existent saga should return None");
-    tx.commit().await.unwrap();
+    let result = tx
+        .update_saga(
```
crodas: We need to update the `update_sage` and similar functions; they should force the user to read get_saga so the saga to update is the latest version.

thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

thesimplekid:
Closing this in favor of a smaller, ordered release stack:

1. [#2337 — require acquired saga rows for updates](https://github.com/cashubtc/cdk/pull/2337)
2. [#2338 — harden melt recovery against unsafe compensation](https://github.com/cashubtc/cdk/pull/2338)
3. [#2339 — persist pending melt recovery handoffs](https://github.com/cashubtc/cdk/pull/2339)

#2338 is the direct replacement for this PR and carries the backend-independent recovery hardening. #2337 separates the database mutation prerequisite, while #2339 isolates the durable `PaymentPending` distinction needed before treating `Unpaid` as safe to compensate.

The advisory-lock, recovery-leader, and generation-fencing work is intentionally deferred from the next release.

PR #2307, opened 2026-08-10
Add CheapStr and replace Arc<str>

crodas:
### Description

`Arc<str>` makes cloning cheap but always allocates, even for a string literal, and it cannot borrow from a buffer the caller already owns. CheapStr covers all three cases in one type: a 'static literal, a borrow, and a refcounted owned string.

Equality, ordering, and hashing are hand-written so they compare by content. The derived versions would key off the variant discriminant, making Static("x") differ from Borrowed("x") and breaking Borrow<str> lookups. Serde is hand-written for the same reason: a derive would emit an externally tagged enum instead of the bare string Arc<str> produces today.

Every replaced site uses CheapStr<'static>. Adding a lifetime parameter to CurrencyUnit or SqlPart would ripple through the workspace for no gain yet.

This idea is borrowed from https://github.com/z-galaxy/zcheapstr/

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/cheap_str.rs
+++ b/crates/cashu/src/cheap_str.rs
@@ -1,0 +1,364 @@
+//! A string that is cheap to clone.
+//!
+//! Covers the three shapes a short, frequently cloned string takes: a `'static`
+//! literal, a borrow from a caller-owned buffer, and a refcounted owned string.
+//! All of them compare, order and hash purely by content, so the variant a value
+//! happens to be in is never observable.
+
+use std::borrow::{Borrow, Cow};
+use std::cmp::Ordering;
+use std::convert::Infallible;
+use std::fmt;
+use std::hash::{Hash, Hasher};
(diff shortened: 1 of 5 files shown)


crodas:
@cdk-bot review

crodas:
@cdk-bot review

thesimplekid:
`CurrencyUnit` is cloned frequently, but its standard variants already contain no string allocation, and `Custom` already uses `Arc<str>`, so those clones are already cheap. The SQL types also already use `Arc<str>`. In the proposed call sites, values would generally become `CheapStr::Owned`, which has essentially the same allocation and cloning behavior.

The `Static` and `Borrowed` variants are not used by any production call site in this PR. Given that, adding a new public type, implementation complexity, and public API churn does not appear justified by a concrete performance problem or benchmark.

PR #2308, opened 2026-08-10
fix(wallet): preserve proofs after ambiguous mint failures

thesimplekid:
Treat already-signed and already-spent responses as indeterminate so replayed swaps do not release inputs the mint may have already consumed.

Reload and claim persisted melt sagas before compensation. Release proofs only when they are owned by that operation and remain reserved or pending. This prevents stale recovery from resurrecting spent proofs or clobbering a newer reservation.

Add regression coverage for replay errors, stale snapshots, concurrent finalization, and database release semantics.

found by project loupe

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/database/wallet/mod.rs
+++ b/crates/cdk-common/src/database/wallet/mod.rs
@@ -177,7 +177,11 @@ where
         operation_id: &uuid::Uuid,
     ) -> Result<(), Err>;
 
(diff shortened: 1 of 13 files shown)


thesimplekid:
@cdk-bot review

crodas (commented):
LGTM

PR #2318, opened 2026-08-10
fix(wallet): read the mint info pubkey as the bytes the store writes

### LINKED ISSUE #2317: Wallet mint info pubkey is written as raw bytes but read as JSON — silently lost on every DB round trip
The wallet database's `mint.pubkey` column has incompatible writer and reader in
  the same file, so any persisted mint info comes back with `pubkey: None`. 
 
  ## Details
 
  In `crates/cdk-sql-common/src/wallet/mod.rs`:
 
  - **Writer** — `add_mint` binds the column as raw SEC1 bytes:
    `pubkey.map(|p| p.to_bytes().to_vec())`
  - **Reader** — `sql_row_to_mint_info` parses JSON only, and swallows failure:
    `pubkey: column_as_nullable_string!(&pubkey, |v| serde_json::from_str(v).ok(), |v| serde_json::from_slice(v).ok())`
   
  A raw 33-byte compressed pubkey (leading `0x02`/`0x03`) is never valid JSON, so
  the parse fails on every row ever written by `add_mint`, and `.ok()` turns that
  into a silent `None`.
  
  ## Symptom
  
  Confusing restart-only failures. A fresh wallet works: mint info is fetched over
  HTTP and held in memory, so the pubkey is present all session while the
  unreadable bytes are persisted in the background. On the next process start the
  metadata cache loads from the database, considers the record populated, and
  serves `pubkey: None` — anything that needs the mint's advertised pubkey (e.g.
  signature schemes that bind the mint identity) fails after the first restart
  while working perfectly on a fresh install. Verified against a real wallet DB:
  the stored column holds the raw 33-byte blob; the loaded `MintInfo` has no 
  pubkey.

  Discovered while testing get mint quote by pubkey, which I guess is the first 
  use of mint identity pubkey? Fixed by retrieving the mint info when pubkey
  is found to be `None`. 

  ## Suggested fix

  Make the reader accept the encoding the writer produces: try
  `PublicKey::from_slice` on the raw bytes first, keeping the JSON parses as
  fallback for any rows written in other formats. (Alternatively switch the
  writer to JSON, but the reader must still accept raw bytes — every existing row
  in the wild holds them.) A round-trip regression test that goes through
  `add_mint` → `get_mint` on a real backend would have caught this; the current
  suites only exercise the in-memory path.

vnprc:
The wallet's `mint` table pubkey column has only ever been written as raw
  compressed bytes — `add_mint` has bound `to_bytes()` since before
  `cdk-sql-common` existed, matching every other key-material column and the
  postgres schema's own `BYTEA` declaration. The reader only attempted JSON
  parses and swallowed the failure, so no stored row has ever round-tripped:
  fresh wallets work from the in-memory fetch, and the first process restart
  serves mint info with the pubkey silently gone.

  Standardize the reader on the one encoding that exists: parse the raw bytes,
  nothing else. The generic `add_and_get_mint` suite test previously
  round-tripped `MintInfo::default()` and asserted only existence (which is how
  this survived); it now stores a pubkey and asserts it reads back intact, so
  every backend running the shared suite inherits the regression net — verified
  against sqlite, redb, and live postgres.
  
  Closes #2317

### DIFF
--- a/crates/cdk-sql-common/src/wallet/mod.rs
+++ b/crates/cdk-sql-common/src/wallet/mod.rs
@@ -2015,8 +2015,8 @@ fn sql_row_to_mint_info(row: Vec<Column>) -> Result<MintInfo, Error> {
 
     Ok(MintInfo {
         name: column_as_nullable_string!(&name),
-        pubkey: column_as_nullable_string!(&pubkey, |v| serde_json::from_str(v).ok(), |v| {
-            serde_json::from_slice(v).ok()
+        pubkey: column_as_nullable_string!(&pubkey, |_| None, |v| {
+            cdk_common::nuts::PublicKey::from_slice(v).ok()
         }),
         version: column_as_nullable_string!(&version).and_then(|v| serde_json::from_str(&v).ok()),
         description: column_as_nullable_string!(description),
--- a/crates/cdk-common/src/database/wallet/test/mod.rs
+++ b/crates/cdk-common/src/database/wallet/test/mod.rs
@@ -181,7 +181,14 @@ where
     DB: Database<crate::database::Error>,
 {
     let mint_url = test_mint_url();
-    let mint_info = MintInfo::default();
+    let pubkey = crate::nuts::PublicKey::from_hex(
+        "02836c831cfff541ba17fbca32dd0f6a54d06305893cbcb2af0d7143e66a609147",
+    )
+    .unwrap();
+    let mint_info = MintInfo {
+        pubkey: Some(pubkey),
+        ..MintInfo::default()
+    };
 
     // Add mint
     db.add_mint(mint_url.clone(), Some(mint_info.clone()))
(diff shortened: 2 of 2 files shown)


thesimplekid:
@cdk-bot review

crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -1966,8 +1966,8 @@ fn sql_row_to_mint_info(row: Vec<Column>) -> Result<MintInfo, Error> {
 
     Ok(MintInfo {
         name: column_as_nullable_string!(&name),
-        pubkey: column_as_nullable_string!(&pubkey, |v| serde_json::from_str(v).ok(), |v| {
-            serde_json::from_slice(v).ok()
+        pubkey: column_as_nullable_string!(&pubkey, |_| None, |v| {
```
crodas, line 2018: We should use `column_as_nullable_binary!`

crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -1966,8 +1966,8 @@ fn sql_row_to_mint_info(row: Vec<Column>) -> Result<MintInfo, Error> {
...
+            cdk_common::nuts::PublicKey::from_slice(v).ok()
         }),
```
crodas, line 2020:
```suggestion
        pubkey: column_as_nullable_binary!(&pubkey).map(cdk_common::nuts::PublicKey::from_slice).transpose()?,
```

thesimplekid (approved):

PR #2319, opened 2026-08-10
fix(wallet): drop the unused mint_quote created_time column

### LINKED ISSUE #2316: Wallet mint_quote.created_time is a dead column — propose dropping it
Updating this issue after some git archaeology: the right fix is probably
  deletion, not population. Mea culpa included — I co-authored the PR that
  created this situation.

  ## History

  The column was added by #1303 ("Wallet get unissued mint quotes",
  `f184795a`, Dec 2025), which I co-authored. The commit message promises
  "add created_time column to mint_quote table for ordering" — but the design
  evolved mid-PR: the shipped `get_pending_mint_quotes` filters by payment
  method and amounts using the composite index, and never orders by age. The
  migration stayed in; the code that would have written or read the column was
  never written. Nothing since has touched it.

  ## Current state

  - `NOT NULL DEFAULT 0` — every row in every cdk wallet stores `0`
  - `add_mint_quote` never binds it; no query reads it; the public
    `cdk_common::wallet::MintQuote` struct has no such field
  - `cdk-supabase`'s `mint_quote` schema never gained the column at all, so the
    backends already disagree about it
   
  ## Proposal: drop it

  Since the column has never contained data, a `DROP COLUMN` migration
  (sqlite + postgres) is lossless today — this is the cheapest moment to
  remove it there will ever be. It also re-unifies the backend schemas for
  free. The plausible future uses are already covered by populated columns:
  `expiry` for validity and `updated_at` for staleness/pruning. If 
  creation-time ordering ever becomes a real feature, re-adding the column
  *with* its plumbing in one coherent PR beats maintaining an empty one for
  years.

vnprc:
Removes the wallet database's `mint_quote.created_time` column. It was added
  for ordering queries that were never implemented: no code path has ever
  written it (every row in every wallet holds the schema default of `0`) and no
  query reads it. The public `MintQuote` struct never carried the field. 

  Deleting it now is lossless — the column contains no information — and it
  restores schema parity with the backends that never gained the column.
  Age- and staleness-based queries already have populated columns to key on
  (`expiry`, `updated_at`); if creation-time ordering ever becomes a real
  feature, the column can return together with its write path and readers in
  one coherent change.
 
  Two migration files (sqlite + postgres); every fresh test database exercises
  the full migration chain (the original ADD followed by this DROP), verified
  on sqlite and live postgres.

  Closes #2316

### DIFF
--- a/crates/cdk-sql-common/src/wallet/migrations/postgres/20260810000000_drop_mint_quote_created_time.sql
+++ b/crates/cdk-sql-common/src/wallet/migrations/postgres/20260810000000_drop_mint_quote_created_time.sql
@@ -1,0 +1,3 @@
+-- Drop the unused created_time column from mint_quote: added for ordering queries
+-- that were never implemented, it has never been written or read.
+ALTER TABLE mint_quote DROP COLUMN created_time;
--- a/crates/cdk-sql-common/src/wallet/migrations/sqlite/20260810000000_drop_mint_quote_created_time.sql
+++ b/crates/cdk-sql-common/src/wallet/migrations/sqlite/20260810000000_drop_mint_quote_created_time.sql
@@ -1,0 +1,3 @@
+-- Drop the unused created_time column from mint_quote: added for ordering queries
+-- that were never implemented, it has never been written or read.
+ALTER TABLE mint_quote DROP COLUMN created_time;


thesimplekid (approved):
Wonder if we should have it. Can always add it later I guess

PR #2320, opened 2026-08-10
feat(cdk-ldk-node): persist BOLT12 quote ID payment lookup

thesimplekid:
### What changed

- Return the melt quote ID as the BOLT12 payment lookup identifier.
- Persist a pre-dispatch sentinel before sending and replace it with the LDK payment ID after dispatch.
- Resolve quote-ID lookups to `Unpaid`, `Unknown`, or the recorded LDK payment state without treating indeterminate dispatch as failure.
- Wire the mint KV store into the LDK backend and regtest setup.

### Why

LDK BOLT12 melt quotes previously had no durable mapping from the mint quote ID to LDK's payment ID. After a crash, a `PaymentAttempted` saga could not determine whether the payment had been dispatched, so it could not be recovered safely.

### Dependency

This PR is based on `main`, but it depends semantically on #2306 preserving `Unknown` as an indeterminate payment state. It should remain draft until #2306 merges, then be rebased onto the updated `main`.

### Validation

- `cargo test -p cdk-ldk-node` (16 passed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1575,6 +1575,7 @@ dependencies = [
  "axum",
  "bip39",
  "cdk-common",
+ "cdk-sqlite",
  "futures",
  "ldk-node",
  "maud",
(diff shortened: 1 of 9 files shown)


thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

PR #2324, opened 2026-08-11
Share one rate-limit budget per mint origin

### LINKED ISSUE #2312: Share one live rate-limit bucket per mint in WalletRepository

## Problem

`WalletRepository` can create separate wallets for different currency units at the same mint. Each wallet currently constructs its own `TokenBucket`. The buckets use the same persistence key, but their in-memory arrival times and writer tasks are independent, so each wallet can consume a full burst and concurrent writers can overwrite one another.

## Expected behavior

Wallets in the same repository that target the same mint origin should share one live rate-limit budget, regardless of currency unit.

## Acceptance criteria

- Wallets in one `WalletRepository` targeting the same mint origin share one `TokenBucket`.
- Wallets targeting different mint origins remain independent.
- A test with two units verifies that they consume one combined burst budget.
- Concurrent persistence writers cannot regress or overwrite a shared mint budget.
- The behavior of independently constructed wallets is explicitly documented.

crodas:
### Description

Fixes #2312

WalletRepository created a wallet per (mint, unit), and each wallet built its own TokenBucket. Wallets for different units at the same mint shared only the persisted KV key, not the live budget, so each could consume a full burst and their independent writer tasks overwrote one another.

Introduce a RateLimiterManager keyed by mint origin (host + non-default port) that owns one shared bucket per origin. WalletRepository holds one manager and injects the shared bucket into every WalletBuilder it builds, so all currency units at a mint pace one combined burst and a single writer persists the budget. Persistence is unchanged: the same origin still maps to the same KV key.

A wallet built directly through WalletBuilder still gets its own bucket unless one is injected, documented on the builder and the runtime setters.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/rate_limit.rs
+++ b/crates/cdk-common/src/rate_limit.rs
@@ -10,6 +10,7 @@
 //! next wallet built for the same mint instead of starting full and bursting
 //! again.
 
+use std::collections::HashMap;
 use std::future::Future;
 use std::num::NonZeroU32;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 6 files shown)


crodas:
@cdk-bot review

thesimplekid (approved):
Can you open PRs from your fork. Otherwise LGTM thanks

PR #2329, opened 2026-08-11
fix(mint): throttle payment backend quote checks

thesimplekid:
Persist the last backend check time for mint quotes and claim checks with an atomic conditional database update. This limits concurrent HTTP and WebSocket status requests to one backend check per quote every ten seconds.

Add SQLite and PostgreSQL migrations plus regression coverage for concurrent checks and the rate-limit boundary.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2330, opened 2026-08-11
feat(cdk-mint-rpc): expose BDK wallet information

thesimplekid:
Add a read-only WalletService for balance, paginated transaction history, and revealed address balances. Keep BDK inspection behind a dedicated provider instead of extending MintPayment, and wire it through mintd only when BDK is configured.

Add matching mint CLI commands and tests for pagination, provider availability, wallet accounting, and mintd wiring.

@orangeshyguy21 

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-bdk/src/wallet_info.rs
+++ b/crates/cdk-bdk/src/wallet_info.rs
@@ -1,0 +1,234 @@
+//! Read-only wallet information for operator interfaces.
+
+use std::collections::HashMap;
+
+use bdk_wallet::bitcoin::Address;
+use bdk_wallet::chain::ChainPosition;
+use bdk_wallet::KeychainKind;
+
+use crate::{CdkBdk, Error};
+
+/// BDK wallet balance split by confirmation status.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WalletBalance {
+    /// Bitcoin network used by the wallet.
+    pub network: String,
+    /// Height of the wallet's latest local chain checkpoint.
+    pub synced_height: u32,
+    /// Confirmed, spendable balance in satoshis.
+    pub confirmed_sat: u64,
+    /// Unconfirmed wallet-created outputs in satoshis.
+    pub trusted_pending_sat: u64,
+    /// Unconfirmed externally-created outputs in satoshis.
+    pub untrusted_pending_sat: u64,
+    /// Immature coinbase outputs in satoshis.
+    pub immature_sat: u64,
+    /// Confirmed plus trusted-pending balance in satoshis.
+    pub trusted_spendable_sat: u64,
+    /// Total wallet balance in satoshis.
+    pub total_sat: u64,
+}
+
+/// A transaction relevant to the BDK wallet.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WalletTransaction {
+    /// Transaction ID.
+    pub txid: String,
+    /// Value received by wallet scripts in satoshis.
+    pub received_sat: u64,
+    /// Value spent from wallet inputs in satoshis.
+    pub sent_sat: u64,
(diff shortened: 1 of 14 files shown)


crates/cdk-bdk/src/wallet_info.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Returns relevant wallet transactions, newest first.
+    pub async fn wallet_transactions(
+        &self,
+        offset: usize,
+        limit: usize,
+    ) -> Result<WalletPage<WalletTransaction>, Error> {
+        let wallet_with_db = self.wallet_with_db.lock().await;
+        let wallet = &wallet_with_db.wallet;
+        let transactions = wallet
+            .transactions_sort_by(|left, right| right.chain_position.cmp(&left.chain_position));
```
orangeshyguy21: Transactions confirmed in the same block sort as equal here.

nondeterminism could lead to some unexpected results like:
* order could change between calls.
* a page boundary could cause the same tx to appear on two pages, or neither.

Worth thinking about ways to break the tie deterministically

crates/cdk-bdk/src/wallet_info.rs
```diff
@@ -0,0 +1,215 @@
...
+                .ok_or_else(|| Error::Wallet("Address balance overflow".to_string()))?;
+            if output.chain_position.is_confirmed() {
+                entry.1 = entry
+                    .1
+                    .checked_add(output.txout.value.to_sat())
+                    .ok_or_else(|| Error::Wallet("Address balance overflow".to_string()))?;
+            }
+        }
+
+        let mut addresses = Vec::new();
```
orangeshyguy21: There might be some wasted work here
Could the offset and limit be applied before this loop?

crates/cdk-mintd/src/lib.rs
```diff
@@ -1079,17 +1213,22 @@ fn configure_fake_wallet_keyset_rotations_once(
...
-) -> Result<MintBuilder> {
+) -> Result<(MintBuilder, ConfiguredWalletInfoProvider)> {
     use config::OnchainBackend;
     #[cfg(feature = "bdk")]
     use setup::OnchainBackendSetup;
 
+    #[cfg(all(feature = "management-rpc", feature = "bdk"))]
+    let mut wallet_info_provider = no_wallet_info_provider();
+    #[cfg(not(all(feature = "management-rpc", feature = "bdk")))]
+    let wallet_info_provider = no_wallet_info_provider();
```
orangeshyguy21, line 1394: clanker seems to think that this fails CI when the `bdk` feature is off.

crates/cdk-mint-rpc/src/bin/mint_rpc_cli.rs
```diff
@@ -247,6 +266,87 @@ async fn main() -> Result<()> {
             let mut keyset_client = KeysetServiceClient::with_interceptor(channel, interceptor);
             subcommands::rotate_next_keyset(&mut keyset_client, &sub_command_args).await?;
         }
+        Commands::GetWalletBalance => {
```
orangeshyguy21, line 259: These inline commands depart from the current conventions for existing cli commands that puts them as unique modules under `mint_rpc_cli/subcommands/`

orangeshyguy21 (commented):
Left a few comments inline.

Initially I was thinking about how the introduction of wallet methods for say, ldk-node, would change things, but ultimately those future decisions don't need to be pulled forward.

We can just use versioning should any breaking change be necessary at all.

thesimplekid:
Thanks for the review @orangeshyguy21. All should be resolved

thesimplekid:
@cdk-bot you up buddy?

PR #2333, opened 2026-08-12
Add advisory locks to the mint transaction traits

### LINKED ISSUE #2332: Add advisory locking mechanism in the mintdb trait
Introduce an advisory locking mechanism, along side the existing row level locking that are implemented in the SQL common trait.

This locking was introduced in  https://github.com/cashubtc/cdk/pull/2273 (https://github.com/cashubtc/cdk/pull/2273/changes#diff-3f9414ef43318f5deeda78597a33c856517dac90a95ef068c74caa626ac1a860R277-R284) but it will be more useful abstracted away

crodas:
### Description

Fixes #2332

Row locks only protect rows that already exist, so operations racing over rows they are about to insert had nothing serializing them. Two swaps sharing proofs, or two batch mints naming the same quotes in a different order, could deadlock against each other.

Transactions can now take a named advisory lock, held until the transaction ends. Melt, mint and swap take theirs before any row lock, quotes before proofs, each group sorted. The keyset lock that was hand-rolled in the SQL keys module goes through the same path.

Postgres backs this with pg_advisory_xact_lock; SQLite is a no-op, since BEGIN IMMEDIATE already serializes writers.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/database/mint/mod.rs
+++ b/crates/cdk-common/src/database/mint/mod.rs
@@ -123,6 +123,26 @@ pub struct LockedMeltQuotes {
     pub all_related: Vec<Acquired<MeltQuote>>,
 }
 
+/// The advisory lock serializing every keyset transaction, which is what makes
+/// a rotation and a peer's reload mutually exclusive. The backend takes it as
+/// the transaction begins, so callers never reach for it themselves.
+///
+/// One global key rather than per-unit: rotations are rare, and a single key
+/// keeps the keyset reads a signatory does during reload consistent with each
+/// other without per-unit lock bookkeeping.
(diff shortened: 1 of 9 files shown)


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

thesimplekid:
I think this PR uses too many locks especially for the proof operations I don't think they are needed however for the quote operations I think there is value so I took it here https://github.com/cashubtc/cdk/pull/2376.

crodas:
Closing in favor of #2376

PR #2334, opened 2026-08-12
Pace requests by destination host, not by wallet

### LINKED ISSUE #2313: Charge only mint-origin HTTP requests to the wallet rate-limit budget

## Problem

`RateLimitedTransport` currently consumes budget for every HTTP URL passed through it. The wallet also uses its transport for requests such as LNURL callbacks and OIDC discovery or token endpoints, which may target hosts unrelated to the mint. Those requests should not consume or be delayed by the mint's budget.

## Acceptance criteria

- Only requests matching the wallet's mint origin consume the mint rate-limit budget.
- LNURL and external OIDC requests bypass that budget.
- Main mint and blind-auth requests continue sharing the same budget.
- Tests cover matching and non-matching schemes, hosts, and effective ports.
- Redirect behavior is documented or tested.

crodas:
### Description

Fixes #2313

RateLimitedTransport held one fixed bucket and paced every request through it, whatever the target. A wallet's transport also carries LNURL fetches and OIDC discovery, token, and JWKS calls, so those burned the mint's budget and were themselves paced by it. A request cap belongs to the host being called, not to the wallet doing the calling.

RateLimiterManager is now what gets injected into the wallet's HTTP clients, and the transport resolves a bucket from each request URL's origin. Wallets at one mint still share that mint's budget; a third-party host gets its own, shared by whoever calls it.

Runtime toggles move to the manager, so disabling rate limiting stops all pacing rather than just the mint's, and is repository-wide for a repository-built wallet. Persistence is unchanged in shape: every origin contacted keeps its budget under the same KV namespace.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-common/src/rate_limit.rs
+++ b/crates/cdk-common/src/rate_limit.rs
@@ -1,14 +1,20 @@
 //! Client-side request rate limiting for the wallet.
 //!
-//! The wallet paces its outbound HTTP requests to a mint so it stays under the
-//! mint's server-side request cap without the caller having to think about it.
-//! Pacing uses the Generic Cell Rate Algorithm (GCRA): the bucket tracks a
-//! single theoretical arrival time (TAT) rather than a refilling token count.
+//! The wallet paces its outbound HTTP requests so it stays under a server's
+//! request cap without the caller having to think about it. Pacing uses the
+//! Generic Cell Rate Algorithm (GCRA): the bucket tracks a single theoretical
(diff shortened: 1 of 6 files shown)


crodas:
@cdk-bot review

crodas:
@cdk-bot review

thesimplekid (approved):

PR #2338, opened 2026-08-13
fix(mint): recover incomplete melt sagas safely

thesimplekid:
### What changed

- Serialize live melt dispatch with payment events and explicit quote checks on
  the same mint instance.
- Recover `PaymentAttempted` sagas by querying the payment backend and trusting
  its fresh terminal status.
- Derive missing recovery identifiers from the BOLT11 payment hash or the mint
  quote ID for BOLT12, on-chain, and custom payments.
- Finalize `Paid` and internally settled melts; compensate fresh `Unpaid` or
  `Failed` results; keep `Pending` and `Unknown` reserved.
- Persist backend lookup IDs returned for pending payments so later checks and
  restarts can continue recovery.
- Use acquired saga rows for recovery mutations and reject stale rollback after
  finalization begins.
- Rely on backend payment streams and explicit quote checks instead of periodic
  five-second status polling.

### Why

Recovery previously had paths that could return proofs after value had already
moved. The earlier follow-up design prevented that with a new
`PaymentPending` state, but it also left `PaymentAttempted + Unpaid` unresolved
indefinitely.

This version uses the payment backend contract directly: `Unpaid` means the
payment cannot settle without a new dispatch, while uncertainty is represented
as `Pending` or `Unknown`. A process-local per-quote guard prevents a live quote
check from observing `Unpaid` before the active task has dispatched.

### Scope

This provides single-instance crash and restart recovery. It deliberately does
not add an idempotency key, a new saga state, advisory locks, database fencing,
or cross-instance recovery coordination. Multi-instance support remains
separate follow-up work.

This replaces [#2306](https://github.com/cashubtc/cdk/pull/2306) and supersedes
[#2339](https://github.com/cashubtc/cdk/pull/2339).

### Stack

This is PR 2 of 2. Merge in this order:

1. [#2337 — acquired saga mutations](https://github.com/cashubtc/cdk/pull/2337)
2. [#2338 — single-instance melt recovery](https://github.com/cashubtc/cdk/pull/2338) (this PR)

This PR targets the branch from #2337 so its diff contains only the recovery
layer. After #2337 merges, retarget this PR to `main`.

### Validation

- `cargo test -p cdk --lib` — 605 passed
- `cargo check -p cdk --all-targets`
- `cargo clippy -p cdk --all-targets -- -D warnings`


thesimplekid:
@cdk-bot review

thesimplekid:
076a9ff09bbbc48c99e17532a965068b0ce1a0c1 Merged

thesimplekid:
Merged https://github.com/cashubtc/cdk/commit/076a9ff09bbbc48c99e17532a965068b0ce1a0c1

PR #2339, opened 2026-08-13
fix(mint): persist pending melt recovery handoffs

thesimplekid:
### What changed

- Add a durable `PaymentPending` melt saga state.
- Persist the backend lookup identifier together with the pending handoff.
- Keep `PaymentAttempted + Unpaid` pending because dispatch may not yet be
  visible to the backend.
- Allow `Unpaid` compensation only after the backend has acknowledged the
  payment through `PaymentPending`.
- Record pending acknowledgements discovered by live processing or restart
  recovery.
- Add focused state-transition and compensation regression tests.

### Why

`PaymentAttempted` proves only that dispatch may have started. A backend can
temporarily return `Unpaid` before it recognizes that request, so recovery
cannot safely return proofs from that observation alone.

`PaymentPending` records the backend acknowledgement that makes a later
terminal `Unpaid` result safe to compensate, without adding an idempotency key
or periodic polling.

### Stack

This is PR 3 of 3. Merge in this order:

1. [#2337 — acquired saga mutations](https://github.com/cashubtc/cdk/pull/2337)
2. [#2338 — core melt recovery hardening](https://github.com/cashubtc/cdk/pull/2338)
3. [#2339 — durable pending-payment handoffs](https://github.com/cashubtc/cdk/pull/2339) (this PR)

This PR targets the branch from #2338 so its diff contains only the durable
handoff layer. After #2338 merges, retarget this PR to `main`.

### Validation

- `cargo test -p cdk --lib` — 604 tests
- `cargo test -p cdk-common --lib` — 130 tests
- `cargo test -p cdk-sqlite --lib` — 138 tests
- `cargo clippy -p cdk -p cdk-common -p cdk-sqlite --all-targets -- -D warnings`

### DIFF
--- a/crates/cdk-common/src/mint.rs
+++ b/crates/cdk-common/src/mint.rs
@@ -164,6 +164,8 @@ pub enum MeltSagaState {
     SetupComplete,
     /// Payment attempted to Lightning network (may or may not have succeeded)
     PaymentAttempted,
+    /// Backend acknowledged an in-flight payment and recovery may poll it
+    PaymentPending,
     /// TX1 committed (proofs Spent, quote Paid) - change signing + cleanup pending
     Finalizing,
 }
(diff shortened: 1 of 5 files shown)


thesimplekid:
@cdk-bot review

thesimplekid:
Superseded by #2338. The release-scoped implementation now trusts a fresh backend `Unpaid` result after serializing live dispatch with same-instance quote checks. This avoids adding `PaymentPending` and prevents definitively unpaid melts from remaining stuck. Cross-instance coordination remains deferred.

Issue #2335, opened 2026-08-13
wallet: mint saga aborts with ConcurrentUpdate after the mint has issued signatures when a quote's secret_key equals seed[..32] (npubcash legacy scrub invalidates the saga's quote snapshot)


kaloudis:
## Summary

With the `npubcash` feature enabled (as in the released cdk-ffi binaries), `Wallet::mint` deterministically fails with `Error::Cdk(50000, "Concurrent update detected")` for any mint quote whose stored `secret_key` is byte-identical to the first 32 bytes of the wallet seed. The failure occurs **after** the mint has already issued the blind signatures, so the quote is consumed on the mint side (state `ISSUED`) while no proofs, transaction, or balance are recorded locally. The funds are only recoverable through a NUT-13 restore scan.

The trigger is not limited to npub.cash quotes. Any application that derives an external quote's NUT-20 key as `seed[0..32]` collides with `is_legacy_npubcash_secret_key`, which treats byte equality with the seed prefix as proof that the key came from cdk's legacy npub.cash scheme. ZEUS hit this on every ZEUS Pay lightning address redemption after upgrading from cdk-ffi 0.14.2 to 0.17.3: ZEUS Pay has locked quotes to the wallet's seed prefix since before cdk had npubcash support, so every redemption stranded the funds until restore.

Genuine legacy npub.cash quotes that carry the stored key should hit the identical crash, since the scrub runs inside the mint saga itself.

## Versions

- cdk 0.17.3 (via cdk-kotlin 0.17.3 and cdk-swift 0.17.3, both of which build cdk-ffi with `npubcash` enabled)
- Reproduced on Android (JVM bindings) and applicable to iOS (same Rust path)
- Wallet store: `WalletSqliteDatabase` (passed as `WalletStore::Custom`; the built-in sqlite store takes the same code path)

## Reproduction

No npub.cash account is needed; only a quote whose `secret_key` equals the seed prefix.

1. Create a `WalletRepository` (or `Wallet`) from a mnemonic. Let `seed` be the BIP39 seed (empty passphrase); let `k = SecretKey::from_slice(&seed[..32])`.
2. At any real mint, create a bolt11 mint quote locked to `k.public_key()` (NUT-20 `pubkey`), pay its invoice out of band, so the quote is `PAID`.
3. Insert the quote into the wallet database the way an external-quote consumer does:

```rust
db.add_mint_quote(MintQuote {
    id: quote_id,
    state: MintQuoteState::Paid,
[...]

PR #2342, opened 2026-08-13
fix(wallet): refresh mint quote after signing-key migration

thesimplekid:
## Summary

- reload mint quotes after signing-key resolution during single-quote preparation
- refresh every quote snapshot during batch preparation
- cover both paths with optimistic-lock regression tests

## Why

Signing-key resolution can migrate legacy quote metadata and advance the quote row version. The mint saga previously retained the older snapshot, then attempted a versioned write after the mint had already issued signatures. That deterministic stale write failed with `ConcurrentUpdate`, leaving issued funds absent from the local wallet.

Refreshing the quote before the external mint request ensures subsequent writes use the persisted version.

This is the first change in a two-PR stack addressing #2335. The follow-up, #2343, removes the ambiguous seed-prefix provenance heuristic.

## Validation

- `cargo clippy -p cdk --features npubcash --all-targets -- -D warnings`
- `cargo test -p cdk --features npubcash` (603 unit tests and 15 doc tests passed; 9 doc tests ignored)

### DIFF
--- a/crates/cdk/src/wallet/issue/saga/mod.rs
+++ b/crates/cdk/src/wallet/issue/saga/mod.rs
@@ -406,6 +406,16 @@ impl<'a> MintSaga<'a, Initial> {
             return Err(Error::SignatureMissingOrInvalid);
         }
 
+        // Resolving a legacy NpubCash signing key can migrate the quote in the
+        // localstore. Reload it so the prepared saga carries the current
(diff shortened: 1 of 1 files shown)


thesimplekid:
Superseded by #2344, recreated from an upstream branch so it can participate in GitHub native stacked pull requests.

PR #2345, opened 2026-08-13
fix(wallet): migrate legacy npubcash keys by provenance

### LINKED ISSUE #2335: wallet: mint saga aborts with ConcurrentUpdate after the mint has issued signatures when a quote's secret_key equals seed[..32] (npubcash legacy scrub invalidates the saga's quote snapshot)

thesimplekid:
## Summary

- resolve NpubCash signing keys only from explicit per-quote provenance markers
- preserve unmarked externally supplied keys, including historical ZEUS Pay seed-prefix keys
- migrate genuine legacy NpubCash quotes only after the legacy NpubCash account establishes provenance
- merge imported server state without regressing local quote state, accounting, reservations, or optimistic-lock versions
- rerun the corrected migration once for wallets that recorded the earlier migration marker

## Why

Byte equality with `seed[..32]` cannot distinguish a legacy NpubCash quote from an independently created external quote. ZEUS Pay used the same derivation before CDK added NpubCash support, so the old heuristic claimed ZEUS quotes and mutated their database rows during minting.

Key lookup is now side-effect free. Legacy key removal happens only in the import path, where provenance is established by fetching the quote through the authenticated legacy NpubCash account.

This preserves existing unclaimed ZEUS quotes, supports ZEUS workaround rows carrying the explicit legacy marker, and safely migrates genuine legacy NpubCash data. Quotes already issued without locally recorded proofs still require NUT-13 restore.

Closes #2335.

## Stack

This is the top change in a two-PR native stack above #2344. Its base is `fix/wallet-mint-saga-quote-version`, so this PR contains only the provenance change.

## Validation

- `cargo clippy -p cdk --features npubcash --all-targets -- -D warnings`
- `cargo test -p cdk --features npubcash` (603 unit tests and 15 doc tests passed; 9 doc tests ignored)

### DIFF
--- a/crates/cdk/src/wallet/issue/mod.rs
+++ b/crates/cdk/src/wallet/issue/mod.rs
@@ -161,35 +161,18 @@ impl Wallet {
 
     /// Resolve the NUT-20 signing key for a mint quote
     ///
-    /// Returns the quote's stored key if present. NpubCash quotes do not
-    /// persist a key; for those the key is re-derived from the wallet seed.
+    /// Re-derives the key for explicitly marked NpubCash quotes. For all other
+    /// quotes, returns the stored key without modifying wallet state.
     pub(crate) async fn mint_quote_signing_key(
         &self,
         quote: &MintQuote,
     ) -> Result<Option<SecretKey>, Error> {
         #[cfg(feature = "npubcash")]
-        if let Some(secret_key) = &quote.secret_key {
-            if self.is_legacy_npubcash_secret_key(secret_key) {
-                self.scrub_legacy_npubcash_quote(quote).await?;
-                return Ok(Some(self.npubcash_quote_secret_key(
-                    crate::wallet::npubcash::NpubCashQuoteKey::LegacySeedPrefix,
-                )?));
-            }
-
-            return Ok(Some(secret_key.clone()));
-        }
-
-        #[cfg(not(feature = "npubcash"))]
-        if let Some(secret_key) = &quote.secret_key {
-            return Ok(Some(secret_key.clone()));
-        }
-
-        #[cfg(feature = "npubcash")]
         if let Some(key) = self.npubcash_quote_key(&quote.id).await? {
             return Ok(Some(self.npubcash_quote_secret_key(key)?));
         }
 
-        Ok(None)
+        Ok(quote.secret_key.clone())
     }
 
     /// Create a mint quote for the given payment method and amount
--- a/crates/cdk/src/wallet/issue/saga/mod.rs
+++ b/crates/cdk/src/wallet/issue/saga/mod.rs
@@ -406,9 +406,9 @@ impl<'a> MintSaga<'a, Initial> {
             return Err(Error::SignatureMissingOrInvalid);
         }
 
-        // Resolving a legacy NpubCash signing key can migrate the quote in the
-        // localstore. Reload it so the prepared saga carries the current
-        // optimistic-lock version into the post-mint persistence step.
+        // Reload the quote after signing-key resolution so the prepared saga
+        // carries the latest optimistic-lock version into the post-mint
+        // persistence step.
         let quote_info = self
             .wallet
             .localstore
@@ -727,9 +727,8 @@ impl<'a> MintSaga<'a, Initial> {
             }
         }
 
-        // Signing-key resolution can migrate legacy NpubCash quotes. Refresh
-        // every snapshot before retaining it in the prepared batch so later
-        // versioned writes do not use pre-migration versions.
+        // Refresh every snapshot after signing-key resolution so later
+        // versioned writes use the latest persisted versions.
         for quote in &mut quote_infos {
             *quote = self
                 .wallet
--- a/crates/cdk/src/wallet/npubcash.rs
+++ b/crates/cdk/src/wallet/npubcash.rs
@@ -7,14 +7,14 @@ use std::sync::Arc;
 
 use bitcoin::bip32::{ChildNumber, DerivationPath, Xpriv};
 use bitcoin::Network;
-use cdk_common::SECP256K1;
+use cdk_common::{database, SECP256K1};
 use cdk_npubcash::{JwtAuthProvider, NpubCashClient, Quote};
 use tracing::instrument;
 
 use crate::error::Error;
 use crate::nuts::SecretKey;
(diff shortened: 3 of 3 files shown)


thesimplekid:
@cdk-bot review

PR #2348, opened 2026-08-13
fix(nut17)!: harden WebSocket notification handling

thesimplekid:
Keep WebSocket streams alive when a notification contains an unsupported or malformed nested value. Correlate shared subscribe and unsubscribe acknowledgements through JSON-RPC request IDs and generate UUID v7 subscription IDs.

BREAKING CHANGE: WsResponseResult is now the shared acknowledgement struct rather than an enum with indistinguishable subscribe and unsubscribe variants.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/subscription.rs
+++ b/crates/cdk/src/wallet/subscription.rs
@@ -202,7 +202,7 @@ impl SubscriptionClient {
             .ok()
     }
 
-    fn get_unsub_request(&self, sub_id: String) -> Option<String> {
+    fn get_unsub_request(&self, sub_id: String) -> Option<(usize, String)> {
         let request: WsRequest<_> = (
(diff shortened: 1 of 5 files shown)


crates/cdk/src/wallet/subscription.rs
```diff
@@ -631,6 +681,45 @@ async fn stream_client(
...
+        }
+    }
+}
+
+fn decode_notification_payload_for_stream(
+    kind: &Kind,
+    sub_id: &str,
+    payload: serde_json::Value,
+) -> Option<NotificationPayload> {
+    match decode_notification_payload(kind, payload) {
```
crodas: I think the content of `decode_notification_payload` can be moved here.

crodas (approved):

PR #2352, opened 2026-08-13
feat(mint): advertise the maximum request array length

robwoodgate:
Implements cashubtc/nuts#425.

Adds the optional [NUT-06](https://github.com/cashubtc/nuts/blob/main/06.md) `max_array_length` to `/v1/info` so wallets can size their batches up front instead of discovering the cap by getting an error partway through a request.

`Mint::mint_info` serves the lower of `max_inputs` and `max_outputs`, so a wallet that respects the figure trips neither. It is derived at read time rather than stored in the config, so it cannot drift from what `check_state`, `restore` and `verification` actually enforce.

### DIFF
--- a/crates/cashu/src/nuts/nut06.rs
+++ b/crates/cashu/src/nuts/nut06.rs
@@ -107,6 +107,9 @@ pub struct MintInfo {
     /// terms of url service of the mint
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tos_url: Option<String>,
+    /// max length the mint accepts for any array in a request
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub max_array_length: Option<u64>,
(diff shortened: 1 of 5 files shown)


crodas (changes requested):
I don't see where this new info can be configured and enforced in the mint. Will this covered by a following PR?

robwoodgate:
> I don't see where this new info can be configured and enforced in the mint. Will this covered by a following PR?

It uses the existing `max_inputs` and `max_outputs` settings, which is used for array length caps, eg in restore

crodas (dismissed):

robwoodgate:
**EDIT**: Updated NUT-06 key name to: `max_array_length ` following NUT feedback

thesimplekid (approved):

PR #2355, opened 2026-08-14
feat(cdk-mint-rpc): add PaymentMethodService

orangeshyguy21:
### Description

Third PR of the mint RPC service split ([#2239](https://github.com/cashubtc/cdk/pull/2239) KeysetService, [#2294](https://github.com/cashubtc/cdk/pull/2294) QuoteService). Adds `PaymentMethodService` (`cdk_mint_payment_method_v1`) next to the legacy `CdkMint` service, on the same port.

Three RPCs:

- `UpdateMintMethod` / `UpdateMeltMethod`  (the old `UpdateNut04`/`UpdateNut05` ). Two behavior improvements: the response returns the settings actually in effect and bolt11 options aimed at a non-bolt11 method are rejected instead of silently stored.
- `UpdateDisabled`  turns minting/melting on or off for the whole mint. In the old RPCs this rode along as a `disabled` field, which looked per-method but actually flipped the whole section. Now it's its own RPC, shaped like `UpdateQuoteTtl`: both flags optional.

Legacy handlers and the new service call the same shared helpers, so the old RPCs behave exactly as before. CLI commands are renamed with the old names kept as aliases.

-----

### Notes to the reviewers

- **One intentional CLI break**: `--disabled` is gone from the method commands (aliases included). It's now the `update-disabled` command. Old scripts fail loudly at argument parsing.
- The options messages are named `Bolt11MintMethodOptions`/`Bolt11MeltMethodOptions` because that's all they model currently.
- Touches some of the same files as #2330 (build.rs, service registration, CLI).
- Two known issues were left alone because fixing them changes legacy behavior: the legacy RPCs still accept bolt11 options on non-bolt11 methods, and nothing validates `min_amount <= max_amount`. Issues should be documented and fixed independently.

-----

#### CHANGED

- cdk-mint-rpc: CLI `update-nut04`/`update-nut05` renamed to `update-mint-method`/`update-melt-method` (old names still work as aliases); `--disabled` moved to the new `update-disabled` command.

#### ADDED

- cdk-mint-rpc: `PaymentMethodService` (`cdk_mint_payment_method_v1`) with `UpdateMintMethod`, `UpdateMeltMethod`, and `UpdateDisabled`, served alongside the legacy `CdkMint` service.

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2356, opened 2026-08-14
feat(cdk-nostr): consolidate Nostr crates 

asmogo:
and add key/NIP-44/NIP-17 inbox primitives

Merge cdk-nwc and cdk-npubcash into a single cdk-nostr crate and add the secp256k1-based primitives foreign-language wallets need so they do not have to bundle a separate secp256k1 library:

- keys: secret key generation/parsing, x-only public key derivation
- nip44: NIP-44 v2 encrypt/decrypt (tested against the official vectors)
- inbox: standing NIP-17 listener that subscribes relays for kind-1059 gift wraps addressed to an identity, unwraps them (NIP-59), and delivers rumors to a NostrInboxListener callback
- nwc / npubcash: the former crates, moved as feature-gated modules

cdk-ffi exposes the new primitives plus the inbox listener (UniFFI callback interface) behind a new default-on 'nostr' feature, and drops its direct nostr-sdk dependency in favor of cdk-nostr's re-export.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


asmogo:
@cdk-bot review

thesimplekid:
@asmo Do we want this in 0.18? I think if we push 0.18 to this Friday we could do it?

asmogo:
I will focus on finishing this branch today (including the changes for npub.cash batch minting that we have been talking about)
We can also move this to the next milestone if I can't make it.

asmogo:
@cdk-bot review

justfile
```diff
@@ -725,7 +734,7 @@ release m="":
     "-p cdk-prometheus"
     "-p cdk-http-client"
     "-p cdk-common"
-    "-p cdk-npubcash"
+    "-p cdk-nostr"
```
thesimplekid, line 764: The `cdk-nwc` crate is also removed by this consolidation, but the release list still contains `-p cdk-nwc` a few lines below. `just release` will therefore fail after it has already published the earlier crates. Should we remove that stale entry as part of the consolidation?

crates/cdk/src/wallet/issue/mod.rs
```diff
@@ -427,50 +460,164 @@ impl Wallet {
...
+                let previous_amounts = mintable
+                    .iter()
+                    .map(|quote| (quote.id.clone(), quote.amount_issued))
+                    .collect::<HashMap<_, _>>();
+                let mintable_ids = mintable
+                    .iter()
+                    .map(|quote| quote.id.as_str())
+                    .collect::<Vec<_>>();
+
                 if let Err(err) = self
```
thesimplekid, line 575: When a mint advertises NUT-29 without `max_batch_size`, the wallet uses 100, but the mint may enforce a smaller implementation-defined limit and return error 11017. This path treats every batch mint error as an ambiguous write: it skips the chunk and eventually returns success, so `claim_npubcash_quotes()` can report success while paid quotes remain unclaimed. Could we distinguish the definitive batch-size error and retry with smaller batches or individual minting, while continuing to leave genuinely ambiguous writes to saga recovery?

crates/cdk/src/wallet/npubcash.rs
```diff
@@ -139,6 +141,21 @@ impl Wallet {
             tracing::warn!("Failed to import legacy NpubCash quotes: {}", e);
         }
 
+        // Best-effort: lock new quotes to this wallet's NpubCash npub so only
+        // this wallet can mint them. Servers/mints without NUT-20 support
+        // reject this; unlocked quotes remain claimable either way.
+        match client.set_quote_locking(true).await {
```
thesimplekid: If locked npub.cash quotes are an invariant, should a failure here abort initialization instead of logging and returning `Ok(())`? The client is published into `npubcash_client` before locking is enabled, so callers can observe an enabled integration even when the server rejected locking. Could we enable and confirm `lock_quote == true` before exposing the client, and return an error when that cannot be established?

crates/cdk/src/wallet/npubcash.rs
```diff
@@ -313,14 +431,56 @@ impl Wallet {
...
+    /// Already-created quotes keep their original lock state; unlocked quotes
+    /// are still claimable (the wallet detects them and mints unsigned).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if NpubCash is not enabled or the update fails
+    #[instrument(skip(self))]
+    pub async fn set_npubcash_quote_locking(
+        &self,
+        lock_quotes: bool,
```
thesimplekid: If CDK only supports newly created locked npub.cash quotes, why expose a boolean that permits callers to disable that invariant? This is also exposed through FFI, allowing new unlocked quotes to be created after successful initialization. Could this setter be removed or made internal while retaining support for claiming already-existing unlocked legacy quotes?

thesimplekid (commented):
A few inline findings from the npub.cash and crate-consolidation review.

crates/cdk/src/wallet/npubcash.rs
```diff
@@ -794,4 +975,105 @@ mod tests {
...
+        let err = wallet
+            .sync_missing_npubcash_quotes()
+            .await
+            .expect_err("sync must fail when NpubCash is not enabled");
+
+        assert!(matches!(err, Error::Custom(_)));
+    }
+
+    #[tokio::test]
+    async fn unlocked_npubcash_quote_never_carries_provenance() {
```
thesimplekid, line 1024: These tests cover the in-memory provenance behavior, but not the actual upgrade boundary. Could we add an upgrade test that persists a pre-migration wallet, reopens the same database and seed, imports and claims both locked and existing legacy-unlocked quotes, and verifies that initialization fails when locking cannot be enabled for new quotes?

asmogo:
@cdk-bot review

thesimplekid:
https://github.com/cashubtc/cdk/pull/2411 puts home back to should be able to remove this last commit.

crates/cdk/src/wallet/npubcash.rs
```diff
@@ -279,6 +304,107 @@ impl Wallet {
...
+    /// Returns the total amount minted across all claimed quotes.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if NpubCash is not enabled or the sync fails
+    #[instrument(skip(self))]
+    pub async fn claim_npubcash_quotes(&self) -> Result<Amount, Error> {
+        self.sync_npubcash_quotes().await?;
+        self.sync_missing_npubcash_quotes().await?;
+        self.mint_unissued_quotes().await
```
thesimplekid: Should this claim only the quote IDs returned by npub.cash? `mint_unissued_quotes()` scans every locally stored unissued quote for this mint and unit, so this call can also mint unrelated quotes created through normal wallet flows. Could we retain the npub.cash quote IDs from reconciliation and pass them to a targeted mint operation? A regression test with one npub.cash quote and one unrelated paid quote would cover this boundary.

crates/cdk-nostr/src/npubcash/client.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// Returns an error if the API request fails or authentication fails.
+    /// Returns `UnsupportedEndpoint` if the server doesn't support this feature.
+    #[instrument(skip(self, mint_url))]
+    pub async fn set_mint_url(
+        &self,
+        mint_url: impl Into<String>,
+    ) -> Result<crate::npubcash::types::UserResponse> {
+        use serde::Serialize;
+
+        const MINT_URL_PATH: &str = "/api/v2/user/mint";
```
thesimplekid, line 221: Should this use the current npub.cash settings API? The documented contract is `PUT /api/v2/settings/mint` with `{"mintUrl": ...}`, but this sends `PATCH /api/v2/user/mint` with `{"mint_url": ...}`. Because `enable_npubcash()` only logs this failure, the wallet can appear enabled while npub.cash continues using its previous/default mint. Could we implement the documented request—optionally with an explicit legacy fallback—and add a request-shape test?

https://docs.npub.cash/docs/api/endpoints.html#update-mint-url

asmogo: docs are wrong.

asmogo:
> #2411 puts home back to should be able to remove this last commit.

I kept the lazy loading for the configuration and moved back to calling `home::home_dir()` instead.

thesimplekid (approved):

PR #2357, opened 2026-08-14
Let callers choose request pacing when a wallet is built

### LINKED ISSUE #2315: Complete wallet rate-limit API and FFI configuration parity

## Problem

The concrete FFI wallet exposes a runtime `set_rate_limit()` method, but FFI `WalletConfig` cannot select default, disabled, or custom pacing during construction. The shared `cdk_common::wallet::Wallet` abstraction and its native/FFI implementations also do not expose rate-limit configuration.

## Acceptance criteria

- Decide whether rate-limit configuration belongs in the shared wallet trait; implement it or document why it is transport-specific.
- Add default/disabled/custom rate-limit configuration to FFI `WalletConfig`.
- Keep native and FFI wallet-trait implementations synchronized when the shared trait is extended.
- Provide equivalent behavior for directly constructed and repository-provided FFI wallets where supported.
- Add binding-level tests for default, disabled, invalid, and custom configurations.

crodas:
### Description

Fixes #2315

The FFI wallet could only change pacing after construction, so every wallet started on the built-in default and burst against a mint that wanted less before the host app could intervene. The shared wallet trait had no rate-limit surface at all, so a trait-only consumer could not reach it.

Rate limiting belongs in the trait next to set_metadata_cache_ttl: both are client policy rather than transport plumbing, and RateLimitConfig already lives in cdk-common beside the trait. The trait takes Option<RateLimitConfig>, where None is off.

FFI WalletConfig and the new WalletRepositoryConfig select default, disabled or custom pacing at construction. Disabled builds the limiter and leaves it off rather than dropping it, so it can be turned back on; a repository shares one limiter, so that is the only disable it could offer, and the two paths now behave alike. WalletBuilder::without_rate_limiting keeps dropping the limiter for callers who want no overhead.

is_rate_limited reports whether pacing is actually on. Without it neither a test nor a host UI could tell default from disabled, and a custom transport leaves the setters silently doing nothing.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/bindings/dart/test/wallet_test.dart
+++ b/bindings/dart/test/wallet_test.dart
@@ -64,6 +64,71 @@ void main() {
     }
   });
 
+  Wallet rateLimitedWallet(RateLimit? rateLimit) => Wallet(
+    mintUrl: 'https://mint.example.com',
+    unit: SatCurrencyUnit(),
+    mnemonic: generateMnemonic(),
+    store: SqliteWalletStore(':memory:'),
+    config: WalletConfig(targetProofCount: null, rateLimit: rateLimit),
+  );
+
+  test('rate limit defaults to pacing', () {
+    // Omitting the field is the backwards-compatible spelling and must keep
+    // selecting the built-in default.
+    final omitted = Wallet(
+      mintUrl: 'https://mint.example.com',
+      unit: SatCurrencyUnit(),
+      mnemonic: generateMnemonic(),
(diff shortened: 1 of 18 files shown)


crodas:
@cdk-bot review

crodas:
@cdk-bot review

thesimplekid (approved):

Issue #2365, opened 2026-08-16
Coin control of BDK wallet


minibits-cash:
I believe that for privacy reasons we need coin control excluding sub-minimum UTXOs from input selection. Such sub-limit deposits are cheap way to introduce address clustering of melts.

PR #2367, opened 2026-08-17
feat(cdk-mint-rpc): expose transaction output quote details

### LINKED ISSUE #2366: Add output addresses to the WalletTransaction record - BDK
To make various operations on top of CDK's grpc BDK Wallet api easier (avoid the need of lookups to bitcoind) it would be great to add output addresses to the rpc transaction list.

thesimplekid:
Add ordered non-change outputs to wallet transaction records with each output's vout, address, amount, and associated quote ID kept together. Resolve incoming quotes from tracked addresses before confirmation and preserve outgoing quote associations after batched sends are finalized.

Closes: https://github.com/cashubtc/cdk/issues/2366

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-bdk/src/wallet_info.rs
+++ b/crates/cdk-bdk/src/wallet_info.rs
@@ -34,6 +34,10 @@ pub struct WalletBalance {
 pub struct WalletTransaction {
     /// Transaction ID.
     pub txid: String,
+    /// Inputs spent by this transaction in input order.
+    pub inputs: Vec<WalletTransactionInput>,
+    /// Non-change payment outputs in transaction output order.
+    pub outputs: Vec<WalletTransactionOutput>,
     /// Value received by wallet scripts in satoshis.
     pub received_sat: u64,
     /// Value spent from wallet inputs in satoshis.
@@ -48,6 +52,32 @@ pub struct WalletTransaction {
     pub confirmation_time: Option<u64>,
     /// First-seen timestamp, when known for an unconfirmed transaction.
     pub first_seen: Option<u64>,
+}
+
+/// An input spent by a wallet transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WalletTransactionInput {
+    /// Transaction ID containing the previous output.
+    pub txid: String,
+    /// Index of the previous output.
+    pub vout: u32,
+    /// Value of the previous output in satoshis, when known.
+    pub amount_sat: Option<u64>,
+    /// Address of the previous output, when it belongs to the wallet.
+    pub address: Option<String>,
+}
+
+/// A payment output belonging to a wallet transaction.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WalletTransactionOutput {
+    /// Transaction output index.
+    pub vout: u32,
+    /// Bitcoin address in network-specific display encoding.
+    pub address: String,
+    /// Output value in satoshis.
+    pub amount_sat: u64,
+    /// Quote associated with this output, when managed by the payment backend.
+    pub quote_id: Option<String>,
 }
 
 /// BDK keychain containing a revealed address.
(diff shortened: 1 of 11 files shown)


thesimplekid:
@cdk-bot review

cd @orangeshyguy21

minibits-cash:
Thank you - targets https://github.com/cashubtc/cdk/issues/2366 very well.

It lets Minibits Watchdog to attribute a deposit to its mint quote directly + per-output `amount_sat` helps to filter-out sub mint's minimal onchain mint amounts.

While you are at it, I identified another extension of transaction data with high analogy:
add WalletTransactionInput inputs alongside outputs.

Use case:
I see sub-limit deposits to the cdk onchain wallet already happening. Those might be harmless, owever they might be easily used for clustering analysis once they become inputs into mint melts. This is what my separate issue (https://github.com/cashubtc/cdk/issues/2365) has an intention to cover.

Adding them to transaction list allows to set any kind of monitoring on them in the meantime.

### Proposed shape

Additive, mirroring `WalletTransactionOutput`:

```proto
message WalletTransactionInput {
    // Previous outpoint spent by this input.
    string txid = 1;
    uint32 vout = 2;
    // Value of the previous output, when known to the wallet.
    optional uint64 amount_sat = 3;
    // Address of the previous output, when it belongs to the wallet.
    optional string address = 4;
}

message WalletTransaction {
    // ...
    // Inputs spent by this transaction, in input order.
    repeated WalletTransactionInput inputs = 10;
}
```

`txid` and `vout` alone would be sufficient for us — matching input outpoints against
known uncredited deposit txids is all the detection needs. The amount and address fields
are conveniences.

This should not require new persistence: the wallet's own transaction already carries
`input[].previous_output`, which is the same source `outputs` is derived from.

thesimplekid:
Added a commit that should address this https://github.com/cashubtc/cdk/pull/2367/commits/3ff6c4e2019219a73cf77194779c5f0a8dda324c

thesimplekid:
@cdk-bot review

crates/cdk-bdk/src/wallet_info.rs
```diff
@@ -109,6 +139,8 @@ impl CdkBdk {
         offset: usize,
         limit: usize,
     ) -> Result<WalletPage<WalletTransaction>, Error> {
+        let (output_quote_ids, receive_quote_ids) =
```
orangeshyguy21: Calls here load everything regardless of the page size requested, repeated on pagination.

Might be worth tracking at a minimum this known inefficiency. We'll likely have more in the same bucket as other get_ endpoints are introduced to the rpc.

crates/cdk-bdk/src/wallet_info.rs
```diff
@@ -133,6 +165,61 @@ impl CdkBdk {
...
+                        }
+                    })
+                    .collect();
+                let outputs = tx
+                    .output
+                    .iter()
+                    .enumerate()
+                    .filter(|(_, output)| {
+                        let is_wallet_output = wallet.is_mine(output.script_pubkey.clone());
+                        is_outgoing != is_wallet_output
```
orangeshyguy21, line 197: Inputs report outputs the wallet owns, outputs don't so a consolidation comes back with an empty output list, and change never shows on a normal payment.

Is that asymmetry deliberate?

orangeshyguy21 (commented):
I left two questions in line for your consideration: 

1) quote lookup costs 
2) how complete the output list should be

crates/cdk-bdk/src/wallet_info.rs
thesimplekid: Resolved in https://github.com/cashubtc/cdk/pull/2367/commits/c2fd5d72cd67e6510c97231ddbe64a8b2d314573

PR #2369, opened 2026-08-18
fix(cdk-bdk): reserve send intent quote ids atomically

thesimplekid:
create_or_retry_failed_send_intent checked the quote-id index with a read-then-write inside one transaction. On backends that do not serialize concurrent transactions, two callers could both observe a missing index entry and each persist an intent for the same quote id.

Add KVStoreTransaction::kv_write_if_absent, an insert-if-absent primitive implemented for the SQL backend as INSERT ... ON CONFLICT DO NOTHING with a rows-affected check, and use it to reserve the quote id in the same transaction as the intent record. The losing transaction now rolls back with DuplicateQuoteId. The retry-failed path is unchanged: an existing index entry still requeues the failed intent under its original id.

Cover the primitive with a concurrent-contender conformance test in the shared mint database test suite and a racing-store regression test in cdk-bdk.

Found by project Loupe

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-bdk/src/lib.rs
+++ b/crates/cdk-bdk/src/lib.rs
@@ -688,23 +688,34 @@ impl MintPayment for CdkBdk {
             _ => return Err(cdk_common::payment::Error::UnsupportedPaymentOption),
         };
 
-        let mut wallet_with_db = self.wallet_with_db.lock().await;
-        let address = wallet_with_db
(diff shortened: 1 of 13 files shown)


thesimplekid:
@cdk-bot review

PR #2376, opened 2026-08-18
fix(mint): coordinate melt dispatch across replicas with quote advisory locks

thesimplekid:
### Description

-----

Builds on the advisory-lock approach from #2333 (credit @crodas), scoped to melt quotes, and adds the saga coordination needed to run multiple mint replicas safely against one Postgres database.

Fixes three cross-replica races:

1. **Duplicate dispatch** — two replicas could pay the same melt quote concurrently. The dispatcher now takes a quote advisory lock, re-reads the saga under it, and aborts unless it is still `PaymentAttempted`.
2. **Refund while in flight** — a status check, failure event, or startup recovery could compensate (return proofs) while a payment was still settling. Reconciliation now happens under the same advisory lock with a fresh backend re-check, and a durable `PaymentPending` saga marker blocks compensation unless a definitive failure event arrives from the backend.
3. **Startup rollback racing live dispatch** — setup rollback is refused unless the saga is still `SetupComplete`.

Reconciliation paths use a non-blocking `pg_try_advisory_xact_lock`: a contended lock proves a live dispatch owns the quote, so reconcilers fail closed (leave pending) instead of pinning a pool connection across the holder's network I/O. Only dispatchers ever wait, at most one per quote.

All outcomes fail closed: worst case is a quote stuck `Pending` until a later poll/event/startup recovery resolves it — never a double payment or a double refund.

SQLite is a no-op for the advisory lock (single-process deployments are already serialized by the process-local quote guard), and behavior there is unchanged. **Multi-replica deployments require Postgres.**

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-sql-common/src/mint/mod.rs
+++ b/crates/cdk-sql-common/src/mint/mod.rs
@@ -13,10 +13,12 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use cdk_common::database::{self, DbTransactionFinalizer, Error, MintDatabase};
+use cdk_common::QuoteId;
 
 use crate::common::migrate;
 use crate::database::{ConnectionWithTransaction, DatabaseExecutor};
-use crate::pool::{DatabasePool, Pool, PooledResource};
+use crate::pool::{DatabaseConfig, DatabasePool, Pool, PooledResource};
+use crate::stmt::query;
 
 mod auth;
 mod completed_operations;
@@ -44,6 +46,7 @@ where
     RM: DatabasePool + 'static,
 {
     pub(crate) pool: Arc<Pool<RM>>,
+    dispatch_pool: Option<Arc<Pool<RM>>>,
 }
 
 /// SQL Transaction Writer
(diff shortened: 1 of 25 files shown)


thesimplekid:
@cdk-bot review

crodas (dismissed):

crates/cdk-sql-common/src/mint/mod.rs
```diff
@@ -80,8 +82,90 @@ where
...
+        Ok(true)
+    }
+
+    /// Attempt quote advisory locks without waiting, in stable key order.
+    async fn try_take_quote_locks(
+        &mut self,
+        quote_ids: &[QuoteId],
+    ) -> Result<database::mint::QuoteLockAttempt, Error> {
+        if quote_ids.is_empty() || RM::Connection::name() != "postgres" {
+            return Ok(database::mint::QuoteLockAttempt::Unsupported);
```
crodas, line 159: For SQLite, it works; it locks everything exclusively.

thesimplekid: SQLite serializes writers, but we cannot hold its write transaction across the payment-backend call without blocking the whole database. The advisory-lock path therefore remains PostgreSQL-only. SQLite is safe for one mint process through the process-local quote guard; multi-replica deployments require PostgreSQL. I added documentation clarifying this, including WAL behavior.

thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

PR #2377, opened 2026-08-18
feat(db): add transactional conditional KV writes

thesimplekid:
### Summary

Add transactional conditional writes to the shared key-value store interface so callers can claim or transition a key without a read-then-write race.

### Root cause

`KVStoreTransaction` only exposed unconditional `kv_write`. Callers that needed write-once ownership had to read a key and then overwrite it, allowing concurrent transactions to observe the same initial state and both proceed.

### Approach

- Add `kv_write_if_absent` for atomic first-writer-wins claims.
- Add `kv_write_if_equals` for compare-and-set transitions.
- Implement SQL claims with `INSERT ... ON CONFLICT DO NOTHING` and affected-row checks.
- Implement SQL transitions with a conditional `UPDATE` and affected-row checks.
- Extend the shared database conformance suite with success, missing-value, mismatched-value, and concurrent-writer cases.
- Update the CLN transactional test store implementation for the expanded trait.

### PR stack

This is the shared prerequisite and should merge first.

- **This PR:** conditional KV primitives → `main`
- #2378: BDK atomic quote/address reservations → this PR
- #2379: CLN/LDK BOLT12 write-once bindings → this PR

After this PR merges, #2378 and #2379 can be retargeted to `main` and merged independently.

### Testing

- `just quick-check`
- Shared conditional-write conformance tests on SQLite
- Relevant SQL, CLN, BDK, and LDK-node targets compile through the full stack

#### ADDED

- Transactional absent-only and compare-and-set operations for `DynKVStore`.

### Checklist

### DIFF
--- a/crates/cdk-cln/src/lib.rs
+++ b/crates/cdk-cln/src/lib.rs
@@ -1270,6 +1270,59 @@ mod tests {
             Ok(())
         }
 
+        async fn kv_write_if_absent(
+            &mut self,
+            primary_namespace: &str,
+            secondary_namespace: &str,
+            key: &str,
+            value: &[u8],
+        ) -> Result<bool, DatabaseError> {
+            let key = memory_kv_key(primary_namespace, secondary_namespace, key);
+
+            let exists = match self.writes.get(&key) {
+                Some(pending) => pending.is_some(),
+                None => {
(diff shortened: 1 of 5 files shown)


thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

PR #2378, opened 2026-08-18
fix(cdk-bdk): reserve quote IDs and receive addresses atomically

thesimplekid:
### Summary

Make the BDK payment store reserve quote IDs and receive addresses atomically, preventing concurrent callers from assigning the same identity to different payments.

### Root cause

Two BDK paths used read-then-write ownership checks:

- send-intent creation could persist more than one intent for the same quote ID when concurrent transactions both observed a missing index entry;
- receive-address tracking could overwrite an existing address-to-quote mapping, allowing another instance sharing the seed and KV store to reassign a tracked address.

### Approach

- Reserve send-intent quote IDs with `kv_write_if_absent` in the same transaction as the intent record.
- Roll back the losing transaction with `DuplicateQuoteId`.
- Preserve retry behavior for the original failed intent.
- Reserve receive-address mappings with `kv_write_if_absent`.
- Keep same-address/same-quote repeats idempotent.
- Reject conflicting mappings and advance to a fresh derivation index.

### PR stack

This PR contains only the BDK consumer changes and depends on #2377.

- #2377: conditional KV primitives → `main`
- **This PR:** BDK atomic quote/address reservations → #2377
- #2379: sibling CLN/LDK BOLT12 hardening → #2377

Review this PR against `kvstore_conditional_writes`; the #2377 changes are intentionally excluded from its diff.

### Testing

- Concurrent send-intent reservation regression test
- Receive-address reservation, idempotency, and conflict tests
- Two-instance shared-seed/shared-store receive-address regression test
- `just quick-check`

#### FIXED

- Prevent concurrent BDK payments from sharing a quote ID or reassigning a tracked receive address.

### Checklist

### DIFF
--- a/crates/cdk-bdk/src/error.rs
+++ b/crates/cdk-bdk/src/error.rs
@@ -199,6 +199,13 @@ pub enum Error {
     #[error("Receive address not found: {0}")]
     ReceiveAddressNotFound(String),
 
+    /// No unreserved receive address was found within the retry limit.
+    #[error("Could not reserve a fresh receive address after {attempts} attempts")]
+    ReceiveAddressReservationExhausted {
+        /// Number of addresses derived before giving up.
+        attempts: usize,
+    },
+
     /// Database
     #[error("Database error")]
     BdkPersist,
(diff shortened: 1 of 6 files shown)


thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

PR #2379, opened 2026-08-18
fix(payment): make BOLT12 quote bindings write-once

thesimplekid:
### Summary

Make CLN and LDK-node BOLT12 quote bindings write-once so retries remain idempotent and concurrent backend calls cannot dispatch more than one payment for the same quote.

The normal mint flow already serializes quote dispatch. This change also enforces that invariant at the payment-backend boundary, making recovery safe for retries, direct backend use, and future call paths.

### Root cause

Both backends stored quote ownership with unconditional KV overwrites:

- CLN mapped a mint-generated `quote_id` to a payment hash after fetching an invoice;
- LDK-node mapped a `quote_id` first to a `Dispatching` sentinel and then to the resolved payment ID.

A duplicate or concurrent backend call could replace those mappings, dispatch another payment, and leave recovery following only the last stored value.

#### CLN

- Claim the quote-to-payment-hash binding with `kv_write_if_absent`.
- Treat an existing identical hash as an idempotent success.
- Reject a different hash before fetching or paying a second invoice.
- Propagate the conflict from `make_payment`.

#### LDK-node

- Claim `Dispatching` with an absent-only write.
- Reject an existing sentinel or resolved payment ID.
- Resolve `Dispatching` to the payment ID with `kv_write_if_equals`.
- Accept an idempotent repeat of the same resolved payment ID and reject conflicts.

### PR stack

This PR contains only the payment-backend changes and depends on #2377.

- #2377: conditional KV primitives → `main`
- #2378: sibling BDK atomic reservations → #2377
- **This PR:** CLN/LDK BOLT12 write-once bindings → #2377

Review this PR against `kvstore_conditional_writes`; it does not depend on #2378. After #2377 merges, this PR can be retargeted to `main`.

### Testing

- CLN repeated, conflicting, and concurrent binding tests
- LDK-node duplicate claim, idempotent/conflicting binding, concurrent claim, and concurrent resolution tests
- `just quick-check`

#### FIXED

- Make CLN and LDK-node BOLT12 payment dispatch idempotent per quote.

### Checklist

Closes project-loupe/audit-cdk#419
Closes project-loupe/audit-cdk#464

### DIFF
--- a/crates/cdk-cln/src/error.rs
+++ b/crates/cdk-cln/src/error.rs
@@ -20,6 +20,12 @@ pub enum Error {
     /// Invalid quote id
     #[error("Invalid quote id")]
     InvalidQuoteId,
+    /// BOLT12 quote is already bound to a different payment hash
+    #[error("BOLT12 quote {quote_id} is already bound to a different payment hash")]
+    Bolt12QuoteBindingConflict {
+        /// Quote id whose binding was rejected
+        quote_id: String,
+    },
     /// Cln Error
     #[error(transparent)]
     Cln(#[from] cln_rpc::Error),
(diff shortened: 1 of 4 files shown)


thesimplekid:
@cdk-bot review

thesimplekid:
@cdk-bot review

PR #2386, opened 2026-08-19
Sign arbitrary payloads with the identity key

### LINKED ISSUE #2387: Sign arbitrary data with the mint private key

crodas:
### Description

Fixes #2387

The Signatory trait only exposed keyset operations, so a mint with something to attest to that is not ecash had no way to get it signed without holding a private key itself.

Add sign(payload), backed by the key whose public half is already published as SignatoryKeysets::pubkey. The payload is not signed bare: it goes through HMAC_SHA256 keyed by a domain tag first, so a signature here is not interchangeable with a NUT-11 or NUT-20 signature over the same bytes. The tag is a public constant, so verifiers recompute the digest with no secret.

Bumps the signatory schema version, so mint and signatory must be upgraded together.

This is an attempt to comply with https://github.com/cashubtc/nuts/pull/416

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-mintd/src/config_service.rs
+++ b/crates/cdk-mintd/src/config_service.rs
@@ -302,27 +302,37 @@ impl ConfigurationService {
 pub(crate) fn discover_signing_identity(
     settings: &Settings,
 ) -> Result<SigningIdentity, ConfigurationServiceError> {
-    let pubkey = if settings.enabled_signatory().is_some() {
+    if settings.enabled_signatory().is_some() {
         return Err(ConfigurationServiceError::SigningIdentity(
             "remote signatory identity requires asynchronous validation".to_owned(),
         ));
-    } else if let Some(seed) = settings
+    }
+    let seed = local_seed(settings)?.ok_or_else(|| {
+        ConfigurationServiceError::SigningIdentity(
+            "no local signing source is configured".to_owned(),
+        )
+    })?;
+    Ok(signing_identity_from_pubkey(root_pubkey(&seed)?))
+}
+
+/// Seed bytes the local signatory would be built from, if one is configured.
+fn local_seed(settings: &Settings) -> Result<Option<Vec<u8>>, ConfigurationServiceError> {
+    if let Some(seed) = settings
         .info
         .seed
         .as_deref()
         .filter(|seed| !seed.is_empty())
     {
-        root_pubkey(seed.as_bytes())?
-    } else if let Some(mnemonic) = settings.info.mnemonic.as_deref() {
-        let mnemonic = Mnemonic::from_str(mnemonic)
-            .map_err(|error| ConfigurationServiceError::SigningIdentity(error.to_string()))?;
-        root_pubkey(&mnemonic.to_seed_normalized(""))?
-    } else {
-        return Err(ConfigurationServiceError::SigningIdentity(
-            "no local signing source is configured".to_owned(),
-        ));
-    };
-    Ok(signing_identity_from_pubkey(pubkey))
+        return Ok(Some(seed.as_bytes().to_vec()));
+    }
+    match settings.info.mnemonic.as_deref() {
+        Some(mnemonic) => {
+            let mnemonic = Mnemonic::from_str(mnemonic)
+                .map_err(|error| ConfigurationServiceError::SigningIdentity(error.to_string()))?;
+            Ok(Some(mnemonic.to_seed_normalized("").to_vec()))
+        }
+        None => Ok(None),
+    }
 }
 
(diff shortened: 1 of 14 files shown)


crodas:
@a1denvalu3, can I have a review of this PR?

a1denvalu3 (approved):

a1denvalu3:
@crodas It looks good. Couldn't find any particular issue.

thesimplekid:
@cdk-bot review

crates/cdk-mintd/src/config_service.rs
```diff
@@ -382,14 +401,35 @@ fn validate_authored_mint_pubkey(
...
-        .is_some_and(|pubkey| pubkey != signing_identity.pubkey)
-    {
-        return Err(ConfigurationServiceError::SigningIdentityChange);
+    let Some(authored) = settings.mint_info.pubkey else {
+        return Ok(());
+    };
+    if authored == signing_identity.pubkey {
+        return Ok(());
+    }
+    if is_legacy_identity(settings, &authored)? {
```
thesimplekid, line 410: Should we also migrate the persisted `ConfigEnvelope.signing_identity` here? This recognizes an authored legacy `mint_info.pubkey`, but an existing envelope still fingerprints the legacy BIP-32 key. `startup()` later compares that fingerprint directly with the new NUT-06 fingerprint and returns `SigningIdentityChange`; `apply()` cannot repair it for the same reason. The same gap affects `initialize()`, which directly rejects an existing database legacy pubkey before `Mint::new` can replace it. Could we verify and atomically migrate the legacy fingerprint for local seeds, provide an explicit migration mechanism for remote signatories, and cover startup and initialization from legacy state?

crates/cdk-signatory/src/proto/signatory.proto
```diff
@@ -17,7 +19,7 @@ service Signatory {
 enum Constants {
   CONSTANTS_UNSPECIFIED = 0;
   // this constant should be sent on the header of the grpc code for verification
-  CONSTANTS_SCHEMA_VERSION = 2;
+  CONSTANTS_SCHEMA_VERSION = 3;
```
thesimplekid, line 22: Should this increment the signatory schema version again? Version 3 was introduced for the parent commit with the master-key/HMAC behavior, while the current head changes both what `Sign` signs and the meaning of `SignatoryKeysets.pubkey`. Mixed versions therefore pass validation despite being cryptographically incompatible. This also contradicts the ADR statement that mixed versions refuse to connect.

crates/cdk/src/mint/mod.rs
```diff
@@ -322,9 +322,16 @@ impl Mint {
             );
         }
 
-        // Persist missing pubkey early to avoid losing it on next boot and ensure stable identity across restarts
+        // Persist the pubkey early to avoid losing it on next boot and ensure stable identity across restarts
         let mut computed_info = mint_info;
-        if computed_info.pubkey.is_none() {
+        if computed_info.pubkey != Some(keysets.pubkey) {
```
thesimplekid, line 327: This reconciles the identity during construction, but should the subscription drain also validate `SignatoryKeysets::pubkey`? It currently extracts only `.keysets`. After a remote signatory reconnects with another identity, signing would use the new key while the mint continues advertising the old persisted key. Could we fail closed on an unexpected identity change and add a reconnect regression test?

crates/cdk-signatory/src/signatory.rs
```diff
@@ -150,6 +151,13 @@ pub trait Signatory {
     /// Verify [`Proof`] meets conditions and is signed by the mint (ignores P2PK/HTLC signatures"
     async fn verify_proofs(&self, proofs: Vec<Proof>) -> Result<(), Error>;
 
+    /// Sign an arbitrary payload with the signatory's identity key.
+    ///
+    /// Verifies against `SignatoryKeysets::pubkey` through
```
thesimplekid, line 156: This documentation still describes the previous domain-separated digest, but the current implementation signs `SHA256(payload)` directly. Could we update it to reflect the new signing semantics?

thesimplekid (commented):
This should not merge as a NUT-06-compliant implementation until cashubtc/nuts#416 settles. The unresolved tagged-hash discussion (https://github.com/cashubtc/nuts/pull/416#discussion_r3913388013) would directly change the signing construction and its vectors, after this PR already changes every mint identity and the signatory wire semantics. The PR description is also stale: it still describes the previous BIP-32 master-key/HMAC construction rather than the current NUT-06-derived key and plain SHA-256 signing behavior.

PR #2390, opened 2026-08-19
Collapse duplicate P2PK keys when building spending conditions

### LINKED ISSUE #2389: P2PK construction accepts duplicate keys that verification rejects, minting unspendable outputs
> This issue was discovered will writing the Nitro cashu bindings to have parity with cashu-ts

## Problem

Spend-time verification collapses pubkeys by x-coordinate and rejects repeats (check_duplicate_pubkeys in `crates/cashu/src/nuts/nut10/mod.rs`, `Error::DuplicatePubkey`). Construction does not agree with it: SpendingConditions::P2PKConditions keeps the data key and the pubkeys tag in separate buckets, and Conditions::new validates num_sigs against the raw key count. SpendingConditions::pubkeys() did dedupe, but on the full 33-byte key, so the same x-coordinate under both parity prefixes survives as two distinct keys.

The result: a caller can build a nominal 2-of-2 whose two keys are the same key. try_into::<Secret>() succeeds, the output gets minted, and the mismatch only shows up at spend time, when verification refuses the secret as a duplicate. Nobody can satisfy the condition, so the funds are stuck.

## Reproduction

```rust
let sk = SecretKey::generate();
let mut bytes = sk.public_key().to_bytes();
bytes[0] = 0x02;
let pk_02 = PublicKey::from_slice(&bytes).unwrap();
bytes[0] = 0x03;
let pk_03 = PublicKey::from_slice(&bytes).unwrap(); // same x-coordinate

let conditions = Conditions {
    locktime: None,
    pubkeys: Some(vec![pk_03]),
    refund_keys: None,
    num_sigs: Some(2),
    sig_flag: SigFlag::SigInputs,
    num_sigs_refund: None,
};

// Accepted: looks like a 2-of-2.
let secret: Secret = SpendingConditions::new_p2pk(pk_02, Some(conditions))
    .try_into()
    .unwrap();

// Rejected at spend time: Error::DuplicatePubkey.
let nut10: Nut10Secret = secret.try_into().unwrap();
get_pubkeys_and_required_sigs(&nut10, unix_time()).unwrap_err();
```

A wallet holding such a proof has no recovery path: the secret is committed in the minted output.

## Expected

Construction and verification should share one view of key identity. A key repeated across data and pubkeys collapses into one slot, and the threshold is checked after the collapse, so the impossible 2-of-2 above is either a valid 1-of-1 or a clear construction error, never a mintable dead output.

crodas:
### Description

Fixes #2389. This fix was extracted from #1982

Spending validation counts pubkeys by x-coordinate and rejects repeats, but construction counted them naively. A caller could build a nominal 2-of-2 whose two keys are the same key, mint the output, and only find out at spend time that nothing can satisfy it.

Dedupe on the same x-only view verification uses, with the P2PK data key sharing the set with the pubkeys tag, and validate the threshold after the collapse rather than before. While here: enforce the NUT-28 slot ceiling, reject refund keys with no locktime since that path never opens, and route with_conditions through the validating conversion it was bypassing.

Amount::split already rejects a keyset advertising a zero denomination. Say so on the function and pin it with a test.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/nut10/spending_conditions.rs
+++ b/crates/cashu/src/nuts/nut10/spending_conditions.rs
@@ -294,6 +294,7 @@ impl Conditions {
 /// the blinded `data` key plus the `pubkeys` and refund tags cannot occupy more
 /// than [`MAX_LOCKING_SLOTS`] entries. Plain NUT-11 and NUT-14 locks have no
 /// such limit, and verification never counts slots.
+#[cfg(feature = "wallet")]
 pub(crate) fn check_locking_slots(
     pubkeys: usize,
     refund_keys: Option<&[PublicKey]>,
--- a/crates/cashu/src/nuts/nut10/spending_conditions.rs
+++ b/crates/cashu/src/nuts/nut10/spending_conditions.rs
@@ -2,7 +2,6 @@
 //!
 //! <https://github.com/cashubtc/nuts/blob/main/10.md>
 
-use std::collections::HashSet;
 use std::str::FromStr;
 
 use bitcoin::hashes::sha256::Hash as Sha256Hash;
@@ -63,9 +62,7 @@ impl SpendingConditions {
                 if let Some(conditions) = conditions {
                     pubkeys.extend(conditions.pubkeys.clone().unwrap_or_default());
                 }
-                // Remove duplicates
-                let unique_pubkeys: HashSet<_> = pubkeys.into_iter().collect();
-                Some(unique_pubkeys.into_iter().collect())
+                Some(pubkeys)
             }
             Self::HTLCConditions { conditions, .. } => conditions.clone().and_then(|c| c.pubkeys),
         }
@@ -128,9 +125,14 @@ impl TryFrom<Nut10Secret> for SpendingConditions {
     }
 }
 
-impl From<SpendingConditions> for super::Secret {
-    fn from(conditions: SpendingConditions) -> super::Secret {
-        match conditions {
+/// The only door from an author-supplied lock to wire bytes, so the
+/// construction rules are enforced here instead of at each caller.
+impl TryFrom<SpendingConditions> for super::Secret {
+    type Error = Error;
+    fn try_from(conditions: SpendingConditions) -> Result<super::Secret, Self::Error> {
+        conditions.validate()?;
+
+        Ok(match conditions {
             SpendingConditions::P2PKConditions { data, conditions } => super::Secret::new(
                 Kind::P2PK,
                 super::SecretData::new(data.to_hex(), conditions),
@@ -139,16 +141,14 @@ impl From<SpendingConditions> for super::Secret {
                 Kind::HTLC,
                 super::SecretData::new(data.to_string(), conditions),
             ),
-        }
+        })
     }
 }
 
 impl TryFrom<SpendingConditions> for Secret {
     type Error = Error;
     fn try_from(conditions: SpendingConditions) -> Result<Secret, Self::Error> {
-        conditions.validate()?;
-        let secret: Nut10Secret = conditions.into();
-        Secret::try_from(secret)
+        Secret::try_from(Nut10Secret::try_from(conditions)?)
     }
 }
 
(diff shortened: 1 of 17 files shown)


thesimplekid:
@cdk-bot review

crodas:
@cdk-bot review

crates/cashu/src/nuts/nut10/spending_conditions.rs
```diff
@@ -266,7 +279,107 @@ impl Conditions {
...
+    data: PublicKey,
+    conditions: Option<Conditions>,
+) -> Result<(PublicKey, Option<Conditions>), Error> {
+    let Some(conditions) = conditions else {
+        return Ok((data, None));
+    };
+
+    let mut primary = vec![data];
+    primary.extend(conditions.pubkeys.clone().unwrap_or_default());
+    let primary = super::dedupe_pubkeys(&primary);
```
thesimplekid: Should duplicate keys return `DuplicatePubkey` instead of being silently removed? NUT-11 defines duplicate keys within a signing pathway as malformed. Normalizing them can hide caller mistakes and silently change the requested signer set—for example, 1-of-[A, A] becomes 1-of-[A], while 2-of-[A, A] errors. Could we validate uniqueness by x-coordinate here, keeping the primary and refund pathways separate, and return an error before constructing or blinding the secret?

crates/cashu/src/nuts/nut10/spending_conditions.rs
```diff
@@ -146,6 +143,7 @@ impl From<SpendingConditions> for super::Secret {
 impl TryFrom<SpendingConditions> for Secret {
     type Error = Error;
     fn try_from(conditions: SpendingConditions) -> Result<Secret, Self::Error> {
+        let conditions = conditions.normalized()?;
```
thesimplekid: Does this actually cover every construction path? `new_p2pk()` remains infallible, `new_htlc()` does not validate its attached `Conditions`, and the public `From<SpendingConditions> for nut10::Secret` conversion above bypasses this `TryFrom` implementation. A caller can therefore still emit a malformed NUT-10 secret with `.into()`. Should the constructors become fallible and the infallible secret/request conversions be replaced or restricted so validation cannot be bypassed?

crates/cashu/src/nuts/nut10/spending_conditions.rs
```diff
@@ -233,10 +231,22 @@ impl Conditions {
             _ => {}
         }
 
+        if self
+            .refund_keys
+            .as_ref()
+            .is_some_and(|keys| !keys.is_empty())
+            && self.locktime.is_none()
+        {
+            return Err(Error::NUT11(crate::nut11::Error::RefundKeysRequireLocktime));
```
thesimplekid: Should this be part of protocol validation? Under NUT-11/NUT-14, an absent locktime creates a permanent lock: the refund pathway is unreachable, but the primary/receiver pathway remains valid. Rejecting this may be a useful authoring safeguard, but it rejects previously accepted, spec-valid conditions. Could we either remove it from general validation or expose it as a separately documented safe-construction policy?

crates/cdk/src/wallet/payment_request.rs
```diff
@@ -772,6 +772,8 @@ impl WalletRepository {
     /// Errors:
     /// - Invalid SHA-256 `hash` strings or invalid HTLC/P2PK parameterizations surface as errors
     ///   from parsing and `SpendingConditions` constructors.
+    /// - Conditions are normalized here so a request advertising duplicate keys
```
thesimplekid: This currently says duplicates fail, but `normalized()` silently removes them unless the resulting signature threshold is impossible. Could we align the documentation with the chosen behavior?

crodas: 8f410977519e73cba26b3960f84bba59c4417633

crates/cashu/src/nuts/nut10/spending_conditions.rs
crodas: f0c2efb80990059ada61ca6985159e3397e19f59

crodas:
@cdk-bot review

crates/cashu/src/nuts/nut10/spending_conditions.rs
crodas: 02691e39b6521dcf042bee403a90217064084d94

thesimplekid:
CI failed in the intial push as well as after review fixes so I fixed it in https://github.com/cashubtc/cdk/pull/2390/commits/bf05c55ba381c2f2975afc8f34926126e55e0422

crodas:
@cdk-bot review

thesimplekid (approved):

PR #2395, opened 2026-08-20
fix(hygiene): escape untrusted terminal and log output

thesimplekid:
Route untrusted strings from mints, NpubCash, RPC, and APIs through control-character and bidi escaping. Redact proof secrets by default and reject unsafe migration filenames before generating Rust source.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2401, opened 2026-08-21
fix(mintd): harden environment config and exports

thesimplekid:
Preserve protected endpoint policies when auth is configured only through environment variables. Reject malformed, non-finite, or out-of-range LND fee percentages instead of silently ignoring or accepting them.

Create new configuration exports with owner-only permissions on Unix while preserving the existing --force replacement behavior.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2410, opened 2026-08-22
feat: remove unchecked math

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas (approved):
I love this, can we clippy to reject unsafe math unless it is explicitly disabled, for those cases when we really know what we are doing (like in increments or unimportant math operations)

thesimplekid:
Does clippy have that option then sure?

PR #2413, opened 2026-08-23
feat(wallet)!: require confirmation for payment requests

thesimplekid:
Expose NUT-18 method and input fees through a prepared payment request. Require callers to inspect and explicitly confirm or cancel the reserved payment instead of using an atomic pay_request API.

Return a structured delivery error when token creation succeeds but transport delivery fails. Preserve the pending operation ID so callers can revoke an unclaimed send instead of paying the request again.

Add method-fee publishing to request creation, preserve advertised transport order, and expose the prepared flow through the CLI and FFI.

BREAKING CHANGE: Remove atomic pay_request methods from Wallet, WalletRepository, wallet traits, and FFI. Call prepare_pay_request and then confirm or cancel the returned PreparedPaymentRequest.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/payment_request.rs
+++ b/crates/cdk/src/wallet/payment_request.rs
@@ -18,6 +18,7 @@ use nostr_sdk::nips::nip19::Nip19Profile;
 use nostr_sdk::prelude::*;
 #[cfg(feature = "nostr")]
 use nostr_sdk::{Client as NostrClient, EventBuilder, FromBech32, Keys, ToBech32};
+use tracing::instrument;
 
 use crate::error::Error;
 use crate::mint_url::MintUrl;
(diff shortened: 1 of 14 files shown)


crates/cdk/src/wallet/payment_request.rs
```diff
@@ -72,21 +275,17 @@ impl Wallet {
...
-                    .iter()
-                    .find(|t| t._type == TransportType::HttpPost)
-            });
+        // NUT-18 transports are sorted by receiver preference. Select the first
+        // transport this build can use without reordering the request.
+        let transport = payment_request_transport(&payment_request.transports)
+            .cloned()
+            .ok_or_else(|| {
+                Error::Custom("No transport available in payment request".to_string())
+            })?;
```
asmogo, line 288: This is a behavioral change that we might want to rethink. Previously were preferring Noster for transport. 

Now the transport selection follows the receiver's advertised order. 
This could result in HTTP requests that reveal the IP address of the client.

crates/cdk/src/wallet/payment_request.rs
```diff
@@ -31,19 +31,204 @@ use crate::wallet::ReceiveOptions;
...
+    }
+
+    /// Confirm and deliver the prepared payment.
+    ///
+    /// If token creation succeeds but transport delivery fails, this returns
+    /// [`Error::PaymentRequestDeliveryFailed`]. The token remains a pending
+    /// send: do not prepare the payment again. Use the error's operation ID
+    /// with [`Wallet::revoke_send`] to reclaim the token if it has not already
+    /// been claimed by the receiver.
+    pub async fn confirm(self) -> Result<(), Error> {
```
asmogo, line 155: nit: crates/cdk/src/wallet/payment_request.rs:153,178,225,334,1214

The new public async methods (`Wallet::prepare_pay_request`, `WalletRepository::prepare_pay_request`, `PreparedPaymentRequest::confirm`/`cancel`, `deliver_payment_request`) have no `#[instrument]` attribute. AGENTS.md asks for `#[instrument(skip_all)]` on most public async methods, and the removed `WalletTrait::pay_request` impl had `#[instrument(skip(self, request))]`. Worth adding for trace continuity.

crates/cdk/src/wallet/payment_request.rs
thesimplekid: Ah good catch. I think go back to nostr? Since the list is not ordered by any preference we can make the choice to prefer nostr for the privacy reason you mention.

PR #2424, opened 2026-08-26
Sign the mint info per NUT-06

### LINKED ISSUE #2388: Sign mint info, to proof the mint owns the public key they claim
cashubtc/nuts#416

crodas:
### Description

Fixes #2388 and depends on #2386

ADR-0005 gave the mint an identity key and no caller for it. Everything
/v1/info returns arrives unattributed, so a wallet cannot tell the mint's
answer from a rewritten one.

The mint now signs the canonical response and the wallet checks it. Signing
needs JCS (RFC 8785), which nothing in the workspace could produce, so this
adds a canonicalizer to the cashu crate. It rejects floats and integers with
no exact double rather than guess at the ECMAScript number rules, neither of
which any Cashu type can reach.

Verification runs on the response as JSON rather than on a deserialized
MintInfo. Serde drops members this crate does not model, and a mint
advertising a NUT that CDK has not implemented signed those too, so a payload
rebuilt from the struct would be missing bytes and fail against a mint that
did nothing wrong.

Signing lives in signed_mint_info rather than mint_info, which quote issuance
and melting call on every request. They need the settings, not an attestation,
and must not start failing when the signatory is unreachable. The cost is that
/v1/info does: a cold start during a signatory outage now serves an error
where it used to serve a database read.

The signature is cached against the payload it covers, so a repeat handshake
costs no signatory round trip. Anything that changes what is served changes
the payload, so the cache needs no explicit invalidation. Excluding time from
the payload is what lets the HTTP layer stamp it after signing.

Neither side persists a signature. The wallet databases already drop
max_array_length, so a stored one could never verify; on the mint it is
derived, so set_mint_info and the constructor clear it.

An unsigned response is warned about rather than rejected, since no deployed
mint signs yet. A signature that is present is always verified and a bad one
is always an error. A wallet that wants the spec's MUST today can ask for it
with set_require_signed_mint_info.

ADR-0005 and the record for this change were one decision told in two halves,
so they are merged. The identity key and what it signs now live together, under
0005, which records the decisions and their consequences.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crodas:
This also needs your review @a1denvalu3

PR #2427, opened 2026-08-27
Store keyset u64 fields as text

### LINKED ISSUE #2426: Store keyset u64 fields as text
There are many u64 keyset fields that should be stored as texts

crodas:
### Description

Fixes #2426 

An oversized input_fee_ppk or final_expiry wrapped negative when narrowed to i64 on write, while the read path rejects negatives. The row committed, and every later keyset read failed for the whole table, so periodic reloads broke and a signatory restart could not reconstruct its keysets. Rotation RPCs accept both values unbounded.

Storing valid_from, valid_to and input_fee_ppk as text removes the narrowing conversion, so the full range round-trips. The read macros already parse text, leaving pre-migration rows readable.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

thesimplekid (changes requested):
Could we cap these fields at i64::MAX instead of storing them as text? I don’t see a practical reason for input_fee_ppk, valid_from, or final_expiry to exceed that range. PostgreSQL would still need a migration from INTEGER to BIGINT, while SQLite’s INTEGER already supports signed 64-bit values. We should use checked i64::try_from conversions so oversized values return an error instead of wrapping.

Also, the auth database has separate migrations and appears to be missed here. Could we apply the same validation and PostgreSQL BIGINT migration there as well?

crodas:
Closed in favour of #2473

PR #2430, opened 2026-08-27
fix(cashu): preserve mint URL path and query data in format_url

thesimplekid:
Parse mint URLs with url::Url instead of ad-hoc splitting. Stop truncating paths at a second "://", preserve query value casing, and strip trailing slashes only from the path so routing data remains intact.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2431, opened 2026-08-27
fix: escape untrusted strings before terminal and log output

thesimplekid:
Route remaining peer- and server-controlled strings (NUT-21 endpoint paths, BIP353 resolver values, mint error bodies in startup recovery, currency units in check-pending, OIDC token display, and gift-wrap mint URLs) through control-character and bidi escaping. Use the shared cdk-common helper for non-CLI sinks, and reject unsafe NUT-21 endpoint paths before formatting deserialization errors.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/src/nuts/auth/nut21.rs
+++ b/crates/cashu/src/nuts/auth/nut21.rs
@@ -43,6 +43,18 @@ impl Settings {
     }
 }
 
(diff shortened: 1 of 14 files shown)


thesimplekid:
@cdk-bot review

PR #2436, opened 2026-08-27
Source keyset unit and expiry from the info

crodas:
### Description

A reload reuses the already-derived MintKeySet keyed by id while re-reading the keyset info from the database, so any field taken from the cached key can go stale. unit and final_expiry were the only two still key-sourced, so a divergence would have been published to wallets.

No supported flow can trigger it today (rotation is the only writer, rows are insert-only, and blind_sign gates on the info), but make the info authoritative for everything but the key material and pin it with tests.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-signatory/src/db_signatory.rs
+++ b/crates/cdk-signatory/src/db_signatory.rs
@@ -728,6 +728,62 @@ mod test {
         );
     }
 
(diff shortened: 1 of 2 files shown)


crodas:
@cdk-bot review

thesimplekid (approved):

PR #2438, opened 2026-08-27
feat(wallet): sign P2PK-locked inputs in prepare_melt_proofs/token

kaloudis:
### Description

`prepare_melt_proofs` and `prepare_melt_token` reserve external proofs as-is, so P2PK or HTLC locked inputs reach the mint without NUT-11 witnesses and get rejected. In practice this means a wallet holding a P2PK-locked token (for example one locked to the wallet's own key by a sender) cannot melt it directly: it has to `receive` first (an extra swap round trip plus input fees on fee-charging mints) and then melt from balance, even though the whole point of `prepare_melt_token` is to go straight from token to Lightning payment.

This PR signs locked inputs during prepare, following the existing `SendOptions`/`ReceiveOptions` pattern:

- New `MeltPrepareOptions { p2pk_signing_keys, preimages, metadata }` in `cdk-common`, with redacted `Debug` like its send/receive counterparts. It lives only in the in-memory prepare flow — nothing new is persisted, no DB migration.
- New `prepare_melt_proofs_with_options` / `prepare_melt_token_with_options` on `cdk::Wallet`, the `cdk_common::wallet::Wallet` trait (where the existing methods become default-impl wrappers), and `cdk-ffi` (which gains `metadata` support over FFI as a bonus). The existing `prepare_melt_proofs` / `prepare_melt_token` signatures are unchanged — no breaking change for current callers or FFI consumers.
- In `MeltSaga::prepare_with_proofs`, caller keys are merged with wallet-known keys via the send saga's `merge_keyring_keys` (moved to `wallet::util` and shared instead of duplicated), then inputs are signed with the existing `util::sign_proofs`, so slot numbering, condition/refund keys, NUT-28 ephemeral derivation, and P2BK metadata stripping stay identical to send/receive.
- HTLC preimages from `options.preimages` are attached to matching inputs (`util::add_htlc_preimages`), mirroring `ReceiveOptions`. Inputs without a matching preimage are left unchanged so refund-path spends after locktime remain possible.
- SIG_ALL-locked inputs are rejected at prepare with `SigAllNotSupportedHere`: a SIG_ALL witness must commit to the melt request's change outputs, which only exist at confirmation. Rejecting early avoids reserving proofs the mint would refuse. Confirm-time `sign_sig_all` can layer on top of the options struct as a follow-up.

Motivation: ZeusLN/zeus#4491 - a mobile wallet sweeping received (P2PK-locked) tokens directly to a Lightning invoice via `prepare_melt_token`, without an intermediate receive swap.

Unit tests: `test_prepare_melt_signs_p2pk_locked_external_proofs` verifies a locked external proof is reserved with a witness attached; `test_prepare_melt_rejects_sig_all_proofs` verifies SIG_ALL inputs error out with nothing reserved.

-----

### Notes to the reviewers

- Signing happens after `initialize_melt` and before the amount check and reservation, so a failed prepare still releases nothing and the token remains spendable elsewhere.
- SIG_INPUTS signing stays at prepare time (rather than confirmation) so reserved proofs are persisted with their witnesses and crash recovery can rebuild the melt request; keys held only in the in-memory `PreparedMelt` would not survive a restart.
- Keys provided by the caller take precedence; wallet DB lookup only fills in pubkeys the caller did not cover, mirroring `receive_proofs` (now literally the same `merge_keyring_keys` code as send).
- I did not add signing to `prepare_melt` (wallet-selected proofs), since wallet-held proofs are unlocked post-swap by construction.
- Local clippy did not run because my clippy is older than the repo's `clippy.toml` (`large-error-ignored`), deferring to CI. fmt and unit tests pass locally.

-----

#### ADDED

- cdk/cdk-common/cdk-ffi: `MeltPrepareOptions` and `prepare_melt_proofs_with_options` / `prepare_melt_token_with_options`, signing P2PK/HTLC-locked melt inputs during prepare using caller-provided keys plus signing keys known to the wallet, and attaching HTLC preimages, so locked proofs and tokens can be melted directly.

----

### Checklist


thesimplekid:
Should we mirror the existing `PreparedSend` options pattern here rather than adding signing keys as a required positional argument?

[`prepare_send`](https://github.com/cashubtc/cdk/blob/main/crates/cdk/src/wallet/send/mod.rs#L159-L176) accepts [`SendOptions`](https://github.com/cashubtc/cdk/blob/main/crates/cdk-common/src/wallet/mod.rs#L450-L470), which carries externally supplied signing keys and other preparation settings. Its confirmation flow combines those keys with keys found in the wallet through [`merge_keyring_keys`](https://github.com/cashubtc/cdk/blob/main/crates/cdk/src/wallet/send/saga/mod.rs#L146-L168).

The melt implementation already performs the same local-key lookup with [`get_signing_key`](https://github.com/cashubtc/cdk/blob/b5b614b7abb1e4260c326e02066984bda6bed967/crates/cdk/src/wallet/melt/saga/mod.rs#L522-L535), so callers only need to supply keys for external proofs whose keys are not stored by the wallet.

Could we introduce a `MeltPrepareOptions` containing `metadata`, `p2pk_signing_keys`, and HTLC `preimages`, and add `prepare_melt_*_with_options` methods while retaining the existing methods as default wrappers? This would:

- keep `PreparedSend` and `PreparedMelt` APIs consistent;
- avoid adding another required positional argument;
- support HTLC preimages, as [`ReceiveOptions`](https://github.com/cashubtc/cdk/blob/main/crates/cdk-common/src/wallet/mod.rs#L522-L533) already does;
- allow signing and verification during confirmation, when `SIG_ALL` change outputs exist.

The options should remain in the in-memory `PreparedMelt` and not be persisted, so this should not require a database migration.

kaloudis:
Agreed. Reworking this to `MeltPrepareOptions` + `prepare_melt_*_with_options`, keeping the existing methods as wrappers.

One caveat: SIG_INPUTS proofs will still be signed at prepare time, so reserved proofs are persisted with witnesses and crash recovery can rebuild the melt request — keys held only in the in-memory `PreparedMelt` wouldn't survive a restart. SIG_ALL inputs get an explicit error at prepare for now (a valid witness must commit to the change outputs, which only exist at confirm); confirm-time `sign_sig_all` can layer on top of the options struct as a follow-up.

crates/cdk/src/wallet/melt/saga/mod.rs
```diff
@@ -514,6 +521,34 @@ impl<'a> MeltSaga<'a, Initial> {
...
+        // need witnesses before the mint will accept them as melt inputs.
+        //
+        // SIG_ALL cannot be satisfied here: its signature must commit to the
+        // melt request's change outputs, which are only created at
+        // confirmation. Reject early rather than reserving proofs the mint
+        // would refuse.
+        let sig_flag = enforce_sig_flag(proofs.clone()).sig_flag;
+        if sig_flag == SigFlag::SigAll {
+            return Err(crate::nuts::nut11::Error::SigAllNotSupportedHere.into());
+        }
```
thesimplekid: Could we move the `initialize_melt` above on line 522 below this block as if we error here we have already reserved the proofs in the `initialize_melt` when we should not as they cannot be used.

thesimplekid:
Rebases fixed a few errors from the ci. 

@cdk-bot review

thesimplekid (approved):

Issue #2439, opened 2026-08-27
Wallet database update_mint_url leaves transactions, melt_quote, and mint rows on the old URL


kaloudis:
The wallet database trait method `update_mint_url` only migrates a subset of the tables that carry a `mint_url` column.

In `cdk-sql-common` the implementation updates `mint_quote` and `proof` only:

https://github.com/cashubtc/cdk/blob/1ef10319/crates/cdk-sql-common/src/wallet/mod.rs#L959-L984

But the schema has `mint_url` columns in more places:

* `transactions.mint_url` (20250401120000_add_transactions_table.sql)
* `melt_quote.mint_url` (20260323000000_melt_quote_mint_url.sql)
* `mint.mint_url` (primary key), with `keyset.mint_url` referencing it via ON UPDATE CASCADE

After a migration, transaction history filtered by the new mint URL misses everything recorded before the move, melt quotes keep referencing the old URL (relevant to saga recovery), and stale `mint`/`keyset` rows for the old URL linger until re-fetched under the new URL. The redb implementation has the same scope (proofs and mint quotes). I have not checked cdk-supabase.

Should the trait contract be that every row keyed by the mint URL moves? If so I am happy to follow up with a PR extending the implementations and the shared database test.

Context: hit while wiring up mint URL migration for ZEUS (following a mint that announces an alternative endpoint via the NUT-06 `urls` field). Related: #1767.

PR #2440, opened 2026-08-27
feat(wallet): add WalletRepository::update_mint_url and expose it over FFI

kaloudis:
### Description

Adds `WalletRepository::update_mint_url(old, new)` and exposes it through the FFI `WalletRepository`.

`Wallet::update_mint_url` exists on the core wallet but is listed in #1767 as missing from the FFI, and exposing it directly would be a footgun for long-lived processes: it migrates the database and updates `wallet.mint_url`, but the stored `MintConnector` keeps hitting the old base URL. That is fine for cdk-cli, which exits right after the call and rebuilds the wallet on the next run, but a mobile wallet holding the instance would keep talking to the dead endpoint.

The repository is the natural home for this: it owns the localstore, seed, and transport configuration, so it can run the database migration once and rebuild every affected wallet (all currency units) with fresh connectors against the new URL. Rebuilt wallets keep their target proof count, the wallets map is rekeyed under the new URL, and the rebuilt wallets are returned. Migrating a mint onto its own URL is rejected.

App-side motivation: ZEUS wants to follow a mint that announces alternative endpoints through the NUT-06 `urls` field (for example clearnet plus onion), verifying that the new endpoint serves the same keysets before migrating. This is the missing primitive on the bindings side.

-----

### Notes to the reviewers

* The database migration runs once per mint rather than per wallet, since the rows are keyed by mint URL across all currency units.
* Wallet clones obtained before the call keep the old URL and stale connectors; the doc comment tells callers to fetch wallets again from the repository. This matches the existing behavior of `set_mint_config`, which also re-creates rather than mutates.
* A wallet created with a custom `mint_connector` is rebuilt on the repository's transport (default, proxy, or Tor), because the custom connector targets the old URL. Callers can install a replacement with `set_mint_config` afterwards.
* While looking at this I noticed the database-level `update_mint_url` only migrates `mint_quote` and `proof`, leaving `transactions`, `melt_quote`, and `mint` rows behind; filed separately as #2439 since it affects all backends and is orthogonal to this exposure.
* Local `clippy` (1.97.1) does not recognize the `large-error-ignored` field in clippy.toml, so I ran lint, clippy, and the unit tests for the touched crates individually rather than `just quick-check`; `cargo fmt --all -- --check` is clean, clippy is clean for `cdk` and `cdk-ffi` with that field masked, and `cdk` plus `cdk-ffi` lib tests pass.

-----

#### ADDED

- cdk/cdk-ffi: `WalletRepository::update_mint_url` migrates persisted state for a mint to a new URL and rebuilds the affected wallets so their connectors target the new endpoint, letting wallets follow a mint that moves or announces an alternative endpoint via the NUT-06 `urls` field.

----

### Checklist

### DIFF
--- a/crates/cdk-ffi/src/wallet_repository.rs
+++ b/crates/cdk-ffi/src/wallet_repository.rs
@@ -278,6 +278,36 @@ impl WalletRepository {
             .map_err(|e| e.into()) // Ensure the inner error can convert to FfiError
     }
 
+    /// Move all wallets for a mint to a new mint URL
+    ///
+    /// Persists the migration through the wallet database, then rebuilds every
+    /// affected wallet against the new URL so their connectors target the new
+    /// endpoint. Wallet objects obtained before this call keep the old URL and
+    /// must be fetched again from the repository. Returns the rebuilt wallets.
+    ///
+    /// Intended for when a mint moves or announces an alternative endpoint,
+    /// for example through the NUT-06 `urls` field. The caller is responsible
+    /// for verifying that the new endpoint serves the same mint (its keysets
+    /// match) before migrating.
+    pub async fn update_mint_url(
+        &self,
+        old_mint_url: MintUrl,
+        new_mint_url: MintUrl,
+    ) -> Result<Vec<Arc<crate::wallet::Wallet>>, FfiError> {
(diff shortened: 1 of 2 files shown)


kaloudis:
Zeus-side tracking issue for the consumer of this API: ZeusLN/zeus#4494

PR #2441, opened 2026-08-27
feat: add reversible mint database migrations

asmogo:
## Summary

- embed paired `.down.sql` files while keeping normal startup forward-only
- add transactional rollback planning with preflight rejection at irreversible migrations
- add `cdk-mintd database rollback --steps N --yes` for SQLite, SQLCipher, and PostgreSQL
- provide backward migrations for the latest three mint schema changes and document the downgrade workflow
- add rollback, schema inspection, forward reapplication, and CLI regression tests

## Verification

- `cargo test -p cdk-sql-common --all-features`
- `cargo test -p cdk-sqlite backward_migrations_are_preflighted_and_can_be_reapplied`
- `cargo test -p cdk-sqlite --features sqlcipher backward_migrations_are_preflighted_and_can_be_reapplied`
- `cargo test -p cdk-mintd`
- PostgreSQL 16 rollback and reapplication test against a disposable live database
- `cargo clippy -p cdk-mintd --features postgres,sqlcipher --all-targets -- -D warnings`


crodas:
There is much more nuance than a simple migration up or down, since migrations are a dangerous but sometimes necessary step in upgrading live software.

I reviewed our migrations, and most add new columns; those are easy to step down.

There are some that are not possible:

```sql
UPDATE transactions
SET id = convert_to(replace(saga_id, '-', ''), 'UTF8')
WHERE saga_id IS NOT NULL;
```


```sql
ALTER TABLE mint_quote DROP COLUMN created_time;
```

I think every time we drop a column, instead of dropping it right away, we simply stop caring about it. We deprecate it in the code, and eventually we drop it.

We do quite a good job of never selecting all columns or inserting into all columns. As long as the codebase is robust, which it is,  to ignore redundant columns, the deprecating approach is fine.


Overall, I like the approach, but we need to enforce discipline in writing migrations, perhaps with AI?

PR #2448, opened 2026-08-30
Re-issue melt change under the active keyset after a rotation

### LINKED ISSUE #2447: Honor reserved change keysets after rotation
`process_melt_change` re-evaluates the change output keyset only after the external payment has completed. If the keyset was active when `setup_melt` accepted and persisted the blank outputs but rotates before line 644, `get_keyset_fee_and_amounts` falls back to a synthetic schedule and `Mint::blind_sign` rejects the now-inactive keyset. TX1 has already marked the quote Paid and its proofs Spent, so every finalization/recovery retry encounters the same stored outputs and fails permanently; the melt request and saga are never cleaned up. An unauthenticated Cashu client can keep otherwise-valid asynchronous melts pending across an observable key rotation and settle them afterward, accumulating permanent recovery failures while also losing its change.

crodas:
### Description

Attempt to fix #2447

A melt's blank change outputs are accepted while their keyset is active but signed only after the payment settles. A rotation in that window made the signatory reject them, and because the quote was already paid, every finalization and recovery retry failed the same way: the melt stayed stuck and the change was never issued.

The mint now retries the signing under the keyset active at signing time, re-splitting against its denominations, and the wallet accepts and unblinds change from a substituted keyset when the mint still publishes it for this unit and it is active or DLEQ-proven. If no keyset can issue the change, the paid melt finalizes without it rather than staying stuck.

Two latent bugs had to go with it: filling a pre-registered blind signature row left the old keyset id beside a C_ from the new key, and melt finalize plus both restore paths unblinded with the wallet's currently active keys instead of the signature's keyset.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid (changes requested):
I am not sure this is even within spec, I guess it doesn't say that we cannot substitute outputs but it does not say that we can. Even if its unclear in spec other wallets will not support or be able to handle this; so while this may fix it for cdk mints and cdk wallets it does not fix it if the wallet is anything else, so this introduces an interoperability issue. 

 I checked both cashu-ts and Coco, and neither currently supports this behavior. cashu-ts explicitly rejects a returned `BlindSignature.id` that differs from the corresponding `BlindedMessage.id`, producing a `MeltChangeError`. Coco uses the original output ID to select the unblinding keyset and then hits the same strict validation in its cashu-ts dependency.

crodas:
The alternative would be to make an exception and sign with expired/rotated keysets. I'll prepare a PR

PR #2460, opened 2026-08-31
Sign with inactive keysets

crodas:
### Description

This solution supersedes #2448 by fixing #2447. This solution works because [expired keysets are not valid to swap from](https://github.com/cashubtc/cdk/blob/beaa3c522ca5b200a8246dd59e2c4d7199fbba05/crates/cdk/src/mint/verification.rs#L145-L151), but [rotated keysets (inactive) are](https://github.com/cashubtc/cdk/blob/beaa3c522ca5b200a8246dd59e2c4d7199fbba05/crates/cdk/src/mint/verification.rs#L145-L151).

A melt commits its change outputs when the request is accepted, but they are signed after settlement. If the keyset rotates in between, the signatory refused to sign and the user lost their change. Expired keysets are still refused: they are also rejected as inputs, so signing one would hand back unspendable proofs.

The mint continues to verify swap and issuance outputs against the active keyset, so this only widens what melt change can land on.

Change on a rotated-out keyset also has to be split with that keyset's own fee and denominations. The fallback set produced amounts the keyset had no key for, failing after the payment had already gone out.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk-signatory/src/db_signatory.rs
+++ b/crates/cdk-signatory/src/db_signatory.rs
@@ -360,9 +360,6 @@ impl Signatory for DbSignatory {
                 } = blinded_message;
 
                 let (info, key) = keysets.by_id.get(&keyset_id).ok_or(Error::UnknownKeySet)?;
-                if !info.active {
-                    return Err(Error::InactiveKeyset);
-                }
                 if info.is_expired() {
                     return Err(Error::ExpiredKeyset);
                 }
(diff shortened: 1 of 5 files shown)


robwoodgate:
Concept ACK on signing change under the reserved keyset rather than substituting it.

My agent noted this:

One wallet-side gap this PR makes reachable. Today the mint never returns change on an inactive keyset, so the wallet's finalize path has not had to care which keys it unblinds with. With this change it does.

[finalize_melt_common](https://github.com/cashubtc/cdk/blob/af4929cc495f1b20a73e2df55e0bc8810376a947/crates/cdk/src/wallet/melt/saga/mod.rs#L101-L105) resolves the wallet's current active keyset and passes those keys to [construct_proofs](https://github.com/cashubtc/cdk/blob/af4929cc495f1b20a73e2df55e0bc8810376a947/crates/cdk/src/wallet/melt/saga/mod.rs#L132-L137). The validation just above it is correct: [validate_mint_response_signatures](https://github.com/cashubtc/cdk/blob/af4929cc495f1b20a73e2df55e0bc8810376a947/crates/cdk/src/wallet/blind_signature.rs#L65) loads sig.keyset_id and checks the DLEQ against those keys. But unblinding then uses a different keyset if the wallet's active one has moved on.

Sequence:

- Wallet melts with blanks on keyset A, A active. Payment is slow to settle.
- Mint rotates. A is inactive, B is active.
- Mint settles and (with this PR) signs change under A. sig.keyset_id == A, DLEQ verifies against A.
- Wallet's keyset cache has refreshed in the meantime (default CacheThenNetwork policy does this once its TTL expires, and a long pending melt is exactly the case where it will). active_keyset_with_policy now returns B.
- construct_proofs(change, rs, secrets, &B.keys) computes C = C_ - r*K_B instead of C_ - r*K_A. Standard keysets share denominations, so amount_key succeeds and nothing errors. The proofs are stored with keyset_id: A and an invalid C, and fail when spent.

The resume path is already right: [restore_outputs_with_result](https://github.com/cashubtc/cdk/blob/af4929cc495f1b20a73e2df55e0bc8810376a947/crates/cdk/src/wallet/recovery.rs#L538) loads keys for the blanks' keyset. Only the happy path uses the active keyset.

**Suggested fix:** in finalize_melt_common, unblind with the keys of the keyset each signature names (all change signatures share one keyset, so wallet.keyset(change[0].keyset_id) is enough, or group by id to be safe). That also keeps the validation and unblinding steps on the same keys.

For reference, cashu-ts had the equivalent problem and now unblinds melt change and restore results with whatever keyset the signature names ([#1081](https://github.com/cashubtc/cashu-ts/issues/1081)). It stays strict for swap and mint outputs.

crodas:
This is blocked until https://github.com/cashubtc/nuts/pull/436 gets resolved

thesimplekid:
I am going to close this as we don't want to sign inactive and we should reopen a new one after more spec discussion, for now we are inline with nutshell.

PR #2463, opened 2026-09-01
refactor(wallet)!: unify the Rust and FFI wallet API

asmogo:
### Description

Replace the manually mirrored Rust/FFI wallet surfaces with one application-facing Rust facade that is exported directly through UniFFI. `cdk::Wallet` remains the advanced protocol engine; `cdk-ffi::portable` becomes the shared product API for Rust, Swift, Kotlin, Python, Dart, and Go.

This is intentionally a breaking redesign. It does not include compatibility shims or migration APIs.

The facade now provides:

* single-mint `Wallet` and multi-mint `CashuWallet` roots;
* workflow-oriented request, session, plan, confirmation, and receipt types;
* durable operation IDs with reconstruction after restart;
* prepare/confirm/cancel flows for sends and payments;
* local/online synchronization and resumable cross-mint transfers;
* stable structured error categories and retryability;
* UI-safe history and redacted debug output that do not expose bearer tokens, proofs, signing material, or payment secrets.

The underlying issue, send, and melt sagas were also hardened so the simpler API is safe to use in real applications. Preparation markers are persisted before reservations, state changes use optimistic claims and wallet-local operation locks, compensations are operation-scoped and error-aware, ambiguous payment outcomes remain recoverable, and arithmetic is checked throughout fee/change accounting.

The portable API checker now validates full signatures and every reachable record/enum shape, preventing either accidental contract growth or legacy protocol objects from leaking back into the facade. The architecture and lifecycle guarantees are documented in `docs/wallet-api.md`.

-----

### Notes to the reviewers

The key boundary is deliberate: common application workflows belong in `crates/cdk-ffi/src/portable.rs`; proof selection, keysets, swaps, and NUT-specific controls remain in the `cdk` engine. There is no second wallet trait to keep synchronized.

Prepared handles are lightweight references to authoritative persisted saga state. Dropping a handle does not release funds; applications persist the operation ID and explicitly confirm, cancel, or synchronize after restart.

A future packaging improvement can publish the portable bindings separately from expert/protocol bindings without introducing another implementation.

Validation completed:

* `nix develop .#stable -c just quick-check`
* `cargo check --workspace --all-targets`
* `just ffi-api-check`
* `just ffi-test`
* focused issue/send/melt/recovery regression suites
* Swift, Kotlin, Dart, Go, and Python binding generation/tests

-----

#### CHANGED

* Replaced the mirrored Rust and FFI wallet APIs with one workflow-oriented portable facade exported directly from Rust through UniFFI.
* Made send, payment, synchronization, and cross-mint workflows durable and restart-safe.

#### ADDED

* Added stable portable error categories, API contract checking, operation reconstruction, and UI-safe history types.

#### REMOVED

* Removed the manually mirrored wallet traits and proof-centric application API surface.

#### FIXED

* Fixed recovery races, orphaned proof/quote reservations, stale state overwrites, secret-bearing debug output, and unchecked wallet amount arithmetic.

----

### Checklist

* [x] If the portable Wallet API changed, I ran `just ffi-api-check`, reviewed the manifest diff, and updated affected binding tests/docs


thesimplekid:
I think we said we would close this a maybe revisit another way. Reopen if not the case.

PR #2466, opened 2026-09-03
Identify mints by internal id

### LINKED ISSUE #2439: Wallet database update_mint_url leaves transactions, melt_quote, and mint rows on the old URL

crodas:
### Description

Fixes #2439

A mint URL is a mutable attribute, not an identity. Keying every wallet table on the URL meant `update_mint_url` had to rewrite each table, and it only rewrote two of them, leaving transactions, melt quotes, keysets and sagas stranded under the old URL after a mint moved.

Wallet tables now reference an internal mint id that never changes, so a move is a single row update. `add_mint` is what creates that identity: writing a row for a URL with no stored mint fails with `UnknownMint`.  Applies to the SQL backends, redb and supabase; the id never leaves the storage layer, so the public trait still speaks `MintUrl` and callers are unchanged.

The id is assigned by the store on insert, as sqlite AUTOINCREMENT, postgres identity and a monotonic counter in redb. No write path needs the generated value back, so nothing has to read it and there is no dialect split. A freed id is never reused: sqlite does not enforce these foreign keys, and reuse would silently reattach a stale row to a new mint.

Migrations backfill a mint row for any URL referenced by an orphan row so no existing data is dropped.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crates/cdk-sql-common/src/wallet/migrations/sqlite/20260902000000_mint_internal_id.sql
```diff
@@ -0,0 +1,104 @@
...
+
+ALTER TABLE transactions ADD COLUMN mint_id INTEGER;
+UPDATE transactions SET mint_id = (SELECT m.id FROM mint m WHERE m.mint_url = transactions.mint_url);
+ALTER TABLE transactions DROP COLUMN mint_url;
+CREATE INDEX IF NOT EXISTS transactions_mint_id_index ON transactions(mint_id);
+
+ALTER TABLE wallet_sagas ADD COLUMN mint_id INTEGER;
+UPDATE wallet_sagas SET mint_id = (SELECT m.id FROM mint m WHERE m.mint_url = wallet_sagas.mint_url);
+ALTER TABLE wallet_sagas DROP COLUMN mint_url;
+CREATE INDEX IF NOT EXISTS wallet_sagas_mint_id_index ON wallet_sagas(mint_id);
```
asmogo, line 102: why are we not using foreign key relations for mint_id here in sqlite ?

asmogo (commented):
one question about foreign keys in the sqlite migration.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-sql-common/src/wallet/migrations/sqlite/20260902000000_mint_internal_id.sql
crodas: Fixed

PR #2471, opened 2026-09-04
Update nostr dep

thesimplekid:
### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cashu/Cargo.toml
+++ b/crates/cashu/Cargo.toml
@@ -14,7 +14,7 @@ readme = "README.md"
 default = ["mint", "wallet"]
 mint = []
 wallet = []
-nostr = ["dep:nostr-sdk"]
+nostr = ["dep:nostr"]
 bench = []
 
 [dependencies]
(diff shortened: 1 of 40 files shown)


crates/cashu/Cargo.toml
```diff
@@ -32,7 +32,7 @@ tracing.workspace = true
 url.workspace = true
 serde_json.workspace = true
 serde_with.workspace = true
-nostr-sdk = { workspace = true, optional = true }
+nostr = { workspace = true, optional = true }
```
crodas, line 35: I think the whole purpose of the cashu crate was to be a simple crate independent of implementation. I guess we need generic self-contained error types, and let have other higher level crates to include the nostr crate.

thesimplekid: The cashu crate has to to implement what is specifically outlined in the nuts.  [NUT-27](https://github.com/cashubtc/nuts/blob/main/27.md) outlines a nostr event and why this dep is used.

thesimplekid:
@cdk-bot review

PR #2473, opened 2026-09-04
Reject unsigned values that do not fit their column

### LINKED ISSUE #2425: SQL layer narrows u64 to i64 with unchecked casts, so one wrapped value makes a whole table unreadable
Across `cdk-sql-common`, `u64` values are bound to the database with unchecked `as i64` casts, while the read path converts back with a checked `try_into`. The asymmetry is the problem. A value above `i64::MAX` wraps to a negative number, the write succeeds and commits, and every later read of that column fails, because the reader rejects the negative it just stored.

The damage is not confined to the bad row. Readers `collect` into a `Result<Vec<_>, _>`, so one undecodable row fails the entire query. A single bad write can make a whole table unreadable until the row is removed by hand.

This affects **both SQLite and Postgres**, since the casts live in the shared crate. Postgres has a second, worse tier on top of it (below).

Here is a list of non amount i64 to u64 casts that should be addressed. We either do the conversion in rust with try_into *or* we store them texts

| Site | Field(s) | Rust type | Value source | Risk |
|---|---|---|---|---|
| `cdk-sql-common/src/wallet/mod.rs:1211` | `input_fee_ppk` | `u64` | Mint's `/v1/keysets` response, unbounded | **High** — remote mint controls it; read checked at `:2054`, collected at `:290`, so one row breaks every keyset read for that mint |
| `cdk-sql-common/src/wallet/mod.rs:1212` | `final_expiry` | `Option<u64>` | Mint's `/v1/keysets` response, unbounded | **High** — same path, read checked at `:2055` |
| `cdk-supabase/src/wallet.rs:2562` | `input_fee_ppk` | `u64` → `i64` field | Mint's `/v1/keysets` response | Low — `as i64` out, `as u64` back at `:2578` round-trips by two's complement, but stores a negative other readers see as garbage |
| `cdk-sql-common/src/keyvalue.rs:84,85,121,122,190,276,393,394` | `created_time`, `updated_time` | `u64` | `unix_time()` | Low — clock-derived |
| `cdk-sql-common/src/mint/saga.rs:142,143,165,209` | `created_at`, `updated_at` | `u64` | `unix_time()` | Low — clock-derived |
| `cdk-sql-common/src/mint/quotes.rs:730,927,948,971,1022,1025-1027,1073,1085,1086,1143,1296,1297` | `expiry`, `created_time`, `updated_at`, `last_checked`, `timestamp`, `paid_time`, `threshold` | `u64` | `unix_time()`, mint-set expiries, payment processor | Low — not caller-controlled. Note `last_checked` is compared numerically at `:1293`, so it cannot move to text storage |
| `cdk-sql-common/src/mint/signatures.rs:130,131,171` | `created_time`, `signed_time` | `u64` | `unix_time()` | Low — clock-derived |
| `cdk-sql-common/src/mint/proofs.rs:180` | `created_time` | `u64` | `unix_time()` | Low — clock-derived |
| `cdk-sql-common/src/wallet/mod.rs:940,1158,1265,1273,1345,1483,1484,1564,1931` | `timestamp`, `mint_time`, `expiry`, `updated_at`, `created_at`, `created_time` | `u64` | `unix_time()`, mint-set expiries | Low — not caller-controlled |
| `cdk-sql-common/src/mint/signatures.rs:132`, `mint/quotes.rs:733` | `order_index` | `usize` | `enumerate()` | None — not a `u64` |
| `cdk-sql-common/src/wallet/mod.rs:1275-1277,1350-1352,1486,1566,1567` | `version`, `new_version`, `expected_version` | `u32` | optimistic lock counter | None — widens losslessly |

crodas:
### Description

Fixes #2425 and supersedes #2427 (as [suggested](https://github.com/cashubtc/cdk/pull/2427#pullrequestreview-5112564514) by @thesimplekid with a better and wider solution)

Amounts are u64 on the wire, but no column could hold one. On Postgres most integer columns were declared int4, and the driver overrides to_sql_checked, so the range guard tokio-postgres normally applies never ran: anything above two billion was truncated before reaching the server. SQLite's only integer is signed 64 bit, so the top half of the u64 range had no representation at all.

The bind path made it worse. It narrowed with an unchecked cast, so a value past the signed maximum wrapped to a negative value and was committed. The read path is checked, so that one bad row then failed every query over its table, not just its own row. Amounts were bound through a conversion that returned an Option, so an oversized amount landed as NULL on a nullable column and was erased silently. The wallet balance aggregate read the other direction with the same lack of care, turning a negative sum into an enormous positive balance.

#### What changed

**Representation**: Amount columns become numeric(20, 0) on Postgres and eight big-endian bytes on SQLite. Big endian is what keeps memcmp over equal-length blobs in numeric order, so equality, ordering, and indexes need no cast. Everything that is not an amount stays an integer, widened to int8 on Postgres where it had outgrown int4 (unix timestamps, keyset valid_from/valid_to/input_fee_ppk, melt quote estimated_blocks/selected_fee_index). Integer columns produced locally and bounded by construction are left alone.

**Types carry the decision**: Value gains Unsigned(u64) and Amount(u64). Conversion into Value has to stay infallible, so an unsigned value is carried unchanged until the statement is rendered, where the placeholder is known and an out-of-range bind can be named in the error and refused. Amount stays distinct all the way to the driver because SQLite cannot tell at bind time which column a bare u64 is headed for. Amount::to_i64 has no caller left and is removed.

**Arithmetic**: The two totals computed in SQL go through u64_add and u64_sum, named the same on both backends so one statement works everywhere. On SQLite, they are registered functions that refuse overflow; on Postgres, they alias arithmetic the server already does, with a check constraint on keyset_amounts refusing a running total past u64::MAX.

**Postgres numeric codec**: tokio-postgres maps numeric to nothing on its own, and the arbitrary-precision crates that fill the gap are a large dependency for the one shape this driver needs. cdk-postgres/src/numeric.rs encodes and decodes a non-negative integer no wider than u64, rejecting negative, fractional, NaN, and oversized values.

#### Migrations

One migration per store (mint, wallet, auth) on each backend, 20260906000000_v0_18_column_types. Postgres changes every column of a table in one ALTER so no table is rewritten twice; SQLite has no ALTER COLUMN, so each table is rebuilt. The branch's earlier separate int4 → int8 widening migrations are folded in, since taking a column to int8 and then to numeric rewrote the same table twice.

#### Tests

Backend-agnostic cases in the shared database test suite, so both drivers run them: keyset u64 column bounds, onchain melt quote u32 bounds, mint quote amounts across the whole u64 range, an issued total summing past the signed range, and a wallet balance doing the same.

#### Dependencies

rusqlite gains the functions feature; cdk-postgres and cdk-sqlite add thiserror.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crates/cdk-sql-common/src/stmt.rs
```diff
@@ -340,6 +340,30 @@ impl Statement {
         self
     }
 
+    /// Binds a given placeholder to a `u64`, or to NULL when it is absent.
```
thesimplekid: Instead of adding a new fn could we add a new variant to the enum https://github.com/crodas/cdk/blob/5ac597edaf94fc9bdced2628d9df3bc58fd2d57a/crates/cdk-sql-common/src/value.rs#L4? This way we do not have to patch every call site and forgetting one now or when adding one in the future is not possible.

crates/cdk-sql-common/src/wallet/migrations/postgres/20260905000000_widen_wallet_integer_columns.sql
thesimplekid, line 1: This comment goes for the mint as well. I think we do not need to migrate all columns to bigint some just don't need it for examples estimated_blocks, version and I think there are more. I think it would be worth being more precise and not just always ing BITINT?

thesimplekid (commented):
Overall I think a good solution just a few comments

crates/cdk-sql-common/src/wallet/migrations/postgres/20260905000000_widen_wallet_integer_columns.sql
crodas: The migration is based on the rust types, but I will over them again

crodas:
@cdk-bot review

crodas:
@cdk-bot review

thesimplekid:
**Fee-index column mismatch** — [mint migration, line 6](https://github.com/cashubtc/cdk/blob/8210481cd51acc27dabe885d41cd31eae88f41a7/crates/cdk-sql-common/src/mint/migrations/postgres/20260905000000_widen_mint_integer_columns.sql#L6)

The wallet’s `melt_quote.fee_index` is widened to `BIGINT`, but the mint’s corresponding `selected_fee_index` remains `INTEGER`. Should we align these? They represent the same backend-assigned `u32` ID, so the wallet can now persist a selection that the mint cannot. A PostgreSQL round-trip test with an ID above `i32::MAX` would cover this.

**Storage and migration impact** — [mint migration, line 22](https://github.com/cashubtc/cdk/blob/8210481cd51acc27dabe885d41cd31eae88f41a7/crates/cdk-sql-common/src/mint/migrations/postgres/20260905000000_widen_mint_integer_columns.sql#L22)

I think widening these columns is justified for correctness. The storage and upgrade costs deserve documenting:

- Mint proofs: approximately **8 extra bytes per row**.
- Mint blind signatures: **12 extra bytes per row**.
- Mint and melt quotes: **20 extra bytes per row**.
- Wallet proofs: **4 extra bytes per row**.
- Wallet transactions: **12 extra bytes per row**.

These assume populated fields and exclude alignment and index overhead. More significantly, upgrading existing PostgreSQL tables requires rewrites and exclusive locks, with temporary disk space needed for the rebuilds. Could we validate the migration against a realistically populated database and document the expected operational impact? I support the change; this would help operators plan the upgrade.

crates/cdk-sql-common/src/wallet/mod.rs
```diff
@@ -610,7 +610,8 @@ where
             .await
             .map_err(|e| Error::Database(Box::new(e)))?;
 
-        let mut query_str = "SELECT COALESCE(SUM(amount), 0) as total FROM proof".to_string();
+        let mut query_str =
+            "SELECT CAST(COALESCE(SUM(amount), 0) AS BIGINT) as total FROM proof".to_string();
```
asmogo: Should the i64::MAX limit apply only to individual proof amounts, or also to the total wallet balance?
Two proofs of 2^62 each pass the new write checks, but their combined balance is 2^63. That fits the u64 returned by get_balance, yet this CAST(... AS BIGINT) makes PostgreSQL reject the query.

crates/cdk-sql-common/src/mint/signatures.rs
```diff
@@ -141,7 +141,7 @@ where
                         DO UPDATE SET total_issued = keyset_amounts.total_issued + EXCLUDED.total_issued
```
asmogo: This can still overflow the total_issued column, even if we check the input values. We might need to check smth. like `existing_total <= i64::MAX - incoming_amount`

asmogo (changes requested):

crates/cdk-sql-common/src/mint/signatures.rs
crodas: Yes, I know. But we can't fix it unless we do all the math in Rust.

crodas:
There are two paths here: either we merge the current approach as a half-baked solution, or I prepare a more sustainable one.

The sustainable approach would be:

1. Use PostgreSQL NUMERIC(20, 0) for the full u64 range, with validation when converting values back to u64.
2. Store u64 values as decimal TEXT in SQLite and provide Rust-backed SQLite functions for checked arithmetic.

That gives us exact u64 semantics on both backends without relying on SQLite's signed i64 arithmetic or risking promotion to floating-point.

The SQLite side would look roughly like this:

```rust
use rusqlite::{
    functions::FunctionFlags,
    Connection,
    Error,
    Result,
};

fn register_u64_functions(conn: &Connection) -> Result<()> {
    conn.create_scalar_function(
        "u64_add",
        2,
        FunctionFlags::SQLITE_UTF8 | FunctionFlags::SQLITE_DETERMINISTIC,
        |ctx| {
            let a = ctx.get::<String>(0)?;
            let b = ctx.get::<String>(1)?;

            let a: u64 = a.parse().map_err(user_error)?;
            let b: u64 = b.parse().map_err(user_error)?;

            let result = a
                .checked_add(b)
                .ok_or_else(|| user_error("u64 overflow"))?;

            Ok(result.to_string())
        },
    )?;

    Ok(())
}

fn user_error<E>(err: E) -> Error
where
    E: std::error::Error + Send + Sync + 'static,
{
    Error::UserFunctionError(Box::new(err))
}
```

crates/cdk-sql-common/src/mint/signatures.rs
crodas: Fixed in f874022e7e2c101918e63063a3807e998a6c8245

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
This solves what https://github.com/cashubtc/cdk/pull/1954 tried to solve, elegantly and efficiently.

crodas:
@cdk-bot review

thesimplekid:
### Correctness issue: SQLite upgrade fails with existing quote references

The [SQLite migration](https://github.com/cashubtc/cdk/blob/d5b05c6022e317fb5cc89fcc7b3705a3c713b4b2/crates/cdk-sql-common/src/mint/migrations/sqlite/20260906000000_v0_18_column_types.sql#L170) drops `mint_quote` while existing payment and issuance rows still reference it. With foreign keys enabled, this fails with `FOREIGN KEY constraint failed` and prevents the mint from starting. Dropping `melt_quote` similarly fails when a melt request exists.

Please handle foreign keys across the rebuild and add an upgrade test using the actual migration with populated, related tables. I reproduced all three cases.

thesimplekid:
### Recommendation: split out the checked-conversion fix

Could we split out the checked-conversion fix and merge it first, without changing storage types? Rejecting values that exceed the existing columns' ranges fixes silent truncation without requiring migrations.

That smaller change should retain the public `Amount::to_i64()` methods and adjust tests to the existing column limits. PostgreSQL widening and full-`u64` storage can then be evaluated separately, with migration tests and upgrade guidance.

PR #2476, opened 2026-09-06
Fix/mint subscription backfill

thesimplekid:
### Description

-----

 Fix three gaps in NUT-17 subscription backfills that prevent wallets from recovering after reconnecting:

  - Missing melt change: Load persisted fee-return signatures and include them in settled melt quote notifications.
  - Missing custom quote snapshots: Send the current state for custom mint and melt subscriptions, including paid amounts, payment proofs, and change.
  - Missed payment broadcasts: Notify all subscribers when a backfill discovers and commits an incoming payment. Previously, concurrent subscribers could remain stuck on UNPAID.
  
  Found by project Loupe

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/mint/subscription.rs
+++ b/crates/cdk/src/mint/subscription.rs
@@ -2,7 +2,7 @@
 
 use std::collections::HashMap;
 use std::ops::Deref;
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 
 use cdk_common::common::PaymentProcessorKey;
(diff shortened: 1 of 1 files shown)


thesimplekid:
@cdk-bot review

crodas (approved):
Although I approve it as-is since it fixes a bug now, I think we need a simpler solution: introduce a past notification table that we can easily read from.

This will avoid maintaining the huge function `get_events_from_db` which reads from multiple sources.

PR #2482, opened 2026-09-07
fix(lnd): reject expired invoices before payment dispatch

thesimplekid:
Return a definitive failure for already-expired invoices so melts do not remain pending when LND rejects them without creating a payment. Preserve existing paid or pending results before checking expiry.

Add regression tests for expiry rejection, valid invoices, and preservation of existing payments.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2485, opened 2026-09-07
fix(bdk): delay wallet DB creation until initial checkpoint succeeds

### LINKED ISSUE #2467: BDK mint won't restart after failed first boot
## Summary

If a BDK mint's **first** startup fails because the node is unreachable, it leaves behind an empty `bdk_wallet.sqlite`. Every restart after that is rejected by the startup check, even once the node is healthy again.

**Version:** 0.18.0 · **Config:** `[bdk] chain_source_type = "bitcoinrpc"`
**Impact:** availability only, operator friction.

## Steps to reproduce

1. Set up a new BDK mint (`cdk-mintd config init --new-mint --allow-new-bdk-wallet`).
2. Stop Bitcoin Core, or point `bitcoind_rpc_port` at a dead port.
3. Start `cdk-mintd` → exits with `Failed to determine the Bitcoin Core chain tip for a fresh wallet`. Expected.
4. Start Bitcoin Core, verify it's reachable, start `cdk-mintd` again.

**Expected:** the mint creates its wallet and starts.
**Actual:** `Persisted BDK wallet database at .../bdk_wallet.sqlite does not contain an initialized wallet`

## Workaround

Delete `<work_dir>/bdk_wallet/bdk_wallet.sqlite` and restart. Simple, but nothing in the error makes this manual repair step obvious.

j-kon:
### Description

Fixes #2467.

A fresh BDK mint could fail to restart after an interrupted first boot if the configured Bitcoin node was unreachable.

`CdkBdk::new` previously created `bdk_wallet.sqlite` before fetching the initial checkpoint from the chain source. If checkpoint retrieval failed, initialization returned an error but could leave an empty SQLite database behind. On a later `cdk-mintd` startup, the existing-wallet preflight detected the uninitialized database and rejected startup, requiring the operator to manually delete the file.

This PR changes the **fresh-wallet path** in `CdkBdk::new` so filesystem/database creation happens only after `initial_checkpoint()` succeeds:

- When no wallet database exists yet, an unreachable chain source fails before `bdk_wallet.sqlite` is created.
- Retrying the same fresh setup after the chain source becomes available can initialize normally.
- Fresh database creation uses `OpenOptions::create_new(true)` so a file that appears concurrently is never overwritten.
- If initialization fails after the current invocation created the fresh database file, cleanup removes only that newly-created file.
- Databases that existed before `CdkBdk::new` preserve the constructor behavior from `main` and are never removed by the fresh-wallet cleanup path.

`cdk-mintd`'s existing-wallet preflight behavior is unchanged.

-----

### Notes to the reviewers

The implementation is strictly scoped to:

`crates/cdk-bdk/src/lib.rs`

No behavioral changes were made to `cdk-mintd`.

Regression coverage includes:

1. An unreachable Bitcoin RPC node on fresh first boot leaves no wallet database behind.
2. A subsequent fresh initialization can succeed in the same directory.
3. An existing valid wallet loads without querying the initial checkpoint.
4. A pre-existing empty SQLite database retains the constructor behavior from `main`.
5. Pre-existing database files are never removed when initialization fails.
6. Cleanup is limited to a database file created by the current fresh initialization attempt.

Fresh file creation uses `create_new(true)` to avoid accidental overwrite or a create-time race. The overall wallet initialization process is not described as atomic.

-----

#### FIXED

- `cdk-bdk`: prevent a failed fresh first boot from leaving an empty `bdk_wallet.sqlite` that blocks later startup (#2467)

----

### Checklist

### DIFF
--- a/crates/cdk-bdk/src/lib.rs
+++ b/crates/cdk-bdk/src/lib.rs
@@ -365,62 +365,6 @@ impl CdkBdk {
     ) -> Result<Self, Error> {
         chain_source.validate()?;
 
-        let storage_dir_path = PathBuf::from(storage_dir_path);
-        let storage_dir_path = storage_dir_path.join("bdk_wallet");
-        fs::create_dir_all(&storage_dir_path)?;
-
-        let mut db = Connection::open(storage_dir_path.join("bdk_wallet.sqlite"))?;
-
-        let xkey: ExtendedKey = mnemonic.into_extended_key()?;
-        let xprv = xkey.into_xprv(network.into()).ok_or(Error::Path)?;
-
-        let descriptor = Bip84(xprv, KeychainKind::External);
-        let change_descriptor = Bip84(xprv, KeychainKind::Internal);
-
-        let wallet_opt = Wallet::load()
-            .descriptor(KeychainKind::External, Some(descriptor.clone()))
-            .descriptor(KeychainKind::Internal, Some(change_descriptor.clone()))
-            .extract_keys()
-            .check_network(network)
-            .load_wallet(&mut db)
-            .map_err(|e| Error::Wallet(e.to_string()))?;
-
-        // A fresh Bitcoin Core wallet should start at the current tip rather
-        // than scanning from genesis. An explicit rescan height overrides the
-        // tip for seed recovery. Fetch this before creating the wallet so an
-        // unreachable or misconfigured node cannot persist a genesis-pinned
-        // wallet by accident.
-        let initial_checkpoint = match wallet_opt.is_none() {
-            true => chain_source.initial_checkpoint()?,
-            false => None,
-        };
-
-        let mut wallet = match wallet_opt {
-            Some(wallet) => wallet,
-            None => {
-                let mut wallet = Wallet::create(descriptor, change_descriptor)
-                    .network(network)
-                    .create_wallet(&mut db)
-                    .map_err(|e| Error::Wallet(e.to_string()))?;
-
-                if let Some(block_id) = initial_checkpoint {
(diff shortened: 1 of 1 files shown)


j-kon:
The follow-up commit now preserves the existing `CdkBdk::new` behavior for databases that predate the constructor call, while limiting cleanup strictly to files created by the fresh initialization path.

CI and the Nutshell integration suite are green. Ready for review.

j-kon:
@cdk-bot review

j-kon:
Hi @thesimplekid, whenever you have some time, could you please take a look at this PR? All CI checks and the Nutshell test suite are green, and it addresses the failed fresh first boot recovery in #2467.

j-kon:
Thanks for the review! Addressed all three findings in commit `6c016135`:

1. **Dead match arm in `wallet_exists`**: Updated non-regular file matches to return `Error::ExistingWalletNotInitialized { path: wallet_path }`, matching the behavior of `validate_existing_wallet` and preventing opaque SQLite errors when a directory is at the path. Added a test (`directory_at_wallet_path_fails_as_not_initialized`) covering this.
2. **Atomic initialization via temporary file**: Switched the fresh-wallet initialization to build and persist into a temporary SQLite file (`.bdk_wallet.sqlite.tmp-<uuid>`), cleaning up only the temporary file on error, and atomically renaming it into `wallet_path` on success.
3. **Elimination of adoption / cleanup races**: Because no 0-byte placeholder is ever created at `bdk_wallet.sqlite`, concurrent processes will never observe or initialize into an uninitialized placeholder at that path, and failed initializations never unlink `wallet_path`.

PR #2489, opened 2026-09-07
ci(kotlin): publish armeabi-v7a and x86 Android libraries again

kaloudis:
### Description

cdk-android releases on Maven Central shipped four Android ABIs through 0.17.5 (arm64-v8a, armeabi-v7a, x86, x86_64). The rewritten kotlin publish pipeline introduced with 0.18.0-rc.0 only builds `aarch64-linux-android` and `x86_64-linux-android`, so the 32-bit jniLibs silently disappeared from the aar.

This restores them:

* two more legs in the `kotlin-build-native` matrix (`armv7-linux-androideabi`, `i686-linux-android`)
* their entries in the `ANDROID_ABI` map used when distributing native libraries to jniLibs

The `kotlin-build` Nix dev shell already provisions the rust std libraries, NDK linkers and `CC`/`AR` env for both 32-bit targets, so no flake changes are needed. Bindings generation is unaffected; it runs off the arm64 library only.

If dropping 32-bit Android was intentional rather than a side effect of the pipeline rewrite, happy to close this, but in that case it would be good to state it in the release notes so downstream packagers can plan. For context, Zeus currently ships per-ABI release APKs including armeabi-v7a and x86 and is weighing whether to retire them; knowing upstream's direction here would inform that call. 0.17.x aars are also noticeably smaller per ABI since the split cdk-jvm/cdk-android layout carried no bundled bindings, so if aar size was the motivation the 32-bit slices add roughly what the old layout carried per ABI.

-----

### Notes to the reviewers

The 16 KB ELF alignment check is gated to the 64-bit targets: Android's 16 KB page size requirement applies to 64-bit ABIs only, and 32-bit Android always uses 4 KB pages, so the check would fail spuriously on the restored targets.

Workflow-only change; I could not exercise the release pipeline from a fork since it needs the deploy key and repo vars, but the YAML parses and the matrix/map shapes mirror the existing 64-bit legs.

-----

#### ADDED

- cdk-android aars on Maven Central again include armeabi-v7a and x86 native libraries, matching the 0.17.x ABI coverage

#### FIXED

----

### Checklist

* [ ] I ran `just quick-check` before committing (workflow-only change, no Rust code touched)
* [x] If the Wallet API was modified (added/removed/changed), I have reflected those changes in the FFI bindings (`crates/cdk-ffi`) (not applicable)

### DIFF
--- a/.github/workflows/kotlin-publish.yml
+++ b/.github/workflows/kotlin-publish.yml
@@ -234,6 +234,12 @@ jobs:
           - target: aarch64-linux-android
             runner: ubuntu-22.04
             artifact: libcdk_ffi_kotlin.so
+          - target: armv7-linux-androideabi
+            runner: ubuntu-22.04
+            artifact: libcdk_ffi_kotlin.so
+          - target: i686-linux-android
+            runner: ubuntu-22.04
+            artifact: libcdk_ffi_kotlin.so
           - target: x86_64-linux-android
             runner: ubuntu-22.04
             artifact: libcdk_ffi_kotlin.so
@@ -313,6 +319,9 @@ jobs:
             bash -c '"$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin/llvm-strip" --strip-all '"$FILE"
 
       - name: Verify Android ELF 16 KB alignment
+        # The 16 KB page size requirement applies to 64-bit ABIs only;
+        # 32-bit Android always uses 4 KB pages
+        if: matrix.target == 'aarch64-linux-android' || matrix.target == 'x86_64-linux-android'
         working-directory: cdk-kotlin/rust
         shell: bash
         run: |
@@ -386,6 +395,8 @@ jobs:
         run: |
           declare -A ANDROID_ABI=(
             ["aarch64-linux-android"]="arm64-v8a"
+            ["armv7-linux-androideabi"]="armeabi-v7a"
+            ["i686-linux-android"]="x86"
             ["x86_64-linux-android"]="x86_64"
           )
 


kaloudis:
Downstream consumer for context: ZeusLN/zeus#4588 (draft) bumps Zeus to 0.18.0 and is currently blocked on the 32-bit question for its armeabi-v7a and x86 release APKs.

kaloudis:
I dug through the history to find where the 32-bit targets went, for anyone landing here with the same question: they were removed in #2223 (`3300cea7`), which rewrote the publish pipeline for the Central Portal. The stated motivation there was deployment surface: six coordinates down to three, the native matrix from eight targets to five, and the projected Central file count from 260 to 56 per release. There's no indication of a technical problem with the 32-bit targets themselves; 0.17.x was still shipping fixes for them a few weeks prior (`1764d320`).

Worth noting for this PR: restoring armeabi-v7a and x86 doesn't touch that motivation. The 32-bit `.so` files are bundled inside the existing cdk-android aar, so this adds zero Maven coordinates and zero files to the Central deployment. The cost is two more legs in the `kotlin-build-native` matrix and a larger aar (roughly what the 0.17.x layout carried per ABI, as noted above).

So the open question is just whether dropping 32-bit Android was deliberate policy or a side effect of the rewrite. If it's policy, happy to close this and we'll plan the deprecation downstream; a line in the release notes would help other packagers do the same. If it was a side effect, this should be everything needed to restore parity with 0.17.x.

kaloudis:
One more option to put on the table, as a middle ground: I can trim this PR to restore `armv7-linux-androideabi` only and leave x86 (`i686-linux-android`) retired.

x86 32-bit Android is effectively emulator-only hardware at this point, and modern emulator images are x86_64. Downstream we see no sign of real x86 users: no x86 devices appear in our Play device catalog, and direct-download numbers for our x86 APK are effectively zero. Zeus is planning to retire its x86 release APK regardless of what happens here.

armeabi-v7a is different: it still has a small but real user base, and not just on decade-old hardware. Current ultra-budget phones (Redmi A3, Galaxy A13 class) ship 32-bit firmware on 64-bit silicon and can only run the armv7 build.

So the compromise would be one extra build leg instead of two, no change to Maven coordinates or Central file count, and the only published ABI with actual users restored. Happy to update the PR to that shape if it's preferable to full 0.17.x parity.

thesimplekid:
Sorry you ran into this. We cut some because of limits being imposed by maven central (size, file count and releases per month), however we cut a bit too aggressively and did not communicate it well and don't want to cut the android support or force you to rush a decision. 

We have room within our limits for both `armv7-linux-androideabi` and `i686-linux-android` I believe, so I think what would be best is bring back both in an 0.18.1 cdk release this way you do not have to change your targets now based on us and we can follow you in removing `i686-linux-android` in a future release with more notice and communication. Would this work for you? 

@crodas Can you check this PR is it everything we need to bring it back?

kaloudis:
> Sorry you ran into this. We cut some because of limits being imposed by maven central (size, file count and releases per month), however we cut a bit too aggressively and did not communicate it well and don't want to cut the android support or force you to rush a decision.
> 
> We have room within our limits for both `armv7-linux-androideabi` and `i686-linux-android` I believe, so I think what would be best is bring back both in an 0.18.1 cdk release this way you do not have to change your targets now based on us and we can follow you in removing `i686-linux-android` in a future release with more notice and communication. Would this work for you?
> 
> @crodas Can you check this PR is it everything we need to bring it back?

Yep, that would be great. We'll skip the v0.18.0 upgrade and wait for v0.18.1.

crodas:
@thesimplekid Everything the build needs is already in place, so the two matrix legs are genuinely all that was missing

crodas (approved):

crodas:
@thesimplekid Another thing: dart-publish.yml:253-260 builds arm64/armv7/x86_64 for Android and has never included i686. Do we add support for it there as well?

thesimplekid:
> @thesimplekid Another thing: dart-publish.yml:253-260 builds arm64/armv7/x86_64 for Android and has never included i686. Do we add support for it there as well?

Sure we don't have the limit issues on dart. if its just another build. Can be another pr

thesimplekid (approved):

PR #2490, opened 2026-09-07
refactor(sql): batch mint quote related data loading

### LINKED ISSUE #2435: Refactor: Use JOIN to get Mint Quotes `payments` and `issuance`
In `/crates/cdk-sql-common/src/mint/quotes.rs`, functions that return's Mint Quote execute more queries to get `payments` and `issuance` from Mint Quotes. 

We can use `JOIN` operators to get everything that we need in just one query, instead execute more queries in different functions.

j-kon:
### Description

Closes #2435

Eliminates the N+1 query pattern when fetching mint quotes in `cdk-sql-common` by batch loading related child records (`mint_quote_payments` and `mint_quote_issued`) using constant queries instead of looping per quote.

Before: $1 + 2N$ queries for $N$ quotes.  
After: $3$ queries constant (1 for parent quotes, 1 for payments belonging to those quotes, 1 for issuances belonging to those quotes).

Key details:
- Avoids multi-table JOINs and Cartesian row multiplication ($P \times I$ rows) between payments and issuance tables.
- Directly uses the loaded parent quote's `CurrencyUnit` for child payments and issuances, avoiding redundant joins back to `mint_quote`.
- Groups child records by `quote_id` in Rust (`HashMap<String, Vec<...>>`) and attaches them to each `MintQuote`.
- Preserves input order, missing `None`, and repeated ID semantics in `get_mint_quotes_inner`.
- Preserves single-quote locking semantics: `SELECT ... FOR UPDATE` acquires row locks on `mint_quote` first, and relations are read while the lock is held.
- Keeps `crates/cdk-sqlite/src/async_sqlite.rs` unchanged.

-----

### Notes to the reviewers

- Added comprehensive database integration tests in `crates/cdk-common/src/database/mint/test/mint.rs` (`mint_quotes_load_payments_and_issuance_without_duplicates`) covering:
  1. Zero payments and zero issuance
  2. Multiple payments only
  3. Multiple issuance only
  4. Multiple payments and multiple issuance (3 payments, 2 issuances)
  5. Cross-quote isolation (ensuring quotes do not mix related records)
  6. Requested ID ordering preservation
  7. Missing IDs returning `None`
  8. Transactional `get_mint_quote` relations loading with row locking

-----

#### CHANGED
- Batch load mint quote payments and issuance in `cdk-sql-common` to resolve N+1 queries (#2435).

----

### Checklist

### DIFF
--- a/crates/cdk-sql-common/src/mint/quotes.rs
+++ b/crates/cdk-sql-common/src/mint/quotes.rs
@@ -119,6 +119,80 @@ where
     .collect()
 }
 
+fn rows_to_payments_map(
+    rows: Vec<Vec<Column>>,
+) -> Result<HashMap<String, Vec<RawQuotePayment>>, Error> {
+    let mut map: HashMap<String, Vec<RawQuotePayment>> = HashMap::new();
+    for row in rows {
+        let quote_id = column_as_string!(&row[0]);
+        let payment_id = column_as_string!(&row[1]);
+        let timestamp: u64 = column_as_number!(row[2].clone());
+        let amount: u64 = column_as_number!(row[3].clone());
+
+        map.entry(quote_id).or_default().push(RawQuotePayment {
+            payment_id,
+            timestamp,
+            amount,
+        });
+    }
+
+    Ok(map)
+}
+
+fn rows_to_issuance_map(
+    rows: Vec<Vec<Column>>,
+) -> Result<HashMap<String, Vec<RawQuoteIssuance>>, Error> {
+    let mut map: HashMap<String, Vec<RawQuoteIssuance>> = HashMap::new();
+    for row in rows {
+        let quote_id = column_as_string!(&row[0]);
+        let amount: i64 = column_as_number!(row[1].clone());
+        let timestamp: u64 = column_as_number!(row[2].clone());
+
+        map.entry(quote_id)
+            .or_default()
+            .push(RawQuoteIssuance { amount, timestamp });
+    }
+
+    Ok(map)
+}
+
+fn attach_relations_to_quotes<'a, I>(
+    quotes: I,
+    mut payments: HashMap<String, Vec<RawQuotePayment>>,
+    mut issuance: HashMap<String, Vec<RawQuoteIssuance>>,
+) where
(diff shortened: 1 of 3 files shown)


crates/cdk-sql-common/src/mint/quotes.rs
```diff
@@ -62,46 +74,235 @@ where
...
     .bind("quote_id", quote_id.to_string())
     .fetch_all(conn)
     .await?
     .into_iter()
     .map(|row| {
         let time: u64 = column_as_number!(row[1].clone());
-        let unit = column_as_string!(&row[2], CurrencyUnit::from_str);
         Ok(Issuance::new(
             Amount::from_i64(column_as_number!(row[0].clone()))
                 .expect("Is amount when put into db")
```
crodas, line 114: Can we throw an error instead of panicking?

crates/cdk-sql-common/src/mint/quotes.rs
```diff
@@ -62,46 +74,235 @@ where
...
+                })
+                .collect();
+        }
+        if let Some(raw_issuances) = issuance.remove(&quote_id_str) {
+            quote.issuance = raw_issuances
+                .into_iter()
+                .map(|i| {
+                    Issuance::new(
+                        Amount::from_i64(i.amount)
+                            .expect("Is amount when put into db")
```
crodas, line 186: Same here

crodas (changes requested):
Overall looks good,  *but* we cannot panic outside of tests

thesimplekid:
@cdk-bot review

j-kon:
Hi @crodas, thanks for the review!

In commit `e0f8789`, I have addressed both points:
- Replaced `.expect(...)` with error propagation via `Amount::from_i64(...).ok_or_else(|| ConversionError::InvalidConversion(...))` in `get_mint_quote_issuance` and `rows_to_issuance_map`.
- `RawQuoteIssuance` now holds the validated `Amount`, making `attach_relations_to_quotes` completely infallible without panics.
- Aligned the test fixture comment on `quote4` per `cdk-bot`'s note.

All unit, SQLite, and clippy checks pass cleanly.

PR #2491, opened 2026-09-07
fix(mintd): enforce seed length only for new mints

thesimplekid:
Require at least 32 resolved seed bytes during new-mint initialization. Allow nonempty legacy seeds when importing existing mints, validating configuration, applying updates, and restarting. Emit an import warning that tells operators to preserve the original seed and enable warning output for config init.

Restore compatibility for Nutshell migrations and CDK 0.17.x upgrades without changing seed bytes or key derivation. Document the 0.18.1 upgrade behavior and cover the initialization boundary and legacy seed lifecycle, including persisted mint state.

Validation: 187 mintd library tests, the updated persisted-state test, cargo check for the mintd binary, and cargo fmt.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

crates/cdk-mintd/src/config_service.rs
```diff
@@ -235,6 +247,29 @@ impl ConfigurationService {
         if database_pubkey.is_some_and(|pubkey| pubkey != signing_identity.pubkey) {
             return Err(ConfigurationServiceError::SigningIdentityChange);
         }
+        if let Some(seed) = resolved
+            .settings
+            .info
+            .seed
+            .as_ref()
+            .filter(|seed| seed.len() < MIN_NEW_MINT_SEED_BYTES)
```
asmogo: I think we should skip this validation if `resolved.settings.enabled_signatory().is_some()`

```suggestion
           .filter(|seed| {
    resolved.settings.enabled_signatory().is_none()
        && seed.len() < MIN_NEW_MINT_SEED_BYTES
})
```

asmogo (commented):
One small finding that might cause issues with remote signatories.

PR #2492, opened 2026-09-07
Fix releases ffi

thesimplekid:
### Description

-----
 All FFI release recipes now dispatch with --ref v<VERSION>, ensuring backport releases use the workflow definitions from their release tag instead of main. This also applies when just release triggers FFI publishing automatically.

Adds documentation for workflow selection and rerunning releases with workflow fixes from a maintenance branch.

Validation: dry-run checks passed for just release and all five FFI release recipes.
### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/bindings/README.md
+++ b/bindings/README.md
@@ -122,6 +122,26 @@ This runs the **FFI - Publish All Bindings** GitHub Actions workflow
 The `release` just recipe calls `ffi-release-all` automatically after publishing
 Rust crates.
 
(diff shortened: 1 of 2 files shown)


crodas (commented):
LGTM, I will approve once we have a green pipeline

thesimplekid:
> I will approve once we have a green pipeline

Worth noting none of these changes are actually tested by the pipeline as they only effect releases. But good practice anyway.

PR #2499, opened 2026-09-08
fix(wallet): preserve exact recipient amounts after redemption fees

### LINKED ISSUE #2493: prepare_pay_request can underfund receiver fees: 21-sat request redeems for 19 sats
A 21-sat Cashu payment request paid through CDK 0.18.0 can produce a token that credits the recipient only **19 sats** when redeemed at a mint with `input_fee_ppk = 1000`. Payment preparation, confirmation, and HTTP delivery succeed; the shortfall appears when a separate wallet redeems the delivered proofs.

This was reproduced with both the Swift and Kotlin native bindings while adding payment integration tests in https://github.com/cashubtc/wallet/pull/360.

**Environment**

- CDK Swift binding: `0.18.0`.
- CDK Android/Kotlin binding: `0.18.0`.
- Mint: Nutshell `cashu==0.20.1`, sat unit, `MINT_INPUT_FEE_PPK=1000`.
- A real Nutshell ledger signs, checks, and swaps proofs. A controlled FakeWallet backend supplies only the initial invoice payment.
- Transport: NUT-18 HTTP POST; no transport faults are injected in this reproduction.
- Independent payer/receiver seeds and SQLite stores; payer initially funded with 100 sats.

**Steps**

1. Fund the payer with 100 sats at the fee-charging mint.
2. Create a Cashu Request for 21 sats, restricted to that mint, with an HTTP POST transport. An amountless request with a custom payment amount of 21 has also exhibited the problem.
3. Call `prepare_pay_request` / `preparePayRequest`, then confirm the prepared payment.
4. Receive the delivered payment payload and reconstruct its token without changing its mint, unit, or proofs.
5. Redeem that token using the independent recipient wallet and check the amount actually credited.

**Expected**

The receiver obtains the full requested 21 sats after redemption fees. `prepare_pay_request` passes `include_fee: true` to `prepare_send`, and `PreparedPaymentRequest::send_fee()` is documented as the fee added so the receiver obtains the full payment amount:

https://github.com/cashubtc/cdk/blob/v0.18.0/crates/cdk/src/wallet/payment_request.rs#L128

https://github.com/cashubtc/cdk/blob/v0.18.0/crates/cdk/src/wallet/payment_request.rs#L310

**Observed**

Sanitized assertions from independent runs:

```text
Swift: XCTAssertEqual failed: ("19") is not equal to ("21")
Kotlin: Payment request must include the receiver's redemption fee expected:<21> but was:<19>
```

The equivalent zero-input-fee control delivers and redeems 21 sats. The shortfall is not identical on every run; proof selection can affect the outcome. This is therefore a demonstrated failing case, not a claim that every 21-sat request underpays.

**Runnable reproduction**

The tests and fixture are pinned at wallet commit `a8599d3b8168db57dd673730087d89905dc7a71c`:

- [Swift regression and shared reproduction](https://github.com/cashubtc/wallet/blob/a8599d3b8168db57dd673730087d89905dc7a71c/CI/IntegrationTests/Tests/PaymentSafetyIntegrationTests.swift#L400)
- [Android regression and shared reproduction](https://github.com/cashubtc/wallet/blob/a8599d3b8168db57dd673730087d89905dc7a71c/android/app/src/androidTest/java/com/cashu/me/liveintegration/PaymentExtendedLocalMintTest.kt)
- [Fixture/runbook, including Android execution](https://github.com/cashubtc/wallet/blob/a8599d3b8168db57dd673730087d89905dc7a71c/CI/payment-tests/README.md)

On macOS with Swift/Xcode and Python 3.10–3.12 available, from a checkout of that wallet commit:

```sh
./CI/setup-nutshell.sh
./CI/start-payment-fixtures.sh

PAYMENT_FIXTURE_URL=http://127.0.0.1:3341 \
PAYMENT_TEST_TIER=full \
PAYMENT_KNOWN_REGRESSIONS=1 \
swift test --package-path CI/IntegrationTests \
  --filter PaymentExtendedIntegrationTests/testCashuRequestReceiverFeeRegression

./CI/stop-payment-fixtures.sh
```

The loopback address above belongs to the local test fixture. The reproduction does not require a Lightning node or real funds. The regression tests retain the intended `received == 21` assertion and are opt-in, so the existing CI pass does not imply this defect is fixed.

The `include_fee` calculation and the final delivered proof denominations/count seem worth inspecting, but the Rust-level root cause has not yet been isolated. A regression test should verify the recipient's **net redeemed amount**, covering both fixed and amountless requests across different proof selections and nonzero input fees.

thesimplekid:
Plan fee-inclusive exact sends from their output proof count and preserve those denominations through confirmation. Validate the final net amount so payment requests deliver the requested value after redemption fees.

Add coverage for fee rates, proof counts, and independent recipient redemption of fixed and amountless payment requests. Document broader follow-up work.

Fixes https://github.com/cashubtc/cdk/issues/2493

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1545,6 +1545,7 @@ version = "0.18.0"
 dependencies = [
  "anyhow",
  "async-trait",
+ "axum",
  "bip39",
  "bitcoin",
  "bitcoincore-rpc",
--- a/crates/cdk-common/src/wallet/mod.rs
+++ b/crates/cdk-common/src/wallet/mod.rs
@@ -463,7 +463,10 @@ pub struct SendOptions {
     pub memo: Option<SendMemo>,
     /// Spending conditions
     pub conditions: Option<SpendingConditions>,
-    /// Amount split target
+    /// Amount split target.
+    ///
+    /// Preparation replaces this with the complete outgoing denomination plan
+    /// for fee-inclusive exact sends; confirmation preserves that plan.
     pub amount_split_target: SplitTarget,
     /// Send kind
     pub send_kind: SendKind,
--- a/crates/cdk-ffi/src/types/wallet.rs
+++ b/crates/cdk-ffi/src/types/wallet.rs
@@ -223,7 +223,7 @@ pub struct SendOptions {
     pub memo: Option<SendMemo>,
     /// Spending conditions
     pub conditions: Option<SpendingConditions>,
-    /// Amount split target
+    /// Amount split target. Fee-inclusive exact sends prepare an exact denomination plan.
     pub amount_split_target: SplitTarget,
     /// Send kind
     pub send_kind: SendKind,
--- a/crates/cdk-integration-tests/Cargo.toml
+++ b/crates/cdk-integration-tests/Cargo.toml
@@ -61,6 +61,7 @@ getrandom = { version = "0.4", features = ["wasm_js"] }
 uuid = { workspace = true, features = ["js"] }
 
 [dev-dependencies]
+axum = { workspace = true }
 bip39 = { workspace = true, features = ["rand"] }
 anyhow.workspace = true
 cdk-axum = { workspace = true }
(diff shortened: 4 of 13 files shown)


thesimplekid:
@cdk-bot

PR #2504, opened 2026-09-08
fix(lnd): timout on lnd check status

thesimplekid:
### Description

-----

 LND payment-status checks waited for a terminal result, blocking startup recovery—and the HTTP listener—when an outgoing payment remained in flight, such as a held invoice.

Request in-flight updates and return Pending immediately. Also bound recovery’s backend status checks to 10 seconds, deferring unresolved sagas while keeping proofs reserved. Later recovery still finalizes paid payments or compensates authoritative failures.

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2507, opened 2026-09-09
fix(cln): reconnect payment event stream after RPC failures

thesimplekid:
Close the stream and clear its active flag when waitanyinvoice fails so that the mint supervisor reconnects with backoff. The CLN RPC client cannot recover a broken socket, so retrying the same client stalls notifications.

Add a Unix-socket regression test covering disconnects, failed reconnects, resumption from the persisted payment index, and cancellation.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2516, opened 2026-09-10
fix(wallet): validate and persist Nostr payment requests

thesimplekid:
Reject payments with mismatched IDs, units, amounts, mint policies, or spending conditions. Require net receipts to cover applicable NUT-18 method fees using the lowest supported requested method fee.

Persist request state and receive-operation linkage for recovery after restart. Expose request recovery and cancellation through CLI and FFI, and prevent duplicate redemption across repository clones.

Cover validation, fee handling, offline delivery, cancellation, and interrupted receive recovery with regression tests.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk/src/wallet/receive/saga/resume.rs
```diff
@@ -90,14 +90,24 @@ impl Wallet {
...
+                matches!(
+                    tx.status,
+                    TransactionStatus::Completed | TransactionStatus::Failed
+                )
+            }) {
+                self.localstore.delete_saga(saga_id).await?;
+                return Ok(RecoveryAction::Recovered);
+            }
+            tracing::warn!(%saga_id, "Receive inputs missing; retaining uncertain saga");
+            return Ok(RecoveryAction::Skipped);
```
asmogo, line 110: Could we recover the case where proof replacement succeeded but the transaction completion write did not? ReceiveSaga::execute saves the new proofs and removes the inputs before separately marking the transaction Completed. A crash between those writes leaves a pending transaction with no reserved inputs, so this branch returns Skipped on every recovery attempt and the Nostr request stays Receiving.

The funds remain spendable, but the request cannot complete or be cancelled. Could we verify the recorded outputs before completing recovery, or make the two writes atomic? A regression test covering this specific interruption point would help preserve the protection against treating genuinely missing inputs as payment.

PR #2517, opened 2026-09-10
fix(axum): enforce batch bounds before quote lookups

thesimplekid:
Validate batch size and reject duplicate quote IDs before per-quote database work on NUT-29 issuance and status routes. Apply the configured issuance limit to batch status requests as well.

Add regression tests for early rejection and acceptance at the limit.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2518, opened 2026-09-10
fix(bdk): box wallet sync future to simplify Send checks

thesimplekid:
Erase the nested chain-sync future at the supervisor boundary so its Send check does not recurse through every backend future type.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2520, opened 2026-09-10
fix(wallet): stop Nostr payment pump when stream is dropped

thesimplekid:
Cancel the pump on drop and receiver closure, and make every send cancellation-aware so malformed messages and full channels cannot keep relay connections alive after the caller stops listening.

Poll the receiver directly to prevent pending reads from ending the stream prematurely. Add regression tests for shutdown, relay cleanup, blocked sends, and repeated polling.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2530, opened 2026-09-10
Deliver NUT-18 payments over the configured transport

### LINKED ISSUE #2525: NUT-18 HTTP payment delivery ignores wallet proxy/Tor settings
NUT-18 HTTP payment delivery does not use the wallet's configured proxy or Tor transport.

The CLI applies `--proxy` and Tor settings through `WalletRepositoryBuilder`, and the repository uses those settings when constructing wallet connectors. However, `Wallet::deliver_payment_request` creates a new default `cdk_common::HttpClient` for `TransportType::HttpPost` and posts to the request's target URL through that client.

Relevant code:
- [CLI transport configuration](https://github.com/cashubtc/cdk/blob/530ab618/crates/cdk-cli/src/main.rs#L236-L258)
- [HTTP payment delivery](https://github.com/cashubtc/cdk/blob/530ab618/crates/cdk/src/wallet/payment_request.rs#L457-L465)

### Expected behavior

HTTP payment delivery should use the wallet's configured transport, including proxy/Tor routing, and return an error if that transport cannot deliver the request.

### Actual behavior

Delivery uses a separate default client. With direct network access available, the recipient can see the payer's ordinary network egress address even though the wallet was configured to use a proxy or Tor. This also means delivery does not inherit the configured transport's redirect behavior.

### Conditions to reproduce

1. Configure the CLI with a working `--proxy`, or use a build with the `tor` feature and Tor enabled.
2. Pay and confirm a valid NUT-18 request offering only HTTP transport, targeting a server whose connection logs can be inspected.
3. Observe that mint requests use the configured transport, while the recipient POST uses the default HTTP client.

Tor defaults to on only in builds with the `tor` feature. OS-level VPNs or enforced network routing still apply. This report is based on source inspection; the network reproduction above has not been run.

### Suggested fix

Reuse the wallet connector's existing transport for recipient delivery instead of constructing a new client. A narrow delivery method on `MintConnector` could expose this, with an unsupported error as the default for custom connectors rather than a direct-network fallback.

The transport needs a JSON POST operation returning a raw response so that successful empty responses remain accepted. Recipient requests must not include mint authentication credentials.

Regression coverage should verify proxy routing, failure without direct fallback, successful empty responses, and absence of mint credentials on recipient requests.

crodas:
### Description

Fixes #2525 

HTTP delivery built a fresh default client, so a wallet configured with --proxy or Tor still reached the recipient over ordinary network egress.  The recipient learned the payer's address even though every mint request went through the configured transport.

Delivery now rides the wallet's mint connector, alongside the other non-mint calls it already makes. A connector without delivery support fails explicitly rather than falling back to a direct request.

The receiver's response is now parsed as JSON, so a success answered with an empty body is reported as a failure. NUT-18 does not define a response body. Delivery also stops following redirects and is paced by the client rate limiter, both inherited from the connector transport.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist

### DIFF
--- a/crates/cdk/src/wallet/payment_request.rs
+++ b/crates/cdk/src/wallet/payment_request.rs
@@ -255,6 +255,12 @@ impl Wallet {
     /// The returned payment must be explicitly completed with
     /// [`PreparedPaymentRequest::confirm`] or released with
     /// [`PreparedPaymentRequest::cancel`].
+    ///
+    /// A request whose only transport is HTTP POST fails here with
+    /// [`Error::PaymentRequestDeliveryUnsupported`] when the wallet's mint
+    /// connector cannot deliver. Capability is knowable before any proofs are
+    /// reserved, and finding out at confirm time would leave the payer with a
+    /// pending send to revoke.
     #[instrument(skip_all)]
     pub async fn prepare_pay_request(
         &self,
@@ -316,6 +322,12 @@ impl Wallet {
             .ok_or_else(|| {
                 Error::Custom("No transport available in payment request".to_string())
             })?;
+
+        if transport._type == TransportType::HttpPost
+            && !self.client.supports_payment_request_delivery()
+        {
+            return Err(Error::PaymentRequestDeliveryUnsupported);
+        }
 
         let prepared_send = self
             .prepare_send(
@@ -947,7 +959,9 @@ mod tests {
         )
         .await
         .expect("store proof");
-        let wallet = create_test_wallet_with_mock(db, Arc::new(MockMintConnector::new())).await;
+        let mock = MockMintConnector::new();
+        mock.set_payment_request_delivery_supported(true);
+        let wallet = create_test_wallet_with_mock(db, Arc::new(mock)).await;
         let request = payment_request_with_http_transport(Amount::from(100), Amount::from(500));
 
         let prepared = wallet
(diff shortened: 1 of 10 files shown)


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

Codepocketdev:
Code review ACK 8b93bed4

On ambiguous delivery failure, `revoke_send` reclaims but there's no path to retry delivery of the same token — is that intended, or is retry planned as follow-up?

crodas:
@cdk-bot review

crodas:
> On ambiguous delivery failure, revoke_send reclaims but there's no path to retry delivery of the same token — is that intended, or is retry planned as follow-up?

It is unrelated, if it is an issue it should be addresed in a follow-up

crates/cdk/src/wallet/payment_request.rs
thesimplekid, line 1615: Should WalletRepository::prepare_pay_request check each candidate’s delivery capability before ranking it? This PR makes HTTP delivery depend on the selected wallet’s connector, but the repository still selects by balance alone. A higher-balance wallet without delivery support causes preparation to fail even when another accepted, funded wallet can deliver

thesimplekid (changes requested):

PR #2533, opened 2026-09-11
Bound websocket and pub/sub resources

### LINKED ISSUE #2532: Resource management for subscription endpoints
Adds operator-tunable resource limits to the websocket endpoint and to the subscription manager, so a mint operator can size the endpoint for their deployment instead of relying on it being unbounded.

No protocol or wire-format change. NUT-17 subscribe and unsubscribe should behave as before within the limits. Covered by unit tests in cdk-axum, cdk-common and cdk-mintd, plus an integration test.

crodas:
### Description

Fixes #2532 

Give each of those a ceiling the operator can tune, refuse an over-quota upgrade before it happens so the client gets a plain HTTP error rather than a socket closed a moment later, and let a subscription own its backfill so dropping it stops the database and payment backend work.

A ceiling bounds concurrency, not rate, so churning subscriptions still stayed under every cap while repeatedly taking the topic lock that event delivery also needs. Charge each connection a budget, per filter on subscribe, refuse what it cannot cover and close a connection that keeps flooding.

Claim the connection slot before verifying auth, since auth reads the database and may reach the signatory, and parse each frame once instead of building a throwaway value so a refused frame costs almost nothing.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-axum/src/ws/subscribe.rs
```diff
@@ -15,42 +24,84 @@ pub(crate) async fn handle(
...
+        tracing::warn!(
+            "WebSocket subscription request exceeds per-connection topic budget: {} > {}",
+            topics_in_use,
+            limits.max_topics_per_connection
+        );
+        return Err(WsError::ServerBusy);
+    }
+
+    let units = u32::try_from(requested).unwrap_or(u32::MAX);
+    if context.budget.charge(units, Instant::now()) != Charge::Accepted {
```
thesimplekid: Could we propagate `Charge::Exhausted` to the WebSocket loop instead of converting it into `ServerBusy` here? A client can keep the one-unit frame charge affordable while repeatedly exceeding its filter budget, bypassing the disconnect threshold. Keeping both charging stages explicit but routing their outcomes through one enforcement path would simplify the control flow and fix this. Could we test repeated filter-budget violations through the full socket request path?

crates/cdk-common/src/pub_sub/pubsub.rs
```diff
@@ -144,22 +181,56 @@ where
...
         let subscribed_to_for_spawn = subscribed_to.clone();
-
-        spawn(async move {
-            // TODO: Ignore topics broadcasted from fetch_events _if_ any real time has been broadcasted already.
+        let backfill_slots = self.backfill_slots.clone();
+
+        let backfill = spawn(async move {
+            // Backfill reads the database and the payment backend once per topic,
+            // so a burst of subscriptions has to queue rather than fan out.
+            let _permit = match backfill_slots.acquire_owned().await {
```
thesimplekid, line 189: Could we add deadlines to the cancellable database reads and payment-backend status checks performed while this permit is held? With the defaults, 32 stalled checks can occupy every slot and prevent unrelated proof-state backfills from running. The quote-check count bounds calls, not their duration. We should preserve payment commits already underway and explicitly fail subscriptions whose backfill times out. Could we test that a queued subscription proceeds after a stalled check reaches its deadline?

crates/cdk-axum/src/ws/limits.rs
```diff
@@ -0,0 +1,707 @@
...
+            });
+        }
+
+        if self.ping_interval.is_zero() {
+            return Err(WsLimitsError::Zero {
+                field: WsLimitsField::PingInterval,
+            });
+        }
+
+        if self.ping_interval > self.idle_timeout {
```
thesimplekid: Could we require the ping interval to be strictly shorter than the idle timeout, or adjust the scheduling to provide a response window? Validation accepts equality, but the loop consumes the immediate interval tick and checks idle expiry before sending its first ping. With both set to 30 seconds, a quiet, healthy connection can close before receiving a ping to answer. Could we cover this boundary with a socket-level test?

crates/cdk-axum/src/ws/subscribe.rs
```diff
@@ -15,42 +24,84 @@ pub(crate) async fn handle(
...
-            }
-        }),
+        SubscriptionSlot {
+            handle: tokio::spawn(async move {
+                while let Some(response) = subscription.recv().await {
+                    // Dropped rather than awaited on purpose: blocking here would
+                    // let one slow socket stall every other subscriber sharing
+                    // this connection's writer.
+                    if let Err(err) =
+                        publisher.try_send((sub_id_for_sender.clone(), response.into_inner()))
```
thesimplekid: Could we close the connection or explicitly require resynchronization when this queue is full? A burst of historical or live events can make `try_send` fail, but the update is discarded and the subscription stays open. If the dropped event is the final payment-state transition, the client can remain stale without knowing it missed anything. Bounded waiting with a deadline, followed by disconnect on failure, would keep memory bounded and make the delivery failure observable.

crates/cdk-mintd/src/lib.rs
```diff
@@ -569,10 +570,171 @@ pub(crate) fn validate_settings(settings: &config::Settings) -> Result<()> {
...
+fn validate_ws_limits(limits: &config::Limits) -> Result<()> {
+    let ws_limits = limits.ws_limits().map_err(|err| {
+        anyhow!("Invalid limits configuration: ws_trusted_client_ip_header is not a valid HTTP header name: {err}")
+    })?;
+
+    let Err(err) = ws_limits.validate() else {
+        return Ok(());
+    };
+
+    let key = ws_config_key(err.field());
```
thesimplekid: Could we simplify this by preserving `WsLimitsError` and adding configuration context? We map errors to fields, translate fields into configuration keys, and then reconstruct each error message here. That gives future validation changes several places to update. Keeping the original error as the source, while still identifying the setting the operator must change, would reduce this plumbing.

thesimplekid (commented):
I think we should drop per-IP limiting from this PR. The reverse proxy is a better place to enforce that policy, while CDK should retain the connection, subscription, and backfill resource limits. We can add application-level IP limiting in a follow-up if we decide we need it.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk-common/src/pub_sub/pubsub.rs
thesimplekid: The deadline now releases the backfill slot, but a timeout only logs an error and leaves the subscription open. A client waiting for an already-settled quote may never receive another event. Could we explicitly fail/close the subscription or provide bounded retry, and test that the affected client can recover?

crates/cdk-common/src/pub_sub/subscriber.rs
```diff
@@ -85,15 +98,26 @@ where
     S: Spec + 'static,
 {
     fn drop(&mut self) {
-        // remove the listener
+        // The backfill walks the database and the payment backend on behalf of a
+        // listener that is now gone. Aborted first so it cannot observe a
+        // half-removed index; it only ever awaits between statements, so a
+        // dropped database transaction rolls back rather than half-committing.
+        self.backfill.abort();
```
thesimplekid, line 105: Could we preserve payment recording when the initiating subscriber disconnects? `self.backfill.abort()` can cancel the transaction even after the backend reports payment; excluding that transaction from the deadline doesn't protect it from task cancellation. Please cover a disconnect during payment recording and verify that another subscriber still receives the committed state.

crates/cdk-common/src/pub_sub/pubsub.rs
```diff
@@ -38,27 +39,57 @@ where
...
     S: Spec + 'static,
 {
-    /// Create a new instance
+    /// Create a new instance with [`PubsubLimits::default`]
     pub fn new(inner: Arc<S>) -> Self {
+        Self::with_limits(inner, PubsubLimits::default())
+    }
+
+    /// Create a new instance whose shared resources are capped by `limits`
+    pub fn with_limits(inner: Arc<S>, limits: PubsubLimits) -> Self {
```
thesimplekid, line 60: Mint construction now validates these limits, but direct `Pubsub::with_limits` and `PubSubManager::with_limits` callers still bypass validation. Could these constructors return a validation error too? Zero concurrency otherwise parks backfills indefinitely, and values above the semaphore maximum panic.

docs/migrations/v0.18.md
```diff
@@ -130,6 +130,47 @@ the resulting format.
 Only bootstrap variables and variables explicitly referenced by `env:` remain
 runtime inputs.
 
+### Review the New WebSocket Limits
```
thesimplekid, line 133: This change targets v0.19. Could we move the new WebSocket and subscription-limit sections to `docs/migrations/v0.19.md`? Please explicitly note that existing configurations receive defaults without adding fields, while the new limits take effect on upgrade.

thesimplekid (changes requested):
Please address the remaining backfill recovery and cancellation gaps before merging, complete validation for direct pub/sub construction, and move the migration notes to v0.19. I reopened the existing timeout thread with the remaining recovery requirement: https://github.com/cashubtc/cdk/pull/2533#discussion_r4074726283

Validation: all 19 cdk-common pub/sub tests and 34 cdk-axum WebSocket tests passed. The full integration suite was not run. Please add regression coverage for timeout recovery, disconnect during payment recording, and invalid limits through the public constructors.

PR #2534, opened 2026-09-11
fix(wallet): validate mint signatures when receiving proofs

asmogo:
### Description

Receiving ecash could accept a mint swap response with an invalid DLEQ proof or a mismatched amount or keyset, credit unusable proofs, and delete the recovery record. Validate returned blind signatures against the requested outputs before constructing or storing proofs.

Invalid responses now leave the receive saga and transaction pending so recovery can obtain valid signatures later. Valid responses remain supported with or without optional DLEQ proofs.

### Notes to the reviewers

Reproduced on upstream main `e294f845` before applying the fix. Five regression tests cover invalid DLEQ proofs, amount/keyset/count mismatches, preserved recovery, and valid responses with and without DLEQ proofs. This is independent of #2516, which adds Nostr payment-request validation and operation linkage.

Local validation:

- `cargo clippy -p cdk --all-targets --offline -- -D warnings`
- `cargo test -p cdk --lib --offline`: 693 tests passed in the sandbox; the three tests requiring local TCP listeners passed when rerun with the necessary permissions (696 total).

#### FIXED

- Validate mint-returned blind signatures when receiving ecash and preserve recovery state after rejecting an invalid response.

### Checklist

- [x] I followed the code style guidelines.
- [ ] I ran `just quick-check` before committing. Scoped formatting, Clippy, and CDK library tests were run as listed above.
- [x] FFI synchronization checked: no public Wallet API signatures or types changed.

### DIFF
--- a/crates/cdk/src/wallet/receive/saga/mod.rs
+++ b/crates/cdk/src/wallet/receive/saga/mod.rs
@@ -51,6 +51,9 @@ use crate::nuts::nut00::ProofsMethods;
 use crate::nuts::nut10::Kind;
 use crate::nuts::{Conditions, Proofs, PublicKey, SecretKey, SigFlag, State};
 use crate::util::hex;
+use crate::wallet::blind_signature::{
+    validate_mint_response_signatures, SignatureAmountValidation,
+};
 use crate::wallet::saga::{
     add_compensation, clear_compensations, execute_compensations, new_compensations, Compensations,
 };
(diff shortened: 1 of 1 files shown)


thesimplekid:
Duplicate of https://github.com/cashubtc/cdk/pull/2511?

thesimplekid:
Few better tests I took from here and put on https://github.com/cashubtc/cdk/pull/2511

PR #2538, opened 2026-09-13
Reach React Native from the cashu crate without hand-written bindings

crodas:
### Description

An attempt at a better solution for #1982 without any handwritten bindings.

This PR introduces a simpler FFI package, `cashu-ffi`, and exports it through [ReactNative with Nitro](https://nitro.margelo.com/).

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
# React Native bindings: hand-written C ABI vs generated from UniFFI

Two React Native Nitro bindings for the `cashu` crate exist in this repository,
built independently. This is the measured comparison used to decide which to
keep.

| | branch | shape |
|---|---|---|
| **handwritten** | `features/nitro` (`056d98ad`, pushed) | a hand-written flat C ABI, `cdk-nitro` crate |
| **generated** | `feat/react-native-nitro-bindings` (`c20c1fdf`, local) | UniFFI plus an in-repo generator, `cashu-ffi` crate |

Both branch from `ff5a3f07` and both sit on the same `crates/cashu`, so the
crypto is identical and every difference measured below is the binding layer.

## Summary

| dimension | winner | margin |
|---|---|---|
| Correctness | tie | 246/246 checks agree, and both agree with cashu-ts |
| Speed | tie | handwritten by 7 to 10 us per crossing, invisible in batches |
| Binary size | handwritten | 716 KB vs 1985 KB, of which only 86 KB is UniFFI |
| Cost of adding a function | generated | 13 lines in 1 file vs 71 lines across 5 |
| Amount range | generated | exact above 2^53, where the handwritten path silently rounds |
| Ready to ship | handwritten | it has CI, a publish workflow and examples; the other has none |

Recommendation: keep the generator, adopt the handwritten branch's packaging.
See [Recommendation](#recommendation).

## Method

A scratch git worktree held `features/nitro` so neither branch's tree was
disturbed. Both crates were built with `--release`. Both are already driven
from Node through koffi, so one script could load both libraries into the same
process and run identical workloads against them, with cashu-ts as the
baseline.

Caveats that apply to every number here:

- koffi is not JSI. Both native columns pay a crossing cost that neither will
  pay in React Native, so the delta between them is the signal, not the
  absolute value.
- macOS arm64, release only, one machine.
- Timing loops run for a fixed time budget rather than a fixed iteration count.

## 1. Correctness

Both bindings and cashu-ts were run over the NUT-13 vector seed (mnemonic
`half depart obvious quality work element tank gorilla view sugar picture
humble`, keyset `009a1f293253e41e`) and compared field by field:

- single deterministic outputs for counters 0 through 20
- deterministic batches for 1, 8, 31, 1023 and 4096 sats
- random batches, compared on count and denominations

246 of 246 checks agree on `B_`, on the secret, on the blinding factor and on
the amounts. Nothing here separates the two.

## 2. Speed

Milliseconds per call, lower is better.

| workload | outputs | cashu-ts | generated | handwritten | gen/hand |
|---|---|---|---|---|---|
| single random output | 1 | 0.258 | 0.034 | 0.027 | 1.24x |
| single deterministic output | 1 | 1.186 | 0.148 | 0.133 | 1.11x |
| random batch, native call only | 10 | 2.670 | 0.285 | 0.274 | 1.04x |
| random batch, split included | 10 | 2.707 | 0.296 | 0.283 | 1.05x |
| deterministic batch, native call only | 10 | 7.090 | 1.307 | 1.287 | 1.02x |
| deterministic batch, split included | 10 | 6.806 | 1.302 | 1.302 | 1.00x |
| NUT-09 restore, 500 counters | 500 | 591.7 | 65.1 | 65.9 | 0.99x |

Microseconds per output:

| workload | cashu-ts | generated | handwritten |
|---|---|---|---|
| single random output | 257.7 | 33.9 | 27.3 |
| single deterministic output | 1185.6 | 148.3 | 133.5 |
| deterministic batch of 10 | 709.0 | 130.7 | 128.7 |
| NUT-09 restore, 500 counters | 1183.3 | 130.1 | 131.7 |

Reading these:

- UniFFI's `RustBuffer` encoding costs 7 to 10 microseconds per crossing. On a
  133 microsecond BIP32 derivation that is about 8 percent, and amortised over
  a batch it disappears.
- The two "split included" rows measure the end to end call a wallet makes.
  They are level, because the generated binding's batch API takes explicit
  denominations and pays a `splitAmount` in JavaScript, while the handwritten
  one splits in Rust. The two costs cancel.
- The restore row is the one place the generated side leads, and it is an API
  difference rather than an ABI one: it has a batch entry point, so 500
  counters cost one crossing instead of 500.

Both are roughly 8 to 9 times faster than cashu-ts, which is the point of the
exercise for either branch.

## 3. Binary size

Stripped host dylib, which is the shape Android ships:

| | size | exported symbols |
|---|---|---|
| floor: one plain C function over `cashu` | 343 KB | 1 |
| floor: one UniFFI function over `cashu` | 429 KB | 15 |
| handwritten `cdk_nitro` | 716 KB | 12 |
| generated `cashu_ffi` | 1985 KB | 128 |

The two floor crates were built specifically for this comparison: identical
`cashu` dependency, one exported function each, differing only in whether that
function goes through UniFFI.

So **UniFFI's fixed overhead is 86 KB**. The remaining 1.2 MB gap is API
surface, not binding technology. The generated crate exports DLEQ verification,
keyset id computation, amount splitting, NUT-09 restore, a stateful factory
object and batch blinding, none of which the handwritten crate has. Trimming
the exports to what a wallet actually calls would close most of that.

Raw iOS static libraries, for reference: 47.69 MB generated, 29.39 MB
handwritten. Static library size is a poor proxy for what ships, since the
linker drops unused objects.

## 4. Cost of adding a function

The generated figure was measured by removing `unblind_signature` from the
Rust crate and regenerating.

**Generated:** 13 lines of safe Rust in one file, then `just nitro-bindings`.
The generator emits 48 lines across nine files. Nothing outside Rust is written
by hand.

```
crates/cashu-ffi/src/crypto.rs                          13   hand-written
bindings/react-native/src/generated/*.nitro.ts           4   generated
bindings/react-native/cpp/generated/*Ffi.hpp             2   generated
bindings/react-native/cpp/generated/*Bridge.{hpp,cpp}   18   generated
bindings/react-native/cpp/generated/Hybrid*.{hpp,cpp}    9   generated
bindings/react-native/nitrogen/generated/**              2   generated
bindings/react-native/test/generated/*.koffi.mjs        13   generated
```

**Handwritten**, counting the equivalent existing function
(`cdk_create_deterministic_blinded_message`) across its layers:

```
rust/src/ffi.rs                        41   hand-written, unsafe
cpp/cdk_nitro.h                         9   hand-written
src/specs/OutputDataCreator.nitro.ts    6   hand-written
cpp/HybridOutputDataCreator.hpp         1   hand-written
cpp/HybridOutputDataCreator.cpp        14   hand-written
```

71 lines across five files, four of them outside Rust, and all five must be
kept in step by hand. For scale, `ffi.rs` is 689 non-test lines of `unsafe` for
eight exported functions, and the branch had to downgrade the workspace's
`unsafe_code = "forbid"` to `deny` to allow it.

## 5. Amount range

The handwritten C ABI takes amounts as `double`. It validates honestly (its
`to_u64` rejects NaN, infinities, negatives and non-integers), but a JavaScript
`number` cannot express an integer above 2^53 in the first place:

```
requesting an output for 9007199254740993   (2^53 + 1)
  generated   -> 9007199254740993
  handwritten -> 9007199254740992
```

The generated binding maps `u64` to `bigint` end to end. In practice this only
matters away from sats, where max supply is about 2^51, but it does matter for
msat-denominated keysets.

## 6. What each branch already has

| | handwritten | generated |
|---|---|---|
| Publish workflow | `nitro-publish.yml`, 408 lines | none |
| CI jobs | 2, wired into nix `.#bindings` | none |
| Examples | 3, including for coco | none |
| Android manifest, cpp-adapter | yes | no |
| Committed lockfile | yes | no |
| JS test lines | about 1350 across 6 files | about 590 |
| C++ test harness | no | yes, 49 checks over the bridge |
| Benchmark | no | yes |
| cashu-ts parity test | yes | yes |
| Reused by Swift, Kotlin, Python | no | yes, through stock UniFFI generators |
| ABI staleness guard | no | yes, contract version and per-function checksums |
| Structured errors | no, `char* error` | yes, typed variants with fields |
| Stateful objects | no | yes, with handle ownership and a close path |

## Recommendation

Keep the generator and adopt the handwritten branch's packaging.

The case for hand-writing a C ABI rested on it being faster and smaller.
Measured, it is faster by 7 microseconds per call and smaller by 86 KB of
binding overhead. Neither is worth maintaining 71 lines across five files, in
`unsafe`, for every function added, when the alternative is 13 lines of safe
Rust and one command.

Concretely:

1. Rebase `feat/react-native-nitro-bindings` onto `features/nitro`.
2. Delete `rust/src/ffi.rs`, `cpp/cdk_nitro.h`, the hand-written
   `src/specs/OutputDataCreator.nitro.ts` and
   `cpp/HybridOutputDataCreator.{hpp,cpp}`. Restore
   `unsafe_code = "forbid"`.
3. Keep the publish workflow, the CI jobs, the nix wiring, the examples and the
   JS tests, repointed at the generated module.
4. Keep the handwritten branch's in-Rust `splitAmount`, which the generated
   batch API currently pushes back into JavaScript.
5. Trim the exported surface to what the wallet calls, if the 1.2 MB matters.

## Reproducing

```sh
git worktree add /tmp/cdk-features-nitro features/nitro
cargo build --release -p cashu-ffi
(cd /tmp/cdk-features-nitro && cargo build --release -p cdk-nitro)
```

Then run the three scripts against both libraries: a parity check over the
NUT-13 vectors, a timing loop over the workloads in the table above, and a size
comparison against the two floor crates. The generated side's koffi harness is
`bindings/react-native/test/generated/CashuCrypto.koffi.mjs`; the handwritten
side's is `bindings/react-native/test/OutputDataCreatorFFI.mjs` on its own
branch. `just bench-nitro` runs the generated side against cashu-ts on
its own.

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

PR #2541, opened 2026-09-13
fix(sql): update keyset fees in a consistent order

thesimplekid:
Sort fee entries by keyset ID before updating keyset_amounts so concurrent completed operations acquire fee row locks in the same order.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2555, opened 2026-09-16
feat(lnd): make self-payments configurable

thesimplekid:
Add an opt-in mint setting and environment variable for circular LND payments. Keep the setting disabled by default and pass it through to SendPaymentV2.

Cover successful self-payment routing and accounting in the LND regtest.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot

PR #2559, opened 2026-09-16
refactor(sql): replace custom connection pools with Deadpool

asmogo:
Replace CDK's custom PostgreSQL and SQLite connection pools with Deadpool. Library pools own capacity, waiting, acquisition timeouts, recycling and connection return. CDK's adapters retain an owned checkout until transaction completion or bounded rollback; failed or cancelled cleanup detaches the connection.

### Changes

- Delete the custom pool and resource-manager interfaces; migrate shared mint, auth, wallet and key-value SQL to the backend/owned-transaction interfaces.
- Support direct PostgreSQL, PgBouncer session pooling and transaction pooling with protocol prepared-statement tracking. Use transaction-local schema selection and serialize schema bootstrap/migrations with transaction-scoped advisory locks.
- Run SQLite operations through `interact()`, preserve SQLCipher initialization, and fail closed if an initialized in-memory connection is lost.
- Apply UniFFI's Tokio compatibility adapter to database trait delegations, preserving runtime ownership and cancellation when foreign callers use the database factory.
- Preserve quoted/escaped connection credentials and upstream's strict TLS validation/refusal behavior. Add structured cleanup reasons and counters while retaining balanced active-connection gauges.
- Reselect proofs after a known, rolled-back reservation conflict during concurrent send/melt preparation. Limit attempts and propagate other database errors without replaying writes, commits or payments; add a synchronized conflict regression.
- Wait for committed change signatures before returning a settled melt response. Add a deterministic regression at the transaction boundary between payment settlement and change finalization.
- Fix the derivation-counter race exposed by asynchronous SQLite execution: reserve and catch up atomically across database instances, restoring the consecutive-index regression test.
- Run normal PostgreSQL suites concurrently and isolate failure/backend-switch tests on a separate two-server proxy. Verify actual backend PID changes and prepared-statement correctness.

### Compatibility

Normal mint/wallet/auth constructors and high-level wallet methods retain their signatures. Existing PostgreSQL/SQLite database formats and migration files are unchanged. PgBouncer transaction mode requires a maintained release supporting protocol prepared statements and nonzero `max_prepared_statements` (documented starting value: 200); statement pooling is unsupported.

Low-level pool/connector users must migrate to `SqlBackend`, `SqlConnection` and `SqlTransaction`. Custom `WalletDatabase` implementations and FFI database callbacks must implement atomic `reserve_derivation_index`. Supabase deployments must apply `20260917000000_reserve_derivation_index.sql` before upgrading clients (schema version 11).

Rebased onto upstream main, retaining its TLS policy and configuration validation. Both regular and MSRV lockfiles contain the pool dependencies without unrelated dependency upgrades.

### Validation

Passed locally on Rust 1.98:

- Workspace compilation and unit tests: 2,490 passed, two existing tests ignored.
- Workspace all-target Clippy and SQL Clippy with metrics, with warnings denied.
- SQLite and SQLCipher suites, including cleanup metrics and atomic counter tests.
- Complete PostgreSQL matrix: direct, PgBouncer session, and PgBouncer transaction endpoints; keyset rotation/second replica, isolated fault recovery, and verified backend switching. The combined SQL/SQLCipher/metrics/matrix run passed 827 test invocations.
- Generated Python bindings: 31 database, key-value, and wallet/repository tests; Rust FFI unit tests: 109 passed.
- FFI compilation with PostgreSQL, mint-only and wallet-only SQL builds, WebAssembly compilation, and Rust 1.85 SQL checks using the MSRV lockfile.
- Rust formatting, Nix formatting, shellcheck and spelling checks.
- Changed database API doctests and documentation builds with warnings denied.

Final validation on `ca1c257b`: [CI](https://github.com/cashubtc/cdk/actions/runs/35199482859) and [Nutshell interoperability](https://github.com/cashubtc/cdk/actions/runs/35199482795) both pass, with 40 successful checks and four workflow-conditioned skips. This includes the previously failing concurrent wallet, SQLite onchain and mint-only jobs.

The CI follow-up fixes also pass 713 CDK unit tests, both mint-only HTTP configurations, all 27 wallet saga integration tests across memory/SQLite/Redb, workspace Clippy and formatting locally. Deterministic regressions cover conflicting proof reservations and the transaction boundary between payment settlement and change finalization.


crates/cdk-postgres/tests/pool_metrics.rs
```diff
@@ -0,0 +1,147 @@
...
+                .get_label()
+                .iter()
+                .any(|label| label.value() == outcome)
+        })
+        .map(|metric| metric.get_counter().value())
+        .sum()
+}
+
+#[tokio::test]
+async fn checkouts_errors_and_cleanup_balance_gauge() {
```
crodas, line 45: Can we move the tests under their namespaces?

crates/cdk-sqlite/src/connection.rs
```diff
@@ -0,0 +1,258 @@
...
+                dirty: false,
+                pending: None,
+                owner,
+                metrics: Some(ConnectionMetrics::acquired()),
+            })),
+            in_transaction: false,
+        }
+    }
+
+    async fn run<T, F>(&self, operation: F) -> Result<T, Error>
```
crodas, line 141: Does run have to be async?

PR #2562, opened 2026-09-17
fix(ci): keep mintd:latest on the newest stable release

robwoodgate:
### Description

Only move the `mintd:latest` Docker tag when the release is the newest stable version.

The manifest job tags `:latest` on every non-prerelease release. Publishing v0.17.7 on 2026-09-16 therefore re-pointed `:latest` from 0.18.0 to 0.17.7, an image that predates `cdk-mintd config init`. Anyone starting from `:latest` with the 0.18 config flow now fails at boot, which is how cashubtc/cashu-ts#1174 found it.

The tag step now asks the GitHub API for the repository's latest release and only sets `is_latest` when the published tag matches. The `Create latest manifest` step keys off that instead of `is_stable`. The versioned and major.minor manifests are unchanged, so patch releases on older lines still publish normally.

-----

### Notes to the reviewers

GitHub's latest-release marker follows the highest semver tag by default and can be overridden with the "Set as the latest release" checkbox, so the maintainer cutting the release keeps the final say. The job already grants `contents: read`, which is enough for the API read.

`:latest` still points at 0.17.7 today. Re-running the publish workflow for v0.18.0 (or cutting the next 0.18.x) puts it right.

-----

#### FIXED

- Docker publish: a patch release on an older line no longer moves `mintd:latest` backwards.

----

### Checklist

* [ ] I ran `just quick-check` before committing (workflow-only change, checked with actionlint)


thesimplekid:
@cdk-bot review

thesimplekid (approved):
Please don't merge main into feature and rebase instead.

Thanks for findings thing LGTM.

PR #2565, opened 2026-09-17
fix(wallet): enforce the fee reserve during paid melt recovery

mmalmi:
### Description

A paid melt can treat missing change as fees and delete its recovery record. With a 32-sat input, an 8-sat payment and a 1-sat fee reserve, restoring only 16 sats incorrectly records an 8-sat fee. Enforce the saved fee reserve before saving change or completing recovery.

### Notes to the reviewers

All 701 wallet/mint unit tests, strict Clippy and formatting pass. Regressions cover incomplete/excessive change, retry with metadata preserved, full-reserve use and unavailable recovery data. No database or public API changes.

Full `just quick-check` passes on the current PR head (`85403f22`) with Rust 1.98.0, `nixpkgs-fmt` 1.3.0 and `typos` 1.46.3. Changed Rust files also pass the repository’s nightly rustfmt import rules.

#### FIXED

- Retain paid-melt recovery when missing change would exceed the quoted fee reserve.

### Checklist

- [x] Followed the code style guidelines.
- [x] Ran `just quick-check` on the current PR head.
- [x] Wallet API unchanged; no FFI update needed.

PR #2566, opened 2026-09-17
fix(wallet): recover melt change from quote signatures

mmalmi:
### Description

Restart recovery ignores change signatures in a paid melt quote and always requests NUT-09 restore. Use the quote signatures in the original output order through the existing validation and proof-construction path. Request restore when the quote omits change.

### Notes to the reviewers

All 701 wallet/mint unit tests, strict Clippy and formatting pass. Regressions cover unavailable restore, omitted change, and malformed signature counts, keysets, denominations and DLEQ proofs. Metadata and derivation indices are preserved. No database or public API changes.

Full `just quick-check` passes on the current PR head (`78e93daf`) with Rust 1.98.0, `nixpkgs-fmt` 1.3.0 and `typos` 1.46.3. Changed Rust files also pass the repository’s nightly rustfmt import rules.

#### FIXED

- Recover melt change from signatures already provided by a paid quote.

### Checklist

- [x] Followed the code style guidelines.
- [x] Ran `just quick-check` on the current PR head.
- [x] Wallet API unchanged; no FFI update needed.


comwanga:
### Finding

High: `crates/cdk/src/wallet/melt/saga/resume.rs:313`

Should this quote response path enforce the persisted fee reserve before committing the recovered proofs and deleting the saga?

`OutputRecoveryMode::Partial` validates each returned signature, but cannot distinguish unused blank outputs from a truncated change array. On the current PR head, I reproduced this with the existing fixture:

* Input: 32 sats
* Payment: 8 sats
* Fee reserve: 1 sat
* Quote change: one valid 16-sat signature

Recovery accepts the response, records an 8-sat fee, leaves the wallet with only 16 of the expected 24 sats, and deletes the saga. The missing change can therefore no longer be recovered through this operation.

This is the same invariant addressed by PR #2565. Please enforce `maximum_change - recovered_change <= fee_reserve` before any proof or terminal-state writes, or stack this PR on #2565, and add the truncated quote-signature case as a regression test.

The prefix ordering and existing signature validation are otherwise correctly reused. I ran the focused reproduction against commit `78e93daf`; it passed and confirmed the terminal state above.

mmalmi:
Thanks for catching this. Fixed in ed590c39: both quote and restore recovery now enforce the saved fee reserve before saving change or finalizing. Truncated responses leave the saga and input proofs intact for retry.

Added the 16-sat regression case and verified that a later complete response recovers all change. All 703 wallet/mint tests, strict Clippy, and formatting pass locally.

thesimplekid:
@cdk-bot review

mmalmi:
Fixed in eea7e8ff: incomplete quote change now tries NUT-09, and duplicate/unrequested restore outputs are rejected before saving proofs. Added regressions for public recovery, failures, retries, and valid signatures.

If neither source supplies sufficient change, retaining the saga is intentional: a paid melt cannot safely be rolled back, and finalizing would discard its recovery data.

All 707 wallet/mint tests, strict Clippy, and formatting pass.

thesimplekid:
Could you address the conflict. Also I notice your commits do not have correct authorship.

mmalmi:
Rebased onto current `main` to resolve the conflict and corrected all five commits to Martti Malmi (@mmalmi) as author and committer. All 710 wallet/mint tests, strict Clippy, and formatting pass on the rebased head (`6701a5e0`).

thesimplekid (approved):

PR #2573, opened 2026-09-18
feat: add the NUT-02 keyset active window

robwoodgate:
Implements the keyset active window proposed in cashubtc/nuts#438. The mint publishes `active_from` and `active_until` alongside `final_expiry`, and the wallet stops minting new outputs onto a keyset that is about to be retired.

## Why

A mint can inactivate a keyset at any moment today, so a wallet can mint outputs onto one that stops accepting them days later. [NUT-02](https://github.com/cashubtc/nuts/blob/main/02.md) closes that by making a mint announce an intended inactivation with `active_until` and keep at least one keyset per unit active for 30 days. Nothing consumed that guarantee: `active_keyset_with_policy` picked the lowest `input_fee_ppk` and broke ties arbitrarily.

## Changes

- Both fields on `KeySet`, `KeySetInfo` and `MintKeySetInfo`, serialised like `final_expiry` and skipped when absent. Neither is part of the keyset ID, so `Id::v2_from_data` is untouched and a mint may extend them without rotating.
- Threaded through the signatory, both protos, mint rotation, the mint and wallet SQL layers with migrations for sqlite and postgres, the Supabase wallet, and FFI. The FFI `KeySetInfo` previously mirrored no timestamps and hardcoded `final_expiry: None` on conversion, so it gains all three and stops discarding data.
- Wallet selection narrows before it ranks: superseded keyset versions are dropped, then keysets that stop being usable inside the next 30 days, then the cheapest of what remains wins, ties going to the latest `active_until` then `final_expiry`. Absent counts as never. The horizon is dropped rather than failed on when the mint offers nothing that long-lived.
- All four selection sites share one function, including the duplicated copy in the send saga and the two lookups in `retry_on_inactive_keyset`, which previously used a rule that could disagree with the actual choice.

## Impact

Selection only changes where the mint offers a healthier keyset than the cheapest one. The Supabase wallet schema version moves to 11, following that crate's convention, so existing Supabase wallets migrate. `Mint::rotate_keyset` takes two more arguments; the admin gRPC surface is unchanged and passes `None`. The blind-auth wallet still takes the first active keyset, unchanged.


j-kon (changes requested):
### Findings

* **critical: `crates/cdk-sql-common/src/mint/migrations/postgres/20260918000000_add_active_window_to_keyset.sql:1-2` and `crates/cdk-sql-common/src/wallet/migrations/postgres/20260918000000_add_active_window_to_keyset.sql:1-2`**
  The Postgres migrations define the new columns as `INTEGER`:
  ```sql
  ALTER TABLE keyset ADD COLUMN active_from INTEGER;
  ALTER TABLE keyset ADD COLUMN active_until INTEGER;
  ```
  In PostgreSQL, `INTEGER` is a 32-bit signed integer (`int4`, max `2,147,483,647`), which will overflow on Unix timestamps past 2038 (and immediately if millisecond timestamps are ever used). Note that `cdk-supabase` correctly used `BIGINT`. These should use `BIGINT` in the Postgres migrations as well:
  ```sql
  ALTER TABLE keyset ADD COLUMN active_from BIGINT;
  ALTER TABLE keyset ADD COLUMN active_until BIGINT;
  ```

* **critical: `fuzz/fuzz_targets/fuzz_short_keyset_id.rs:21`**
  Adding `active_from` and `active_until` to `cashu::nuts::nut02::KeySetInfo` without default struct construction causes the fuzz target compilation to fail:
  ```text
  error[E0063]: missing fields `active_from` and `active_until` in initializer of `cashu::nuts::nut02::KeySetInfo`
  ```
  Because `fuzz/` is excluded from the root workspace `Cargo.toml`, local workspace `cargo check` does not catch this, but it causes `CI/Fuzz targets compile` to fail.
  Suggested fix:
  ```rust
      let v1_info = KeySetInfo {
          id: v1_id,
          unit: CurrencyUnit::Sat,
          active: true,
          input_fee_ppk: 0,
          active_from: None,
          active_until: None,
          final_expiry: None,
      };
  ```

* **critical: `crates/cdk-ffi/src/types/keys.rs:18`**
  Adding `active_from: Option<u64>` and `active_until: Option<u64>` to `KeySetInfo` in UniFFI without default values breaks generated Python constructor signatures, leading to the failing `CI/FFI Python tests`:
  ```text
  TypeError: KeySetInfo.__init__() missing 2 required positional arguments: 'active_from' and 'active_until'
  ```
  Annotating optional fields with `#[uniffi(default = None)]` preserves backwards compatibility in foreign language bindings:
  ```rust
      #[uniffi(default = None)]
      pub active_from: Option<u64>,
      #[uniffi(default = None)]
      pub active_until: Option<u64>,
  ```

### Summary

Reviewed NUT-02 active window specification changes across `cashu`, `cdk-common`, `cdk`, `cdk-sql-common`, `cdk-supabase`, FFI bindings, and tests. Keyset selection logic based on the active window and input fee is clean and well-tested. However, the Postgres column types risk 32-bit timestamp truncation, and missing fields in the fuzz target and UniFFI bindings are currently causing CI failures.

### Verdict

CHANGES_REQUESTED

PR #2575, opened 2026-09-18
fix(mint): wait for change finalization before exposing paid melts

thesimplekid:
A quote can be Paid before its change signatures commit. Polling, subscription snapshots, and the async completion waiter could expose that intermediate state, allowing a wallet to finalize without collecting change.

Keep responses Pending while the melt-request record exists. Finalization deletes that record atomically with change persistence, including when no change is due. Check it before reading signatures and preserve the durable Paid state for recovery.

The transaction split originated in PR #954:
https://github.com/cashubtc/cdk/commit/64f7b0785569425aca174531f5387a44089267ef

The async waiter inherited the race when generic payment events were added: https://github.com/cashubtc/cdk/commit/00e7e525b55afe916ba4508eaa849716c86e4c67

Add deterministic coverage at the transaction boundary and subscription snapshot coverage for all payment methods, including zero-change completion.

Validation: 100 melt tests, 7 subscription tests, and cargo fmt checks pass.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crates/cdk/src/mint/melt/mod.rs
```diff
@@ -121,10 +121,41 @@ fn validate_custom_quote_fields(request: &str, extra: &serde_json::Value) -> Res
 }
 
 impl Mint {
+    /// Load a quote, keeping its response pending until change and cleanup commit.
+    pub(super) async fn load_melt_quote_for_response(
```
crodas, line 125: Why is this a static method instead of a public/crate function?

crates/cdk/src/mint/melt/mod.rs
```diff
@@ -121,10 +121,41 @@ fn validate_custom_quote_fields(request: &str, extra: &serde_json::Value) -> Res
...
+                if tx
+                    .get_melt_request_and_blinded_messages(quote_id)
+                    .await?
+                    .is_some()
+                {
+                    quote.state = MeltQuoteState::Pending;
+                }
+            }
+        }
+        tx.rollback().await?;
```
crodas, line 150: Why are you rolling back? This seems to be quite inefficient. 

I mean, if you need a transaction, can this function's signature include a DB transaction? Because this function would break from SQLite if call it form a context where there is an ongoing db transaction

PR #2580, opened 2026-09-18
Set BOLT12 mint offer absolute_expiry from mint_ttl

### LINKED ISSUE #2579: BOLT12 mint offers omit absolute_expiry while the quote expires at mint_ttl
## Summary

BOLT12 mint quotes expire at `[info.quote_ttl] mint_ttl` (default 600s), but the **offer** CDK asks the LN backend to mint is created with `unix_expiry: None`. LDK/CLN therefore emit an `lno1` **without** `absolute_expiry`.

Wallets (`bolt12-cli decode`, Lexe, …) see a perpetual offer. The mint has already dropped the quote. Payers keep trying; `ldk-server` never sees a usable InvoiceRequest tied to a live quote. Looks like a Lightning failure. It is a TTL mismatch.

## Observed

Hedwig mint: `cdk-mintd/0.18.0-rc.3` + `cdk-ldk-server-processor` + ldk-server.

Quote in sqlite:

- `payment_method = bolt12`
- `expiry = 0` (no-expiry sentinel on the **offer** side)
- mint still applies `mint_ttl = 600` to the quote

`bolt12-cli decode` of a 5-day-old `lno1`:

```json
{
  "description": "Nice this is a payment for the Cashu wallet",
  "issuer_id": "02096160e8ef5aa13e…",
  "paths": [{ "introduction_node": { "type": "short_channel_id", "…" } }],
  "type": "offer"
}
```

No `absolute_expiry`. Offer still decodes. Paying it cannot credit the mint.

Same mint: 103 unpaid BOLT12 quotes vs 4 ever paid (last paid BOLT12 mint: 1 sat, days earlier).

## Code

`crates/cdk/src/mint/issue/mod.rs` — Bolt11 vs Bolt12 incoming options:

```rust
// Bolt11
let mint_ttl = self.quote_ttl().await?.mint_ttl;
let quote_expiry = unix_time() + mint_ttl;
// ...
unix_expiry: Some(quote_expiry),

// Bolt12
let bolt12_options = Bolt12IncomingPaymentOptions {
    // ...
    unix_expiry: None,   // <-- offer has no absolute_expiry
};
```

`cdk-ldk-node` then:

```rust
let time = unix_expiry.map(|t| t.checked_sub(unix_time()) ...).transpose()?;
self.inner.bolt12_payment().receive(..., time, None)
// time == None → no absolute_expiry on the offer
```

NUT-25 types mint-quote `expiry` as the Unix time the **quote** is valid. #2195 / #2211 fixed serializing `expiry: 0` vs `null` on the **JSON quote**. That is a different layer: wallets that only look at the `lno1` never see quote JSON.

## What I think should happen

When creating a BOLT12 incoming payment for a mint quote, pass **`unix_expiry: Some(quote_expiry)`** (`unix_time() + mint_ttl`), same as Bolt11.

Then:

- offer `absolute_expiry` ≈ quote expiry
- `bolt12-cli decode` shows the offer is dead
- Lexe / other wallets stop paying a quote the mint will not credit

If the intent is a long-lived reusable offer, that should be explicit and **not** bound to a 600s mint quote. Today it is the worst of both: quote dies, offer looks forever.

## Related

- #2195 — Bolt12 mint quote serializes `expiry: 0` instead of null (JSON quote, closed in 0.18)
- Not the compact-intro / private-channel BOLT12 path issue (rust-lightning #4828). This is only TTL on the offer vs the quote.

vincenzopalazzo:
## Summary
- BOLT12 mint quotes expired at `mint_ttl` while the LN offer was created with `unix_expiry: None`, so wallets saw a perpetual `lno1` for a quote the mint had already dropped.
- Root cause: `get_mint_quote` passed `unix_expiry: None` for Bolt12 (Bolt11 and custom methods already used `unix_time() + mint_ttl`).
- Pass `unix_expiry: Some(quote_expiry)` for Bolt12 mint quotes and stamp the same timestamp as `absolute_expiry` on FakeWallet offers so tests can decode it.

Fixes #2579

A follow-up will drop `mint_ttl` on Bolt12 so the quote follows a long-lived offer instead.

## Test plan
- [x] `cargo test -p cdk --lib -- batch_mint_tests -- --test-threads 1`
- [x] `cargo test -p cdk-fake-wallet --lib -- --test-threads 1`
- [x] `cargo clippy -p cdk -p cdk-fake-wallet --all-targets -- -D warnings`
- [ ] CI passes
- [ ] Regression: `bolt12_mint_quote_sets_offer_absolute_expiry_from_mint_ttl` asserts offer `absolute_expiry` equals quote expiry and tracks `mint_ttl`
- [ ] Regression: `bolt12_incoming_offer_includes_absolute_expiry` asserts FakeWallet encodes `unix_expiry` on the offer


crates/cdk-fake-wallet/src/lib.rs
```diff
@@ -841,6 +841,13 @@ impl MintPayment for FakeWallet {
                 let offer_builder = OfferBuilder::new(secret_key.public_key(&secp_ctx))
                     .description(description.clone());
 
+                let offer_builder = match expiry {
+                    Some(unix_expiry) => {
+                        offer_builder.absolute_expiry(Duration::from_secs(unix_expiry))
+                    }
+                    None => offer_builder,
+                };
```
crodas, line 849: Can we use map and unwrap_or ?

vincenzopalazzo: Not a bug -- `OfferBuilder` is consumed by value and is not `Clone`, so `expiry.map(|t| offer_builder.absolute_expiry(...)).unwrap_or(offer_builder)` moves `offer_builder` twice and does not compile. `match` is how we optionally consume it. Happy to change this if I am missing a builder API that takes `&mut self`.

thesimplekid:
Left a comment here https://github.com/cashubtc/cdk/pull/2583#issuecomment-5792285555. 

I would just merge this one now as a bug fix, but I can't push to your fork since its a org account and this needs to be rebased and signed.

PR #2581, opened 2026-09-19
Do not block when dropping an owned runtime from async context

crodas:
### Description

Bug found while testing a target application with Python to test the release of #2582

Foreign callers construct a store from a sync thread, so RuntimeGuard creates and owns a Tokio runtime that the object keeps alive. uniffi runs exported async methods on a runtime of its own and can release that object from those threads, where a runtime's blocking shutdown panics. Every binding saw the panic on teardown even though the calls themselves succeeded.

Detach the runtime when we are already inside one, and keep the graceful blocking shutdown for the ordinary sync drop, where it still waits for in-flight work.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

thesimplekid (approved):

PR #2582, opened 2026-09-19
Package and release Python bindings as wheels

crodas:
### Description

Python was the only binding with no package and no release path, and cdk-python held sources importing a module nobody built. It now gets the same treatment as the others.

bindings/python is a setuptools project with a cdk-ffi-python wrapper crate, a pytest suite porting the scenarios the Dart, Go, Kotlin and Swift suites already run, runnable examples, and a build.sh that rebuilds the wheel from a published cdk-ffi without this repo. The release workflow syncs that tree to cdk-python, builds the synced clone the way cdk-dart does so CI compiles what a downstream user compiles, then commits the generated bindings and the prebuilt wheels into the repo and attaches them to its release, as cdk-go does. Wheels also go to PyPI.

Each wheel is py3-none: the library is loaded with ctypes rather than linked against the CPython ABI, so one per platform is enough. Two things are easy to get wrong here. uniffi names the library it loads after the cdk-ffi namespace rather than the wrapper crate, so the cdylib is renamed on the way into the package. And it reads binding metadata out of the ELF symbol table, so the release profile must not strip before generation or generation silently emits nothing.

Here is my [own fork](https://github.com/crodas/cdk)creating a [Python release](https://github.com/crodas/cdk/actions/runs/35400044668) which is stored in https://github.com/crodas/cdk-python

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
[Here](https://github.com/crodas/cdk/actions/runs/35446828848/job/105907151454) the last commit doing a deploy

crodas:
@cdk-bot review

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,722 @@
...
+        if: ${{ inputs.publish_target == 'pypi' && steps.pypi_token.outputs.present == 'true' }}
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: wheelhouse/
+          password: ${{ secrets.PYPI_TOKEN }}
+
+  python-publish:
+    name: "python: Publish to cdk-python"
+    runs-on: ubuntu-latest
+    needs: [python-package]
```
thesimplekid, line 639: **Wait for PyPI before finalizing the release**

Could we make GitHub release creation, merging, and cleanup wait for successful PyPI publishing when it is requested? Kotlin completes Maven publishing before creating its release. Here the two publishing jobs run independently, so a failed PyPI upload can still leave a finalized GitHub release and a deleted release branch. GitHub-only releases should remain possible when explicitly selected.

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,722 @@
...
+
+      - name: Check for a PyPI token
+        # A release that cannot reach PyPI should still ship the wheels it built,
+        # so a missing token is a warning rather than a failed release. Read
+        # through the environment so the secret is never spliced into a command.
+        id: pypi_token
+        env:
+          PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+        run: |
+          if [ -n "${PYPI_TOKEN}" ]; then
```
thesimplekid, line 614: **Fail when requested publishing cannot happen**

Could we fail when `publish_target` requests PyPI or TestPyPI but `PYPI_TOKEN` is missing? This currently reports success without uploading the package. Matching Kotlin’s registry publishing behavior would make configuration failures visible; `publish_target: none` already provides an explicit GitHub-only option.

justfile
```diff
@@ -1318,3 +1318,72 @@ test-swift:
...
+
+  echo "🧪 Running offline Python examples..."
+  for example in examples/wallet_setup.py examples/token_inspect.py examples/transaction_history.py; do
+    echo "  → $example"
+    "{{PYTHON}}" "$example" > /dev/null
+  done
+  echo "✅ Offline examples ran successfully!"
+
+# Trigger Python Bindings release workflow
+ffi-release-python VERSION:
```
thesimplekid, line 1374: **Align individual and aggregate release commands**

Could `ffi-release-python` select the same publishing target as `ffi-publish-all`? The documented individual release command currently omits `publish_target`, so it defaults to `none` and skips PyPI even for stable releases. A maintainer should get the same publishing behavior through either entry point.

.github/workflows/ffi-publish-all.yml
```diff
@@ -78,3 +78,17 @@ jobs:
...
+  python:
+    name: "Trigger Python bindings"
+    needs: detect-prerelease
+    uses: ./.github/workflows/python-publish.yml
+    with:
+      release_tag: ${{ inputs.release_tag }}
+      cdk_version: ${{ inputs.cdk_version }}
+      cdk_ref: ${{ inputs.cdk_ref }}
+      prerelease: ${{ needs.detect-prerelease.outputs.prerelease == 'true' }}
+      # A pre-release still gets wheels on the cdk-python release, just not a
```
thesimplekid, line 91: **Align prerelease publishing policy**

Could we follow Kotlin’s distinction between nightlies and versioned prereleases? Kotlin publishes non-nightly prereleases to Maven, whereas Python skips PyPI for all prereleases. The workflow already translates alpha, beta, and RC versions into PEP 440 versions. Unless there is a Python-specific reason, publishing those versions while keeping nightlies GitHub-only would make the targets consistent.

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,722 @@
...
+          cat checksums.sha256
+
+      - name: Upload wheels
+        uses: actions/upload-artifact@v7
+        with:
+          name: python-wheels
+          path: wheelhouse/
+          if-no-files-found: error
+
+  python-publish-pypi:
```
thesimplekid, line 590: **Serialize publishing of the same version**

Could we add concurrency protection keyed by the release tag, as Kotlin does for Maven publishing? The individual and aggregate workflows can otherwise attempt to upload the same immutable PyPI version concurrently. This should serialize matching versions without cancelling an in-progress publish.

bindings/python/build.sh
```diff
@@ -0,0 +1,102 @@
...
+  fi
+  PY="$BUILD_VENV/bin/python"
+fi
+
+echo "Building wheel..."
+"$PY" -m build --wheel
+
+# The library is loaded with ctypes rather than linked as a CPython extension,
+# so the wheel is valid for any Python 3 and is retagged to say so.
+PLATFORM=$("$PY" -c 'import sysconfig; print(sysconfig.get_platform().replace("-", "_").replace(".", "_"))')
```
thesimplekid, line 93: **Match local macOS wheel tags to the compiled library**

Could we derive the macOS wheel architecture and minimum OS version from the Rust build rather than Python’s configuration? A universal2 Python can label this wheel as supporting both architectures even though Cargo builds only one. The tag needs to describe the native library actually packaged.

bindings/python/build.sh
```diff
@@ -0,0 +1,102 @@
...
+if python3 -c 'import build, wheel' 2>/dev/null; then
+  PY=python3
+else
+  BUILD_VENV=".build-venv"
+  # Check that the venv actually works rather than that it merely exists: a
+  # copied or half-installed one has absolute paths pointing somewhere else.
+  if ! "$BUILD_VENV/bin/python" -c 'import build, wheel' 2>/dev/null; then
+    rm -rf "$BUILD_VENV"
+    python3 -m venv "$BUILD_VENV"
+    "$BUILD_VENV/bin/pip" install --quiet --upgrade pip "build" "wheel>=0.42"
```
thesimplekid, line 83: **Remove Windows support for now**

Could we just remove Windows support for now? The script handles Windows library names, but assumes `bin/python` and `bin/pip`; native Windows Python creates these under `Scripts`. A clean Windows source build therefore fails when it reaches the venv fallback, and `test-python` has the same assumption. We can remove the Windows release matrix entry, wheel mappings, and advertised support together, updating the expected wheel count accordingly.

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,722 @@
...
+            echo "present=false" >> "$GITHUB_OUTPUT"
+            echo "::warning::PYPI_TOKEN is not set, skipping the upload. The wheels are still committed to cdk-python and attached to its release."
+          fi
+
+      - name: Publish to TestPyPI
+        if: ${{ inputs.publish_target == 'test-pypi' && steps.pypi_token.outputs.present == 'true' }}
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: wheelhouse/
+          password: ${{ secrets.PYPI_TOKEN }}
```
thesimplekid, line 626: **Document the publishing setup and separate registry credentials**

Could we document what maintainers need to configure before publishing?

- `PYPI_TOKEN`: a GitHub Actions secret available to `cashubtc/cdk`, containing a PyPI token authorized to publish `cdk-python`.
- `TEST_PYPI_TOKEN`: a separate secret for TestPyPI, with the workflow selecting the appropriate token. Both upload steps currently use `PYPI_TOKEN`, but the two registries require separate credentials.
- `CDK_PYTHON_REPO`: the repository variable identifying the downstream repository, such as `cashubtc/cdk-python`.
- `FFI_DEPLOY_KEY`: ensure the existing GitHub credential can push branches, create releases, and upload assets in that downstream repository.

thesimplekid (changes requested):
Could we bring Python’s release flow into line with Kotlin and Swift? The source sync, CI gate, and downstream release structure already follow the same pattern. The main differences to resolve are registry publishing, failure handling, and the individual release command.

.github/workflows/python-publish.yml
crodas: It does publishes to git, similar to what we do with golang, kotlin and others http://github.com/crodas/cdk-python

I did this for testing, but I can make sure the PyPI publishing is required

PR #2583, opened 2026-09-19
Keep BOLT12 mint quotes as long-lived offers

vincenzopalazzo:
## Summary
- BOLT12 mint quotes are reusable long-lived offers, not `mint_ttl` invoices.
- Do not pass `unix_expiry` when creating the LN offer, so it has no `absolute_expiry`. Quote JSON expiry follows the offer (`none` / `0`).
- Wallet `expiry == 0` is the no-expiry sentinel so `get_active_mint_quotes` does not drop perpetual BOLT12 quotes.

This is independent of #2580, which aligns Bolt12 quote expiry with `mint_ttl` so the offer is not perpetual while the quote is already dead. Land that first if the desired behavior is TTL-bounded offers; this PR is the alternative of making the quote follow a long-lived offer.

Related: #2579, #2580

## Test plan
- [x] `cargo test -p cdk --lib -- bolt12_mint_quote_follows_long_lived_offer_expiry`
- [x] `cargo test -p cdk-common --lib -- mint_quote_zero_expiry_is_not_expired`
- [ ] CI passes


crates/cdk/src/wallet/issue/mod.rs
crodas, line 676: @thesimplekid, are we expecting multiple matches?  If so, we should explore moving the logic behind the `retain` into the storage layer so it can do a pre-filter or proper filtering.

crodas (approved):
LGTM. Left a question unrelated to the changes introduced here, as a follow up.

thesimplekid:
I think we should combine https://github.com/cashubtc/cdk/pull/2580 and this PR. I don't think we want to take the expiry away completely but should allow it to be set independently from the bolt11 one. And as https://github.com/cashubtc/cdk/pull/2580 does the offer and quote should match. Ideally we go this direction where the wallet can request an expiry and the mint just sets a max expiry https://github.com/cashubtc/nuts/pull/415.

PR #2584, opened 2026-09-19
fix(regtest): work around LDK BOLT12 final CLTV failures

thesimplekid:
Set CLN cltv-final to zero in the regtest launcher. Lightning 0.2.6 omits the advertised final delta for single-hop blinded paths, causing payments to fail when the random shadow-route offset is too small. Keep this workaround confined to regtest and remove it once LDK honors the advertised delta.

Validated with 20 consecutive LDK BOLT12 melts and all eight BOLT12 tests against each of the CLN and LDK mints, with no backend payment failures or CLTV rejections.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas (approved):

thesimplekid:
Thanks for the approval but given the test this was supposed to make less flaky failed on the first run it did not work so going to close this.

PR #2585, opened 2026-09-19
fix: preserve quoted fake custom payment amounts

### LINKED ISSUE #2539: Fake custom payments can execute a different amount than quoted
The fake custom-payment backend can quote one amount and execute a different amount when the typed amount and `extra_json.amount` disagree. This makes the test backend inconsistent with the quote and can hide payment-accounting regressions.

### Current behavior

In `crates/cdk-fake-wallet/src/lib.rs`, `custom_outgoing_amount` prefers the typed amount over the JSON amount. `get_payment_quote` returns that selected amount but preserves the conflicting JSON. When execution options are reconstructed from the melt quote, the typed amount is omitted and the JSON is replayed, so `make_payment` selects a different amount.

For example:

1. Request a custom payment quote with typed amount `20 sat` and `extra_json = {"amount":30}`.
2. The quote amount is `20 sat`.
3. Execute using the same quote ID and returned JSON, with the typed amount omitted as in `OutgoingPaymentOptions::from_melt_quote_with_fee`.
4. The fake backend reports `31 sat` spent (30 plus its simulated fee), rather than `21 sat`.

### Expected behavior

Execution should retain the amount selected during quoting. Normalize the returned backend data or retain the selected amount by quote ID so conflicting input cannot change it between quote and execution. Preserve legitimate JSON-only amount support.

### Regression coverage

Exercise the full quote-to-execution round trip with conflicting typed/JSON amounts, matching amounts, and JSON-only amounts. Assert that the executed amount and simulated fee agree with the quote, including after reconstructing execution options.

Source reviewed at `ff5a3f075ad853ab3cc27f85eea88ebc47c1047a`; reproduction test has not been run. Scope is fake-backend correctness; no external funds are transferred by this backend.

Found by Project Loupe.

comwanga:
## Summary

Fixes #2539.

The fake custom payment backend could execute a different amount from the one it quoted when a typed amount conflicted with extra_json.amount.

For example, a payment with a typed amount of 20 sat and {amount:30} was quoted for 20 sat. After the melt quote was persisted and converted back into OutgoingPaymentOptions, the selected amount was not carried into execution. make_payment() then fell back to the JSON amount and reported 31 sat spent, including the simulated 1 sat fee, instead of 21 sat.

Custom payment reconstruction now carries the authoritative amount selected by the backend and persisted in MeltQuote. This avoids process-local quote state, survives backend or mint recreation, and preserves method-specific extra JSON unchanged. Direct custom payments that do not originate from a persisted melt quote retain the existing typed-amount and JSON-fallback behavior.

## Tests

Added a regression covering the full flow from get_payment_quote() through MeltQuote reconstruction and make_payment() using a newly created FakeWallet for execution.

Before the fix:

```text
left:  Amount { value: 31, unit: Sat }
right: Amount { value: 21, unit: Sat }
```

The tests verify that:

- A typed amount of 20 sat with a JSON amount of 30 spends 21 sat.
- The reconstructed execution options contain the persisted quoted amount.
- Execution remains correct after recreating the fake backend.
- A JSON-only amount of 30 spends 31 sat.
- Unrelated custom JSON fields survive reconstruction.
- Existing fake-wallet behavior remains unchanged.

Validation performed:

- cargo test -p cdk-common -p cdk-fake-wallet
- cargo fmt --all -- --check
- cargo clippy -p cdk-common -p cdk-fake-wallet --all-targets -- -D warnings
- git diff --check


j-kon (commented):
### Findings

* **warning: `crates/cdk-fake-wallet/src/lib.rs:355`**
  `quoted_custom_amounts` is stored in an unbounded `Arc<Mutex<HashMap<QuoteId, Amount<CurrencyUnit>>>>`. Entries are inserted during `get_payment_quote` and queried in `make_payment` via `.get(&custom_options.quote_id).cloned()`.
  
  Because entries are not removed upon payment and unpaid/expired quotes are never cleared, this map will grow monotonically in long-running test suites or local regtest daemons.
  
  Should we remove the quote on successful payment in `make_payment`? E.g.:
  ```rust
  let quoted_amount = self
      .quoted_custom_amounts
      .lock()
      .await
      .remove(&custom_options.quote_id);
  ```
  Or consider bounding/evicting older quotes similar to the `SecondaryRepaymentQueue` pattern if retries are expected.

### Summary

Reviewed the custom payment reconstruction fix and unit tests in `cdk-fake-wallet`. Preserving the quoted typed amount and extra JSON resolves the shortfall bug during custom payment execution. The only residual concern is the unbounded growth of the `quoted_custom_amounts` map.

### Verdict

COMMENT

comwanga:
Good point. I think the safer approach is to replace the unbounded `HashMap` with a small bounded FIFO `VecDeque`, evicting the oldest quote when capacity is reached while keeping `make_payment()` lookups non-destructive.

That bounds memory for both paid and never-paid quotes without making the quote correlation one shot. Removing the entry during `make_payment()` could cause a retry with the same `quote_id` to fall back to `extra_json.amount` again.

I’ll keep the change scoped to `cdk-fake-wallet`.

thesimplekid:
@cdk-bot review

thesimplekid (approved):

PR #2590, opened 2026-09-21
Refuse to take in more proofs than a keyset ever issued

### LINKED ISSUE #2591: Refuse to take in more proofs than a keyset ever issued
The mint tracked total_issued and total_redeemed per keyset, but only as accounting

crodas:
### Description

Fixes #2591

The mint tracked `total_issued` and `total_redeemed` per keyset, but only as accounting: nothing read them on the spending path. A leaked signing key was therefore unbounded. Forged proofs carry valid signatures and fresh secrets, so swap and melt saw nothing to object to and an attacker could redeem without limit.

Make keyset_amounts a ledger. It gains `total_reserved`, for proofs the mint is holding but has not burned, and the invariant becomes redeemed plus reserved never overtaking issued. Taking custody reserves, handing proofs back releases, and burning them moves the amount across. The check therefore runs when the mint takes custody, which is where it has to be: a melt reaches the spend only after its invoice is paid, so refusing there refuses money that is already gone and leaves the quote unfinalizable.

Only a movement that raises what a keyset owes can be refused. Burning proofs the mint already holds leaves the total unchanged, so a melt that paid can always settle, and issuance can never be turned away.

A migration re-derives all three counters from blind_signature and proof, so the cap starts from a balance the database can vouch for rather than one kept incrementally. A keyset that still owes more than it issued has its issued total raised, because an incomplete issuance history must not freeze every proof held against it. That repair is a startup step rather than part of opening the database, so inspecting a mint's database no longer rewrites its accounting.

Issuance moves into the same keyset-ordered pass, which also settles the lock order it used to take. The spend credits only proofs that were not already spent, since this layer validates no transitions of its own. The refusal names the keyset in the log and says nothing on the wire, because the detail is what someone probing a leaked key wants to learn.

This bounds the damage rather than preventing it: an attacker holding the key can still reach the keyset's outstanding float, since forged proofs are indistinguishable from honest ones. It stops the unbounded case and makes the breach visible.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

thesimplekid (commented):
I support the overall approach. Checking the per-keyset issuance limit when we take custody of proofs, before a melt pays out, makes sense. Tracking reservations also accounts for concurrent operations without blocking settlement of proofs we already accepted. My concerns are around transaction-wide lock ordering and the migration path.

crates/cdk/src/mint/swap/swap_saga/mod.rs
```diff
@@ -361,56 +365,55 @@ impl SwapSaga<'_, Signed> {
...
+        let mut proofs = match tx.get_proofs(&self.state_data.ys).await {
+            Ok(proofs) => proofs,
+            Err(err) => {
+                tx.rollback().await?;
+                self.compensate_all().await?;
+                return Err(err.into());
+            }
+        };
+
+        if let Err(err) = Mint::update_proofs_state(&mut tx, &mut proofs, State::Spent).await {
```
thesimplekid, line 386: Could we acquire all affected keyset accounting rows in a consistent order across the whole transaction? This changes swap finalization to lock input keysets before output keysets, while melt change finalization updates output issuance before input-keyset fees. When those keysets differ and the melt has nonzero input fees, the paths can deadlock on PostgreSQL. Sorting each individual ledger pass doesn't resolve that ordering conflict. Could we add a PostgreSQL concurrency test covering both paths?

crates/cdk-sql-common/src/mint/migrations/postgres/20260921000001_recompute_keyset_amounts.sql
```diff
@@ -0,0 +1,33 @@
+-- Re-derive every counter from the rows that are authoritative for it, so the
+-- cap starts from a balance the database can vouch for instead of one that has
+-- only ever been maintained incrementally. fee_collected is not derivable and
+-- is never named here.
+--
+-- The zeroing pass is needed because the join below cannot reach a keyset_amounts
+-- row whose source rows have all gone away.
+UPDATE keyset_amounts SET total_issued = 0, total_redeemed = 0, total_reserved = 0;
```
thesimplekid: Why do we need to recompute both `total_issued` and `total_redeemed` instead of preserving them and only backfilling `total_reserved`? Is there known drift we need to correct?

We should also account for this in the Nutshell-to-CDK migration tooling. We likely need to update it to populate the ledger correctly and ensure migrated mints have the history this backfill expects.

crates/cdk-sql-common/src/mint/migrations/postgres/20260921000000_add_keyset_total_reserved.sql
```diff
@@ -0,0 +1,4 @@
+-- Proofs the mint is holding but has not yet burned. Tracking them lets the
+-- per-keyset cap be checked when the mint takes custody, rather than after a
+-- melt has already paid the invoice.
+ALTER TABLE keyset_amounts ADD COLUMN total_reserved BIGINT NOT NULL DEFAULT 0;
```
thesimplekid, line 4: With larger production mints running CDK now, could we validate these migrations against a realistically populated database and document the expected upgrade time and locking impact, including the historical counter rebuild?

How does the new counter and its arithmetic fit with the unresolved column-type discussion in #2473 and #2568? Could we coordinate the migration sequence to avoid unnecessary repeated migration work?

crodas: It would be the same as we are doing now. An i64 can therefore hold about 4,392 times Bitcoin’s maximum supply in sats, so we are safe.

| Asset | Maximum|
|--|--|
| Maximum Bitcoin supply | 21,000,000 BTC
| Satoshis per BTC | 100,000,000
| Maximum supply | 2,100,000,000,000,000 sats
| i64 maximum | 9,223,372,036,854,775,807

I think #2473 and #2568 pursue correctness for other assets that people may want to represent.

crodas:
@cdk-bot review

crates/cdk-sql-common/src/mint/migrations/postgres/20260921000000_add_keyset_total_reserved.sql
thesimplekid: > 4,392 times Bitcoin’s maximum

This assumes sats an msat mint would be less. However, I meant more around constancy in the other PR we suggested maybe using `NUMERIC` if we use `NUMERIC` there we should probably also use it here.

crates/cdk/src/mint/swap/swap_saga/mod.rs
crodas: You're right that sorting inside a pass can't help, because the conflict is between passes. So the fix was to stop having several. Every movement (issue, reserve, release, redeem, fee) is staged on the transaction as a `BTreeMap<Id, LedgerMove>` and applied in one keyset-ordered pass immediately before `COMMIT`, in `keyset_ledger::commit_moves`. No call site writes keyset_amounts as it goes, so none can leave the ordering. The swap no longer locks inputs before outputs; it locks neither first: both fold into the same sorted map, and so do the melt's change issuance and input fees.

 That gives something stronger than a consistent order. The keyset rows are the last locks a transaction takes, so a transaction holding one never waits for anything else, and no cycle can form through them.

Added the test `swap_and_melt_change_take_keyset_rows_in_one_order`.

PR #2593, opened 2026-09-21
Return the JSON-RPC error code that fits the failure

### LINKED ISSUE #2592: Improve JSON-RPC error
Return the JSON-RPC error code that fits the failure

crodas:
### Description

Fixes #2592

The WebSocket server answered every rejected frame with -32602 "Invalid params", whether the client sent malformed JSON, named a method that does not exist, or sent a well-formed request with bad parameters. WsError already defined the other codes; nothing constructed them.

Classify the failure instead: -32700 for bytes that are not JSON, -32600 for a frame that is not a valid request object, -32601 for an unknown method, and -32602 only for parameters. The envelope is judged from the parsed document rather than from serde's wording, except for telling an unknown method from bad params, which only serde can answer; a test pins that wording.

A frame too malformed to carry an id was previously answered with nothing at all, leaving -32700 unreachable. WsErrorResponse.id becomes optional so that reply can carry the null id JSON-RPC requires. Clients parsing id as a number will need to accept null, though they receive no reply at all today.

Also correct a pub/sub subscribe failure reported as a parse error, which discarded the cause along the way, and drop the unused ServerError variant.

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crodas:
@cdk-bot review

crates/cdk/src/wallet/subscription.rs
```diff
@@ -652,15 +652,15 @@ async fn stream_client(
                         );
                     }
                     RawWsMessageOrResponse::ErrorResponse(error) => {
-                        match pending_requests.remove(&error.id) {
+                        match error.id.and_then(|id| pending_requests.remove(&id)) {
```
thesimplekid, line 655: **warning: Uncorrelated errors terminate all WebSocket subscriptions**

Should the `None` branch log and `continue` instead of reaching the unconditional `return Err`?

Changing `WsErrorResponse.id` to `Option<usize>` means errors with `"id": null` now deserialize successfully; previously they were ignored. They now terminate the shared WebSocket stream despite matching no pending request, interrupting all active subscriptions and triggering polling/reconnection backoff.

Please add a regression test showing that an `id: null` error does not prevent subsequent subscription notifications from being delivered.

crates/cdk-axum/src/ws/mod.rs
```diff
@@ -37,21 +39,97 @@ async fn process(
...
+/// with which JSON-RPC error, and which request id can still be echoed back.
+///
+/// The happy path deserializes straight from the text. Only a rejected frame
+/// pays for the `serde_json::Value` tree, which is several times the size of the
+/// message and is needed solely to classify the failure and recover the id. Any
+/// category but [`Category::Data`] means the bytes were not JSON at all, so
+/// there is no document to read an id out of.
+fn deserialize_request(text: &str) -> Result<WsRequest, Rejection> {
+    let err = match serde_json::from_str::<WsRequest>(text) {
+        Ok(request) => return Ok(request),
```
thesimplekid, line 61: **warning: Version validation only runs when deserialization fails**

Should `request.jsonrpc` be validated before returning `Ok(request)`?

A request with `"jsonrpc":"1.0"` and valid parameters still reaches the handler. For example, `{"jsonrpc":"1.0","method":"unsubscribe","params":{"subId":"existing"},"id":5}` can remove a subscription instead of returning `-32600`.

The existing version test uses invalid parameters, masking this bypass. Please cover an unsupported version with otherwise valid parameters. This is an incomplete fix for pre-existing behavior, rather than a new regression.

thesimplekid (commented):
Two findings below: a wallet stream regression and incomplete request-version validation. All 14 focused WebSocket tests passed; an isolated parser reproduction confirmed the version-validation bypass.

PR #2595, opened 2026-09-22
fix(payment-processor): stop event forwarding on client disconnect

thesimplekid:
Cancel subscription setup, event forwarding, and retry waits when the client disconnects or the server shuts down. Exit the task on send failure instead of resubscribing.

Add regression coverage for cancellation during setup, idle streams, and retry waits.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

PR #2596, opened 2026-09-22
feat(mint-rpc): add melt quote inspection and resolution

thesimplekid:
Add ListMeltQuotes with filters and pagination for inspecting stored quotes and recovery operations without triggering payment recovery.

Add ResolveMeltQuote and CLI commands for operator-confirmed payment finalization or compensation, with operation and saga-state guards, durable resolution records, and idempotent retries. Include usage documentation and validation, filtering, and recovery tests.

### Description

-----

### Notes to the reviewers

-----

#### FIXED

----

### Checklist


thesimplekid:
@cdk-bot review

crates/cdk-mint-rpc/src/proto/server/quote.rs
```diff
@@ -124,6 +128,204 @@ impl MintRPCServer {
...
+        &self,
+        request: Request<crate::quote::ResolveMeltQuoteRequest>,
+    ) -> Result<Response<crate::quote::ResolveMeltQuoteResponse>, Status> {
+        use cdk::mint::{MeltQuoteResolution, MeltQuoteResolutionAction, MeltQuoteResolutionError};
+        use cdk_common::mint::MeltFinalizationData;
+        use cdk_common::payment::PaymentIdentifier;
+
+        use crate::quote::resolve_melt_quote_request::Resolution;
+
+        self.ensure_mutation_allowed().await?;
```
orangeshyguy21: Melt quote resolution doesn't have an explicit opt-in like mint quote payment overrides do. 

Should we generalize the flag to cover both? Both allow an operator to assert a payment outcome without backend verification, so applying the same default-disabled policy may be appropriate.

crates/cdk/src/mint/melt_resolution.rs
```diff
@@ -0,0 +1,837 @@
...
+                    || matches!(
+                        saga.state,
+                        SagaStateEnum::Melt(MeltSagaState::PaymentFailed) | SagaStateEnum::Swap(_)
+                    )
+                {
+                    return Err(Conflict(
+                        "operation cannot be finalized from its current state".to_owned(),
+                    ));
+                }
+                if saga.state == SagaStateEnum::Melt(MeltSagaState::Finalizing) {
```
orangeshyguy21: Should this allow operator-supplied payment details when a legacy `Finalizing` saga has no `finalization_data`? 

Currently `None != Some(payment)` always rejects it.

crates/cdk-mint-rpc/src/proto/quote.proto
```diff
@@ -12,6 +17,124 @@ service QuoteService {
...
+message MeltQuoteSaga {
+    // Correlation ID for locating the operation in mint logs.
+    string operation_id = 1;
+    // Persisted recovery stage: setup_complete, payment_attempted,
+    // payment_pending, payment_failed, or finalizing.
+    string state = 2;
+    // Unix timestamps in seconds; updated_at is the last persisted saga update,
+    // not the time of the last payment backend check.
+    uint64 created_at = 3;
+    uint64 updated_at = 4;
```
orangeshyguy21: Could we expose the saved finalization result here, including total_spent?

Resolving an already-finalizing operation requires those exact details, but inspection doesn't return them. 

A client implementation of the rpc cannot offer a simple “finish cleanup” action unless it already retained those details or asks the operator to supply them.

crates/cdk-mint-rpc/src/proto/quote.proto
```diff
@@ -12,6 +17,124 @@ service QuoteService {
...
+}
+
+message ResolveMeltQuoteResponse {
+    string quote_id = 1;
+    string operation_id = 2;
+    // PAID after finalization or UNPAID after compensation, with cleanup complete.
+    MeltQuoteState state = 3;
+}
+
+message ListMeltQuotesRequest {
```
orangeshyguy21: A couple filters that are useful for mint data exploration:

* reversed (show newest items first)
* unit
* creation_date_start (items <= stamp)
* creation_date_end (items >= stamp)

There's work to be done to allow efficient exploration of the mints database, so these could be added after optimizations. 

But, adding a reversed flag should probably be prioritized. paginating backwards from now() is more common pattern than paginating forward from birth - as a troubleshooting efforts are biased towards the present.

orangeshyguy21 (commented):
Great foundation for an operator UI. 
Specific feedback and question to ponder were provided in the diffs

As we continue the RPC work, there are two potential follow-ups:

- **Database-side filtering and pagination.** The README acknowledges that listing loads and filters the full melt quote history in memory. We should move this into the database so listing remains practical as history grows, and useful filtering options can be added without db burden.
- **Expose the saved resolution decision and whether cleanup remains pending.** If an interruption occurs before response is collected, a client should be able to inspect what was accepted and continue cleanup. Currently, retrying requires the exact original request, including its reason and expected state, so a client would need to persist that request before submission and retain it across reloads. These decision logs could also be a useful list on their own.

